Issue #15, opened 2024-04-08
Create block_state for block_state_legacy for snapshot during transition


heifner:
- Similar to how block_state data for SHiP is created during transistion

PR #45, opened 2024-04-17
Create block_state for block_state_legacy for snapshot correctly during transition

linh2931:
* Create Transition Savanna block for snapshots during Savanna transition.
* Read both Transition Legacy block and Transition Savanna block for snapshot created during transition.
* Use Transition Savanna block as the starting point to recreate proper block
* Add integration tests for snapshots created in both regular and irreversible mode.

Resolved https://github.com/AntelopeIO/spring/issues/15


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Improve support for snapshots during transition by detecting transition and reading both legacy and transition blocks.
Note:end

PR #54, opened 2024-04-19
IF: Add the beginning of a savanna disaster recovery test

heifner:
Integration test with 4 finalizers (A, B, C, and D).
   - The 4 nodes are cleanly shutdown in the following state:
     - A has LIB N. A has a finalizer safety information file that locks on a block after N.
     - B, C, and D have LIB less than N. They have finalizer safety information files that lock on N.

   All nodes but A lose their reversible blocks and restart from an earlier snapshot.

   A is restarted and replays up to block N after restarting from snapshot. Block N is sent to the other
   nodes B, C, and D after they are also started up again.

### DIFF
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -222,8 +222,8 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
    };
 
    // compute strong and weak accumulated weights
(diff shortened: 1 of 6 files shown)


ericpassmore:
Note:start
group: IF
category: TEST
summary: Disaster recovery test with four finalizers. Ensure block N on one node may be recovered after losing reservsible blocks and starting from snapshot.
Note:end

PR #57, opened 2024-04-22
IF: pack size of bitset

heifner:
Change the `pack`/`unpack` scheme for `boost::dynamic_bitset` used by `quorum_certificate`.

Before this PR the number of blocks of the underlying storage was packed. On `unpack` the size of the dynamic bitset was the size of the number of bits in the underlying blocks. 

With this PR, the size of the bitset is pack/unpack so that the bitset can be restored to its original size.

Included in this PR is the introduction of `fc::dynamic_bitset` defined to be `boost::dynamic_bitset<uint8_t>`. Using `uint8_t` minimizes space used for packed format and provides a more expected packed format where only the bytes that are needed are packed/unpacked.

`fc::dynamic_bitset` is used for votes in `quorum_certificate` which is included in most blocks.


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Improve pack/unpack function which is used for votes in quorum_certificate.
Note:end

PR #60, opened 2024-04-23
IF: Do not allow set_finalizers during transition

heifner:
Setting finalizers during transition causes the transition to fail.


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Do not allow set_finalizers during transition.
Note:end

Issue #38, opened 2024-04-16
Test Failure: nodeos_read_terminate_at_block_if_lr_test 


linh2931:
https://github.com/AntelopeIO/spring/actions/runs/8707646627/job/23883711651#step:4:711

PR #61, opened 2024-04-23
Fix occasional terminate_at_block failures

linh2931:
https://github.com/AntelopeIO/spring/actions/runs/8707646627/job/23883711651#step:4:729 reports a failure of nodeos_read_terminate_at_block_test. It was triggered by https://github.com/AntelopeIO/spring/blob/053b53b94d6f977bbcef9b1a43b4b4e7ce4c27c2/tests/nodeos_read_terminate_at_block_test.py#L114
with `AssertionError: head 76 termAtBlock 75`. This indicates nodoes did not stop at the specified terminate-at block, but a block after it.

The PR fixes the problem by `log_irreversible` checking each block which is to become LIB and make sure it not pass the terminate-at block.

Resolved https://github.com/AntelopeIO/spring/issues/38


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Correctly terminate-at-block when running in irreversible mode.
Note:end

libraries/chain/controller.cpp
```diff
@@ -3441,6 +3442,12 @@ struct controller_impl {
                      const trx_meta_cache_lookup& trx_lookup ) {
       try {
          try {
+               if( conf.terminate_at_block > 0 && conf.terminate_at_block <= bsp->block_num() ) {
+                  ilog("Block ${n} reached configured maximum block ${num}; terminating", ("n", bsp->block_num())("num", conf.terminate_at_block) );
+                  shutdown();
+                  return;
+               }
+
```
greg7mdp, line 3450: indentation.

greg7mdp (approved):

libraries/chain/controller.cpp
linh2931: Thanks.

Issue #62, opened 2024-04-23
IF: Test Failure: snapshot_in_svnn_transition_test 


heifner:
https://github.com/AntelopeIO/spring/actions/runs/8801519365/job/24155565450

The request for the snapshot of block 93 came in too late to the node so it was not processed. Seems the test should not request a specific block or it should request a block more than 1 in the future to avoid this race condition.

PR #64, opened 2024-04-23
Fix scheduled snapshot requests coming too late in snapshot_in_svnn_transition_test

linh2931:
In race conditions, scheduled snapshot requests can be processed too late, which results in no snapshots taken. Changed to send snapshot requests directly.

Resolved https://github.com/AntelopeIO/spring/issues/62.


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Improve snapshot scheduling to avoid race conditions, and fixes issues where no snapshot was taken during transition to Savanna.
Note:end

PR #68, opened 2024-04-24
IF: Do not set new proposer policy unless different

### LINKED ISSUE #6: Avoid adding new proposer policy if there is no change
The host function should check to see if the policy to be proposed is different than the last proposal policy in the queue (not the currently active one). If it is the same, it should avoid actually adding the policy to the queue (it acts like a no-op). 

The host function would uint32_t max as the version number if the proposed policy is not invalid. If it is invalid (e.g. empty producers in the policy) it would return -1; this is the only time is returns -1. This means that if the caller of the host function proposes a policy that is identical to the one it last proposed, the host function will still return successfully (not with a -1 return value).

The diff in the block extension is relative to the previous policy proposed and represented in a block extension (as a diff) in the history of the chain.

This is a change the protocol, so if we want to do it we need to get it done before rc1. But it also is not necessary to make this change to the protocol prior to release.

heifner:
- Avoid adding new proposer policy if there is no change.
- Returns the same version from host function if set proposer policy is ignored because there is no diff from the active proposer policy or the next proposer policy.

Resolves #6

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -492,24 +492,27 @@ struct building_block {
       uint32_t get_block_num() const { return block_num; }
 
       // returns the next proposer schedule version and true if different
-      //         if producers is not different then returns the current schedule version (or next schedule version)
-      // uses current building_block timestamp
+      // if producers is not different then returns the current schedule version (or next schedule version)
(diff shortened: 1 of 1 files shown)


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Check and test to ensure proposer policy is only set when there is a change.
Note:end

Issue #59, opened 2024-04-22
Add integration test for configuration where producer and finalizers nodes are separate


arhag:
Let's make sure we explicitly test that the vote from a finalizer key on a node without a producer is incorporated into the QC.

PR #71, opened 2024-04-25
Integration test for configuration where producer and finalizer nodes are separate

linh2931:
The test configures 2 producer nodes and 3 non-producer nodes; each of them has a BLS finalizer key. This results in a threshold of 4. The test verifies LIB advances, which implies at least 2 non-producer finalizer participates in the finalization process. It also checks the votes in `qc_extention` are from the 5 finalizers.

Resolved https://github.com/AntelopeIO/spring/issues/59

### DIFF
--- a/tests/separate_prod_fin_test_shape.json
+++ b/tests/separate_prod_fin_test_shape.json
@@ -36,6 +36,7 @@
           "bios",
           "testnet_01",
           "testnet_02",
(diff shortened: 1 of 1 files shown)


ericpassmore:
Note:start
group: IF
category: TEST
summary: Integration test when producer and finalizer nodes are separate.
Note:end

PR #72, opened 2024-04-25
IF: Disaster_recovery scenario 2 test

heifner:
## Disaster Recovery Scenario 2

Integration test with 5 nodes (A, B, C, D, and P). Nodes A, B, C, and D each have one finalizer but no proposers. Node P has a proposer but no finalizers. The finalizer policy consists of the four finalizers with a threshold of 3. The proposer policy involves just the single proposer P.

A, B, C, and D can be connected to each other however we like as long as blocks sent to ndoe A can traverse to the other nodes B, C, and D. However, node P should only be connected to node A.

At some point after IF transition has completed and LIB is advancing, block production on node P should be paused. Enough time should be given to allow and in-flight votes on the latest produced blocks to be delivered to node P. Then, the connection between node P and node A should be severed, and then block production on node P resumed. The LIB on node P should advance to but then stall at block N. Then shortly after that, node P should be cleanly shut down.

Verify that the LIB on A, B, C, and D has stalled and is less than block N. Then, nodes A, B, C, and D can all be cleanly shut down.

Then, reversible blocks from all nodes should be removed. All nodes are restarted from an earlier snapshot (prior to block N).

P is restarted and replays up to block N after restarting from snapshot. Blocks up to and including block N are sent to the other nodes A, B, C, and D after they are also started up again.


tests/disaster_recovery_2.py
```diff
@@ -0,0 +1,132 @@
...
+    for node in [node0, node1, node2, node3, node4]:
+        node.removeReversibleBlks()
+        node.removeState()
+
+    for i in range(5):
+        isRelaunchSuccess = cluster.getNode(i).relaunch(chainArg=" -e --snapshot {}".format(node0.getLatestSnapshot()))
+        assert isRelaunchSuccess, f"node {i} relaunch from snapshot failed"
+
+    for node in [node0, node1, node2, node3, node4]:
+        assert node.waitForLibToAdvance(), "Node did not advance LIB after relaunch"
```
greg7mdp, line 125: I think we should make sure that lib advances past libN.

tests/disaster_recovery_2.py
```diff
@@ -0,0 +1,132 @@
...
+    assert ret is not None, "Snapshot creation failed"
+    ret_head_block_num = ret["payload"]["head_block_num"]
+    Print(f"Snapshot head block number {ret_head_block_num}")
+
+    Print("Wait for snapshot node lib to advance")
+    node0.waitForBlock(ret_head_block_num+1, blockType=BlockType.lib)
+    assert node1.waitForLibToAdvance(), "Ndoe1 did not advance LIB after snapshot of Node0"
+
+    assert node0.waitForLibToAdvance(), "Node0 did not advance LIB after snapshot"
+    currentLIB = node0.getIrreversibleBlockNum()
```
greg7mdp, line 91: Is this line needed (redundant with line 94 I think)

greg7mdp (approved):

tests/disaster_recovery_2.py
```diff
@@ -0,0 +1,138 @@
...
+import os
+import shutil
+import signal
+import time
+from TestHarness import Cluster, TestHelper, Utils, WalletMgr
+from TestHarness.Node import BlockType
+
+###############################################################
+# disaster_recovery - Scenario 2
+#
```
linh2931, line 11: I like the detailed description here. It will be even better for each of scenarios to have one sentence describing what the purpose of the test.

I think our new tests should follow this test's pattern.

tests/disaster_recovery_2.py
```diff
@@ -0,0 +1,138 @@
...
+    ret = node0.createSnapshot()
+    assert ret is not None, "Snapshot creation failed"
+    ret_head_block_num = ret["payload"]["head_block_num"]
+    Print(f"Snapshot head block number {ret_head_block_num}")
+
+    Print("Wait for snapshot node lib to advance")
+    node0.waitForBlock(ret_head_block_num+1, blockType=BlockType.lib)
+    assert node1.waitForLibToAdvance(), "Ndoe1 did not advance LIB after snapshot of Node0"
+
+    assert node0.waitForLibToAdvance(), "Node0 did not advance LIB after snapshot"
```
linh2931, line 90: Is this the same as line 87?

tests/disaster_recovery_2.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Print("Pause production on Node0")
+    lib = node0.getIrreversibleBlockNum()
+    ret_json = node0.processUrllibRequest("producer", "pause")
```
linh2931, line 94: `ret_json` is not used.

Will a race condition occurs where `pause` is processed very fast such that node0 has not produced a block after `lib`, and no lib would advance.

The same for `disconnect` below.

tests/disaster_recovery_2_test_shape.json
```diff
@@ -0,0 +1,124 @@
...
+          }
+        ],
+        "peers": [],
+        "producers": [
+          "eosio"
+        ],
+        "dont_start": false
+      },
+      "testnet_00":{
+        "name": "testnet_00",
```
linh2931, line 25: It might be better to make the names match with node names in the test. But we can do this next time.

tests/disaster_recovery_2.py
heifner: No, this is just that LIB continues to advance. Probably not really needed, but doesn't hurt to verify.

tests/disaster_recovery_2.py
heifner: `pause` does not interrupt a block. It only prevents the next block from being produced. So should always produce the current block. There is a chance that this happens right after block is produced but before the start of the next block. Not sure how likely that is in practice.

tests/disaster_recovery.py
```diff
@@ -9,6 +9,8 @@
 ###############################################################
 # disaster_recovery - Scenario 1
 #
+# Verify that if one node in network says locked blocks then consensus can continue.
```
linh2931, line 12: `says`?

heifner: has

linh2931 (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Add a test covering disaster recovery with 5 nodes (A, B, C, D, and P).
Note:end

PR #74, opened 2024-04-26
IF: Update new proposer policy algorithm

### LINKED ISSUE #6: Avoid adding new proposer policy if there is no change

heifner:
`set_proposed_producers[_ex]` host functions after Savanna activation check to see if the policy to be proposed is different than the last proposal policy in the queue (not the currently active one). If it is the same, it avoids adding the policy to the queue (it acts like a no-op). 

The host functions now return uint32_t max as the version number if the proposed policy is valid. If it is invalid (e.g. empty producers in the policy) it returns `-1`; this is the only time is returns `-1`. This means that if the caller of the host function proposes a policy that is identical to the one it last proposed, the host function will still return successfully (not with a -1 return value, but with a uint32_t max).

The diff in the block extension is relative to the previous policy proposed and represented in a block extension (as a diff) in the history of the chain. (diff is begin worked under #5).

Resolves #6


linh2931 (dismissed):

greg7mdp:
I'm actually not inderstanding that whole lookup by `active_time` in `proposer_policies`. 
It looks that the `active_time` will always be `detail::get_next_next_round_block_time(bb.timestamp())` (see https://github.com/AntelopeIO/spring/blob/c4bbe578cc47d0817bd5aa42132eebd629af0c62/libraries/chain/controller.cpp#L3150).

Will `if (itr != parent.proposer_policies.begin()) {` ever be true? What are the cases where we have more than one entry in this map?

heifner:
> It seems inconsistent. I'm not quite understanding why we used `proposer_policies.end())` before?

That was an error, corrected in this PR.

heifner:
> Are we returning a policy which will become active later? If that's the case, even if the one proposed in the current block is identical, shouldn't we take that one into account rather than making it a no-op?

The same block is handled in the `set_proposed_producers` method.

heifner:
> Will `if (itr != parent.proposer_policies.begin()) {` ever be true? What are the cases where we have more than one entry in this map?

` round A [1,2,..12], next_round B [1,2,..12], next_next_round C [1,2,..12], D [1,2,..12]`

If set in `A1, A2, .. A12` becomes active in `C1`
If set in `B1, B2, .. B12` becomes active in `D1`

e.g. If one is set in `A2` and another in `B2` then there can be 2 in the map at `B3`.

greg7mdp:
I think there is still an issue:

suppose that:
- on `A2`, we did set `P1` which will become active on `C1`. 
- on `B2`, we did set `P2` which will become active on `D1`, and got added to the `proposer_policies` map when assembling block `B2`

Now we are on block `B3`. map contains `{ C1:P1, D1:P2 }`

We propose `P3`, which would also become active on `D1`. 

`get_next_sched` returns `P1`. Suppose it compares the same as `P3`, so we decide to not add a new policy in the `gpo`. Which means that the `proposer_policies` map will still contain `D1:P2` when it shouldn't.

greg7mdp:
Same issue with:
```
// no previous to what will be replaced, use active
return active_proposer_policy->proposer_schedule;
```

The map may already contain a policy set to activate on D1, and if the one we propose is different we better take that into account.

greg7mdp:
I think that optimization should *not* be done in `controller_impl::set_proposed_producers`, but in  `finish_next` in `block_header_state`.

greg7mdp:
> > I think that optimization should _not_ be done in `controller_impl::set_proposed_producers`, but in `finish_next` in `block_header_state`.
> 
> The requirement is to return the correct version from the host function to the contract.

It doesn't seem possible to do though as if a policy is different, you return `version+1`, but if later in the same block you call the host function again with a policy identical to what was set previously, you'd return `version`?

greg7mdp:
Instead of a `flat_map`, I think it would be easier to use a `std::array<proposer_policy_ptr, 2>`

heifner:
Added additional test cases and added a fix. Ultimately, it is rather complicated to calculate what the version will be ahead of time. If you can think of additional test cases please let me know.

greg7mdp:
> Added additional test cases and added a fix. Ultimately, it is rather complicated to calculate what the version will be ahead of time. 

I don't see how it is possible to return what the version will be, since a further call to `set_proposed_producers` could revert back to an earlier version.

I wonder:
- why do we need to return what the version will be? What is it used for? Is it a problem if successive calls (potentially in different blocks) return decreasing versions?
- I feel that the optimization to "not set new proposer policy unless different" is adding a lot of complexity, and causing more potential issues than the benefit it provides. 
- Is this optimization a protocol improvement which will have any significant impact in the community? If this optimization is not 100% flawless and results in errors in proposer_policy changes, could this have detrimental repercussions in the community?

With my current linited understanding of the pros and cons, I think we should revert to not doing it.

libraries/chain/controller.cpp
```diff
@@ -491,38 +491,30 @@ struct building_block {
 
       uint32_t get_block_num() const { return block_num; }
 
-      // returns the next proposer schedule version and true if different
+      // returns the next proposer schedule version and true if producers should be proposed in block
       // if producers is not different then returns the current schedule version (or next schedule version)
       std::tuple<uint32_t, bool> get_next_proposer_schedule_version(const vector<producer_authority>& producers) const {
```
greg7mdp, line 496: I don't think this function is needed anymore. 
If `set_proposed_producers` is called in a block, the version should be the version last proposed in a previous block + 1.
The version doesn't have to be stored in the `global_property_object`.  I think it could be tracked in the `block_header_state` just like `finalizer_policy_generation`. Whenever a block has at least one `set_proposed_producers` call, the new `proposer_policy` is included in the `if_extension` with the version being `bb.parent.proposer_policy_generation + 1`

libraries/chain/controller.cpp
```diff
@@ -5216,25 +5208,39 @@ int64_t controller_impl::set_proposed_producers( vector<producer_authority> prod
...
+                     gpo.proposed_schedule.producers.begin(), gpo.proposed_schedule.producers.end())) {
+         return std::numeric_limits<uint32_t>::max(); // the proposed producer schedule does not change
+      }
+      // clear gpo proposed_schedule as we may determine no diff between proposed producers and next proposer schedule
+      db.modify( gpo, [&]( auto& gp ) {
+         gp.proposed_schedule_block_num.reset();
+         gp.proposed_schedule.version = 0;
+         gp.proposed_schedule.producers.clear();
+      });
+   }
```
greg7mdp, line 5224: not needed anymore I think, we can just overwrite the last one.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -68,8 +68,18 @@ struct block_header_state {
...
+   //   round A [1,2,..12], next_round B [1,2,..12], next_next_round C [1,2,..12], D [1,2,..12]
+   //   If proposed in A1, A2, .. A12 becomes active in C1
+   //   If proposed in B1, B2, .. B12 becomes active in D1
+   // This is a `deque` because the same active block time can contain up to 12 entries (one per block)
+   // for the entry proposed in a block:
+   //   For example, proposer policy proposed: P1 in A1, P2 in A7, P3 in A12, P4 in B2, P5 in B12 then the map contains:
+   //       [C1 -> P1],[C1 -> P2],[C1 -> P3],[D1 -> P4],[D1 -> P5]
+   //   At C1 P1,P2,P3 are applied causing an active policy of P3
+   //   At D1 P4,P5 are appliced causing an acive policy of P5
+   std::deque<std::pair<block_timestamp_type, proposer_policy_ptr>> proposer_policies;
```
greg7mdp, line 82: I cannot think of any reason why we wouldn't just store the one that will become active at any given time.

libraries/chain/controller.cpp
```diff
@@ -5216,25 +5208,39 @@ int64_t controller_impl::set_proposed_producers( vector<producer_authority> prod
...
    ilog( "proposed producer schedule with version ${v}", ("v", sch.version) );
 
-   // overwrite any existing proposed_schedule set earlier in this block
+   // store schedule in gpo so it will be rolledback if transaction fails
    auto cur_block_num = chain_head.block_num() + 1;
-   auto& gpo = db.get<global_property_object>();
    db.modify( gpo, [&]( auto& gp ) {
       gp.proposed_schedule_block_num = cur_block_num;
       gp.proposed_schedule = sch;
    });
```
greg7mdp, line 5241: Same we can just overwrite the last one I think.

greg7mdp (changes requested):

libraries/chain/controller.cpp
```diff
@@ -491,38 +491,30 @@ struct building_block {
...
-
             // none currently in-flight, use active
             return active_proposer_policy->proposer_schedule;
          };
 
          const producer_authority_schedule& lhs = get_next_sched();
 
-         if (std::ranges::equal(lhs.producers, producers)) {
-            return {lhs.version, false};
+         decltype(lhs.version) v = lhs.version;
```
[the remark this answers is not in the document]
heifner: `auto` will work now, previous impl needed the `decltype` because it was outside an `if`.

libraries/chain/controller.cpp
heifner: We don't overwrite if it reverts back to the current active or previously proposed schedule.

greg7mdp: I don't follow. I thought we agreed that whener a new policy was proposed in a block, it will have a new generation number, even if the diff is empty. 
Which means that we can always update the `global_property_object` withing the block (`version` filed is irrelevant). 
After executing the transactions, if the `global_property_object` does contain a new policy for this block, we increment the version tracked in `parent` and store the new policy in the if_extension.

libraries/chain/block_header_state.cpp
```diff
@@ -106,18 +106,23 @@ void finish_next(const block_header_state& prev,
    if (!prev.proposer_policies.empty()) {
       auto it = prev.proposer_policies.begin();
       // +1 since this is called after the block is built, this will be the active schedule for the next block
-      if (it->first.slot <= next_header_state.header.timestamp.slot + 1) {
+      while (it != prev.proposer_policies.end() && it->first.slot <= next_header_state.header.timestamp.slot + 1) {
```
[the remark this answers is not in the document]
heifner: The active schedule will be used for the next block, not this current block. This current block used the current active schedule. This is called during assemble_block, so the active schedule will be used for the next block not this block hence the +1.

libraries/chain/controller.cpp
heifner: We have to store in `global_property_object` in case the transaction is aborted, we need the chainbase rollback to remove the change to the global_property_object. A rollback of a transaction does not rollback any changes in the `block_header_state`.

libraries/chain/controller.cpp
heifner: That was not my understanding. Why put an empty diff in a block header extension? My understanding was that we prevented no-change proposals in block header extensions, but don't bother to try and avoid set/unset in subsequent blocks. So block header will always contain some kind of diff (or nothing). But a proposal of b1->P1v1, b2->P2v2, b3-P1v3 are all placed in block headers even though the end result is no change to the schedule but does result in a change to version.

libraries/chain/include/eosio/chain/block_header_state.hpp
heifner: Agree. Reverted back.

libraries/chain/controller.cpp
```diff
@@ -491,38 +491,30 @@ struct building_block {
...
       uint32_t get_block_num() const { return block_num; }
 
-      // returns the next proposer schedule version and true if different
+      // returns the next proposer schedule version and true if producers should be proposed in block
       // if producers is not different then returns the current schedule version (or next schedule version)
       std::tuple<uint32_t, bool> get_next_proposer_schedule_version(const vector<producer_authority>& producers) const {
          assert(active_proposer_policy);
 
          auto get_next_sched = [&]() -> const producer_authority_schedule& {
             // if there are any policies already proposed but not active yet then they are what needs to be compared
```
linh2931, line 500: Update this comment that you are returning the last proposed policy.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -69,6 +69,11 @@ struct block_header_state {
    proposer_policy_ptr                 active_proposer_policy;  // producer authority schedule, supports `digest()`
 
    // block time when proposer_policy will become active
+   // current algorithm only two entries possible, for for the next,next round and one for block round after that
```
linh2931, line 72: Correct `current algorithm only two entries possible, for for`

linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -491,38 +491,27 @@ struct building_block {
...
-                     return (--itr)->second->proposer_schedule;
-                  }
-                  // no previous to what will be replaced, use active
-                  return active_proposer_policy->proposer_schedule;
-               }
-               // will not replace any proposed policies, use next to become active
-               return parent.proposer_policies.begin()->second->proposer_schedule;
+            if (!parent.proposer_policies.empty()) { // proposed in-flight
+               // return the last proposed policy to use for comparison
+               return (--parent.proposer_policies.end())->second->proposer_schedule;
```
greg7mdp, line 502: Wait, that's not right. We should compare with the one it will replace, not necessarily the last one (which may be the policy becoming active on the `next` round, not the `next_next` one).
This would be easier if we had an `array<2>`, so we would just check `proposer_policies[1]`

greg7mdp (changes requested):

libraries/chain/controller.cpp
```diff
@@ -491,38 +491,27 @@ struct building_block {
 
       uint32_t get_block_num() const { return block_num; }
 
-      // returns the next proposer schedule version and true if different
+      // returns the next proposer schedule version and true if producers should be proposed in block
       // if producers is not different then returns the current schedule version (or next schedule version)
-      std::tuple<uint32_t, bool> get_next_proposer_schedule_version(const vector<producer_authority>& producers) const {
+      std::tuple<uint32_t, bool> get_next_proposer_schedule_version(const shared_vector<shared_producer_authority>& producers) const {
```
greg7mdp, line 496: This could return `std::optional<uint32_t>`, i.e. return a new version if one is needed.

libraries/chain/controller.cpp
heifner: You want to always compare to the last one as that will be the one active before your new proposal.

greg7mdp: Got it, makes sense, thanks!

greg7mdp (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Avoid adding new proposer policy if there is no change.
Note:end

PR #75, opened 2024-04-26
IF: Avoid using fork in log messages that are not about micro-forks

### LINKED ISSUE #73: Change Net Plugin Log Messages To Be Less Alarming
Net_plugin has log message which use the word `fork`. This may cause unneeded alarm as many people associate fork with a divergence in block chain across nodes. This issue is a request to use a different noun in place of fork in net_plugin. 

Example:  `updating chain info lib ${lib}, fork ${fork}`
https://github.com/AntelopeIO/spring/blob/29396aab19a93869e3753bf849dd1bb3e4b407c8/plugins/net_plugin/net_plugin.cpp#L3258-L3284

heifner:
Use of `fork` in log messages in `net_plugin` caused confusion as they are not indicating micro-forking.

Resolves #73

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2386,9 +2386,9 @@ namespace eosio {
       if( req.req_blocks.mode == catch_up ) {
          {
             fc::lock_guard g( sync_mtx );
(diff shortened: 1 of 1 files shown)


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Avoid using fork in log messages that are not related to micro-forks.
Note:end

PR #76, opened 2024-04-26
Transition to BSL 1.1

wanderingbort:
This PR relicenses the Spring software from the temporary beta license to [BSL 1.1](https://mariadb.com/bsl11/)

The change license is MIT 2.0 

There is an Additional Use Grant intended to make sure the EOS community and developers are unencumbered by the new licensing terms. For all other uses, the BSL provides terms and custom arrangements can be discussed with licensing@eosnetwork.com


ericpassmore:
Note:start
group: IF
category: CHORE
summary: Sets licensing terms for Spring 1.0
Note:end

PR #86, opened 2024-04-29
IF: Move create and send vote off the main thread

### LINKED ISSUE #58: Allow node with finalizer keys (but no producers) to have no vote-threads
In this case, the node would not validate received votes. However, the votes created by the finalizer would still be propagated to its peers. We can go either way on whether the vote created by the finalizer is integrated in the pending_qc or not in the case when vote-threads is 0.

heifner:
- Moves create and send vote off the main thread.
- Also adds test to verify `vote-threads=0` works for a finalizer without a producer.

Resolves #58

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3228,7 +3228,13 @@ struct controller_impl {
          }
 
          if ( s == controller::block_status::incomplete || s == controller::block_status::complete || s == controller::block_status::validated ) {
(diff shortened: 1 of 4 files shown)


greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Move create and send vote off the main thread.
Note:end

Issue #20, opened 2024-04-10
Re-enable the ship_streamer_if_test and ship_streamer_test


arhag:
Those tests were temporarily disabled because they are failing frequently due to a bug in SHiP tracked by https://github.com/AntelopeIO/leap/issues/2323.

Once https://github.com/AntelopeIO/leap/issues/2323 is resolved, these tests need to be re-enabled.

PR #96, opened 2024-04-30
modify (simplify) state_history plugin's thread logic; add `get_status_request_v1`

spoonincode:
### User Facing Changes
Add `get_status_request_v1` and `get_status_result_v1` for completeness and consistency. It's a minor thing, this just provides access to the range of blocks with finality data available.

Some state_history log messages have been removed, most notably the log that occurs on every block. That seemed rather unnecessary (and was not present in 2.0/3.1) but this change can certainly be revisited.

### Internal Changes
The entirety of the connection code has been reworked to simplify the threading logic. This new design makes all decisions about what blocks to send on the main thread, and then conveys that information (while holding appropriate handles) to the ship thread which does the log reading, decompression, and websocket operations. This rework is expected to fix AntelopeIO/leap#2323 because the prior implementation would sometimes not send forks properly due to thread timing.

#### Blocking Issues
Unfortunately with the way the log implementation relies on mutexes the new approach is currently subject to deadlocks. Log implementation will need to be refactored to be free of mutexes next.


ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Significant improvement of thread logic for state history plugin.
Note:end

libraries/state_history/abi.cpp
```diff
@@ -21,6 +24,19 @@ extern const char* const state_history_plugin_abi = R"({
...
         },
+        {
+            "name": "get_status_result_v1", "fields": [
+                { "name": "head", "type": "block_position" },
+                { "name": "last_irreversible", "type": "block_position" },
+                { "name": "trace_begin_block", "type": "uint32" },
+                { "name": "trace_end_block", "type": "uint32" },
+                { "name": "chain_state_begin_block", "type": "uint32" },
+                { "name": "chain_state_end_block", "type": "uint32" },
+                { "name": "chain_id", "type": "checksum256$" },
```
spoonincode, line 35: I just noticed this is no good; can't have extension here

arhag (dismissed):
Please ship_streamer_if_test and ship_streamer_test and make sure they pass. Hopefully this resolves https://github.com/AntelopeIO/spring/issues/20.

Also, besides that change, is this PR ready for review? Or is there more work still required?

spoonincode:
really sorry for the force push at the end; pushed the log changes here that I wanted to keep as a separate PR

spoonincode:
actually, since review hasn't started, I'm just going to reopen a PR without the force pushes... sorry about that...

PR #100, opened 2024-05-01
replace `experimental-binaries` with `spring-experimental-binaries` to avoid version conflict

### LINKED ISSUE #50: `experimental-binaries` version roll back tag conflict
The `experimental-binaries` package is at the org level and as of right now both leap & spring publish packages there tagged based on the version. Once spring reaches v3.1+ there will be a tag conflict. Perhaps instead of `experimental-binaries` spring should place these in to `spring-experimental-binaries` (ultimately all users will need to be touched anyways to prepare for the difference in spring.deb vs leap.deb)

spoonincode:
Change the package name that dev.deb packages are stored in to avoid a version conflict in a few releases, fixes #50

Since we have to touch all downstream users of `leap-dev.deb` -> `spring-dev.deb` anyways, this also makes another modification to eliminate a long standing inconsistency. Previously, all dev packages corresponding to a release were stored in `experimental-binaries` as if using it as a "bucket", but then the dev packages from a CI run would be stored in separate artifacts such as `leap-dev-ubuntu20-amd64` and `leap-dev-ubuntu22-amd64`. But the thing is, an artifact is actually a bucket too. So for consistency just use a single artifact: `spring-dev-deb-amd64`.

`asset-artifact-download-action` seems prepared to handle multiple files in an artifact,
https://github.com/AntelopeIO/asset-artifact-download-action/blob/87f9193fddc71d0515ee1d3cac929e0b4d6bd503/main.mjs#L79-L88

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -115,7 +115,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: matrix.platform != 'reproducible'
         with:
-          name: spring-dev-${{matrix.platform}}-amd64
+          name: spring-dev-deb-amd64
           path: build/spring-dev*.deb
           compression-level: 0
       - name: Upload spring package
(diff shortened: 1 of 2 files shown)


spoonincode:
oh that's right, v4 doesn't allow this any more. oh well guess will stick with how it is

PR #112, opened 2024-05-06
template `fc::listener`'s Executor

spoonincode:
There is a bit of a threading trap with `fc::listener`'s current implementation. Consider usage something along the lines of,
```c++
named_thread_pool<struct ship>   thread_pool;  //running on multiple threads
std::list<connection> connections;

fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
   connections.emplace_back(new connection(std::move(socket)));
});
```
While the callback will never be called across multiple threads at once (ensuring `connections` will never be touched by multiple threads simultaneously _here at this location_), presumably somewhere else in the code needs to remove items from `connections` which, without any form of additional synchronization, would be a problem.

What one might be tempted to do is something very asio-like via a `strand` and `bind_executor`,
```c++
named_thread_pool<struct ship>   thread_pool;  //running on multiple threads
std::list<connection> connections;
boost::asio::strand<...> connections_strand = boost::asio::make_strand(thread_pool.get_exexcutor());

fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "", boost::asio::bind_executor(connections_strand, [this](Protocol::socket&& socket) {
   connections.emplace_back(new connection(std::move(socket)));
}));
```
Despite this compiling and running, this does not work as expected because `fc::listener` does not honor the associated executor of its callback.

So instead what could work is something like,
```c++
named_thread_pool<struct ship>   thread_pool;  //running on multiple threads
std::list<connection> connections;
boost::asio::strand<...> connections_strand = boost::asio::make_strand(thread_pool.get_exexcutor());

fc::create_listener<Protocol>(connections_strand, _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
   connections.emplace_back(new connection(std::move(socket)));
});
```
but this won't compile currently because `fc::listener` assumes only an `io_context` not any executor type.

Template the executor type passed to `fc::listener` so a strand can be passed.

### DIFF
--- a/libraries/libfc/include/fc/network/listener.hpp
+++ b/libraries/libfc/include/fc/network/listener.hpp
@@ -59,8 +59,8 @@ struct listener_base<boost::asio::local::stream_protocol> {
 /// detail for fc::create_listener().
 ///
 /////////////////////////////////////////////////////////////////////////////////////////////
-template <typename Protocol, typename CreateSession>
-struct listener : listener_base<Protocol>, std::enable_shared_from_this<listener<Protocol, CreateSession>> {
+template <typename Protocol, typename Executor, typename CreateSession>
+struct listener : listener_base<Protocol>, std::enable_shared_from_this<listener<Protocol, Executor, CreateSession>> {
  private:
    typename Protocol::acceptor      acceptor_;
    boost::asio::deadline_timer      accept_error_timer_;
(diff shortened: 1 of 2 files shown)


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Make fc:listener more robust by allowing binding of executor. 
Note:end

PR #113, opened 2024-05-06
use a buffer for state_history's boost iostreams de/compression pipeline improving performance

spoonincode:
Without a buffer size passed to these components in the de/compression pipeline, boost iostreams seems to send data through byte-by-byte. Adding a buffer improves performance nearly double for populating initial state. In leap 5.0: ~12m30s, after #33: ~6m20s, this PR: ~3m15s.

Certainly open to being schooled on how to better utilize boost iostreams if I missed a better way; this very pessimistic default buffering behavior was surprising.

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -78,6 +78,8 @@ static constexpr int state_history_log_header_serial_size = sizeof(state_history
                                                             sizeof(state_history_log_header::payload_size);
 static_assert(sizeof(state_history_log_header) == state_history_log_header_serial_size);
 
+static constexpr unsigned ship_log_iostreams_buffer_size = 64*1024;
+
 namespace state_history {
    struct prune_config {
       uint32_t                prune_blocks;                  //number of blocks to prune to when doing a prune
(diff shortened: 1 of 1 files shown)


heifner (approved):
The zlib de/compressor of transaction.cpp could likely benefit from a buffer as well.

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -433,9 +435,9 @@ class state_history_log {
          detail::counter cnt;
          {
             bio::filtering_ostreambuf buf;
-            buf.push(boost::ref(cnt));
-            buf.push(bio::zlib_compressor(boost::iostreams::zlib::no_compression));
-            buf.push(bio::file_descriptor_sink(stream.fileno(), bio::never_close_handle));
+            buf.push(boost::ref(cnt), ship_log_iostreams_buffer_size);
+            buf.push(bio::zlib_compressor(bio::zlib::no_compression, ship_log_iostreams_buffer_size));
+            buf.push(bio::file_descriptor_sink(stream.fileno(), bio::never_close_handle), ship_log_iostreams_buffer_size);
```
linh2931, line 440: Will the buffer be flushed in case of crashes?

spoonincode:
fwiw I looked at this a little more and the reason buffering is being disabled is because the `counter` indicates,
https://github.com/AntelopeIO/spring/blob/929533127d3032052565d6ff0187d3dfade29b31/libraries/state_history/include/eosio/state_history/log.hpp#L254
which means this gets called,
https://github.com/AntelopeIO/spring/blob/929533127d3032052565d6ff0187d3dfade29b31/libraries/state_history/include/eosio/state_history/log.hpp#L260
which effectively disables buffering (I guess just going in to the `counter`?). Otherwise filters should have a default buffer of 128 bytes, and devices 4K. It's quite possible for ship's large data structures the bigger 64KB buffer is healthier than 128/4K anyways though.

libraries/state_history/include/eosio/state_history/log.hpp
spoonincode: Since the buffer is entirely owned by the `bio::filtering_istreambuf` it seems that the buffer would need to be emptied (really all buffers emptied and flushed through the pipeline) before destruction of `bio::filtering_istreambuf` otherwise data would be lost no matter crash or not.

I think this happens in the streambuf's close_impl(),
https://github.com/boostorg/iostreams/blob/develop/include/boost/iostreams/detail/streambuf/indirect_streambuf.hpp#L390
where sync() will then call sync_impl(),
https://github.com/boostorg/iostreams/blob/develop/include/boost/iostreams/detail/streambuf/indirect_streambuf.hpp#L410
which looks like it's writing out anything left in the 'put buffer'

linh2931 (approved):

greg7mdp (approved):

spoonincode:
Since I've already tested these changes, going to go ahead and merge as-is even though it seems like the lack of buffering by default is due to improper `counter` implementation.
> The zlib de/compressor of transaction.cpp could likely benefit from a buffer as well.

It would be curious to benchmark it now that it appears that usage would have a (small?) buffer by default

ericpassmore:
Note:start
group: STABILITY
category: PERFORMACE
summary: Improve state history performance by using a buffer for the de/compression pipeline.
Note:end

PR #118, opened 2024-05-07
IF: Use finalizer diffs in instant_finality_extension

heifner:
Instead of placing the entire finalizer set in the `instant_finality_extension`, only put a diff of the active finalizer policy in the extension.


ericpassmore:
Note:start
group: IF
category: PERFORMANCE
summary: Instead of placing the entire finalizer set in the instant_finality_extension, only put a diff of the active finalizer policy in the extension.
Note:end

greg7mdp:
I think `ordered_diff` should be templatized by size_type, because for policy diffs `uint16_t` is probably plenty enough to store in the `remove_indexes` and `insert_indexes` vectors.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -0,0 +1,99 @@
...
+      std::ptrdiff_t offset = 0;
+      for (size_t index : diff.remove_indexes) {
+         source.erase(source.begin() + index + offset);
+         --offset;
+      }
+
+      // Insert into the source based on diff.insert_indexes
+      for (auto& t : diff.insert_indexes) {
+         size_t index = std::get<0>(t);
+         auto&  value = std::get<1>(t);
```
greg7mdp, line 92:
```suggestion
      for (auto& [index, value] : diff.insert_indexes) {
```

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -0,0 +1,99 @@
...
+            ++t;
+         }
+      }
+
+      return result;
+   }
+
+   template <typename X>
+   requires std::same_as<std::decay_t<X>, diff_result>
+   static void apply_diff(Container<T>& source, X&& diff) {
```
linh2931, line 81: The method will modify `source`. Should its name reflect that?

libraries/chain/include/eosio/chain/finality/finalizer_policy.hpp
```diff
@@ -2,13 +2,37 @@
...
 #include <eosio/chain/types.hpp>
 #include <eosio/chain/finality/finalizer_authority.hpp>
+#include <fc/container/ordered_diff.hpp>
 
 namespace eosio::chain {
 
+   using finalizers_differ = fc::ordered_diff<finalizer_authority>;
+   using finalizers_diff_t = fc::ordered_diff<finalizer_authority>::diff_result;
+
+   struct finalizer_policy_diff {
```
linh2931, line 12: Not sure if `finalizer_policy_diff` is needed, as `generation` and `threshold` are not really differed. Only finalizers are differed.

heifner:
> I think `ordered_diff` should be templatized by size_type, because for policy diffs `uint16_t` is probably plenty enough to store in the `remove_indexes` and `insert_indexes` vectors.

Good idea. proposers can be `uint8_t` even.

linh2931 (commented):
I started to wonder if it is worthy the trouble (time to diff and apply_diff, and more code) to pack diff only in the extension. Another side effect is one cannot just look at the extension to see the finalizers; previous extensions need to be checked.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -129,6 +129,24 @@ struct block_header_state {
...
+      if (finalizer_policies.empty()) {
+         return active_finalizer_policy->create_diff(new_policy);
+      }
+      finalizer_policy fin_policy = *active_finalizer_policy;
+      for (const auto& e : finalizer_policies) {
+         if (e.second.state == finalizer_policy_tracker::state_t::pending) {
+            fin_policy.apply_diff(e.second.policy_diff);
+         }
+      }
+      for (const auto& e : finalizer_policies) {
```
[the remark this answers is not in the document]
heifner: I don't think so, at least my understanding of how they are stored. @greg7mdp ?

libraries/libfc/include/fc/container/ordered_diff.hpp
heifner: Renamed to `container` and added comments

libraries/chain/include/eosio/chain/finality/finalizer_policy.hpp
heifner: Provides a struct for easy `fc::pack` into `instant_finality_extension`.

libraries/libfc/include/fc/container/ordered_diff.hpp
greg7mdp: I think a somewhat clearer API would be:

`static Container<T> apply_diff(Container<T>&& container, X&& diff)`

making it clear that it consumes a rvalue and produces a new value.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -0,0 +1,111 @@
...
+ * @param Container container type for ordered diff and for diff_result
+ */
+template <typename T, typename SizeType = size_t, template<typename Y, typename...> typename Container = std::vector>
+requires std::equality_comparable<T> && std::random_access_iterator<typename Container<T>::iterator>
+class ordered_diff {
+public:
+   struct diff_result {
+      Container<SizeType>                remove_indexes;
+      Container<std::pair<SizeType, T>>  insert_indexes;
+   };
```
greg7mdp, line 30: Why not have `apply_diff` here as a member function

`Container<T> apply(Container<T>&& source) &&;`

libraries/chain/include/eosio/chain/finality/finalizer_policy.hpp
```diff
@@ -2,13 +2,38 @@
...
+                 .threshold = target.threshold,
+                 .finalizers_diff = finalizers_differ::diff(finalizers, target.finalizers)};
+      }
+
+      template <typename X>
+      requires std::same_as<std::decay_t<X>, finalizer_policy_diff>
+      void apply_diff(X&& diff) {
+         generation = diff.generation;
+         threshold = diff.threshold;
+         finalizers_differ::apply_diff(finalizers, std::move(diff.finalizers_diff));
```
greg7mdp, line 35: That's wrong, it should be `std::forward<X>(diff).finalizers_diff`.

`std::move` will move unconditionally from the in-flight policy diffs stored in  `block_header_state:: finalizer_policies` in `calculate_finalizer_policy_diff`

greg7mdp (changes requested):

libraries/libfc/include/fc/container/ordered_diff.hpp
heifner: It might be a bit too clever, but the current implementation works for both a `const diff_result` and a rvalue `diff_result`. Not sure if it is possible to share implementation between a const member function and a `&&` member function.

greg7mdp: Ah, yes, makes sense. We'll be able to support this in a member finction with c++23 [deducing_this](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0847r6.html#deduplicating-code).

libraries/chain/block_header_state.cpp
```diff
@@ -67,6 +67,48 @@ const vector<digest_type>& block_header_state::get_new_protocol_feature_activati
...
+   }
+   for (const auto& e : finalizer_policies) {
+      if (e.second.state == finalizer_policy_tracker::state_t::proposed) {
+         result = *e.second.policy;
+         result.apply_diff(diff);
+         return result;
+      }
+   }
+   assert(false);
+}
```
greg7mdp, line 110: I believe these two functions don't do the right thing. They should return the **last** policy that will become active, not the the **first** one.

Also, both of them could use a utility function in `block_header_state`:

```c++
#include <boost/range/adaptor/reversed.hpp>

   const finalizer_policy_ptr& get_finalizer_policy_tail() const {
      if (!finalizer_policies.empty()) {
         auto reversed_finalizer_policies = boost::adaptors::reverse(finalizer_policies);
         for (const auto& e : reversed_finalizer_policies) {
            if (e.second.state == finalizer_policy_tracker::state_t::proposed)
               return e.second.policy;
         }
         return reversed_finalizer_policies.begin()->second.policy;
      }
      return active_finalizer_policy;
   }
```

greg7mdp (changes requested):

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -134,7 +134,7 @@ static_assert(maximum_tracked_dpos_confirmations >= ((max_producers * 2 / 3) + 1
 /**
  * Maximum number of finalizers in the finalizer set
  */
-const static size_t max_finalizers = 64*1024;
+const static size_t max_finalizers = 65535; // largest allowed finalizer policy diff
```
greg7mdp, line 137: I don't think this change is necessary.

heifner: I think we should change this to 1024 or maybe 512 or even 128.

libraries/chain/block_header_state.cpp
```diff
@@ -78,7 +120,8 @@ const vector<digest_type>& block_header_state::get_new_protocol_feature_activati
 void finish_next(const block_header_state& prev,
                  block_header_state& next_header_state,
                  vector<digest_type> new_protocol_feature_activations,
-                 instant_finality_extension if_ext) {
+                 instant_finality_extension if_ext,
+                 std::optional<finalizer_policy> new_finalizer_policy) {
```
greg7mdp, line 124: I would prefer not pass this additional parameter (there will be two when we do `proposer_policies` as well).
I understand that it would require an extra `apply_diff` for the producer, but I don't think it would make any significant difference, and the code would be cleaner and more streamlined. The `calculate_finalizer_policy` could be moved to `finish_next`.

libraries/chain/block_header_state.cpp
heifner: I completely agree. I don't understand how our existing tests didn't catch this. Need to add a new test case.

greg7mdp: > I don't understand how our existing tests didn't catch this. Need to add a new test case.

Because in the current finalizer_update_tests, we run as a proposer, and because we pass the original uncompressed policy to `finish_next()` from `next(block_header_state_input&)`. We don't `push_block()`.

So the compression/decompression is never activated in the test. If you extract the diff from the extension in `finish_next()` instead of taking the uncompressed policy, I think the tests would fail.

heifner: The test is running a validator controller, so surprised that didn't catch it. I was looking into validating the diff, I think what you suggest is at least one way to do that.

greg7mdp: Yes, you are right it does a `push_block` to a `unique_ptr<controller> validating_node;`, but we don't check anything on that `validating_node` when it comes to finalizer_policy status.

heifner: Calculating the policy instead of passing and the tests still pass. I can tell in the test that the if_ext has the wrong diff. It must be getting the right answer because the inverse of the diff is used with the same wrong value on both nodes. I've added a test that verifies the if_ext is correct.

greg7mdp: > Calculating ...  It must be getting the right answer because the inverse of the diff is used with the same wrong value on both nodes.

I don't think so. The issue is that we calculate a diff with the wrong source policy, but then we also use the same wrong source policy to apply the diff and restore the original `finalizer_policy`.

heifner: I think we just said the same thing.

We do "compare" between both nodes because block ids have to match which includes the block header extensions.

greg7mdp: Oh, got it. Glad we agree. I understand your "compare" point now.

unittests/finalizer_update_tests.cpp
```diff
@@ -71,6 +71,18 @@ BOOST_AUTO_TEST_CASE(savanna_set_finalizer_multiple_test) { try {
    size_t num_keys    = 50u;
    size_t finset_size = 21u;
 
+   auto verify_block_finality_generation = [](const signed_block_ptr& block, uint32_t gen, const bls_public_key& key) {
```
greg7mdp, line 74: Good check! 

No big deal, but I think this is a more accurate name.

```suggestion
   auto verify_block_finality_policy_diff = [](const signed_block_ptr& block, uint32_t gen, const bls_public_key& key) {
```

greg7mdp (changes requested):

libraries/chain/include/eosio/chain/finality/finalizer_policy.hpp
```diff
@@ -2,13 +2,38 @@
 
 #include <eosio/chain/types.hpp>
 #include <eosio/chain/finality/finalizer_authority.hpp>
+#include <fc/container/ordered_diff.hpp>
 
 namespace eosio::chain {
 
+   static_assert(std::numeric_limits<uint16_t>::max() <= config::max_finalizers);
```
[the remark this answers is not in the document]
heifner: Yes, already had that in my changes for the next PR. Note you can't use `max() + 1` because that overflows. So `max_finalizers - 1`.

greg7mdp (approved):

PR #120, opened 2024-05-08
libtester: return transaction trace for `setfinalizer` host function. 

greg7mdp:
This is needed for IBC proof tests.

### DIFF
--- a/libraries/testing/include/eosio/testing/tester.hpp
+++ b/libraries/testing/include/eosio/testing/tester.hpp
@@ -147,7 +147,7 @@ namespace eosio::testing {
    struct produce_block_result_t {
       signed_block_ptr                   block;
       transaction_trace_ptr              onblock_trace;
-      std::vector<transaction_trace_ptr> unapplied_transaction_traces; // unapplied from previous `push_block` not on head branch
+      std::vector<transaction_trace_ptr> unapplied_transaction_traces; // only traces of any unapplied transactions
    };
 
    /**
--- a/libraries/testing/include/eosio/testing/tester.hpp
+++ b/libraries/testing/include/eosio/testing/tester.hpp
@@ -147,7 +147,7 @@ namespace eosio::testing {
    struct produce_block_result_t {
       signed_block_ptr                   block;
       transaction_trace_ptr              onblock_trace;
-      std::vector<transaction_trace_ptr> traces;         // transaction traces
+      std::vector<transaction_trace_ptr> unapplied_transaction_traces; // unapplied from previous `push_block` not on head branch
    };
 
    /**
@@ -271,19 +271,23 @@ namespace eosio::testing {
          transaction_trace_ptr       set_producer_schedule(const vector<producer_authority>& schedule);
          transaction_trace_ptr       set_producers_legacy(const vector<account_name>& producer_names);
 
+         struct set_finalizers_output_t {
+            transaction_trace_ptr        setfinalizer_trace;
+            std::vector<bls_private_key> privkeys;  // private keys of **local** finalizers
+            std::vector<bls_public_key>  pubkeys;   // public keys of all finalizers in the policy
+         };
+
          // libtester uses 1 as weight of each of the finalizer, sets (2/3 finalizers + 1)
          // as threshold, and makes all finalizers vote QC
-         std::pair<transaction_trace_ptr, std::vector<fc::crypto::blslib::bls_private_key>>
-         set_finalizers(std::span<const account_name> finalizer_names);
+         set_finalizers_output_t set_finalizers(std::span<const account_name> finalizer_names);
 
-         std::pair<transaction_trace_ptr, std::vector<fc::crypto::blslib::bls_private_key>>
-         set_finalizers(const std::vector<account_name>& names) {
+         set_finalizers_output_t set_finalizers(const std::vector<account_name>& names) {
             return set_finalizers(std::span{names.begin(), names.end()});
          }
 
          void set_node_finalizers(std::span<const account_name> finalizer_names);
 
-         std::vector<fc::crypto::blslib::bls_public_key> set_active_finalizers(std::span<const account_name> finalizer_names);
+         set_finalizers_output_t set_active_finalizers(std::span<const account_name> finalizer_names);
 
          // Finalizer policy input to set up a test: weights, threshold and local finalizers
          // which participate voting.
@@ -297,7 +301,7 @@ namespace eosio::testing {
             uint64_t                    threshold {0};
             std::vector<account_name>   local_finalizers;
          };
-         std::pair<transaction_trace_ptr, std::vector<fc::crypto::blslib::bls_private_key>> set_finalizers(const finalizer_policy_input& input);
+         set_finalizers_output_t set_finalizers(const finalizer_policy_input& input);
 
          std::optional<finalizer_policy> active_finalizer_policy(const block_id_type& id) const {
             return control->active_finalizer_policy(id);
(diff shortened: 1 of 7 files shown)


libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -297,7 +283,32 @@ namespace eosio::testing {
             uint64_t                    threshold {0};
             std::vector<account_name>   local_finalizers;
          };
-         std::pair<transaction_trace_ptr, std::vector<fc::crypto::blslib::bls_private_key>> set_finalizers(const finalizer_policy_input& input);
+
+         struct set_finalizers_output_t {
+            transaction_trace_ptr        setfinalizer_trace;
```
heifner, line 288: Seems should just be `trace` since it is in struct `set_finalizers_output_t`

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -147,7 +147,7 @@ namespace eosio::testing {
    struct produce_block_result_t {
       signed_block_ptr                   block;
       transaction_trace_ptr              onblock_trace;
-      std::vector<transaction_trace_ptr> traces;         // transaction traces
+      std::vector<transaction_trace_ptr> unapplied_transaction_traces; // unapplied from previous `push_block` not on head branch
```
heifner, line 150: Undo this change. These are the transaction traces of the block. They are not:  `unapplied from previous push_block not no head branch`. Although, it is not really clear exactly what that is trying to convey. They are not unapplied transactions.

heifner (changes requested):

libraries/testing/include/eosio/testing/tester.hpp
greg7mdp: Actually, these are the traces for the `unapplied_transactions` indeed. see [here](https://github.com/AntelopeIO/spring/blob/92a21962db0144c2174e1f66d308f72123382f08/libraries/testing/tester.cpp#L409-L417)

greg7mdp: traces from normal transaction within the block are returned by [push_transaction](https://github.com/AntelopeIO/spring/blob/92a21962db0144c2174e1f66d308f72123382f08/libraries/testing/include/eosio/testing/tester.hpp#L221)

heifner: Those are traces returned from `push_transaction` https://github.com/AntelopeIO/spring/blob/92a21962db0144c2174e1f66d308f72123382f08/libraries/testing/tester.cpp#L410

libraries/testing/include/eosio/testing/tester.hpp
greg7mdp: In general, I would agree, but because `set_finalizers()` does other things, I actually think it is clearer to specify what the trace is of. And Gnome already uses this branch I believe.

libraries/testing/include/eosio/testing/tester.hpp
heifner: Oh, you are not capturing the traces of other `push_transaction` calls. This is not at all expected behavior. I think you have to go with the signal to capture all transaction traces.

heifner: The unapplied transaction queue is going to be empty under most circumstances in the tests.

greg7mdp: yes, I also misunderstood what was happening here. I think with the renaming it is clear what it does, and replaces the other API `produce_block( std::vector<transaction_trace_ptr>& traces );` which the previous PR removed (it was used only in `forked_tests.cpp`). I agree it will be empty in most cases.

heifner: I see now that it is only used in the forked_tests. Seems like the forked_tests could have done that via `applied_transaction` signal and not pollute the tester with this odd/confusing functionality.

libraries/testing/include/eosio/testing/tester.hpp
greg7mdp: It is the trace for the [setfinalizer](https://github.com/AntelopeIO/spring/blob/92a21962db0144c2174e1f66d308f72123382f08/libraries/testing/tester.cpp#L1254-L1256) action.

libraries/testing/include/eosio/testing/tester.hpp
heifner: If you keep this then change the comment to something like: `// only traces of any unapplied transactions`
unapplied transactions can come from a fork (which is what the forked test is using this for) or aborted blocks.
My thought is this guy should have never added this confusing interface: https://github.com/EOSIO/eos/pull/8786

libraries/testing/tester.cpp
```diff
@@ -413,14 +413,14 @@ namespace eosio::testing {
...
-            res.traces.emplace_back( std::move(trace) );
+            res.unapplied_transaction_traces.emplace_back( std::move(trace) );
          }
 
          vector<transaction_id_type> scheduled_trxs;
          while ((scheduled_trxs = get_scheduled_transactions()).size() > 0 ) {
             for( const auto& trx : scheduled_trxs ) {
                auto trace = control->push_scheduled_transaction( trx, fc::time_point::maximum(), fc::microseconds::maximum(), DEFAULT_BILLED_CPU_TIME_US, true );
-               res.traces.emplace_back( trace );
+               res.unapplied_transaction_traces.emplace_back( trace );
```
heifner, line 423: Remove this line. These are NOT unapplied transactions.

libraries/testing/include/eosio/testing/tester.hpp
greg7mdp: I agree that the interface is really confusing (confused both of us), but I'll cut this guy some slack :-)
I'm happy to update `forked_tests` and remove this confusing functionality if you'd like.

libraries/testing/tester.cpp
greg7mdp: For sure.

libraries/testing/include/eosio/testing/tester.hpp
heifner: Seems best to remove it if not too much effort.

greg7mdp: Or I could add in the return value `scheduled_transactions_traces` to return traces of the scheduled transactions? Maybe these could be handy?

heifner: Without a use-case (especially now that scheduled transactions are disabled) see no reason to add that.

greg7mdp: OK, I'll give the cleanup a try.

greg7mdp: These `unapplied_transactions` in libtester are also used when we produce an empty block [here](https://github.com/AntelopeIO/spring/blob/2f1b39f6b5e435e3c0f4b9c995fd88e1fc5a8b09/libraries/testing/include/eosio/testing/tester.hpp#L607-L610).

Which is weird because then we call `_produce_block` with `skip_pending_trxs = true`, which means we don't call `control->push_transaction()` on the `unapplied_transactions`.

I'm not quite understanding the purpose of this?

heifner: Processing the unapplied transactions (default of `skip_pending_trxs = false`) mimics what the produce plugin does. Aborted blocks (an empty block aborts any transactions so that it is empty), and forked out blocks put there transactions in the unapplied transaction queue. The producer plugin always processed unapplied transaction as the first step of producing a block. The idea is that any forked out trxs or aborted block transactions are imediately attempted again in the next block before the end of `start_block` and other incoming transactions can be processed.

greg7mdp: OK, thanks, I understand a little bit better I think.

I think it is better to leave this PR  as-is. Also `forked_tests.cpp` verifies that transactions on fork are reported by `push_block` in order (and at the time the transactions have not been executed), so we do need this `unapplied_transaction_queue` for that purpose. I don't know that I can meaningfully simplify what we have now.

heifner: Sounds good. Please do update the comment.

greg7mdp: Thanks. Comment is updated.

heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Add transaction traces to support IBC proof tests.
Note:end

Issue #123, opened 2024-05-08
Add proposed finalizer policy to SHiP finality_data stream


arhag:
The `finality_data` should have a new optional field to contain the full policy (including generation number) of the finalizer policy proposed in the associated block. If no new finalizer policy is proposed in a block, then this field should be `std::nullopt`. The SHiP logs will have to contain this additional field.

PR #124, opened 2024-05-09
Add proposed finalizer policy to SHiP finality_data stream

linh2931:
Add proposed finalizer policy to SHiP finality_data stream

Resolves https://github.com/AntelopeIO/spring/issues/123

### DIFF
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -339,7 +339,7 @@ finality_data_t block_state::get_finality_data() {
    }
 
    // Check if there is any proposed finalizer policy in the block
-   std::optional<finalizer_policy> proposed_finalizer_policy = std::nullopt;
+   std::optional<finalizer_policy> proposed_finalizer_policy;
    auto range = finalizer_policies.equal_range(block_num());
    for (auto itr = range.first; itr != range.second; ++itr) {
       if (itr->second.state == finalizer_policy_tracker::state_t::proposed) {
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -353,7 +353,7 @@ finality_data_t block_state::get_finality_data() {
       .active_finalizer_policy_generation = active_finalizer_policy->generation,
       .action_mroot                       = action_mroot,
       .base_digest                        = *base_digest,
-      .proposed_finalizer_policy          = proposed_finalizer_policy
+      .proposed_finalizer_policy          = std::move(proposed_finalizer_policy)
    };
 }
 
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -340,11 +340,16 @@ finality_data_t block_state::get_finality_data() {
 
    // Check if there is any proposed finalizer policy in the block
    std::optional<finalizer_policy> proposed_finalizer_policy;
-   auto range = finalizer_policies.equal_range(block_num());
-   for (auto itr = range.first; itr != range.second; ++itr) {
-      if (itr->second.state == finalizer_policy_tracker::state_t::proposed) {
-         proposed_finalizer_policy = *itr->second.policy;
-         break;
+   if (is_savanna_genesis_block()) {
+      // For Genesis Block, use the acttive finalizer policy which was proposed in the block.
+      proposed_finalizer_policy = *active_finalizer_policy;
+   } else {
+      auto range = finalizer_policies.equal_range(block_num());
+      for (auto itr = range.first; itr != range.second; ++itr) {
+         if (itr->second.state == finalizer_policy_tracker::state_t::proposed) {
+            proposed_finalizer_policy = *itr->second.policy;
+            break;
+         }
       }
    }
 
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -337,11 +337,23 @@ finality_data_t block_state::get_finality_data() {
    if (!base_digest) {
       base_digest = compute_base_digest(); // cache it
    }
+
+   // Check if there is any proposed finalizer policy in the block
+   std::optional<finalizer_policy> proposed_finalizer_policy = std::nullopt;
+   auto range = finalizer_policies.equal_range(block_num());
+   for (auto itr = range.first; itr != range.second; ++itr) {
+      if (itr->second.state == finalizer_policy_tracker::state_t::proposed) {
+         proposed_finalizer_policy = *itr->second.policy;
+         break;
+      }
+   }
+
    return {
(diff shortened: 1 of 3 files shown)


libraries/chain/block_state.cpp
```diff
@@ -337,11 +337,23 @@ finality_data_t block_state::get_finality_data() {
...
+   }
+
    return {
       // other fields take the default values set by finality_data_t definition
       .active_finalizer_policy_generation = active_finalizer_policy->generation,
-      .action_mroot = action_mroot,
-      .base_digest  = *base_digest
+      .action_mroot                       = action_mroot,
+      .base_digest                        = *base_digest,
+      .proposed_finalizer_policy          = proposed_finalizer_policy
```
greg7mdp, line 356:
```suggestion
      .proposed_finalizer_policy          = std::move(proposed_finalizer_policy)
```

greg7mdp (changes requested):

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -64,6 +64,7 @@ struct finality_data_t {
    uint32_t     active_finalizer_policy_generation{0};
    digest_type  action_mroot{};
    digest_type  base_digest{};
+   std::optional<finalizer_policy> proposed_finalizer_policy{std::nullopt}; // finalizer policy, if proposed in the block
```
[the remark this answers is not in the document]
linh2931: The explicit initialization is there to indicate the intended default value, although it is redundant but easier for the readers.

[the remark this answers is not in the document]
greg7mdp: Isn't it a given that the default value of an optional is to be empty? And that if you use an optional in a struct, it means that it is empty until you assign something to it?

greg7mdp (dismissed):

libraries/chain/block_state.cpp
```diff
@@ -337,11 +337,23 @@ finality_data_t block_state::get_finality_data() {
...
+
    return {
       // other fields take the default values set by finality_data_t definition
       .active_finalizer_policy_generation = active_finalizer_policy->generation,
-      .action_mroot = action_mroot,
-      .base_digest  = *base_digest
+      .action_mroot                       = action_mroot,
+      .base_digest                        = *base_digest,
+      .proposed_finalizer_policy          = std::move(proposed_finalizer_policy)
    };
```
heifner, line 357: I don't think this does the correct thing for IF Genesis Block. The IF Genesis Block has a finalizer set proposed in it (that is what makes it the IF Genesis Block), but that finalizer policy is not put in the tracker I don't think. Is that an issue or is IF Genesis Block handled specially for this finality data?

heifner (changes requested):

libraries/chain/block_state.cpp
linh2931: Should `create_if_genesis_block` set the first proposed policy in `finalizer_policies`?

libraries/chain/include/eosio/chain/block_state.hpp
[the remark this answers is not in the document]
linh2931: I'll make the change as both of you think the other way is better. Thanks.

libraries/chain/block_state.cpp
linh2931: It's decided at the standup meeting we will treat this as a special case here and return the active policy for Savanna Genesis block. Thanks @heifner

libraries/chain/block_state.cpp
```diff
@@ -337,11 +337,28 @@ finality_data_t block_state::get_finality_data() {
    if (!base_digest) {
       base_digest = compute_base_digest(); // cache it
    }
+
+   // Check if there is any proposed finalizer policy in the block
+   std::optional<finalizer_policy> proposed_finalizer_policy;
+   if (is_savanna_genesis_block()) {
+      // For Genesis Block, use the acttive finalizer policy which was proposed in the block.
```
greg7mdp, line 344:
```suggestion
      // For Genesis Block, use the active finalizer policy which was proposed in the block.
```

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -119,6 +119,9 @@ struct block_header_state {
    digest_type compute_base_digest() const;
    digest_type compute_finality_digest() const;
 
+   // Returns true if the block is a Savanna Genesis Block
+   bool is_savanna_genesis_block() const { return core.is_genesis_block_num(block_num()); }
```
greg7mdp, line 123: I am not understanding how this can possibly work.  As far as I understand, we don't create a `block_header_state` or a `core` for the genesis block.

libraries/chain/block_state.cpp
```diff
@@ -337,11 +337,28 @@ finality_data_t block_state::get_finality_data() {
    if (!base_digest) {
       base_digest = compute_base_digest(); // cache it
    }
+
+   // Check if there is any proposed finalizer policy in the block
+   std::optional<finalizer_policy> proposed_finalizer_policy;
+   if (is_savanna_genesis_block()) {
```
greg7mdp, line 343: If we are in a `block_state`, can this be the genesis block?

libraries/chain/include/eosio/chain/block_header_state.hpp
linh2931: We do create `block_header_state` and a `core` for the genesis block https://github.com/AntelopeIO/spring/blob/c7b60aad19f5b7de646c2237d3c9d79e51b41277/libraries/chain/block_state.cpp#L71

libraries/chain/block_state.cpp
linh2931: A transition (and genesis) block can be reclassified as Savanna block. `get_transition_savanna_block()` does that.

libraries/chain/include/eosio/chain/block_header_state.hpp
linh2931: I will add a comment to explain is_savanna_genesis_block() is only applicable for reclassified transition blocks.

greg7mdp: Thanks Lin, my bad, I mis-remembered!

heifner (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Extend finality data to include full policy (including generation number) of the finalizer policy proposed in the associated block. This is used in SHiP data stream.
Note:end

greg7mdp (approved):

PR #126, opened 2024-05-09
Increase max proposers to match max finalizers

heifner:
Increase the maximum number of proposer (producers) in Savanna from 125 to 64K. This matches the maximum number of allowed finalizers.

### DIFF
--- a/libraries/chain/include/eosio/chain/config.hpp
+++ b/libraries/chain/include/eosio/chain/config.hpp
@@ -127,7 +127,7 @@ const static uint32_t   default_abi_serializer_max_time_us = 15*1000; ///< defau
  */
 const static int producer_repetitions = 12;
 const static int max_producers = 125; // pre-savanna producer (proposer) limit
-const static int max_proposers = 64*1024; // savanna max proposer (producer) limit
+const static int max_proposers = 64*1024; // savanna proposer (producer) limit
 
 const static size_t maximum_tracked_dpos_confirmations = 1024;     ///<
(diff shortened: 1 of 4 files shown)


libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -126,7 +126,8 @@ const static uint32_t   default_abi_serializer_max_time_us = 15*1000; ///< defau
  *  The number of sequential blocks produced by a single producer
  */
 const static int producer_repetitions = 12;
-const static int max_producers = 125;
+const static int max_producers = 125; // pre-savanna producer (proposer) limit
+const static int max_proposers = 64*1024; // savanna proposer (producer) limit
```
greg7mdp, line 130: If we expect them to always be the same, should we use one variable (for ex `max_proposers_finalizers`?)

linh2931 (approved):

libraries/chain/include/eosio/chain/config.hpp
heifner: I don't think we do expect them to be the same. We have reason to believe we may want to reduce the max finalizers, where the max proposers seems like could remain at 64K.

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: PERFORMACE
summary: Allow a larger number of proposer (producers) in Savanna from 125 to 64K. This matches the maximum number of allowed finalizers.
Note:end

PR #131, opened 2024-05-10
Correctly support when finality advances multiple blocks at a time.

### LINKED ISSUE #130: bug: When finality advances multiple blocks at a time, duplicate policies can become pending for the same block.
In which case the wrong policy may become active when that block becomes final.

greg7mdp:
In that case, duplicate policies can become pending for the same block.

Resolves #130.

Also added a test which demonstrates the issue when the fix is not present.


heifner (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Resolves duplicate polices from advancing. Finality can advance multiple blocks, previously this may cause duplicate policies to advance.  
Note:end

linh2931 (approved):

PR #132, opened 2024-05-11
Replace active_finalizer_policy_digest with last_pending_finalizer_policy_digest

### LINKED ISSUE #128: Replace active_finalizer_policy_digest with last_pending_finalizer_policy_digest in finality digest computation
Change the `active_finalizer_policy_digest` used in the calculation of the finality digest to instead be `last_pending_finalizer_policy_digest`. This is the digest of the finalizer policy (which includes the generation number in it) with the greatest generation number in the history of the blockchain so far that is not in the proposed state (so it is either in the pending or active state). This is something that should always be determined from the `block_header_state`. For efficiency reasons, it will probably be worth caching the `last_pending_finalizer_policy_digest` in the `block_header_state` and ensuring it is updated as necessary when producing the next `block_header_state`.

This change supports IBC by allowing a proof to be provided to the IBC contract for the contents of a new proposed finalizer policy associated with a particular generation number so that it can then later be used for validating QCs on a block that uses that finalizer policy as the active finalizer policy.

The active finalizer policy generation number used separately in the calculation of the finality digest should still remain as is. That is always the generation number of the finality policy that is active for the block, i.e. the one to be used for validating the votes in the QC for that block.

So, in summary, the finality digest would be defined as follows:

The finality digest is the SHA256 digest of a serialization of a structure that is versioned according to the light header protocol version (this is a new versioning scheme consisting of major and minor version numbers that is separate from protocol features that only gets updated if a protocol feature causes a breaking change to light block header validation). For now the only light header protocol version supported is 1.0. The serialization should include that major version number 1 (as a 32-bit number), the minor version number 0 (as a 32-bit number), and the rest of the serialization of the versioned structure. 

That version structure for the version 1.0 protocol consists of the active finalizer policy generation number followed by two digests. The first digest is the Finality Tree Root of the block which is returned by the new `finality_mroot()` function added to `block_header_state`. The second is a digest computed as the SHA256 hash of the serialization of two other digests: the `last_pending_finalizer_policy_digest` and the `base_digest`.

The `last_pending_finalizer_policy_digest` is the SHA256 hash of some serialization of the last pending finalizer policy in the history of the blockchain so far. Note that if there are no pending finalizer policies in the queue, then the last pending finalizer policy is the active finalizer policy.

The `base_digest` is the SHA256 hash of some serialization of the remaining (non-cached) fields of the `block_header_state`. So this includes: `header`, `core`, `finalizer_policies`, `active_proposer_policy`, `proposer_policies`, and `activated_protocol_features`.

heifner:
Replace `active_finalizer_policy_digest` with `last_pending_finalizer_policy_digest` in finality digest computation

Resolves #128

### DIFF
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -43,13 +43,11 @@ digest_type block_header_state::compute_base_digest() const {
 }
 
 digest_type block_header_state::compute_finality_digest() const {
-   assert(active_finalizer_policy);
-   auto active_finalizer_policy_digest = fc::sha256::hash(*active_finalizer_policy);
    auto base_digest = compute_base_digest();
-
-   std::pair<const digest_type&, const digest_type&> active_and_base{ active_finalizer_policy_digest, base_digest };
+   std::pair<const digest_type&, const digest_type&> active_and_base{ last_pending_finalizer_policy_digest, base_digest };
    auto afp_base_digest = fc::sha256::hash(active_and_base);
 
+   assert(active_finalizer_policy);
    finality_digest_data_v1 finality_digest_data {
       .active_finalizer_policy_generation      = active_finalizer_policy->generation,
       .finality_tree_digest                    = finality_mroot(),
@@ -67,7 +65,7 @@ const vector<digest_type>& block_header_state::get_new_protocol_feature_activati
    return detail::get_new_protocol_feature_activations(header_exts);
 }
 
-// The last proposed finalizer policy if none proposed or pending is the active finalizer policy
+// The last proposed finalizer policy if none proposed or pending then the active finalizer policy
 const finalizer_policy& block_header_state::get_last_proposed_finalizer_policy() const {
    if (!finalizer_policies.empty()) {
       for (auto ritr = finalizer_policies.rbegin(); ritr != finalizer_policies.rend(); ++ritr) {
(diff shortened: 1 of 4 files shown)


greg7mdp (approved):
~~For the name, I would prefer `newest_pending_...` to `last_pending_...` as I find it less ambiguous.~~ not sure anymore  `newest` is much better.

systemzax (approved):
Looks good to me. :thumbsup:

I've merged this PR into the branch I'm working on, tested it, and as far as I can tell, it produces the expected results.

libraries/chain/include/eosio/chain/snapshot_detail.hpp
```diff
@@ -135,6 +136,7 @@ namespace eosio::chain::snapshot_detail {
          , proposer_policies(bs.proposer_policies)
          , finalizer_policies(bs.finalizer_policies)
          , finalizer_policy_generation(bs.finalizer_policy_generation)
+         , last_pending_finalizer_policy_digest(bs.last_pending_finalizer_policy_digest)
```
greg7mdp, line 139: Should we add a comment on top of this function:

When adding a member initialization here, also update `block_state::block_state(snapshot_detail::snapshot_block_state_v7&& sbs)`.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Update to use the last pending finalizer policy in the history of the blockchain, if there are no pending finalizer policies in the queue, then the last pending finalizer policy is the active finalizer policy.
Note:end

greg7mdp (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -79,6 +77,30 @@ const finalizer_policy& block_header_state::get_last_proposed_finalizer_policy()
    return *active_finalizer_policy;
 }
 
+// The last pending finalizer policy if none pending then the active finalizer policy
+// Used to populate last_pending_finalizer_policy_digest which is expected to be the highest generation pending
+const finalizer_policy& block_header_state::get_last_pending_finalizer_policy() const {
+   if (!finalizer_policies.empty()) {
+      [[maybe_unused]] auto highest_pending_generation = [this]() {
```
linh2931, line 84: You can add a comment indicating thus lambda is used only when DEBUG is turned on.

libraries/chain/include/eosio/chain/snapshot_detail.hpp
```diff
@@ -119,12 +119,14 @@ namespace eosio::chain::snapshot_detail {
...
       flat_multimap<block_num_type, finalizer_policy_tracker> finalizer_policies;
       uint32_t                                            finalizer_policy_generation;
+      digest_type                                         last_pending_finalizer_policy_digest;
 
       // from block_state
       std::optional<valid_t>                              valid;
 
       snapshot_block_state_v7() = default;
 
+      // When adding a member initialization here also update block_state(snapshot_block_state_v7) constructor
```
linh2931, line 129: Will some test or static_assert fail if `snapshot_block_state_v7` is forgotten to update?

linh2931 (approved):

libraries/chain/include/eosio/chain/snapshot_detail.hpp
heifner: The `switchToNoBlockLogWithIrrModeSnapshot` test of `nodeos_irreversible_mode_test.py` failed.

linh2931 (approved):

Issue #11, opened 2024-02-16
Create IF versions of applicable unittests


heifner:
- protocol_feature_tests.cpp
- restart_chain_tests.cpp
- block_tests.cpp
- chain_tests.cpp
- database_tests.cpp
- forked_tests.cpp

Consider if other tests that IF should make no difference if they should only be run in IF mode.
For example: dry_run_trx_tests.cpp, eosio.system_tests.cpp, etc.

arhag:
For the DeepMind test, we can keep it as one test with one log. And we should activate the IF hardfork at the beginning. But we should do what exists in the current test before transitioning to Savanna. Then we extend the test to transition to Savanna, update the proposer schedule, and produce some blocks.

linh2931:
### Summary
#### Tests Updated
`api_tests.cpp, auth_tests.cpp, block_tests.cpp, chain_tests.cpp, database_tests.cpp, dry_run_trx_tests.cpp, eosio.token_tests.cpp, misc_tests.cpp, payloadless_tests.cpp, protocol_feature_tests.cpp, read_only_trx_tests.cpp, restart_chain_tests.cpp, snapshot_tests.cpp, special_accounts_tests.cpp, wasm_tests.cpp
`
#### Tests in PR
`eosio.system_tests.cpp, ram_tests.cpp, wasm_config_tests.cpp`

#### Tests in progress
`currency_tests.cpp`

#### Tests Failed
`partitioned_block_log_tests.cpp, state_history_tests.cpp`

#### Tests To Do
`deep_mind_tests.cpp, forked_tests.cpp, whitelist_blacklist_tests.cpp`

#### The rest are not applicable (Already Savanna specific or not using libtester)

arhag:
Tests that still need to be individually addressed are tracked by separated issues.

PR #145, opened 2024-05-15
Add support to run unit tests in both Legacy and Savanna (Part 1)

linh2931:
This PR (and subsequent ones) addresses those requirements:
* for any unit test which can be run in both Legacy and Savanna (majority of our unit tests are), we would like CICD to run it in both modes,
* for any tests which can be only run in Legacy or Savanna, place in a separate file,
* we want to easily turn off and remove Legacy testing some time in the future

This is the first PR which demonstrates how running both modes can be achieved. After it is approved, it will be applied to other tests.
 
Both `BOOST_AUTO_TEST_CASE_TEMPLATE` and `BOOST_FIXTURE_TEST_CASE_TEMPLATE` work. Either involves about the same amount of coding. If an original test already use `tester chain;` declaration, `BOOST_AUTO_TEST_CASE_TEMPLATE` requires fewer changes; if it uses a fixture, `BOOST_FIXTURE_TEST_CASE_TEMPLATE` can be used, but it requires adding `this` or `T` in front of methods.

```
BOOST_AUTO_TEST_CASE_TEMPLATE( missing_sigs, T, validating_testers ) { try {
   T chain;
   chain.create_accounts( {"alice"_n} );

BOOST_FIXTURE_TEST_CASE_TEMPLATE( missing_sigs, T, validating_testers, T ) { try {
   this->create_accounts( {"alice"_n} );
   or
   T::create_accounts( {"alice"_n} );
```
   

### DIFF
--- a/unittests/auth_tests.cpp
+++ b/unittests/auth_tests.cpp
@@ -17,37 +17,43 @@ using namespace eosio::testing;
 
 BOOST_AUTO_TEST_SUITE(auth_tests)
 
-BOOST_FIXTURE_TEST_CASE( missing_sigs, validating_tester ) { try {
-   create_accounts( {"alice"_n} );
-   produce_block();
+BOOST_AUTO_TEST_CASE_TEMPLATE( missing_sigs, TESTER, validating_testers ) { try {
+   TESTER chain;
 
-   BOOST_REQUIRE_THROW( push_reqauth( "alice"_n, {permission_level{"alice"_n, config::active_name}}, {} ), unsatisfied_authorization );
-   auto trace = push_reqauth("alice"_n, "owner");
+   chain.create_accounts( {"alice"_n} );
+   chain.produce_block();
 
-   produce_block();
-   BOOST_REQUIRE_EQUAL(true, chain_has_transaction(trace->id));
+   BOOST_REQUIRE_THROW( chain.push_reqauth( "alice"_n, {permission_level{"alice"_n, config::active_name}}, {} ), unsatisfied_authorization );
+   auto trace = chain.push_reqauth("alice"_n, "owner");
+
+   chain.produce_block();
+   BOOST_REQUIRE_EQUAL(true, chain.chain_has_transaction(trace->id));
 
 } FC_LOG_AND_RETHROW() } /// missing_sigs
 
-BOOST_FIXTURE_TEST_CASE( missing_multi_sigs, validating_tester ) { try {
-    produce_block();
-    create_account("alice"_n, config::system_account_name, true);
-    produce_block();
+BOOST_AUTO_TEST_CASE_TEMPLATE( missing_multi_sigs, TESTER, validating_testers ) { try {
+    TESTER chain;
 
-    BOOST_REQUIRE_THROW(push_reqauth("alice"_n, "owner"), unsatisfied_authorization); // without multisig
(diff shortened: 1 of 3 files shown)


unittests/auth_tests.cpp
```diff
@@ -17,81 +17,90 @@ using namespace eosio::testing;
...
    //todoBOOST_REQUIRE_EQUAL(true, chain_has_transaction(trace->id));
 
 } FC_LOG_AND_RETHROW() }
 
 
 BOOST_AUTO_TEST_CASE(update_auths) {
 try {
-   validating_tester chain;
+   savanna_validating_tester chain;
+
```
linh2931, line 103: `chain.find` on line 120 below causes compilation failure when `BOOST_AUTO_TEST_CASE_TEMPLATE` is used. We test Savanna for now.

heifner: Probably just need `chain.template find<permission_object, by_owner>(boost::make_tuple(name("alice"), name("owner")));`

unittests/auth_tests.cpp
```diff
@@ -17,81 +17,90 @@ using namespace eosio::testing;
 
 BOOST_AUTO_TEST_SUITE(auth_tests)
 
-BOOST_FIXTURE_TEST_CASE( missing_sigs, validating_tester ) { try {
-   create_accounts( {"alice"_n} );
-   produce_block();
+BOOST_AUTO_TEST_CASE_TEMPLATE( missing_sigs, TESTER, validating_testers ) { try {
+   TESTER chain;
```
heifner, line 21: Would be nice if there was a way to combine `BOOST_AUTO_TEST_CASE_TEMPLATE` and `BOOST_FIXTURE_TEST_CASE` so that `TESTER chain;` was not needed. I don't see a way to do that, but maybe @spoonincode might know a way.

heifner (approved):

unittests/auth_tests.cpp
linh2931: Thank! It worked.

unittests/auth_tests.cpp
spoonincode: I believe you can use the (sparsely documented?[^1]) `BOOST_FIXTURE_TEST_CASE_TEMPLATE` for example
```cpp
BOOST_AUTO_TEST_CASE_TEMPLATE( missing_sigs, T, validating_testers, T ) { try {
```
but there is a problem with dependent lookup so you're going to get stuck with doing either
```cpp
BOOST_AUTO_TEST_CASE_TEMPLATE( missing_sigs, T, validating_testers, T ) { try {
   this->create_accounts( {"alice"_n} );
   T::produce_block();
```
so it's not immediately clear to me if it's any better or not.

[^1]: I only found it documented at the very bottom of this page https://www.boost.org/doc/libs/1_85_0/libs/test/doc/html/boost_test/tests_organization/decorators/explicit_decorator_declaration.html

linh2931: Thanks for looking into this. Either way is fine. But I think `chain.create_accounts` is more readable. We can just keep it as is.

spoonincode (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Run unit tests in both Legacy DPOS and Savanna Consensus Modes, part 1.
Note:end

PR #147, opened 2024-05-16
[1.0-beta1] Test: P2P: Add different sync-fetch-span for catchup test

heifner:
Adds test for #146 

Use different sync-fetch-span options for syncing up.

Fails before #146: https://github.com/AntelopeIO/spring/actions/runs/9104723740

PR #151, opened 2024-05-16
[1.0-beta1] P2P: Fix stuck in peer syncing state

### LINKED ISSUE #150: Test failure: lib_advance_test
https://github.com/AntelopeIO/spring/actions/runs/9114438126/job/25059003107?pr=149

node_03 is stuck in lib catchup. Since blocks can be processed inside controller during a fork and don't call `sync_master->sync_recv_block` with `blk_applied==true`.

heifner:
Test failure: https://github.com/AntelopeIO/spring/actions/runs/9114438126/job/25059003107?pr=149

The producing node (Node03) was stuck in LIB catchup. Since blocks can be processed inside controller during a fork and not call `sync_master->sync_recv_block` with `blk_applied==true`. 

This is similar to the issue addressed in #27. In the #27, a handshake with the wrong information was sent. In this case no handshake was sent at all as it become synced up during a fork switch. I believe this issue was masked by #147 because it was incorrectly reporting blocks as applied. Without this improper reporting of applied blocks a fork switch that applies up to synced would not call  `sync_master->sync_recv_block` with `blk_applied==true` and not move node out of LIB catchup.

Now also send out handshakes if we determine we are synced and there is a pending handshake to send.

Related to #27 
Related to #147 

Resolves #150

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -269,6 +269,7 @@ namespace eosio {
                             const fc::microseconds& blk_latency );
       void recv_handshake( const connection_ptr& c, const handshake_message& msg, uint32_t nblk_combined_latency );
       void sync_recv_notice( const connection_ptr& c, const notice_message& msg );
+      void send_handshakes_if_synced(const fc::microseconds& blk_latency);
    };
 
    class dispatch_manager {
@@ -534,7 +535,7 @@ namespace eosio {
       uint32_t get_chain_head_num() const;
 
       void on_accepted_block_header( const signed_block_ptr& block, const block_id_type& id );
-      void on_accepted_block();
+      void on_accepted_block( const signed_block_ptr& block, const block_id_type& id );
       void on_voted_block( uint32_t connection_id, vote_status stauts, const vote_message_ptr& vote );
 
       void transaction_ack(const std::pair<fc::exception_ptr, packed_transaction_ptr>&);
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -4007,7 +4018,8 @@ namespace eosio {
       });
    }
 
-   void net_plugin_impl::on_accepted_block() {
+   void net_plugin_impl::on_accepted_block( const signed_block_ptr& block, const block_id_type& id ) {
```
linh2931, line 4021: `id` is not used.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4007,7 +4018,8 @@ namespace eosio {
...
+      sync_master->send_handshakes_if_synced(fc::time_point::now() - block->timestamp);
```
linh2931, line 4022: You can add an assert `now` is greater than `timestamp`.

heifner: We can not assume that `now()` is greater than `block->timestamp`.

spoonincode (approved):

plugins/net_plugin/net_plugin.cpp
linh2931: Oh I see `fc::microseconds` checks negative input.

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: BUG
summary: Add handshakes to cover additional P2P scenarios, including during a fork and when there are pending handshakes after a sync complete. 
Note:end

PR #153, opened 2024-05-17
Disaster recovery scenario 3

### LINKED ISSUE #13: Disaster recovery integration test
## Scenario 1

Create integration test with 4 nodes (A, B, C, and D) which each have their own producer and finalizer. The finalizer policy consists of the four finalizers with a threshold of 3. The proposer policy involves all four proposers.

The 4 nodes are cleanly shutdown in the following state:
- A has LIB N. A has a finalizer safety information file that locks on a block after N.
- B, C, and D have LIB less than N. They have finalizer safety information files that lock on N.

Nodes B, C, and D lose their reversible blocks. All nodes restart from an earlier snapshot.

A is restarted and replays up to its last reversible block (which is a block number greater than N) after restarting from snapshot. Blocks N and later is sent to the other nodes B, C, and D after they are also started up again.

## Scenario 2

Create an integration test with 5 nodes (A, B, C, D, and P). Nodes A, B, C, and D each have one finalizer but no proposers. Node P has a proposer but no finalizers. The finalizer policy consists of the four finalizers with a threshold of 3. The proposer policy involves just the single proposer P.

A, B, C, and D can be connected to each other however we like as long as blocks sent to node A can traverse to the other nodes B, C, and D. However, node P should only be connected to node A.

At some point after IF transition has completed and LIB is advancing, block production on node P should be paused. Enough time should be given to allow and in-flight votes on the latest produced blocks to be delivered to node P. Then, the connection between node P and node A should be severed, and then block production on node P resumed. The LIB on node P should advance to but then stall at block N. Then shortly after that, node P should be cleanly shut down.

Verify that the LIB on A, B, C, and D has stalled and is less than block N. Then, nodes A, B, C, and D can all be cleanly shut down.

Then, reversible blocks from all nodes should be removed. All nodes are restarted from an earlier snapshot (prior to block N).

P is restarted and replays up to block N after restarting from snapshot. Blocks up to and including block N are sent to the other nodes A, B, C, and D after they are also started up again.

## Scenario 3

Create integration test with 4 nodes (A, B, C, and D) which each have their own producer and finalizer. The finalizer policy consists of the four finalizers with a threshold of 3. The proposer policy involves all four proposers.

- At least two of the four nodes should have a LIB N and a finalizer safety information file that locks on a block after N. The other two nodes should have a LIB that is less than or equal to block N.

All nodes are shut down. The reversible blocks on all nodes is deleted. Then restart all nodes from an earlier snapshot.

All nodes eventually sync up to block N. Some nodes will consider block N to LIB but others may not.

Not enough finalizers should be voting because of the lock in their finalizer safety information file. Verify that LIB does not advance on any node.

Cleanly shut down all nodes and delete their finalizer safety information files. Then restart the nodes.

Verify that LIB advances on all nodes and they all agree on the LIB. In particular, verify that block N is the same ID on all nodes as the one before nodes were first shutdown.

heifner:
## Scenario 3

Create integration test with 4 nodes (A, B, C, and D) which each have their own producer and finalizer. The finalizer policy consists of the four finalizers with a threshold of 3. The proposer policy involves all four proposers.

- At least two of the four nodes should have a LIB N and a finalizer safety information file that locks on a block after N. The other two nodes should have a LIB that is less than or equal to block N.

All nodes are shut down. The reversible blocks on all nodes is deleted. Then restart all nodes from an earlier snapshot.

All nodes eventually sync up to block N. Some nodes will consider block N to LIB but others may not.

Not enough finalizers should be voting because of the lock in their finalizer safety information file. Verify that LIB does not advance on any node.

Cleanly shut down all nodes and delete their finalizer safety information files. Then restart the nodes.

Verify that LIB advances on all nodes and they all agree on the LIB. In particular, verify that block N is the same ID on all nodes as the one before nodes were first shutdown.

Resolves #13


tests/TestHarness/Node.py
```diff
@@ -558,6 +563,11 @@ def removeReversibleBlks(self):
         reversibleBlks = os.path.join(dataDir, "blocks", "reversible")
         shutil.rmtree(reversibleBlks, ignore_errors=True)
 
+    def removeFinalizersSafetyFile(self):
```
linh2931, line 566: Probably this should be named `removeFinalizersSafetyDir`.

tests/disaster_recovery_3.py
```diff
@@ -0,0 +1,131 @@
...
+import shutil
+import signal
+import time
+from TestHarness import Cluster, TestHelper, Utils, WalletMgr
+from TestHarness.Node import BlockType
+
+###############################################################
+# disaster_recovery - Scenario 3
+#
+# Create integration test with 4 nodes (A, B, C, and D) which each have their own producer and finalizer. The finalizer
```
linh2931, line 12: Should be `each have their` be `each has its`?

tests/disaster_recovery_3.py
```diff
@@ -0,0 +1,131 @@
...
+    Print("Restart nodes from snapshot")
+    for i in range(4):
+        isRelaunchSuccess = cluster.getNode(i).relaunch(chainArg=" -e --snapshot {}".format(node0.getLatestSnapshot()))
+        assert isRelaunchSuccess, f"node {i} relaunch from snapshot failed"
+
+    Print("Verify LIB does not advance on any node")
+    for node in [node0, node1, node2, node3]:
+        assert not node.waitForLibToAdvance(), "Node advanced LIB after relaunch when it should not"
+
+    Print("Shutdown all nodes to remove finalizer safety data")
```
linh2931, line 105: You don't remove finalizer safety data here.

tests/disaster_recovery_3.py
```diff
@@ -0,0 +1,131 @@
...
+    for node in [node0, node1, node2, node3]:
+        node.kill(signal.SIGTERM)
+    for node in [node0, node1, node2, node3]:
+        assert not node.verifyAlive(), "Node did not shutdown"
+
+    for node in [node0, node1, node2, node3]:
```
linh2931, line 111: Print remove finalizer safety data here

tests/disaster_recovery_3.py
```diff
@@ -0,0 +1,131 @@
...
+from TestHarness import Cluster, TestHelper, Utils, WalletMgr
+from TestHarness.Node import BlockType
+
+###############################################################
+# disaster_recovery - Scenario 3
+#
+# Create integration test with 4 nodes (A, B, C, and D) which each have their own producer and finalizer. The finalizer
+# policy consists of the four finalizers with a threshold of 3. The proposer policy involves all four proposers.
+#
+# - At least two of the four nodes should have a LIB N and a finalizer safety information file that locks on a block
```
linh2931, line 15: Probably you should say something like "make a condition such that at least two of the four nodes should have ..."

ericpassmore:
Note:start
group: IF
category: TEST
summary: Four node disaster recovery tests with finalizer lock and recovery from previous state. 
Note:end

tests/disaster_recovery_3.py
heifner: I rewrote the paragraph.

tests/disaster_recovery_3.py
heifner: We do on line 112, that is why we are shutting down.

tests/disaster_recovery_3.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Print("Shutdown two nodes")
+    for node in [node0, node1]:
+        node.kill(signal.SIGTERM)
+    for node in [node0, node1]:
+        assert not node.verifyAlive(), "Node did not shutdown"
+
+    Print("Wait for lib to advance on other nodes")
+    for node in [node2, node3]:
+        assert node.waitForBlock(currentLIB-1, timeout=None, blockType=BlockType.lib), "Node did not advance LIB after shutdown of node0 and node1"
```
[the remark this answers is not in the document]
linh2931: I think you can print and add a comment in the code where those conditions are met (like which nodes are two of four, which are the other two).

tests/disaster_recovery_3.py
```diff
@@ -0,0 +1,142 @@
...
+    for node in [node0, node1, node2, node3]:
+        node.removeReversibleBlks()
+        node.removeState()
+
+    Print("Restart nodes from snapshot")
+    for i in range(4):
+        isRelaunchSuccess = cluster.getNode(i).relaunch(chainArg=" -e --snapshot {}".format(node0.getLatestSnapshot()))
+        assert isRelaunchSuccess, f"node {i} relaunch from snapshot failed"
+
+    Print("Verify LIB does not advance on any node")
```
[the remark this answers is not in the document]
heifner: Added verification.

tests/disaster_recovery_3.py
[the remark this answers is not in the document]
heifner: Fixed an off-by-one and updated the test to be clearer.

linh2931 (approved):

arhag (approved):

PR #158, opened 2024-05-17
[1.0-beta1] Vote processing fix

### LINKED ISSUE #139: Test Failure: svnn_ibc_unit_test_eos-vm-jit
https://github.com/AntelopeIO/spring/actions/runs/9084138059/job/24964976733?pr=133#step:5:1356

heifner:
- Do not post to vote thread pool if thread pool is not active.
  - This fixes a memory leak when there are no vote threads; which is why this PR is targeted to `1.0-beta1` instead of `main`
- A producer should notify vote processor that it has produced a block
  - I believe this resolves issue #139 
- Update the tester wait on votes to be consistent

Resolves #139

### DIFF
--- a/libraries/testing/tester.cpp
+++ b/libraries/testing/tester.cpp
@@ -499,7 +499,7 @@ namespace eosio::testing {
          // wait for this node's vote to be processed
          size_t retrys = 200;
          while (!c.node_has_voted_if_finalizer(c.head_block_id()) && --retrys) {
-            std::this_thread::sleep_for(std::chrono::milliseconds(10));
+            std::this_thread::sleep_for(std::chrono::milliseconds(5));
          }
          FC_ASSERT(retrys, "Never saw this nodes vote processed before timeout");
       }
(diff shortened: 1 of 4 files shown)


spoonincode (approved):

ericpassmore:
Note:start
group: IF
category: BUG
summary: Do not post to vote thread pool if thread pool is not active.
Note:end

libraries/testing/tester.cpp
```diff
@@ -499,7 +499,7 @@ namespace eosio::testing {
          // wait for this node's vote to be processed
          size_t retrys = 200;
          while (!c.node_has_voted_if_finalizer(c.head_block_id()) && --retrys) {
-            std::this_thread::sleep_for(std::chrono::milliseconds(10));
+            std::this_thread::sleep_for(std::chrono::milliseconds(5));
```
linh2931, line 502: It would be better if `retries` and the sleep time are defined as constants so that they can be kept consistent in the future. But you don't need to make this change for now.

linh2931 (approved):

PR #160, opened 2024-05-17
[1.0-beta1.1] Bump version to beta1.1

heifner:


linh2931 (approved):
So we use a new branch for `beta1.1`?

heifner:
> So we use a new branch for `beta1.1`?

That was my understandnig. @arhag

PR #163, opened 2024-05-18
Improve fork database logging

heifner:
- Change a `warn` level log to `info` since it does not necessarily indicate any issue.
- Add block ids to fork database log output.
- Add additional fork database log output at shutdown.

### DIFF
--- a/libraries/chain/fork_database.cpp
+++ b/libraries/chain/fork_database.cpp
@@ -209,6 +209,9 @@ namespace eosio::chain {
    void fork_database_impl<BSP>::close_impl(std::ofstream& out) {
       assert(!!head && !!root); // if head or root are null, we don't save and shouldn't get here
 
+      ilog("Writing fork_database with root ${rn}:${r} and head ${hn}:${h}",
+           ("rn", root->block_num())("r", root->id())("hn", head->block_num())("h", head->id()));
+
       fc::raw::pack( out, *root );
 
       uint32_t num_blocks_in_fork_db = index.size();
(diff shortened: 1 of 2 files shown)


libraries/chain/fork_database.cpp
```diff
@@ -719,8 +722,7 @@ namespace eosio::chain {
       auto in_use_value = in_use.load();
       // check that fork_dbs are in a consistent state
       if (!legacy_valid && !savanna_valid) {
-         wlog( "fork_database is in a bad state when closing; not writing out '${filename}', legacy_valid=${l}, savanna_valid=${s}",
-               ("filename", fork_db_file)("l", legacy_valid)("s", savanna_valid) );
+         ilog("No fork_database to persist, not writing out: ${f}", ("f", fork_db_file));
```
linh2931, line 725: Since we are not writing out, we don't need to log the file name `fork_db_file`.

heifner: I thought it didn't hurt to include the file. If you think it is confusing, I can remove it.

linh2931: I think it is a little confusing. The user might think an empty file is there. But  this is very minor, you don't need to change.

linh2931 (approved):

libraries/chain/fork_database.cpp
heifner: Makes sense, I'll remove it

spoonincode (approved):

ericpassmore:
Note:start
group: STABILITY
category: LOGGING
summary: Improve fork database logging, additional data added at shutdown.
Note:end

PR #166, opened 2024-05-20
Make unit tests run in both Legacy and Savanna (Part 3, snapshot_tests)

linh2931:
Update `snapshot_tests` to run in Legacy and Savanna modes.

This is done as a separate PR since the handling is a little different than previous tests: combinations of modes (`Legacy` and `Savanna`) and snapshot suites (`variant_snapshot_suite, buffered_snapshot_suite, json_snapshot_suite`) need to be considered.

### DIFF
--- a/unittests/snapshot_tests.cpp
+++ b/unittests/snapshot_tests.cpp
@@ -481,7 +481,7 @@ void compatible_versions_test()
    //    now the test should pass.
    // 5. add the 3 new snapshot files in git.
    // -------------------------------------------------------------------------------------------------------------
-   // Save one copy of snapshots under Savanna only
+   // Only want to save one snapshot, use Savanna as that is the latest
    if constexpr (std::is_same_v<TESTER, savanna_tester>) {
       if (save_snapshot)
       {
--- a/unittests/snapshot_tests.cpp
+++ b/unittests/snapshot_tests.cpp
@@ -186,9 +186,10 @@ namespace {
    }
 }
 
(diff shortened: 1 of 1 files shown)


unittests/snapshot_tests.cpp
```diff
@@ -459,17 +481,26 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_compatible_versions, SNAPSHOT_SUITE, snapshot
...
    // -------------------------------------------------------------------------------------------------------------
-   if (save_snapshot)
-   {
-      // create a latest snapshot
-      auto latest_writer = SNAPSHOT_SUITE::get_writer();
-      base_chain.control->write_snapshot(latest_writer);
-      auto latest = SNAPSHOT_SUITE::finalize(latest_writer);
-
-      SNAPSHOT_SUITE::write_to_file("snap_" + current_version, latest);
+   // Save one copy of snapshots under Savanna only
```
heifner, line 484: Took me a minute to figure out why you only want this under Savanna. The comment might be better as something like: `// Only want to save one snapshot, use Savanna as that is the latest`

heifner (approved):

unittests/snapshot_tests.cpp
linh2931: Thanks for your suggestion! I was struggling too about how to make it clear, trying to place the `only` in different places...

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: TESTS
summary: Have snapshot unit tests run in both legacy and savanna consensus modes (Part III).
Note:end

spoonincode (approved):

PR #172, opened 2024-05-20
Make unit tests run in both Legacy and Savanna (Part 5, dry_run_trx_tests)

linh2931:
Update `dry_run_trx_tests` to run in both Legacy and Savanna.

This update uses a slightly different approach from previous ones as `dry_run_trx_tests` defines its own test fixture. That fixture is templated to support both Legacy and Savanna. This approach will be used for other tests which have their own fixtures.

### DIFF
--- a/unittests/dry_run_trx_tests.cpp
+++ b/unittests/dry_run_trx_tests.cpp
@@ -12,33 +12,34 @@ using namespace fc;
 
 using mvo = fc::mutable_variant_object;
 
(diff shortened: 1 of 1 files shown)


heifner (approved):

spoonincode (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Run unit tests in both Legacy and Savanna (Part 5).
Note:end

PR #177, opened 2024-05-21
Update protocol_feature_tests to run in both legacy and savanna

heifner:
Part of #11


spoonincode (approved):

linh2931 (approved):
We only transition to Savanna if `setup_policy` is `full`. I think there is no harm to run tests of other `setup_policy` in both Legacy and Savanna.

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Update protocol features tests to support different consensus modes.
Note:end

PR #192, opened 2024-05-23
Test: Add integration test of setfinalizers after transition

heifner:
- Add test of `setfinalizer` after transition to Savanna
- Fix double logging of finalizer change

### DIFF
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -124,7 +124,8 @@ const proposer_policy& block_header_state::get_last_proposed_proposer_policy() c
 void finish_next(const block_header_state& prev,
                  block_header_state& next_header_state,
                  vector<digest_type> new_protocol_feature_activations,
-                 instant_finality_extension if_ext) {
+                 instant_finality_extension if_ext,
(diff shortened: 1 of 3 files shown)


libraries/chain/block_header_state.cpp
```diff
@@ -124,7 +124,8 @@ const proposer_policy& block_header_state::get_last_proposed_proposer_policy() c
 void finish_next(const block_header_state& prev,
                  block_header_state& next_header_state,
                  vector<digest_type> new_protocol_feature_activations,
-                 instant_finality_extension if_ext) {
+                 instant_finality_extension if_ext,
+                 bool log) { // only log on assembled blocks, to avoid double logging
```
linh2931, line 128: It would be nicer if we could move the logging out of this method to the caller, without adding this parameter just for logging. But detailed information we want to log is not accessible in the caller.

linh2931 (approved):

spoonincode (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add integration test of setfinalizers after transition.
Note:end

Issue #210, opened 2024-05-28
test failure: wasm_part2_unit_test_eos-vm-oc timeout


spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/9259956086/job/25473056230

linh2931:
This is to be fixed by https://github.com/AntelopeIO/spring/pull/205

PR #218, opened 2024-05-29
Set cost (priority) to long running tests and split api_tests and wasm_tests

linh2931:
* Set cost (priority) to long running tests so that they can start earlier
* Split api_tests and wasm_tests further so that each can finish within CICD time limits

Resolves https://github.com/AntelopeIO/spring/issues/210


spoonincode:
since memory_trap_1_check_throw_unit_test_eos_vm_oc is hitting the 7 minute timeout, maybe we can try increasing it to 8 minutes
https://github.com/AntelopeIO/spring/blob/e3da612c0c5efcc2e213c3501fdd08d6110324da/.github/workflows/build.yaml#L159
just based on the time ratio of other tests, 8 minutes might do it. But it's also cutting it close to the back end max of 10 minutes (which of course we can change if needed). All of this isn't ideal but that could be the easy quick fix to get green for now.

linh2931:
> since memory_trap_1_check_throw_unit_test_eos_vm_oc is hitting the 7 minute timeout, maybe we can try increasing it to 8 minutes
> 
> https://github.com/AntelopeIO/spring/blob/e3da612c0c5efcc2e213c3501fdd08d6110324da/.github/workflows/build.yaml#L159
> 
> 
> just based on the time ratio of other tests, 8 minutes might do it. But it's also cutting it close to the back end max of 10 minutes (which of course we can change if needed). All of this isn't ideal but that could be the easy quick fix to get green for now.

Wow, `memory_trap_1_check_throw_unit_test_eos-vm-oc` takes `449` seconds, dangerously close to the `480` seconds limit.

spoonincode (approved):

ericpassmore:
Note:start
group: STABILITY
category: CHORE
summary: Update test schedule to finish within CICD limits.
Note:end

heifner (approved):

PR #247, opened 2024-06-06
Update finalizer safety info file based on received QC

### LINKED ISSUE #135: Update finalizer safety information file based on vote evidence from received blocks
Depends on https://github.com/AntelopeIO/spring/issues/259.

If a node deletes their finalizer safety information file, they may lose critical safety information such as the block they should be locked on. But in some cases, this information can be recovered based on the evidence in the QCs attached to blocks that the node receives from peers.

When processing a QC in a block, the node can check to see if any of the finalizer keys it manages are represented in the QC as a strong vote. If so, it can determine (by looking at the history of QC claims within that branch, which is stored in the `finality_core`) which block (ID and timestamp) the node should have been locked on due to casting that strong vote. If that block is more recent than the current lock in the finalizer safety information file for that finalizer key, then the lock should be updated.

heifner:
If a node deletes their finalizer safety information file, they may lose critical safety information such as the block they should be locked on. But in some cases, this information can be recovered based on the evidence in the QCs attached to blocks that the node receives from peers.

When processing a QC in a block, node checks to see if any of the finalizer keys it manages are represented in the QC as a strong vote. If that block is more recent than the current lock in the finalizer safety information file for that finalizer key, then update the lock.

Resolves #135

### DIFF
--- a/libraries/chain/finality/finalizer.cpp
+++ b/libraries/chain/finality/finalizer.cpp
@@ -136,6 +136,10 @@ void my_finalizers_t::maybe_update_fsi(const block_state_ptr& bsp, const valid_q
    if (finalizers.empty())
       return;
 
+   // once we have voted, no reason to continue evaluating incoming QCs
+   if (has_voted.load(std::memory_order::relaxed))
+      return;
+
    assert(bsp->active_finalizer_policy);
 
    // see comment on possible optimization in maybe_vote
--- a/libraries/chain/finality/finalizer.cpp
+++ b/libraries/chain/finality/finalizer.cpp
@@ -86,6 +86,19 @@ finalizer::vote_result finalizer::decide_vote(const block_state_ptr& bsp) {
 }
 
 // ----------------------------------------------------------------------------------------
+// finalizer has voted strong on bsp, update finalizer safety info if more recent than the current lock
+bool finalizer::maybe_update_fsi(const block_state_ptr& bsp) {
+   auto& final_on_strong_qc_block_ref = bsp->core.get_block_reference(bsp->core.final_on_strong_qc_block_num);
+   if (final_on_strong_qc_block_ref.timestamp > fsi.lock.timestamp) {
+      fsi.lock = { final_on_strong_qc_block_ref.block_id, final_on_strong_qc_block_ref.timestamp };
+      fsi.last_vote             = { bsp->id(), bsp->timestamp() };
+      fsi.last_vote_range_start = bsp->core.latest_qc_block_timestamp();
+      return true;
+   }
+   return false;
+}
+
+// ----------------------------------------------------------------------------------------
 vote_message_ptr finalizer::maybe_vote(const bls_public_key& pub_key,
                                        const block_state_ptr& bsp,
                                        const digest_type& digest) {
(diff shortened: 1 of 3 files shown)


libraries/chain/finality/finalizer.cpp
```diff
@@ -85,6 +85,19 @@ finalizer::vote_result finalizer::decide_vote(const block_state_ptr& bsp) {
    return res;
 }
 
+// ----------------------------------------------------------------------------------------
+// finalizer has voted strong on bsp, update finalizer safety info if more recent than the current lock
+bool finalizer::maybe_update_fsi(const block_state_ptr& bsp) {
+   auto& final_on_strong_qc_block_ref = bsp->core.get_block_reference(bsp->core.final_on_strong_qc_block_num);
+   if (final_on_strong_qc_block_ref.timestamp > fsi.lock.timestamp) {
```
greg7mdp, line 92: I think we should do the update iff `fsi,lock.empty()`, so not compare the timestamp.

libraries/chain/finality/finalizer.cpp
```diff
@@ -105,6 +118,41 @@ vote_message_ptr finalizer::maybe_vote(const bls_public_key& pub_key,
...
+      return;
+
+   assert(bsp->active_finalizer_policy);
+
+   // see comment on possible optimization in maybe_vote
+   std::lock_guard g(mtx);
+
+   bool updated = false;
+   for (auto& f : finalizers) {
+      if (has_voted_strong(bsp->active_finalizer_policy->finalizers, received_qc, f.first)) {
```
greg7mdp, line 146: We need to check `has_voted_strong` only iff `f.second.fsi.lock.empty()`

libraries/chain/finality/finalizer.cpp
heifner: I understood the requirement to be if more recent. @arhag

libraries/chain/finality/finalizer.cpp
```diff
@@ -105,6 +118,41 @@ vote_message_ptr finalizer::maybe_vote(const bls_public_key& pub_key,
...
+}
+
+void my_finalizers_t::maybe_update_fsi(const block_state_ptr& bsp, const valid_quorum_certificate& received_qc) {
+   if (finalizers.empty())
+      return;
+
+   assert(bsp->active_finalizer_policy);
+
+   // see comment on possible optimization in maybe_vote
+   std::lock_guard g(mtx);
```
linh2931, line 142: Can this be moved down right before line 152 `save_finalizer_safety_info()`?

heifner: The mutex guards modifications of `finalizer.fsi`. It could be moved down before the first call to `maybe_update_fsi` but didn't think the extra complexity was worth it. Would need to then see if the mutex was locked on each iteration of the loop as you could not unlock it until the file is written.

libraries/chain/finality/finalizer.cpp
greg7mdp: Actually, what you have is probably correct.
I realize we have not implemented the `Initialization on Leap startup` part of https://github.com/AntelopeIO/leap/issues/2070#issuecomment-1941901836

libraries/chain/finality/finalizer.cpp
greg7mdp: Yes, you are right, my bad.

libraries/chain/finality/finalizer.cpp
```diff
@@ -85,6 +85,19 @@ finalizer::vote_result finalizer::decide_vote(const block_state_ptr& bsp) {
    return res;
 }
 
+// ----------------------------------------------------------------------------------------
+// finalizer has voted strong on bsp, update finalizer safety info if more recent than the current lock
+bool finalizer::maybe_update_fsi(const block_state_ptr& bsp) {
+   auto& final_on_strong_qc_block_ref = bsp->core.get_block_reference(bsp->core.final_on_strong_qc_block_num);
+   if (final_on_strong_qc_block_ref.timestamp > fsi.lock.timestamp) {
+      fsi.lock = { final_on_strong_qc_block_ref.block_id, final_on_strong_qc_block_ref.timestamp };
+      fsi.last_vote             = { bsp->id(), bsp->timestamp() };
```
greg7mdp, line 94: I thgink maybe we should update `fsi.last_vote` only if `fsi.last_vote.timestamp` < `bsp->timestamp()` to make sure we don't break the monotony check.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Mechanism to update node in the event finalizer safety data is removed.
Note:end

greg7mdp:
Some more thoughts:
1. I think we should consider updating the safety info of a finalizer if and only if a safety file was not present for this finalizer when nodeos launched.
2. I think that, once a finalizer has voted strong and updated the lock, we should stop considering the received votes for a potential update of this finalizer.

arhag:
@greg7mdp:

> 1. I think we should consider updating the safety info of a finalizer if and only if a safety file was not present for this finalizer when nodeos launched.
> 2. I think that, once a finalizer has voted strong and updated the lock, we should stop considering the received votes for a potential update of this finalizer.

I don't think we should do 1. We still want to handle the case where nodeos is using an old finalizer safety information file.

I do think we should do 2, after we have implemented the changes of https://github.com/AntelopeIO/spring/issues/259.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Extend scenarios where a deleted safety data file may be recovered.
Note:end

PR #262, opened 2024-06-10
Avoid voting until caught up with network

### LINKED ISSUE #259: On startup, avoid voting until caught up with network
When nodeos starts, the finalizers will not vote. There is a boolean that must be turned on before any of the finalizers can vote. Once that boolean is switched, it will not switch back for the remainder of the nodeos process lifetime.

The switch can be determined based on the difference in the timestamp of the block and the wall-clock time. For example, after startup, while the blocks being processed have a timestamp that is older than the wall-clock time minus 30 seconds, then the boolean remains unchanged. The first time a block is processed with a timestamp that is within 30 seconds of the wall-clock time, the boolean should be switched on and voting can begin to be considered for that block and any other blocks processed later by nodeos (regardless of what their timestamp was).

This change is simply to avoid spamming their peers with unnecessary votes when nodeos is first syncing old blocks.

Another improvement along these lines would be to not bother sending to votes to any peer that we are still syncing from.

heifner:
- Do not consider voting until within 30 seconds of head block timestamp
- Do not broadcast votes if syncing from peer

Resolves #259

### DIFF
--- a/libraries/chain/include/eosio/chain/finality/finalizer.hpp
+++ b/libraries/chain/include/eosio/chain/finality/finalizer.hpp
@@ -130,7 +130,9 @@ namespace eosio::chain {
          return std::ranges::all_of(std::views::keys(finalizers), std::forward<F>(f));
       }
 
-      void    set_keys(const std::map<std::string, std::string>& finalizer_keys); // only call on startup
+      /// only call on startup
+      /// @param enable_immediate_voting if true enable immediate voting on startup (for testing)
+      void    set_keys(const std::map<std::string, std::string>& finalizer_keys, bool enable_immediate_voting);
       void    set_default_safety_information(const fsi_t& fsi);
 
       // following two member functions could be private, but are used in testing, not thread safe
--- a/libraries/chain/include/eosio/chain/finality/finalizer.hpp
+++ b/libraries/chain/include/eosio/chain/finality/finalizer.hpp
@@ -71,7 +71,6 @@ namespace eosio::chain {
       using fsi_map = std::map<bls_public_key, fsi_t>;
 
    private:
-      const block_timestamp_type        t_startup;             // nodeos startup time, used for default safety_information
       const std::filesystem::path       persist_file_path;     // where we save the safety data
       mutable std::mutex                mtx;
       mutable fc::datastream<fc::cfile> persist_file;          // we want to keep the file open for speed
@@ -81,9 +80,8 @@ namespace eosio::chain {
       mutable bool                      inactive_safety_info_written{false};
 
    public:
-      my_finalizers_t(block_timestamp_type startup_time, const std::filesystem::path& persist_file_path)
-         : t_startup(startup_time)
-         , persist_file_path(persist_file_path)
+      explicit my_finalizers_t(const std::filesystem::path& persist_file_path)
+         : persist_file_path(persist_file_path)
       {}
 
       template<class F> // thread safe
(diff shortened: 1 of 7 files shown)


libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -71,41 +71,44 @@ namespace eosio::chain {
...
+         const auto& fin_pol = *bsp->active_finalizer_policy;
+
          std::vector<vote_message_ptr> votes;
          votes.reserve(finalizers.size());
 
          // Possible improvement in the future, look at locking only individual finalizers and releasing the lock for writing the file.
          // Would require making sure that only the latest is ever written to the file and that the file access was protected separately.
          std::unique_lock g(mtx);
 
+         if (!enable_voting && bsp->timestamp() < fc::time_point::now() - fc::seconds(30))
```
linh2931, line 104: Add a comment here saying why this is needed and `enable_voting` will never go back to `false` once enabled.

libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -71,41 +71,44 @@ namespace eosio::chain {
...
+            return;
+         enable_voting = true;
```
linh2931, line 106: Maybe you can re-order the condition above such that `enable_voting = true;` is not always executed once it is true.

linh2931 (approved):

libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -71,41 +71,46 @@ namespace eosio::chain {
...
 
          // Possible improvement in the future, look at locking only individual finalizers and releasing the lock for writing the file.
          // Would require making sure that only the latest is ever written to the file and that the file access was protected separately.
          std::unique_lock g(mtx);
 
+         if (!enable_voting) { // Avoid extra processing while syncing. Once caught up, consider voting
+            if (bsp->timestamp() < fc::time_point::now() - fc::seconds(30))
+               return;
+            enable_voting = true;
+         }
```
greg7mdp, line 108: move before line 97 to avoid extra memory allocation+free when syncing

libraries/testing/tester.cpp
```diff
@@ -1253,7 +1253,7 @@ namespace eosio::testing {
                ("pop", pop.to_string()));
       }
 
-      control->set_node_finalizer_keys(local_finalizer_keys);
```
greg7mdp, line 1256: Why is the `enable_immediate_voting` needed?

heifner: I believe because the blocks are produced with an old (near genesis) timestamp.

greg7mdp: That's odd! Shouldn't we fix that instead of having this `enable_immediate_voting` workaround?

heifner: We could look into it, but I suspect it would require a newly generated genesis for each test.

libraries/chain/include/eosio/chain/finality/finalizer.hpp
greg7mdp: Yes, I think `atomic` may be best. Even without `atomic`, it seems that the worse that could happen is for a couple extra votes to go though.

greg7mdp: or move the reserve within the mutex.

greg7mdp: Maybe even better would be to do this check on the main thread before we call `consider_voting(bsp, use_thread_pool_t::yes);`, so we don't post unneeded tasks on the thread pool.

libraries/testing/tester.cpp
greg7mdp: Oh, I get why this happens. We want blocks to be produced every 500ms by default, so that each producer does its 12 blocks, regard;less of the clock time.

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Avoid spamming peers with unnecessary votes when nodeos is first syncing old blocks.
Note:end

PR #269, opened 2024-06-11
Do not post to thread pool while syncing/replaying

heifner:
As pointed out in review of #262,
address https://github.com/AntelopeIO/spring/pull/262#discussion_r1633799054


libraries/chain/controller.cpp
```diff
@@ -3978,18 +3980,20 @@ struct controller_impl {
       // 3. Otherwise, consider voting for that block according to the decide_vote rules.
 
       if (!my_finalizers.empty() && bsp->core.final_on_strong_qc_block_num > 0) {
-         if (use_thread_pool == use_thread_pool_t::yes) {
-            boost::asio::post(thread_pool.get_executor(), [this, bsp=bsp]() {
+         if (bsp->is_recent() || my_finalizers.is_active()) {
```
greg7mdp, line 3983: A little annoying that we need `|| my_finalizers.is_active()`, but I guess it is needed to ensure that once we start voting, we don't stop.

greg7mdp (approved):

libraries/chain/controller.cpp
heifner: Yeah, not really sure we need that restriction, but it is current requirements.

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: When syncing/replaying optimize voting-start by processing on main thread.
Note:end

PR #270, opened 2024-06-11
Test: Increase timeout for votes

### LINKED ISSUE #235: test failure: `finality_unit_test_eos-vm`
https://github.com/AntelopeIO/spring/actions/runs/9372030591/job/25803128010
```
10 assert_exception: Assert Exception
retrys: Never saw this nodes vote processed before timeout
    {}
    unit_test  tester.cpp:524 _wait_for_vote_if_needed

unknown location(0): fatal error: in "finality_tests/out_of_order_votes": std::runtime_error: Caught Unexpected Exception
```

heifner:
Although unlikely, it does seem possible with running so many unittests concurrently that a particular test might not receive necessary CPU for 1 second. Increase timeout to 5 seconds. If this fails again, we can then assume this is something real (a vote is actually getting lost). 

Resolves #235


greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: For better test stability increase timeout for tests.
Note:end

Issue #264, opened 2024-06-10
libtester voting optimizations


arhag:
By default only use 1 finalizer key.

Additionally, consider the following optimization for calls to produce_blocks:

- If the number of blocks to produce is 3 or less, continue voting on each block as is currently done.
- If the number of blocks to produce is more than 3, only vote on the last 3 blocks.

This saves time on unnecessary BLS math operations for tests that need to produce a lot of empty blocks as part of the test.

PR #273, opened 2024-06-11
Test: Remove production of unneeded blocks

heifner:
Remove extra block production not needed that only slow down test execution.

Partially #264

### DIFF
--- a/unittests/eosio_system_tester.hpp
+++ b/unittests/eosio_system_tester.hpp
@@ -27,12 +27,12 @@ public:
    eosio_system_tester(Lambda setup) {
       setup(*this);
 
-      T::produce_blocks( 2 );
+      T::produce_block();
 
       T::create_accounts({ "eosio.token"_n, "eosio.ram"_n, "eosio.ramfee"_n, "eosio.stake"_n,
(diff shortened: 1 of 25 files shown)


unittests/eosio_system_tester.hpp
```diff
@@ -538,7 +538,8 @@ class eosio_system_tester : public T {
             BOOST_REQUIRE_EQUAL( T::success(), regproducer(p) );
          }
       }
-      T::produce_blocks( 250);
+      T::produce_block();
+      T::produce_block(fc::seconds(1000));
```
greg7mdp, line 542: Why `1000`? I would have used `125`, but maybe it doesn't matter?

greg7mdp (approved):

unittests/eosio_system_tester.hpp
heifner: 125 would have been fine. Doesn't matter as the performance is the same.

linh2931 (approved):
Should we keep or make a single test to test producing multiple blocks?

heifner:
> Should we keep or make a single test to test producing multiple blocks?

There are some remaining that produce multiple blocks.

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Remove production of unneeded blocks to improve performance of tests. 
Note:end

PR #279, opened 2024-06-12
Test: Optimize test by minimizing voting

### LINKED ISSUE #264: libtester voting optimizations

heifner:
- Only vote on the last three empty blocks of a set to optimize test time
- Update tests to produce empty blocks where possible to speed up test time

Resolves #264


libraries/testing/tester.cpp
```diff
@@ -529,8 +529,13 @@ namespace eosio::testing {
    signed_block_ptr base_tester::produce_blocks( uint32_t n, bool empty ) {
       signed_block_ptr res;
       if( empty ) {
-         for( uint32_t i = 0; i < n; ++i )
+         for( uint32_t i = 0; i < n; ++i ) {
```
linh2931, line 532: Skip voting only when producing empty blocks?

unittests/producer_schedule_if_tests.cpp
```diff
@@ -174,7 +174,7 @@ BOOST_FIXTURE_TEST_CASE( proposer_policy_progression_test, legacy_validating_tes
                                };
 
    // start a round of production
-   produce_blocks(config::producer_repetitions-1);
+   produce_blocks(config::producer_repetitions-1, true);
```
linh2931, line 177: It seems clearer using a different name for possible skipping voting than passing a `true` argument.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -669,8 +673,6 @@ namespace eosio::testing {
             return;
          }
          try {
-            if( num_blocks_to_producer_before_shutdown > 0 )
```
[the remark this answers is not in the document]
heifner: No idea, but it is not used, so no reason to have it.

libraries/testing/tester.cpp
heifner: For some tests you need to vote on every block, for example the block log tests that require blocks to be finalized. This just adds additional analogous functionality to "empty".

unittests/producer_schedule_if_tests.cpp
heifner: If you think it would be better to have a separate argument I can add one, just don't see the point. There are not very many tests that need the functionality so seemed best to roll it into the existing empty arg.

linh2931: No need to change.

linh2931 (approved):

libraries/testing/tester.cpp
```diff
@@ -529,8 +529,13 @@ namespace eosio::testing {
...
+            if (n > 3 && i < n - 3) // for performance, only vote on the last three to move finality
+               control->allow_voting(false);
             res = produce_empty_block();
+            if (n > 3 && i >= n - 3)
+               control->allow_voting(true);
+         }
```
greg7mdp, line 538:
```suggestion
            if (n > 3) // for performance, only vote on the last three to move finality
               control->allow_voting(i >= n-3);
            res = produce_empty_block();
         }
```

greg7mdp: Also don't require `empty` for the optimization.

libraries/testing/tester.cpp
```diff
@@ -529,11 +529,19 @@ namespace eosio::testing {
...
       signed_block_ptr res;
       if( empty ) {
-         for( uint32_t i = 0; i < n; ++i )
+         for( uint32_t i = 0; i < n; ++i ) {
+            // for performance, only vote on the last four to move finality
+            // This is 4 instead of 3 because the extra block has to be produced to log_irreversible
+            if (n > 4)
+               control->allow_voting(i >= n - 4);
             res = produce_empty_block();
+         }
```
greg7mdp, line 538: I think it would be nicer to have a single loop 
```suggestion
         for( uint32_t i = 0; i < n; ++i ) {
            // for performance, only vote on the last four to move finality
            // This is 4 instead of 3 because the extra block has to be produced to log_irreversible
            if (n > 4)
               control->allow_voting(i >= n - 4);
            res = empty ? produce_empty_block() : produce_block();
         }
```

greg7mdp (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Speed up tests by avoiding unneeded BLS calculations. Part 1.
Note:end

PR #282, opened 2024-06-13
[1.0-beta2] P2P: Prevent syncing from running ahead of chain head

### LINKED ISSUE #278: P2P: Syncing into network runs ahead
```
info  2024-06-12T18:03:56.409 net-1     net_plugin.cpp:2193           operator()           ] ["xxx:9876:blk - 80a0431" - 4 x.x.x.x:9876] requesting range 70638042 to 70640541, head 70638171, lib 70494876
info  2024-06-12T18:03:56.480 nodeos    controller.cpp:3477           log_applied          ] Received block 1432512fc8f1f146... #70495000 @ 2023-04-08T19:50:23.000 signed by junglesweden [trxs: 2, lib: 70638641, net: 288, cpu: 791, elapsed: 315, time: 815, latency: 37232013480 ms]
```

Requesting blocks way ahead of HEAD.

v1.0.0-beta2.1 syncing into Jungle from genesis.

heifner:
P2P syncing was using fork database head for a limited factor on how far ahead to request blocks. With the recent change to ingest blocks quickly into the fork database (https://github.com/AntelopeIO/leap/pull/2385), this allowed the fork database to load hundred of thousands of blocks ahead of chain head.

Use chain head instead of fork database head for sync ahead calculation.

Resolves #278

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -9,7 +9,6 @@
 #include <eosio/chain/plugin_interface.hpp>
 #include <eosio/chain/thread_utils.hpp>
 #include <eosio/producer_plugin/producer_plugin.hpp>
-#include <eosio/chain/contract_types.hpp>
 
 #include <fc/bitutil.hpp>
 #include <fc/network/message_buffer.hpp>
@@ -512,10 +511,12 @@ namespace eosio {
 
 
       struct chain_info_t {
-         uint32_t      lib_num = 0;
          block_id_type lib_id;
-         uint32_t      head_num = 0;
+         uint32_t      lib_num = 0;
          block_id_type head_id;
+         uint32_t      head_num = 0;
+         block_id_type fork_head_id;
+         uint32_t      fork_head_num = 0;
       };
 
       
@@ -533,6 +534,7 @@ namespace eosio {
       chain_info_t get_chain_info() const;
       uint32_t get_chain_lib_num() const;
       uint32_t get_chain_head_num() const;
+      uint32_t get_fork_head_num() const;
 
       void on_accepted_block_header( const signed_block_ptr& block, const block_id_type& id );
       void on_accepted_block( const signed_block_ptr& block, const block_id_type& id );
@@ -1499,7 +1501,7 @@ namespace eosio {
 
    // called from connection strand
    void connection::blk_send_branch( const block_id_type& msg_head_id ) {
-      uint32_t head_num = my_impl->get_chain_head_num();
+      uint32_t head_num = my_impl->get_fork_head_num();
 
       peer_dlog(this, "head_num = ${h}",("h",head_num));
       if(head_num == 0) {
@@ -2145,11 +2147,11 @@ namespace eosio {
       auto chain_info = my_impl->get_chain_info();
 
       fc_dlog( logger, "sync_last_requested_num: ${r}, sync_next_expected_num: ${e}, sync_known_lib_num: ${k}, sync_req_span: ${s}, head: ${h}, lib: ${lib}",
-               ("r", sync_last_requested_num)("e", sync_next_expected_num)("k", sync_known_lib_num)("s", sync_req_span)("h", chain_info.head_num)("lib", chain_info.lib_num) );
+               ("r", sync_last_requested_num)("e", sync_next_expected_num)("k", sync_known_lib_num)("s", sync_req_span)("h", chain_info.fork_head_num)("lib", chain_info.lib_num) );
 
-      if( chain_info.head_num + sync_req_span < sync_last_requested_num && sync_source && sync_source->current() ) {
+      if( chain_info.fork_head_num + sync_req_span < sync_last_requested_num && sync_source && sync_source->current() ) {
          fc_dlog( logger, "ignoring request, head is ${h} last req = ${r}, sync_next_expected_num: ${e}, sync_known_lib_num: ${k}, sync_req_span: ${s}, source connection ${c}",
-                  ("h", chain_info.head_num)("r", sync_last_requested_num)("e", sync_next_expected_num)
+                  ("h", chain_info.fork_head_num)("r", sync_last_requested_num)("e", sync_next_expected_num)
                   ("k", sync_known_lib_num)("s", sync_req_span)("c", sync_source->connection_id) );
          return;
       }
@@ -2157,7 +2159,7 @@ namespace eosio {
       if (conn) {
          // p2p_high_latency_test.py test depends on this exact log statement.
          peer_ilog(conn, "Catching up with chain, our last req is ${cc}, theirs is ${t}, next expected ${n}, head ${h}",
-                   ("cc", sync_last_requested_num)("t", sync_known_lib_num)("n", sync_next_expected_num)("h", chain_info.head_num));
+                   ("cc", sync_last_requested_num)("t", sync_known_lib_num)("n", sync_next_expected_num)("h", chain_info.fork_head_num));
       }
 
       /* ----------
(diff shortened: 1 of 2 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3153,10 +3156,10 @@ namespace eosio {
          pending_message_buffer.advance_read_ptr( message_length );
          return true;
       }
-      peer_dlog( this, "received block ${num}, id ${id}..., latency: ${latency}ms, head ${h}",
+      peer_dlog( this, "received block ${num}, id ${id}..., latency: ${latency}ms, head ${h}, fdb_head ${f}",
```
heifner, line 3159: Used `fdb_head` as having `fork` in the logs can causes confusion.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -512,10 +511,12 @@ namespace eosio {
 
 
       struct chain_info_t {
-         uint32_t      lib_num = 0;
          block_id_type lib_id;
-         uint32_t      head_num = 0;
+         uint32_t      lib_num = 0;
          block_id_type head_id;
+         uint32_t      head_num = 0;
+         block_id_type fork_head_id;
```
linh2931, line 518: Why not `forkdb_head_num` and `forkdb_head_id` to be clear? Too verbose?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3287,29 +3290,35 @@ namespace eosio {
...
    // call only from main application thread
    void net_plugin_impl::update_chain_info(const block_id_type& lib) {
       controller& cc = chain_plug->chain();
-      uint32_t lib_num = 0, head_num = 0;
+      uint32_t lib_num = 0, head_num = 0, fork_head_num = 0;
       {
          fc::lock_guard g( chain_info_mtx );
          chain_info.lib_num = lib_num = block_header::num_from_id(lib);
          chain_info.lib_id = lib;
          chain_info.head_id = cc.fork_db_head_block_id();
```
linh2931, line 3314: Should `cc.fork_db_head_block_id()` be `cc.head_block_id()`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1499,7 +1501,7 @@ namespace eosio {
 
    // called from connection strand
    void connection::blk_send_branch( const block_id_type& msg_head_id ) {
-      uint32_t head_num = my_impl->get_chain_head_num();
+      uint32_t head_num = my_impl->get_fork_head_num();
```
linh2931, line 1504: Why change to `get_fork_head_num`? Several other places also change to fork_head. PR description says using chain_head.

Which change is for switching to chain_head? I could not find it.

plugins/net_plugin/net_plugin.cpp
heifner: I just went with the old original naming, see https://github.com/EOSIO/eos/blob/v1.0.0/plugins/net_plugin/net_plugin.cpp#L485
The important thing is it is a fork head as opposed to the chain head.

plugins/net_plugin/net_plugin.cpp
heifner: Thanks, not sure how I missed that!

plugins/net_plugin/net_plugin.cpp
heifner: See `sync_recv_block`.
There is now both `get_chain_head_num` and `get_fork_head_num`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2554,6 +2556,7 @@ namespace eosio {
                }
             }
 
+            // use chain head instead of fork head so we do not get too far ahead of applied blocks
```
linh2931, line 2559: Nothing change here, only a comment added?

heifner: `get_chain_head_num()` now returns `chain_head` instead of fork head.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3287,29 +3290,33 @@ namespace eosio {
...
          chain_info.lib_id = cc.last_irreversible_block_id();
          chain_info.lib_num = lib_num = block_header::num_from_id(chain_info.lib_id);
-         chain_info.head_id = cc.fork_db_head_block_id();
+         chain_info.head_id = cc.head_block_id();
          chain_info.head_num = head_num = block_header::num_from_id(chain_info.head_id);
+         chain_info.fork_head_id = cc.fork_db_head_block_id();
+         chain_info.fork_head_num = fork_head_num = block_header::num_from_id(chain_info.fork_head_id);
       }
-      fc_dlog( logger, "updating chain info lib ${lib}, head ${h}", ("lib", lib_num)("h", head_num) );
+      fc_dlog( logger, "updating chain info lib ${lib}, head ${h} fdb_head ${f}", ("lib", lib_num)("h", head_num)("f", fork_head_num) );
```
greg7mdp, line 3303: Maybe we should be consistent in log messages and use `fdb_head` when we print the fork_db head?

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4182,13 +4194,13 @@ namespace eosio {
...
       // nothing as changed since last handshake and one was sent recently, so skip sending
-      if (chain_info.head_id == hello.head_id && (hello.time + hs_delay > now))
+      if (chain_info.fork_head_id == hello.head_id && (hello.time + hs_delay > now))
          return false;
       hello.network_version = net_version_base + net_version;
       hello.last_irreversible_block_num = chain_info.lib_num;
       hello.last_irreversible_block_id = chain_info.lib_id;
-      hello.head_num = chain_info.head_num;
-      hello.head_id = chain_info.head_id;
+      hello.head_num = chain_info.fork_head_num;
```
linh2931, line 4202: So handshake sending fork_head?

heifner: Yes. You want to know the max blocks they have.

linh2931 (approved):

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1600,10 +1600,10 @@ namespace eosio {
             c->last_handshake_sent.generation = ++c->sent_handshake_count;
             auto last_handshake = c->last_handshake_sent;
             g_conn.unlock();
-            peer_ilog( c, "Sending handshake generation ${g}, lib ${lib}, head ${head}, id ${id}",
+            peer_ilog( c, "Sending handshake generation ${g}, lib ${lib}, fhead ${h}, id ${id}",
                        ("g", last_handshake.generation)
                        ("lib", last_handshake.last_irreversible_block_num)
-                       ("head", last_handshake.head_num)("id", last_handshake.head_id.str().substr(8,16)) );
+                       ("h", last_handshake.head_num)("id", last_handshake.head_id.str().substr(8,16)) );
```
greg7mdp, line 1606: should it be: `last_handshake.fork_head_num`

heifner: Changing the member name of `handshake_message` is likely ok as it is sent in binary form. It would change the FC_REFLECT variant/json format which shouldn't matter, but that is why I didn't originally change it. What do you think?

greg7mdp: Now the naming is so consistent, that I feel it would be better to update this as well, but up to you if you have any reservations at all.

heifner: I don't think I do now that I have thought through it. Seems like now is the time to make these types of changes before RC1.

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: P2P prevent syncing from running ahead o chain head, use chain head instead of fork database head for sync ahead calculation.
Note:end

PR #283, opened 2024-06-13
add test that leap3 and eosio state history log entries can be read

### LINKED ISSUE #225: ship refactor followup: unit test for pre-leap3/4 ship logs
Leap4 modified the ship log format in a non-backward compatible way. That is, once Leap4+ (or Spring1+) writes to a ship log, it is no longer readable with with Leap3. Furthermore, Leap3 modified the ship log such that it is not backward compatible with any versions of eosio.

However, Spring1 is expected to read all 3 log entry formats (eosio, leap3, leap4+). There needs to be some tests that manually craft at least a leap3 log and ensure spring1 reads it properly. I'm less concerned about the eosio format since leap3 is so similar to the eosio format, but it'd be a nice to have.

spoonincode:
Manually craft a leap 3.x structured state history log and then read it back with current log reading code. Resolves #225. Both eosio and leap3 formats are treated identical in the code so don't see much of a need to craft an eosio formatted log.

A reminder on the differences,
https://github.com/AntelopeIO/spring/blob/3f78e0f134a819f5977c1d0747fbdb141392a83a/libraries/state_history/include/eosio/state_history/log.hpp#L194-L203

### DIFF
--- a/tests/ship_log.cpp
+++ b/tests/ship_log.cpp
@@ -1173,4 +1173,52 @@ BOOST_DATA_TEST_CASE(split_forks, bdata::xrange(1u, 6u), fork_size) try {
    }
 } FC_LOG_AND_RETHROW();
 
+//(manually) fabricate a leap 3.x ship log format and make sure it's readable
+BOOST_AUTO_TEST_CASE(old_log_format) try {
+   //just opens and closes an empty log a few times
+   const temp_directory tmpdir;
+
+   const unsigned begin_block = 2;
+   const unsigned end_block = 45;
+
+   std::map<block_num_type, sha256> wrote_data_for_blocknum;
+
+   {
+      random_access_file file(tmpdir.path() / "old.log");
+      for(unsigned blocknum = begin_block; blocknum < end_block; ++blocknum) {
+         const size_t insertpos = file.size();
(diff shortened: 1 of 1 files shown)


tests/ship_log.cpp
```diff
@@ -1173,4 +1173,51 @@ BOOST_DATA_TEST_CASE(split_forks, bdata::xrange(1u, 6u), fork_size) try {
    }
 } FC_LOG_AND_RETHROW();
 
+//(manually) fabricate a leap 3.x ship log format and make sure it's readable
+BOOST_AUTO_TEST_CASE(old_log_format) try {
+   const temp_directory tmpdir;
+
+   const unsigned begin_block = 2;
+   const unsigned end_block = 45;
```
heifner, line 1181: Any reason to do 45 instead of say 5?

heifner (approved):

tests/ship_log.cpp
spoonincode: Quite arbitrary, honestly.

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add test consuming manually crated Leap3 state history log.
Note:end

PR #285, opened 2024-06-13
TestHarness Enhanced Documentation 

ericpassmore:
- added description to TestHarness python classes
- corrected error codes specified in docs
- additional overview to PerformanceHarness README 
- oneline description of python files in TestHarness 
- example call and output from `trx_generator`

### DIFF
--- a/tests/PerformanceHarness/example_trx_generator_run.md
+++ b/tests/PerformanceHarness/example_trx_generator_run.md
@@ -18,7 +18,7 @@ ${BUILD_TEST_DIR}/trx_generator/trx_generator --generator-id $GENERATORID \
 ```
 
 ## Output
-Why is this 0us 
+Why is this 0us
 ```
 info  2024-06-12T21:14:31.379 trx_gener main.cpp:128                  main                 ] Initializing accounts. Attempt to create name for purplepurple
 info  2024-06-12T21:14:31.379 trx_gener main.cpp:128                  main                 ] Initializing accounts. Attempt to create name for orangeorange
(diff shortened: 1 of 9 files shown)


tests/PerformanceHarness/example_trx_generator_run.md
```diff
@@ -0,0 +1,97 @@
...
+     --priv-keys $COMMA_SEP_KEYS \
+     --last-irreversible-block-id $LIB_ID \
+     --log-dir $TRX_LOG_DIR \
+     --peer-endpoint-type p2p \
+     --peer-endpoint $P2P_HOST \
+     --port $PEER2PEERPORT
+```
+
+## Output
+Why is this 0us
```
heifner, line 21: It is `0` because this is logging `_total_us` which is never updated. Just initialized to 0. Seems you found a bug. I don't guess anyone ever noticed.

heifner (changes requested):

tests/PerformanceHarness/example_trx_generator_run.md
ericpassmore: forgot I had that question, converted PR to draft. I'll continue to add documentation updates, and better document the why 0us. 

This example is from Jungle Spring Beta2 test net. I believe an intermediate node is filtering the submitted transactions to the P2P endpoint, and they are not making onto the blockchain. Need some additional tests to verify.

ericpassmore: Time is zero, because `_total_us` is never updated.

tests/PerformanceHarness/example_trx_generator_run.md
```diff
@@ -0,0 +1,97 @@
+
+## Command
+Sleep is needed otherwise subjective transaction may reject transactions as too soon.
+```
+CHAIN_ID=$(cleos --url $HTTP_URL get info | grep chain_id | cut -d:  -f2 | sed 's/[ ",]//g')
```
spoonincode, line 5: I'm not suggesting that you need to change this, but if you're open to using `jq` you can do this shorter via `cleos get info | jq -r .chain_id`

PR #287, opened 2024-06-13
rename debian package to 'antelope-spring' to avoid conflict

### LINKED ISSUE #213: `spring` deb package naming conflict
The main (default) Ubuntu package repo contains a package named `spring` (an RTS game interestingly enough, https://github.com/spring/spring/). It would probably be wise for us to not name our package the same as a package available through the main Ubuntu package repo. (I'd be much less worried about clashing with a third party repo)

Besides "bad form", there may be ways to encounter some unexpected behavior because the spring RTS package is version 106.0 ('newer' than our spring 1.0). Something like
```
apt install ./spring_1.0.0_amd64.deb
apt-get upgrade
```
will not disturb our package, but either a `apt-get --with-new-pkgs upgrade` or `apt-get dist-upgrade` with both remove our spring package and install the RTS game. I'm not significantly familiar with the wide variety of ways to manage packages and upgrades in the Debian & Ubuntu world so there may be other traps we're laying as well.

I think maybe we should use something like `antelope-spring` for our package name instead.

spoonincode:
Renames the Debian package from `spring` to `antelope-spring` to avoid clashing with an already named `spring` package in Debian & Ubuntu standard package repo. Resolves #213

Unfortunately, will need to change both cdt & reference-contracts after merging this to update the expected package name.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -101,12 +101,12 @@ jobs:
           zstdcat build.tar.zst | tar x
           cd build
           cpack
-          ../tools/tweak-deb.sh spring_*.deb
+          ../tools/tweak-deb.sh antelope-spring_*.deb
       - name: Install dev package
         if: matrix.platform != 'reproducible'
         run: |
           apt-get update && apt-get upgrade -y
-          apt-get install -y ./build/spring_*.deb ./build/spring-dev*.deb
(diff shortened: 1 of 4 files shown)


heifner (approved):

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: CHORE
summary: Rename deb package to antelope-spring.
Note:end

PR #288, opened 2024-06-13
change `experimental-binaries` to `antelope-spring-experimental-binaries` to avoid version overlap with leap packages

### LINKED ISSUE #50: `experimental-binaries` version roll back tag conflict

spoonincode:
Since we're going to have to change downstream users of the dev package anyways due to #287, might as well make this change now too so we can resolve #50

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,7 +46,7 @@ jobs:
           password: ${{github.token}}
       - run: echo "REPOSITORY_OWNER_LOWER=${GITHUB_REPOSITORY_OWNER,,}" >> "${GITHUB_ENV}"
       - name: Build and push antelope-spring-experimental-binaries
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v5
         with:
           push: true
           tags: ghcr.io/${{env.REPOSITORY_OWNER_LOWER}}/antelope-spring-experimental-binaries:${{github.ref_name}}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -49,5 +49,5 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           push: true
-          tags: ghcr.io/${{env.REPOSITORY_OWNER_LOWER}}/experimental-binaries:${{github.ref_name}}
+          tags: ghcr.io/${{env.REPOSITORY_OWNER_LOWER}}/spring-experimental-binaries:${{github.ref_name}}
```
heifner, line 52: Should this be `antelope-spring-experimental-binaries` to match `antelope-spring` name?

spoonincode: I could go either way on this.. `antelope-spring` matches the names of the packages that are placed in here, but `spring` is the name of the repo.

I think maybe since we'll use this string along with other strings that are `antelope-spring` (the package name and the artifact name),
https://github.com/AntelopeIO/cdt/blob/9ecdca0f080b65e1339318cd2451a5f2593c4ee5/.github/workflows/build.yaml#L111-L115
it probably does make sense to make it `antelope-spring` just to keep everything looking similar in that block above. I'll update in a moment.

greg7mdp (approved):

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: CHORE
summary: More specific name for experimental binaries to avoid naming conflict starting at version 3.1.
Note:end

PR #301, opened 2024-06-16
[1.0-beta3] clear state history's finality log if savanna transition is forked out

### LINKED ISSUE #289: a fork removing the start of savanna transition will cause a state history failure with `finality-data-history` enabled
Notice how ship plugin writes to the finality data log,
https://github.com/AntelopeIO/spring/blob/391be54bc06b9690d0f2753e59eb9d791ed429d4/plugins/state_history_plugin/state_history_plugin.cpp#L189-L200

If a `finality_data_t` is returned, it is written to the finality data log. Otherwise, nothing is done with the log.

Remember though that ship log entries must be continuous: there can be no gaps in them.

Consider that savanna transition begins on block 200, and continues for another block 201. This will write two entries in to ship's finality data log: block 200 and 201. Now consider that a new fork is switched to back on block 198, and that this fork continues without savanna activation; say it continues on to block 220.

At this point, ship will continue to indicate a finality data log of blocks 200 through 201 and even return finality data for those blocks even though that never happened from the standpoint of the current fork being followed.

But then consider on block 221 there is another attempt to transition to savanna. This will cause ship to attempt to write to block 221 of the finality log, but this will be flagged as a discontinuity in ship due to the gap between 201 and 221. Ultimately this will cause the node to shut down.

spoonincode:
Avoid state history finality log failure and node shutdown if savanna transition is forked out. Resolves #289. This is a fairly simple change that erases all finality data log files and reopens a fresh head log when savanna finality data is removed (which can only happen upon the transition being forked out).

An even simpler alternative would be to always write the `std::optional<finality_data_t>` out, which we could do instead, but that puts a little more burden on consumers.

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log_catalog.hpp
+++ b/libraries/state_history/include/eosio/state_history/log_catalog.hpp
@@ -156,6 +156,16 @@ public:
       return first == second;
    }
 
+   void clear() {
+      if(empty())
+         return;
+
+      while(!retained_log_files.empty())
+         delete_bundle(retained_log_files.extract(retained_log_files.begin()).value().path_and_basename);
+      delete_head_log();
+      open_head_log();
+   }
+
 private:
    template<typename F>
    typename std::invoke_result_t<F,state_history_log&&> call_for_log(const uint32_t block_num, F&& f) {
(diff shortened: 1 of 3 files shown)


greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Clear out the state history log when Savanna transition is forked out.
Note:end

PR #303, opened 2024-06-17
Liveness rule only valid for the same finalizer policy generation

### LINKED ISSUE #293: Liveness rule in decide_vote is too permissive
Liveness rule should only apply for the same finalizer policy generation.

`finality_core` should be updated to store the finalizer policy generation number alongside the timestamp for each block tracked within `finality_core`.

In `decide_vote`, when evaluating the liveness rule, the active finalizer policy generation number for the block that is being considered for a vote should be compared to the generation number stored within `finality_core` for the block referenced by `latest_qc_block_num`. If those generation numbers are not the same, then the liveness rule is automatically not satisfied regardless of how the timestamps compare.

heifner:
- Liveness check
  - Same finalizer policy generation and lock_block_timestamp < latest_qc_claim_block_timestamp
    - When the policy changes, say from generation 1 to 2, the new finalizers in policy generation 2 could run into a situation where they lock on a block (which may turn out to be finalized later) in one fork branch but then weak vote on another block in a conflicting fork branch. Then by building off of that block with a weak QC, they could end up finalizing it as well, causing a finality violation. They could end up in this situation without breaking any of the rules. That is because the liveness condition to decide to carry out the weak vote would be satisfied not by a QC that they participated in, but one formed by the finalizers of policy generation 1. And those policy generation 1 finalizers wouldn’t have done anything wrong either. Therefore, include the finalizer policy generation in the liveness check unless the next condition is met.
   - OR lock_block_timestamp <= last_final_block_timestamp
     - This can be true when a finalizer that was previously active becomes active again. This condition allows the newly active finalizer to satisfy liveness even when the finalizer policy generation does not match.

Resolves #293


libraries/chain/block_header_state.cpp
```diff
@@ -188,9 +189,13 @@ void finish_next(const block_header_state& prev,
          while (it != prev.finalizer_policies.end() && it->first <= lib) {
             const finalizer_policy_tracker& tracker = it->second;
             if (tracker.state == finalizer_policy_tracker::state_t::pending) {
-               // new finalizer_policy becones active
+               // new finalizer_policy becomes active
                // -----------------------------------
                next_header_state.active_finalizer_policy.reset(new finalizer_policy(*tracker.policy));
+               // new finalizer policy generation
+               parent_block.finalizer_policy_generation = next_header_state.active_finalizer_policy->generation;
```
greg7mdp, line 196: Why do we update the `finalizer_policy_generation` in `parent_block`?

heifner: Because we use it on the next line to calculate the next core.

heifner: Calculating the core again is less than ideal, but seemed like the best option. Alternatively we could modify the core directly to update the `finalizer_policy_generation`, but re-calculating the next core seemed better than making the core mutable.

libraries/chain/finality/finalizer.cpp
```diff
@@ -68,12 +76,12 @@ finalizer::vote_result finalizer::decide_vote(const block_state_ptr& bsp) {
...
       }
 
-      fsi.last_vote             = { bsp->id(), bsp->timestamp() };
+      fsi.last_vote             = { bsp->id(), bsp->timestamp(), bsp->core.latest_qc_block_finalizer_policy_generation() };
       fsi.last_vote_range_start = p_start;
 
       auto& final_on_strong_qc_block_ref = bsp->core.get_block_reference(bsp->core.final_on_strong_qc_block_num);
       if (voting_strong && final_on_strong_qc_block_ref.timestamp > fsi.lock.timestamp) {
-         fsi.lock = { final_on_strong_qc_block_ref.block_id, final_on_strong_qc_block_ref.timestamp };
+         fsi.lock = { final_on_strong_qc_block_ref.block_id, final_on_strong_qc_block_ref.timestamp, final_on_strong_qc_block_ref.finalizer_policy_generation };
```
greg7mdp, line 84:
```suggestion
         fsi.lock = final_on_strong_qc_block_ref;
```

libraries/chain/block_header_state.cpp
greg7mdp: I don't have a problem with caIculating the core again.
I don't see where `finalizer_policy_generation` is used in `finality_core::next()`. I must be missing something.

heifner: https://github.com/AntelopeIO/spring/blob/885a574013ac5617a69c8ce44fa11cda4e18ebe4/libraries/chain/finality/finality_core.cpp#L359-L359

greg7mdp: Thanks!
But that was not the active policy for **that** block which refers to the parent block!

heifner: You are correct, this is not needed.

libraries/chain/finality/finality_core.cpp
```diff
@@ -262,6 +286,7 @@ finality_core finality_core::next(const block_ref& current_block, const qc_claim
    
    assert(refs.empty() || (refs.back().block_num() + 1 == current_block.block_num())); // Satisfied by precondition 2.
    assert(refs.empty() || (refs.back().timestamp < current_block.timestamp)); // Satisfied by precondition 2.
+   assert(refs.empty() || (refs.back().finalizer_policy_generation <= current_block.finalizer_policy_generation)); // Satisfied by precondition 2.
```
greg7mdp, line 289: I think we could have:
```suggestion
   assert(refs.empty() || (current_block.finalizer_policy_generation - refs.back().finalizer_policy_generation <= 1)); // Satisfied by precondition 2.
```

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Scope liveness rule to the same finalizer generation. 
Note:end

PR #315, opened 2024-06-18
[1.0-beta2] P2P: Set peer_syncing_from_us to false when a recent block is received

### LINKED ISSUE #302: Test failure: lib_advance_if_test
https://github.com/AntelopeIO/spring/actions/runs/9547360011/job/26312762056

```
  File "/__w/spring/spring/build/tests/lib_advance_test.py", line 102, in <module>
    assert prodA.waitForNextBlock(timeout=6), "Production node A should continue to advance, even after bridge node is killed"
AssertionError: Production node A should continue to advance, even after bridge node is killed
```

heifner:
A node will set `peer_syncing_from_us` to `false` on a `request_message` of `none` or if it receives a handshake indicating it is synced. This PR sets `peer_syncing_from_us` to `false` if a node receive a recent block from the network as blocks are only relayed after they are validated. Therefore, if we receive a recent block from a peer we know that peer can't be syncing from us even if they have failed to send us a handshake letting us know they are up to date.

Resolves #302

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1064,7 +1064,6 @@ namespace eosio {
       void blk_send_branch( const block_id_type& msg_head_id );
       void blk_send_branch( uint32_t msg_head_num, uint32_t lib_num, uint32_t fork_head_num );
       void blk_send(const block_id_type& blkid);
-      void stop_send();
 
       void enqueue( const net_message &msg );
       size_t enqueue_block( const signed_block_ptr& sb, bool to_sync_queue = false);
@@ -1582,10 +1581,6 @@ namespace eosio {
       } catch( ... ) {
          peer_elog( this, "caught other exception fetching block id ${id}", ("id", blkid) );
       }
-   }
-
-   void connection::stop_send() {
-      peer_syncing_from_us = false;
    }
 
    void connection::send_handshake() {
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2472,13 +2467,17 @@ namespace eosio {
...
                 ("d", blk_applied ? "applied" : "got")("bn", blk_num)("id", blk_id.str().substr(8,16))
-                ("l", blk_latency == fc::microseconds::maximum() ? 0 : blk_latency.count()/1000) );
+                ("l", blk_latency.count()/1000) );
       if( app().is_quiting() ) {
          c->close( false, true );
          return;
       }
       c->latest_blk_time = std::chrono::system_clock::now();
       c->block_status_monitor_.accepted();
+      if (blk_latency.count() < config::block_interval_us && c->peer_syncing_from_us) {
```
linh2931, line 2477: You probably don't need `&& c->peer_syncing_from_us`.

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Add cases where we update peer syncing status to more accurately reflect peer syncing from us status.
Note:end

PR #324, opened 2024-06-27
IBC Contract bitset tests

systemzax:
This PR provides basic constructor coverage for the smart contract bitset implementation , which is used in IBC and finality violation proofs contract.


unittests/svnn_ibc_tests.cpp
```diff
@@ -18,6 +18,46 @@ using namespace eosio::testing;
 
 using mvo = mutable_variant_object;
 
+std::string binary_to_hex(const std::string& bin) {
+   size_t pad_bin_length = (4 - (bin.length() % 4)) % 4;
```
linh2931, line 22: Is the last `% 4` needed? `(4 - (bin.length() % 4))` will be less than 4.

unittests/svnn_ibc_tests.cpp
```diff
@@ -18,6 +18,46 @@ using namespace eosio::testing;
...
+    std::string padded_bin = std::string(pad_bin_length, '0') + bin;
+    std::stringstream hex_stream;
+
+    for (size_t i = 0; i < padded_bin.length(); i += 4) {
+        std::string segment = padded_bin.substr(i, 4);
+        int value = std::stoi(segment, nullptr, 2);
+        hex_stream << std::hex << value;
+    }
+
+    size_t pad_out_length = (2 - (hex_stream.str().length() % 2)) % 2;
```
linh2931, line 33: The same comment as above.

unittests/svnn_ibc_tests.cpp
```diff
@@ -18,6 +18,46 @@ using namespace eosio::testing;
...
+    }
+
+    return reversed_hex;
+
+}
+
+std::string bitset_to_binary(const boost::dynamic_bitset<unsigned char>& bitset) {
+   std::string result;
+   result.reserve(bitset.size());
+   for (std::size_t i = bitset.size(); i > 0; --i) {
```
linh2931, line 50: You can start from `bitset.size() - 1`. Then you can use `bitset[i]` instead of `bitset[i - 1]` below.

Will `bitset` never be empty?

unittests/svnn_ibc_tests.cpp
```diff
@@ -575,4 +607,57 @@ BOOST_AUTO_TEST_SUITE(svnn_ibc)
...
+      std::string bitset_6 = binary_to_hex("000000111111111111111");
+
+      chain.push_action("ibc"_n, "testbitset"_n, "ibc"_n, mvo()
+         ("bitset_string", "00")
+         ("bitset_vector", bitset_1)
+         ("finalizers_count", 1)
+      );
+
+      chain.push_action("ibc"_n, "testbitset"_n, "ibc"_n, mvo()
+         ("bitset_string", "30")
```
linh2931, line 632: Maybe you can add a comment what those parameters are testing, like "30" is the reverse of bitset_2 or something.

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Tests for the smart contract bitset implementation , which is used in IBC and finality violation proofs contract.
Note:end

unittests/svnn_ibc_tests.cpp
greg7mdp: But then you have to be careful to not use `i >= 0` which ia always true for an unsigned like `std::size_t`

unittests/svnn_ibc_tests.cpp
```diff
@@ -18,6 +18,46 @@ using namespace eosio::testing;
...
+   for (std::size_t i = bitset.size(); i > 0; --i) {
+       result += bitset[i - 1] ? '1' : '0';
+   }
+   return result;
+}
+
+std::string bitset_to_input_string(const boost::dynamic_bitset<unsigned char>& bitset) {
+   std::string result = bitset_to_binary(bitset);
+   return binary_to_hex(result);
+}
```
greg7mdp, line 59: Probably it doesn't matter, but that's a pretty inefficient way to implement `bitset_to_input_string`. For fun, I implemented a version which converts directly from the bitset to an hex string (and I checked it provides exactly the same result as the existing version): 

```
std::string bitset_to_input_string(const boost::dynamic_bitset<unsigned char>& bitset) {
   static const char* hexchar = "0123456789abcdef";

   boost::dynamic_bitset<unsigned char> bs(bitset);
   bs.resize((bs.size() + 7) & ~0x7);
   assert(bs.size() % 8 == 0);

   std::string result;
   result.resize(bs.size() / 4);
   for (size_t i = 0; i < bs.size(); i += 4) {
      size_t x = 0;
      for (size_t j = 0; j < 4; ++j)
         x += bs[i+j] << j;
      auto slot = i / 4;
      result[slot % 2 ? slot - 1 : slot + 1] = hexchar[x]; // flip the two hex digits for each byte
   }
   return result;
}
```

unittests/svnn_ibc_tests.cpp
```diff
@@ -18,6 +18,46 @@ using namespace eosio::testing;
...
+    return reversed_hex;
+
+}
+
+std::string bitset_to_binary(const boost::dynamic_bitset<unsigned char>& bitset) {
+   std::string result;
+   result.reserve(bitset.size());
+   for (std::size_t i = bitset.size(); i > 0; --i) {
+       result += bitset[i - 1] ? '1' : '0';
+   }
```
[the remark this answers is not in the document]
greg7mdp: PS: I verified it does the exact same thing in the test with:

```
      for (size_t sz=0; sz<100; ++sz) {
         boost::dynamic_bitset<unsigned char> bs;
         bs.resize(sz);
         for (size_t i=0; i<sz; ++i)
            bs[i] = rand() % 2;
         std::cout << bs << '\n';
         BOOST_REQUIRE_EQUAL(bitset_to_binary(bs), bitset_to_binary2(bs));
         BOOST_REQUIRE_EQUAL(bitset_to_input_string(bs), bitset_to_input_string2(bs));
      }
```

greg7mdp (approved):

linh2931 (approved):

PR #325, opened 2024-06-27
Report forkdb root for lib instead of calculated lib

### LINKED ISSUE #319: LIB regression on Savanna transition
image

As shown in the screenshot above, in some cases, the last irreversible block (as reported in nodeos logs) regresses  upon Savanna activation.

To replicate :

3x producer nodes on different machines, each configured with a producer key and finalizer key
3x P2P nodes on different machines
1x ABP (appointed block producer) on another machine, running eosio account, used to boot the chain, create and elect BPs, regfinkey and activate savanna consensus

Producer nodes are each connected to their respective P2P node, and the P2P nodes are connected between each other.

ABP node is launched, connecting to one of the P2P node, and starts producing blocks.

Boot script is executed, deploying system contracts, issuing initial tokens, initializing the system contract and activating protocol features.

Producer / finalizer accounts are created, and chain is activated (vote with 150m tokens for the 3 block producers).

Immediately after (without waiting for LIB to advance beyond the eosio handoff to BPs), finalizer keys are registered and svnn consensus is activated.

heifner:
To avoid confusion, report actual LIB (the root of the fork database) instead of the calculated LIB.

Resolves #319

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1406,12 +1406,6 @@ struct controller_impl {
       }
    }
 
-   block_num_type latest_known_lib_num() const {
-      block_id_type irreversible_block_id = if_irreversible_block_id.load();
-      block_num_type savanna_lib_num = block_header::num_from_id(irreversible_block_id);
-      return savanna_lib_num > 0 ? savanna_lib_num : fork_db_head_irreversible_blocknum();
-   }
-
    void log_irreversible() {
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Improve reporting of block number by reporting the actual LIB. Previously reported a calculated LIB block number. 
Note: end

PR #327, opened 2024-06-27
Test: SHiP test: make test more reliable

### LINKED ISSUE #321: Test failure: ship_streamer_test
From Matt:

> ship client outputs look sane afaict. The test error is ERROR: Bridge did not shutdown which means that a nodeos that was expected to shutdown didn't. But the nodeos in question log says nodeos successfully exiting. That message is pretty late — after appbase is destroyed even. It's certainly possible something got wedged afterwards, but I'd also be suspect of any sort of race in this aspect of the test (what if nodeos just happens to be slow to shutdown, for example)

https://github.com/AntelopeIO/spring/actions/runs/9625928471/job/26551439097?pr=300

heifner:
Make test more reliable by waiting on next producer (`c`) to produce instead of only waiting on existing producer to produce. By waiting for producer `c`, this gives plenty of time for the non producing node to see producer `b` and shutdown.

Resolves #321

### DIFF
--- a/tests/ship_streamer_test.py
+++ b/tests/ship_streamer_test.py
@@ -159,17 +159,15 @@ try:
         Print(f"Client {i} started, Ship node head is: {shipNode.getBlockNum()}")
 
     # Generate a fork
+    nonProdNode.waitForProducer("defproducera")
     prodNode3Prod= "defproducerd"
     preKillBlockNum=nonProdNode.getBlockNum()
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve reliability of SHiP tests in continuous integration testing.
Note:end

PR #329, opened 2024-06-27
Test: P2P throttle test: Wait for begin block before measuring sync

### LINKED ISSUE #326: Test Failure: p2p_sync_throttle_test
https://github.com/AntelopeIO/spring/actions/runs/9699528062/job/26769268659?pr=325#step:4:77519

heifner:
Before measuring the throttled node sync throttling make sure it has the begin block.

Resolves #326

### DIFF
--- a/tests/p2p_sync_throttle_test.py
+++ b/tests/p2p_sync_throttle_test.py
@@ -58,7 +58,7 @@ try:
 
     Print("Stand up cluster")
     extraNodeosArgs = '--plugin eosio::prometheus_plugin --connection-cleanup-period 3'
-    # Custom topology is a line of singlely connected nodes from highest node number in sequence to lowest,
+    # Custom topology is a line of singly connected nodes from the highest node number in sequence to lowest,
     # the reverse of the usual TestHarness line topology.
     if cluster.launch(pnodes=pnodes, unstartedNodes=2, totalNodes=total_nodes, prodCount=prod_count, 
                       topo='./tests/p2p_sync_throttle_test_shape.json', delay=delay, activateIF=activateIF,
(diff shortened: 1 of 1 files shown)


greg7mdp (changes requested):
This will be executed only if __debug__ is true. Is this always the case?
see [this](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)

heifner:
> This will be executed only if **debug** is true. Is this always the case? see [this](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)

> variable [__debug__](https://docs.python.org/3/library/constants.html#debug__) is True under normal circumstances, False when optimization is requested (command line option [-O](https://docs.python.org/3/using/cmdline.html#cmdoption-O))

linh2931 (approved):

greg7mdp:
> variable [**debug**](https://docs.python.org/3/library/constants.html#debug__) is True under normal circumstances, False when optimization is requested (command line option [-O](https://docs.python.org/3/using/cmdline.html#cmdoption-O))

So we probably shouldn't use assert as the wait will not be executed if `-O` is used.

heifner:
> So we probably shouldn't use assert as the wait will not be executed if `-O` is used.

We use assert in many integration tests. If you think that, then should create an issue to update all of them.

greg7mdp:
> We use assert in many integration tests. If you think that, then should create an issue to update all of them.

fair enough, will do.

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve measurement of sync by waiting for begin block.
Note:end

PR #335, opened 2024-07-01
Update fork choice rules to use timestamp instead of block_num.

greg7mdp:
As the Savanna consensus algorithm and finalizer voting safety rules are based mostly on timestamps rather than block numbers, it was decided to update the fork choice rules to use timestamps rather than block numbers.

This PR updates the fork choice rules to use:

- `last_final_block_timestamp` instead of `last_final_block_num`
- `latest_qc_block_timestamp` instead of `latest_qc_block_num`

`block_state::latest_qc_block_num` used to be named `block_state::last_qc_block_num`, but was renamed to match the name in `finality_core`.


heifner (approved):

libraries/chain/fork_database.cpp
```diff
@@ -56,8 +56,8 @@ namespace eosio::chain {
 
    // match comparison of by_best_branch
    bool first_preferred( const block_state& lhs, const block_state& rhs ) {
-      return std::make_tuple(lhs.last_final_block_num(), lhs.last_qc_block_num(), lhs.timestamp()) >
-             std::make_tuple(rhs.last_final_block_num(), rhs.last_qc_block_num(), rhs.timestamp());
+      return std::make_tuple(lhs.last_final_block_num(), lhs.latest_qc_block_num(), lhs.timestamp()) >
+             std::make_tuple(rhs.last_final_block_num(), rhs.latest_qc_block_num(), rhs.timestamp());
```
arhag, line 60: Why is this still using `last_final_block_num()` and `latest_qc_block_num()` when the index for the `by_best_branch` tag has changed?

greg7mdp: OK, thanks for the catch, fixing it and also [log_fork_comparison](https://github.com/AntelopeIO/spring/blob/e9a3c4ae6b6d89d702a1763a651ddb0eb8133065/libraries/chain/fork_database.cpp#L35-L42).

I wonder why `first_preferred` doesn't have the `id` as a final criteria. I think we should add it (switching to `sha256_greater` for the `by_best_branch` type.

greg7mdp: After I updated `first_preferred to use `timestamp`, I hit  an assert in `finality_core`:

image

This happens when calling `mark_valid` at line 129 below:

https://github.com/AntelopeIO/spring/blob/b48c9349a40349a16512fbe74575aceeb478dfe9/unittests/fork_db_tests.cpp#L127-L136

`first_preferred` calls `last_final_block_timestamp`, which calles `get_block_reference`:

https://github.com/AntelopeIO/spring/blob/b48c9349a40349a16512fbe74575aceeb478dfe9/libraries/chain/finality/finality_core.cpp#L69-L72

and the assert fails in  `get_block_reference`, both `block_num` and `current_block_num()` values are 10.

greg7mdp: Actually, the `refs` vector is empty. This is most likely a test issue. The `block_state` are manually constructed and no qc_claim is received to advance finality. 
We have only on qc_link in the vector, which is `{.source_block_num=10, ,target_block_num=10, .is_link_strong=false}`

greg7mdp: I think that we may need to create a valid `ref` for the genesis block in `finality_core::create_core_for_genesis_block`, so maybe it should take a `block_state_ptr`, or `id + timestamp + finalizer_policy_generation`

libraries/chain/include/eosio/chain/finality/finality_core.hpp
```diff
@@ -33,6 +33,7 @@ struct qc_claim_t
    block_num_type  block_num {0};
    bool            is_strong_qc {false};
 
+   // [greg] I think we should use only `block_num` in the comparison
```
arhag, line 36: I don't think so. We want a strong QC claim on a block to be able to override a weak QC claim on the same block. This could be possible if after putting the weak QC in a block, a few extra strong votes arrive on a pending QC of a block; then the next block has the opportunity to provide the strong QC assuming it doesn't have an even better QC available by then.

arhag (approved):

libraries/chain/include/eosio/chain/finality/finality_core.hpp
greg7mdp: Thanks, seems right, I'm not sure what I was thinking.

PR #337, opened 2024-07-02
fix `Boost::unit_test_framework` linkage -- don't mix header & static library usage variants

spoonincode:
Boost unit test framework can be used via header-only or by linking to a static library (or dynamic library). See https://www.boost.org/doc/libs/1_85_0/libs/test/doc/html/boost_test/usage_variants.html. Mixing these up can cause linker failures and/or ODR violations. If this sounds familiar, we previously had a mix ups fixed in AntelopeIO/leap#807. It appears we've mixed these up again in leap5 with the boost submodularization.

For example, in fc,
https://github.com/AntelopeIO/spring/blob/09d979b0e250b4bdefc693aeaee471b1dbd112f5/libraries/libfc/CMakeLists.txt#L105-L106
The `Boost::unit_test_framework` target will link the Boost unit test static library. But then in `test_fc` the header-only usage variant is used,
https://github.com/AntelopeIO/spring/blob/09d979b0e250b4bdefc693aeaee471b1dbd112f5/libraries/libfc/test/main.cpp#L1-L2

(of course, libfc linking to `Boost::unit_test_framework` is also undesirable on its own -- there is no reason `nodeos` should be linking to boost unit test framework)

This PR removes the `Boost::unit_test_framework` linkage from a number of base libraries -- appbase, fc, and chainbase -- and then adds the appropriate header-only `Boost::included_unit_test_framework` target where it is used. (curiously, `chainbase_test` appears to be the only remaining test leveraging the static library approach. I've left it as is, there is nothing wrong with that.)

Requires,
* AntelopeIO/appbase#33
* AntelopeIO/chainbase#49

### DIFF
--- a/benchmark/bls.cpp
+++ b/benchmark/bls.cpp
@@ -6,6 +6,13 @@
 #include <bls12-381/bls12-381.hpp>
 #include <random>
 
+//This program isn't unit tests. But libtester, because of the way it depends on boost test, will ultimately require implementation of
+// various boost test detail/impl bits that are only implemented by including a boost test usage variant. The most obvious reason for this
+// is because libtester inspects the master_test_suite, but the master_test_suite is only defined after a usage variant is used.
+#define BOOST_TEST_MODULE benchmark dummy module
+#define BOOST_TEST_NO_MAIN
+#include <boost/test/included/unit_test.hpp>
+
 using namespace eosio;
 using namespace eosio::chain;
 using namespace eosio::testing;
--- a/libraries/custom_appbase/tests/CMakeLists.txt
+++ b/libraries/custom_appbase/tests/CMakeLists.txt
@@ -7,7 +7,7 @@ endif()
 
 file(GLOB UNIT_TESTS "*.cpp")
 add_executable( custom_appbase_test ${UNIT_TESTS} )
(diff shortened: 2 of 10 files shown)


benchmark/bls.cpp
```diff
@@ -6,6 +6,13 @@
 #include <bls12-381/bls12-381.hpp>
 #include <random>
 
+//This program isn't unit tests. But libtester, because of the way it depends on boost test, will ultimately require implementation of
+// various boost test detail/impl bits that are only implemented by including a boost test usage variant. The most obvious reason for this
+// is because libtester inspects the master_test_suite, but the master_test_suite is only defined after a usage variant is used.
+#define BOOST_TEST_MODULE benchmark dummy module
+#define BOOST_TEST_NO_MAIN
+#include <boost/test/included/unit_test.hpp>
+
```
spoonincode, line 15: This is ugly but with the way libtester is currently written, anything it is linked to needs to have various boost unit test symbols defined.. and this is the best way without further refactor afaict

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Consistently include Boost Test Framework, linking to library. 
Note: end

Issue #332, opened 2024-06-27
integration tests: do not use `assert` for test required side effects, such as waits.


greg7mdp:
For example when the test is run with the `-O` option, the following `cluster.getNode(2).waitForBlock`:

https://github.com/AntelopeIO/spring/blob/a9e4dcfbca2464a255fc453341034df48d193554/tests/block_log_util_test.py#L154

will not be executed.

PR #339, opened 2024-07-02
Test: Finality Status: Transaction can be forked out multiple times

### LINKED ISSUE #338: Test failure: trx_finality_status_forked_if_test
I'm not sure if this failure is aggravated by changes for #333. But does appear to be an issue with the test and not spring.

```
  File "/__w/spring/spring/build/tests/trx_finality_status_forked_test.py", line 250, in <module>
    assert state == irreversibleState, \
AssertionError: ERROR: getTransactionStatus didn't return "IRREVERSIBLE" state.
```

https://github.com/AntelopeIO/spring/actions/runs/9766033963/job/26958589642

heifner:
The transaction finality status forked test allowed for a transaction to be forked out more than once, but not more than twice. The test failed as it was forked out four times.

Resolves #338

### DIFF
--- a/tests/trx_finality_status_forked_test.py
+++ b/tests/trx_finality_status_forked_test.py
@@ -183,6 +183,7 @@ try:
         blockNum = getBlockNum(retStatus)
         if state == forkedOutState or ( info['head_block_producer'] == 'defproducerd' and info['last_irreversible_block_num'] > blockNum ):
             break
+        prodD.waitForNextBlock()
 
     if state == irreversibleState:
         Print(f"Transaction became irreversible before it could be found forked out: {json.dumps(retStatus, indent=1)}")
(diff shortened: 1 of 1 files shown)


tests/trx_finality_status_forked_test.py
```diff
@@ -183,6 +183,7 @@ def getBlockID(status):
         blockNum = getBlockNum(retStatus)
         if state == forkedOutState or ( info['head_block_producer'] == 'defproducerd' and info['last_irreversible_block_num'] > blockNum ):
             break
+        prodD.waitForNextBlock()
```
heifner, line 186: This was not needed, but just noticed the test was spamming the node with requests. A sleep would have worked here as well.

greg7mdp (approved):

spoonincode (approved):

tests/trx_finality_status_forked_test.py
```diff
@@ -228,11 +229,16 @@ def getBlockID(status):
...
-    retStatus = prodD.getTransactionStatus(transId)
-    if afterForkBlockId != getBlockID(retStatus): # might have been forked out, if so wait for new block to become LIB
-        assert prodD.waitForBlock(getBlockNum(retStatus), timeout=120, blockType=BlockType.lib), \
+    while True: # might have been forked out, if so wait for new block to become LIB
+        retStatus = prodD.getTransactionStatus(transId)
+        currentBlockId = getBlockID(retStatus)
+        currentBlockNum = getBlockNum(retStatus)
+        if afterForkBlockId == currentBlockId:
+            break
+        assert prodD.waitForBlock(currentBlockNum, timeout=120, blockType=BlockType.lib), \
```
spoonincode, line 238: I can't tell.. is this one of those things we want to avoid (#332)?

heifner: Yes, according to #332, but it is pervasive in our tests. I think when #332 is worked we need to find an easy to use alternative. Likely we will just go with: `if not blah: exitError("blah")` but we might add our own `assert` function.

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve transaction finality fork test by supporting more than 2 forks.
Note:end

PR #341, opened 2024-07-03
fix instance of missed `antelope-spring` package rename in `reproducible.Dockerfile`

spoonincode:
This small Dockerfile stage is only used during manual building of the reproducible builds -- not in CI -- and was missed during #287


greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: RELEASE
category: CHORE
summary: Fixed missing antelope-spring package rename in reproducible Dockerfile.
Note:end

Issue #334, opened 2024-06-29
Reject received QCs if the bitset size(s) is not the same as the number of finalizers for the policy it refers to.


greg7mdp:
Also add test cases.

![image](https://github.com/AntelopeIO/spring/assets/1496907/2e66d9ac-93b9-4c37-a953-fbd78bc8b7a1)

PR #346, opened 2024-07-08
Reject a QC if bitset size of its strong_votes or weak_votes is invalid

linh2931:
If the bitset size of strong_votes or weak_votes in q QC does not match the number of finalizers for the factive finalizer policy the QC refers to, reject the QC.

This PR also adds  tests for the validation.

Resolves https://github.com/AntelopeIO/spring/issues/334

### DIFF
--- a/unittests/block_state_tests.cpp
+++ b/unittests/block_state_tests.cpp
@@ -304,6 +304,44 @@ BOOST_AUTO_TEST_CASE(verify_qc_test) try {
       BOOST_CHECK_EXCEPTION( bsp->verify_qc(qc), block_validate_exception, eosio::testing::fc_exception_message_starts_with("weak quorum is not met") );
    }
 
+   {  // strong QC bitset size does not match number of finalizers in the policy
+
+      // construct vote bitset with a size greater than num_finalizers
+      vote_bitset strong_votes(num_finalizers + 1);
+
+      // vote by finalizer 0
+      strong_votes[0] = 1;
+
+      // aggregate votes
+      bls_aggregate_signature agg_sig;
+      bls_signature sig = private_key[0].sign(strong_digest.to_uint8_span());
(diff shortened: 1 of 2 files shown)


heifner (approved):

unittests/block_state_tests.cpp
```diff
@@ -304,6 +304,44 @@ BOOST_AUTO_TEST_CASE(verify_qc_test) try {
...
+
+      // aggregate votes
+      bls_aggregate_signature agg_sig;
+      bls_signature sig = private_key[0].sign(strong_digest.to_uint8_span());
+      agg_sig.aggregate(sig);
+
+      // create a valid_quorum_certificate
+      valid_quorum_certificate qc(strong_votes, {}, agg_sig);
+
+      BOOST_CHECK_EXCEPTION( bsp->verify_qc(qc), block_validate_exception, eosio::testing::fc_exception_message_starts_with("vote bitset size is not the same as the number of finalizers") );
```
greg7mdp, line 323: Why is the exception not `invalid_qc_claim` that is trigerred by the `EOS_ASSERT`?

greg7mdp (approved):

unittests/block_state_tests.cpp
linh2931: Thanks.

`invalid_qc_claim` is derived from `block_validate_exception`: https://github.com/AntelopeIO/spring/blob/d361a93bb548c8d79996b37e447569e95f9b1159/libraries/chain/include/eosio/chain/exceptions.hpp#L258.
Used `block_validate_exception` for keeping consistent with existing usage.

But it is better to be more specific. Changed all of the usages.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Validate QC by bitsize of votes.
Note:end

PR #347, opened 2024-07-09
Refactor fork database

### LINKED ISSUE #333: Remove distinction between fork_db pending_head and head; persist if_irreversible_block_id in fork_db; add fork_db_pending_lib_num
On startup, when loading the fork database from disk, nodeos should either reconstruct the `if_irreversible_block_id` using the QC data stored in the fork database to what it should have been prior to shutdown, or it should simply restore a saved `if_irreversible_block_id`.

Prior to https://github.com/AntelopeIO/spring/pull/325/, we had a function `latest_known_lib_num` which return the LIB number regardless of whether the node was in a pre-Savanna state or post-Savanna state. But it had a problem during the transition where it would return the `dpos_irreversible_block_num` (of the legacy consensus algorithm) without capping it to the IF Genesis block number.

Additionally, we still have `fork_db_head_irreversible_blocknum` which returns the LIB number from the perspective of the best head (whether Savanna or legacy). But this still has the flaw above during the transition and also for a Proper IF block it has the further flaw that it may not be the best known LIB that is tracked by `if_irreversible_block_id` since it does not include the effect of any integrated QC that was pulled from an invalid block.

We should have a function to complement `fork_db_head_block_id` that returns the block ID of the block known by the fork database (including the `if_irreversible_block_id` which can kind of be considered as part of the fork database, and may make sense to officially move into the fork database) to be the best LIB to move towards (this might be called `fork_db_pending_lib_id`). This is not the same as the current LIB since that is still driven by the root block of the fork database. Once we have `fork_db_pending_lib_id` (and we could add a convenience function `fork_db_pending_lib_num` to complement it if we wish), then we can get rid of `fork_db_head_irreversible_blocknum`.

There is no need for a distinction between `pending_head` and `head` in fork database. There is only a single `head` which can be thought of as the pending head. This means that the `chain_head` may not be at the `head` of the fork database. This can be because it is in the middle of advancing the `chain_head` (or doing a fork switch to it), or it can be because it is in irreversible mode. If nodeos is in irreversible mode, then `chain_head` is either at the `fork_db_pending_lib_id` or it is shortly behind it and in the middle of advancing towards it.

heifner:
- Remove distinction between fork database `pending_head` and `head`
  - `head()` is now the old `pending_head()`
  - Track pending head as the head of the fork database best branch index
- Remove `valid` as consideration of best fork
  - Set block state valid outside of fork database
  - Remove `mark_valid()` method and `mark_valid_t` arg of `add()`
  - Remove `rollback_head_to_root()` as block state `is_valid()` is no longer part of what fork database tracks
- Persist `if_irreversible_block_id` in fork database as `pending_savanna_lib_id`
- Add `pending_lib_id()` method to fork database
- Rename `conroller` `set_if_irreversible_block_id` to `set_savanna_lib_id`
- Remove `fork_db_head_or_pending()` from controller and use explicit calls to `irreversible_mode()` where needed

Developer note: fork database file format changed, restart from snapshot will be required

Resolves #333


libraries/chain/controller.cpp
```diff
@@ -1938,18 +1920,17 @@ struct controller_impl {
 
       if( check_shutdown() ) return;
 
-      // At this point head != nullptr && fork_db.head() != nullptr && fork_db.root() != nullptr.
+      // At this point chain_head != nullptr && fork_db.pending_head() != nullptr && fork_db.root() != nullptr.
```
greg7mdp, line 1923:
```suggestion
      // At this point chain_head != nullptr && fork_db.head() != nullptr && fork_db.root() != nullptr.
```

libraries/chain/fork_database.cpp
```diff
@@ -24,12 +24,12 @@ namespace eosio::chain {
...
       static bool is_valid(const block_state& bs) { return bs.is_valid(); }
-      static void set_valid(block_state& bs, bool v) { bs.validated.store(v); }
+      static void set_valid(block_state& bs, bool v) { bs.set_valid(v); }
    };
 
    struct block_state_legacy_accessor {
       static bool is_valid(const block_state_legacy& bs) { return bs.is_valid(); }
-      static void set_valid(block_state_legacy& bs, bool v) { bs.validated = v; }
+      static void set_valid(block_state_legacy& bs, bool v) { bs.set_valid(v); }
    };
```
greg7mdp, line 33: At this point, there is no difference between the two, so we don't need the `block_state_accessor` anymore I believe. 
For access we can just make the `fork_database_impl<>` friend in block_state and block_state_legacy, just like we do `controller_impl`.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -72,18 +66,18 @@ namespace eosio::chain {
...
-      void add( const bsp_t& next_block, mark_valid_t mark_valid, ignore_duplicate_t ignore_duplicate );
+      void add( const bsp_t& next_block, ignore_duplicate_t ignore_duplicate );
 
       void remove( const block_id_type& id );
 
       bool is_valid() const; // sanity checks on this fork_db
 
-      bool has_root() const;
+      bool   has_root() const;
       bsp_t  root() const; // undefined if !has_root()
```
linh2931, line 77: Add a comment explaining `root`, `head`, and `pending_savabba_lib`.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -171,6 +163,27 @@ namespace eosio::chain {
       // see fork_database_t::fetch_branch(forkdb->head()->id())
       block_branch_t fetch_branch_from_head() const;
 
+      block_id_type pending_lib_id(const block_id_type& head_id) const {
+         if (in_use.load() == in_use_t::legacy) {
```
linh2931, line 167: Why not put this implementation in fork_database.cpp?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1498,10 +1498,10 @@ namespace eosio {
 
    // called from connection strand
    void connection::blk_send_branch( const block_id_type& msg_head_id ) {
-      uint32_t fork_head_num = my_impl->get_fork_head_num();
+      uint32_t head_num = my_impl->get_chain_head_num();
 
-      peer_dlog(this, "fhead_num = ${h}",("h",fork_head_num));
-      if(fork_head_num == 0) {
+      peer_dlog(this, "fhead_num = ${h}",("h",head_num));
```
linh2931, line 1503: `fhead_num` should be `head_num`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2489,15 +2489,15 @@ namespace eosio {
          bool set_state_to_head_catchup = false;
          my_impl->connections.for_each_block_connection( [&null_id, blk_num, &blk_id, &c, &set_state_to_head_catchup]( const auto& cp ) {
             fc::unique_lock g_cp_conn( cp->conn_mtx );
-            uint32_t fork_head_num = cp->fork_head_num;
-            block_id_type fork_head_id = cp->fork_head;
+            uint32_t fork_head_num = cp->conn_fork_head_num;
```
linh2931, line 2492: Maybe name `fork_head_num` as `conn_fork_head_num` to be consistent. Same for `fork_head_id`

libraries/chain/fork_database.cpp
```diff
@@ -533,6 +440,11 @@ namespace eosio::chain {
 
    template<class BSP>
    BSP fork_database_impl<BSP>::search_on_head_branch_impl( uint32_t block_num, include_root_t include_root ) const {
+      auto head = head_impl(include_root_t::no);
+      if (!head && include_root == include_root_t::yes && block_num == root->block_num())
+         return root;
+      if (!head)
+         return {};
       return search_on_branch_impl(head->id(), block_num, include_root);
```
greg7mdp, line 448: I think this is equivalent and somewhat simpler.

```suggestion
   auto head = head_impl(include_root);
   if (!head || head == root)
      return head;
   return search_on_branch_impl(head->id(), block_num, include_root);
```

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -171,6 +163,27 @@ namespace eosio::chain {
...
+            block_state_legacy_ptr head;
+            if (head_id.empty()) {
+               head = fork_db_l.head();
+            } else {
+               // maintain legacy only advancing LIB via validated blocks
```
greg7mdp, line 172: I'm not quite understanding how this comment is accurate, since at line 170 above we use `fork_db_l.head();` (which I think doesn't require the blocks to be validated anymore)?

greg7mdp: But if the function is called with an empty id like line 170, then we use `fork_db_l.head()` which may not be valid.

plugins/net_plugin/net_plugin.cpp
heifner: The point of renaming connection variable was to avoid local variables shadowing the connection variable; hence the diff name.

greg7mdp (approved):

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -74,8 +74,22 @@ namespace eosio::chain {
...
+
+      /**
+       * The best branch head of blocks in the fork database, can be null if include_root_t::no and forkdb is empty
+       * @param include_root yes if root should be returned if no blocks in fork database
+       */
       bsp_t  head(include_root_t include_root = include_root_t::no) const;
+
+      /**
+       * The calculated pending savanna LIB ID that will become LIB or is currently LIB
+       */
```
linh2931, line 92: Those comments are very helpful!

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Refactor ForkDB to consolidate concept of HEAD and better track the pending irreversible block.
Note:end

PR #350, opened 2024-07-10
Move savanna pending lib calculation to forkdb add()

### LINKED ISSUE #333: Remove distinction between fork_db pending_head and head; persist if_irreversible_block_id in fork_db; add fork_db_pending_lib_num

heifner:
Move tracking of Savanna pending LIB into fork database. The fork database is in charge of tracking and persisting cached current pending Savanna LIB.

Resolves #333

### DIFF
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -298,6 +298,14 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
                invalid_qc_claim, "signature validation failed" );
 }
 
+qc_claim_t block_state::extract_qc_claim() const {
+   auto itr = header_exts.lower_bound(instant_finality_extension::extension_id());
+   if (itr == header_exts.end())
+      return {};
+   const auto& if_ext = std::get<instant_finality_extension>(itr->second);
+   return if_ext.qc_claim;
+}
+
 valid_t block_state::new_valid(const block_header_state& next_bhs, const digest_type& action_mroot, const digest_type& strong_digest) const {
    assert(valid);
    assert(next_bhs.core.last_final_block_num() >= core.last_final_block_num());
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3255,23 +3255,6 @@ struct controller_impl {
          emit( accepted_block, std::tie(chain_head.block(), chain_head.id()), __FILE__, __LINE__ );
 
          if( s == controller::block_status::incomplete ) {
-            fork_db.apply_s<void>([&](auto& forkdb) {
-               assert(std::holds_alternative<std::decay_t<decltype(forkdb.root())>>(cb.bsp.internal()));
-               const auto& bsp = std::get<std::decay_t<decltype(forkdb.root())>>(cb.bsp.internal());
-
-               uint16_t if_ext_id = instant_finality_extension::extension_id();
-               assert(bsp->header_exts.count(if_ext_id) > 0); // in all instant_finality block headers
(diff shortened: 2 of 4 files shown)


libraries/chain/block_state.cpp
```diff
@@ -298,6 +298,14 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
                invalid_qc_claim, "signature validation failed" );
 }
 
+qc_claim_t block_state::extract_qc_claim() const {
+   auto itr = header_exts.lower_bound(instant_finality_extension::extension_id());
+   if (itr == header_exts.end())
+      return {};
+   const auto& if_ext = std::get<instant_finality_extension>(itr->second);
+   return if_ext.qc_claim;
+}
```
greg7mdp, line 307: We really should have a generic function to extract an extension (like `template<typename Ext> Ext extract_extension()`, but checking if the extension is present and returning a std::optional).

greg7mdp (approved):

libraries/chain/block_state.cpp
```diff
@@ -298,6 +298,14 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
                invalid_qc_claim, "signature validation failed" );
 }
 
+qc_claim_t block_state::extract_qc_claim() const {
+   auto itr = header_exts.lower_bound(instant_finality_extension::extension_id());
+   if (itr == header_exts.end())
+      return {};
+   const auto& if_ext = std::get<instant_finality_extension>(itr->second);
```
linh2931, line 305: Should we change this long name `instant_finality_extension` to `savanna_extension` or something?

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Move tracking of Savanna pending LIB into fork database. The fork database is in charge of tracking and persisting cached current pending Savanna LIB.
Note:end

PR #354, opened 2024-07-10
Deep-mind: Log active proposer policy and active finalizer policy with each block

### LINKED ISSUE #343: Provide full finalizer/proposer policy (not diff) in deep-mind logs
When a new finalizer policy or proposer policy is proposed in a block, the deep-mind logs should contain an entry associated with the block that contains the full policy (not just the diff). The keys within these policies should have string encoding.

We should also always include in the deep-mind logs for each block the active proposer policy (with version within) and active finalizer policy (with generation within). These policies should also be represented in their full form with string encoding of keys as well.

heifner:
`ACCEPTED_BLOCK_V2 ${id} ${num} ${lib} ${blk} ${fd} ${pp} ${fp}`
where 
  `${id}` is block id
  `${num}` is block number
  `${lib}` is last irreversible block number
  `${fd}` is `finality_data_t`
  `${pp}` is active `proposer_policy` in packed hex
  `${fp}` is active `finalizer_policy_with_string_key` in packed hex

```c++
struct finality_data_t {
   uint32_t     major_version{light_header_protocol_version_major};
   uint32_t     minor_version{light_header_protocol_version_minor};
   uint32_t     active_finalizer_policy_generation{0};
   uint32_t     final_on_strong_qc_block_num{0};
   digest_type  action_mroot{};
   digest_type  base_digest{};
   std::optional<finalizer_policy> pending_finalizer_policy; // finalizer policy if one is promoted to pending in the block
};

struct finalizer_authority_with_string_key {
      std::string  description;
      uint64_t     weight = 0;
      std::string  public_key;
};

struct finalizer_policy_with_string_key {
      uint32_t                                         generation = 0;
      uint64_t                                         threshold = 0;
      std::vector<finalizer_authority_with_string_key> finalizers;
};

struct proposer_policy {
   // Useful for light clients, not necessary for nodeos
   block_timestamp_type        active_time; // block when schedule will become active
   producer_authority_schedule proposer_schedule;
};
```

Resolves #343


libraries/chain/controller.cpp
```diff
@@ -4164,9 +4170,11 @@ struct controller_impl {
...
                }
             } else if (!branches.first.empty()) {
-               ilog("applying ${n} fork db blocks from ${cbn}:${cbid} to ${nbn}:${nbid}",
-                    ("n", branches.first.size())("cbid", (*branches.first.rbegin())->id())("cbn", (*branches.first.rbegin())->block_num())
-                    ("nbid", new_head->id())("nbn", new_head->block_num()));
+               if (fc::time_point::now() - new_head->timestamp() < fc::minutes(5)) {
+                  ilog("applying ${n} fork db blocks from ${cbn}:${cbid} to ${nbn}:${nbid}",
+                       ("n", branches.first.size())("cbid", (*branches.first.rbegin())->id())("cbn", (*branches.first.rbegin())->block_num())
+                       ("nbid", new_head->id())("nbn", new_head->block_num()));
+               }
```
heifner, line 4177: Didn't actually mean to commit this, but not sure it is worth creating a diff PR for it.
It reduces log spam while syncing.

linh2931: It's OK to come along with this PR.

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: When a new finalizer policy or proposer policy is proposed in a block, the deep-mind logs should contain an entry associated with the block that contains the full policy (not just the diff). The keys within these policies should have string encoding.
Note:end

PR #359, opened 2024-07-11
Fix starting from snapshot created during transition

### LINKED ISSUE #345: Add a test for restarting from snapshot after blocks are removed during transition to Savanna
From @ericpassmore

_Looks like this test snapshot_in_svnn_transition_test.py (https://github.com/AntelopeIO/spring/blob/main/tests/snapshot_in_svnn_transition_test.py) mostly satisfies the last 2 scenarios for Wave 2 Testing (https://drive.google.com/file/d/1ZrzDuNWiQtPVSftupRvcEv-5_HJeAs_s/view?pli=1). Should we add another test that removes the blocks and does a restart from snapshot verifying Savanna transition completes?_

heifner:
Fix starting from snapshot created during transition when no block log is provided.
Add new test case that starts from a snapshot created during transition after removing the block log.

Resolves #345


libraries/chain/controller.cpp
```diff
@@ -1359,8 +1359,25 @@ struct controller_impl {
       assert(!!legacy_root);
       assert(read_mode == db_read_mode::IRREVERSIBLE || !legacy_branch.empty());
       ilog("Transitioning to savanna, IF Genesis Block ${gb}, IF Critical Block ${cb}", ("gb", legacy_root->block_num())("cb", chain_head.block_num()));
-      auto new_root = block_state::create_if_genesis_block(*legacy_root);
-      fork_db.switch_from_legacy(new_root);
+      if (chain_head_trans_svnn_block) {
```
linh2931, line 1362: Can you add some comments why this section of new code fixes restarting without blocks.log?

tests/snapshot_in_svnn_transition_test.py
```diff
@@ -101,20 +104,31 @@ def takeSnapshot(node):
     info = cluster.biosNode.getInfo(exitOnError=True)
     assert (info["head_block_num"] - info["last_irreversible_block_num"]) < 9, "Instant finality enabled LIB diff should be small"
 
-    def restartWithSnapshot(node):
+    def restartWithSnapshot(node, rmBlocks, addSnapshot=True):
        Print("Shut down node")
        node.kill(signal.SIGTERM)
        Print("Restart node with snapshot")
-       node.removeState()
+       Utils.rmNodeDataDir(node.nodeId, rmBlocks=rmBlocks)
```
linh2931, line 111: Should `rmNodeDataDir()` belong to `node` class? If too many places to change, then  leave as is.

tests/snapshot_in_svnn_transition_test.py
```diff
@@ -101,20 +104,31 @@ def takeSnapshot(node):
...
+    Print("Restart snapshot node with snapshot with blocks")
+    restartWithSnapshot(nodeSnap, rmBlocks=False)
+
+    Print("Restart Irreversible node with snapshot with blocks")
+    restartWithSnapshot(nodeIrr, rmBlocks=False)
 
-    Print("Restart snapshot node with snapshot")
-    restartWithSnapshot(nodeSnap)
+    Print("Restart snapshot node with snapshot without blocks")
+    restartWithSnapshot(nodeSnap, rmBlocks=True, addSnapshot=False)
```
linh2931, line 128: How does `addSnapshot=False` make it restart with snapshot?

heifner:
Something not right, as it failed: https://github.com/AntelopeIO/spring/actions/runs/9897805335/job/27343685063

libraries/chain/controller.cpp
```diff
@@ -1359,8 +1359,31 @@ struct controller_impl {
...
+         if (legacy_root->id() == chain_head_trans_svnn_block->id()) {
+            // setup savanna forkdb with the block_state from the snapshot
+            fork_db.switch_from_legacy(chain_head_trans_svnn_block);
+         } else {
+            // root has moved from chain_head_trans_svnn_block, so transition the legacy root
+            // legacy_root can be one pass the snapshot start block when running in irreversible mode as LIB is advanced
+            // before transition_to_savanna is called.
+            const bool skip_validate_signee = true; // validated already
+            dlog("Create irreversible transition block ${bn}", ("bn", legacy_root->block_num()));
+            auto new_bsp = block_state::create_transition_block(
```
linh2931, line 1375: `new_bsp` can be named as `new_root`. But not a nbig deal.

linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -1359,8 +1359,31 @@ struct controller_impl {
...
+      if (chain_head_trans_svnn_block) {
+         // chain_head_trans_svnn_block is set if started from a snapshot created during the transition
+         // If the snapshot is from during transition then the IF genesis block should not be created, instead
+         // chain_head_trans_svnn_block contains the block_state to build from
+         if (legacy_root->id() == chain_head_trans_svnn_block->id()) {
+            // setup savanna forkdb with the block_state from the snapshot
+            fork_db.switch_from_legacy(chain_head_trans_svnn_block);
+         } else {
+            // root has moved from chain_head_trans_svnn_block, so transition the legacy root
+            // legacy_root can be one pass the snapshot start block when running in irreversible mode as LIB is advanced
```
greg7mdp, line 1371:
```suggestion
            // legacy_root can be one past the snapshot start block when running in irreversible mode as LIB is advanced
```

tests/snapshot_in_svnn_transition_test.py
```diff
@@ -73,10 +73,13 @@
     nodeSnap=cluster.getNode(snapshotNodeId)
     nodeIrr=cluster.getNode(irrNodeId)
 
-    # Active Savanna without waiting for activatation is finished so that we can take a
+    # Active Savanna without waiting for activation is finished so that we can take a
```
greg7mdp, line 76:
```suggestion
    # Activate Savanna without waiting for activation to be finished so that we can take a
```

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Proper snapshot during transition when block log is not present.
Note:end

linh2931 (approved):

libraries/chain/block_state_legacy.cpp
```diff
@@ -84,4 +84,26 @@ namespace eosio::chain {
    {
    }
 
+   block_num_type block_state_legacy::irreversible_blocknum() const {
+      if (std::optional<block_num_type> gen_block_num = savanna_genesis_block_num()) {
+         return std::min<block_num_type>(*gen_block_num, dpos_irreversible_blocknum);
```
greg7mdp, line 89: shouldn't it be `std::max()`?

heifner: We don't want LIB in legacy to advance past savanna genesis block.

libraries/chain/controller.cpp
```diff
@@ -1424,18 +1456,35 @@ struct controller_impl {
...
 
       if( new_lib_num <= lib_num )
          return;
 
-      bool savanna_transition_required = false;
       auto mark_branch_irreversible = [&, this](auto& forkdb) {
          assert(!irreversible_mode() || forkdb.head());
          const auto& head_id = irreversible_mode() ? forkdb.head()->id() : chain_head.id();
-         auto branch = forkdb.fetch_branch( head_id, new_lib_id);
+         auto branch = forkdb.fetch_branch( head_id, new_lib_id); // verifies lib is on head branch
```
greg7mdp, line 1487: When in Savanna mode, should we mark_irreversible only if `new_lib_id` is an ancestor of `head_id`? If so, why?

greg7mdp (approved):

libraries/chain/controller.cpp
heifner: Added additional comments

greg7mdp (approved):

libraries/chain/controller.cpp
greg7mdp: Thanks. I feel that at this point, we know that `head_id` is on the wrong fork (since `lib_id` is *not* an ancestor), so it seems that we would want to take action now, rather than delaying to when `maybe_switch_fork` is called, however I'm not advocation for a change now since this is delicate code. 

I can't help but feel this code could be rearranged in a way that makes it simpler and easier to follow (at least for me :-).

PR #364, opened 2024-07-12
Rename instant_finality_extension to finality_extension

### LINKED ISSUE #353: Clean up instant finality names
Instead of instant_finality we can either use finality or savanna.

heifner:
Resolves #353

### DIFF
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -1,6 +1,6 @@
 #include <eosio/chain/block_header_state.hpp>
 #include <eosio/chain/block_header_state_utils.hpp>
-#include <eosio/chain/finality/instant_finality_extension.hpp>
+#include <eosio/chain/finality/finality_extension.hpp>
 #include <eosio/chain/finality/proposer_policy.hpp>
 #include <eosio/chain/exceptions.hpp>
 #include <limits>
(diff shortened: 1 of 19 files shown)


libraries/chain/block_header_state.cpp
```diff
@@ -183,7 +183,7 @@ void evaluate_finalizer_policies_for_promotion(const block_header_state& prev,
 void finish_next(const block_header_state& prev,
                  block_header_state& next_header_state,
                  vector<digest_type> new_protocol_feature_activations,
-                 instant_finality_extension if_ext,
+                 finality_extension if_ext,
```
linh2931, line 186: Maybe we should stop using `if_ext`, using `fin_ext` or `f_int` instead.

linh2931 (approved):

libraries/chain/block_header_state.cpp
greg7mdp: Personally, I'd prefer avoiding such naming changes, I don't find `if_ext` objectionable, but I guess opinions may vary.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Rename extension to `finality_extension`.
Note: end

PR #365, opened 2024-07-12
SHiP: Pull block by block id instead of by block number

### LINKED ISSUE #363: Test failure: ship_streamer_if_test
https://github.com/AntelopeIO/spring/actions/runs/9913312540/job/27390423148

```
  File "/__w/spring/spring/build/tests/ship_streamer_test.py", line 215, in <module>
    assert isinstance(i['get_blocks_result_v1']['block'], str) # verify block in result
```

```
{ "get_blocks_result_v1":
{
    "head": {
        "block_num": 220,
        "block_id": "000000DCABB6F9B3CDEA683452030DDCC74EE8AA78D98FB368FE876D64CA43B3"
    },
    "last_irreversible": {
        "block_num": 215,
        "block_id": "000000D7BD770A24441393B9F823E6840CDF3B3E255BF176C9FD1D62136E7344"
    },
    "this_block": {
        "block_num": 220,
        "block_id": "000000DCABB6F9B3CDEA683452030DDCC74EE8AA78D98FB368FE876D64CA43B3"
    },
    "prev_block": {
        "block_num": 219,
        "block_id": "000000DB755625BEBAE3BCBCFD74D40D23E0DAEC68F35C15AF40E5A93ED55B7B"
    },
    "block": null,
    "traces": "0B00C6B2C05C85B2B394151DCCE77CD868D8C7B129190FC15DCF01ADE4DCCC2A820000640000000000000000000000000000000000000000000101010001000000000000EA30556AF623BD40AF9E0D95694CF91033FA53C43729DD23DC391C3F5033249D39C4D400080000000000007F04000000000000010000000000EA3055160100000000000002020000000000EA30550000000000EA305500000000221ACFA4010000000000EA305500000000A8ED32327E5491485C90AE423AD15B974A0000000000DABA7ADE708C722457461F599154406A43228FD2A38BD494F65D59DC3DF7EDFE0D10DC5851631D6AA5B6F813E9634CFA02F9FB01176777B346B1369CA4A289501E993FDDC2181670614CBAA0AF152E9C5B8E0FC1AA73595C97AF2D907F000000800001020007D90000000100000000000000000000000000000000000000000100000000000000000000000000000000000074DD9F7834C49AFABBB2F6AC99470557AFB618728624A7303C9001CAB6864BB700BA000000180000000000000000C0000000000000000002010100010000408C7A02EA30555DDD07C2D2AE40AA66982BDEA851FC020CAAE9529DD171CA8FDF37B9A5CE1B9C0108000000000000750300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A303A3838333A373339303834323338363532393634393038333A31373230383134363334303131353835010000000000000000000000000001020001000000000000EA305572C149E32751024CF2282F294E34606DF30A1100156F8AAE90489E10EBF1358F020800000000000080040000000000000160AE423AD15B974AB90100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0160AE423AD15B974A00000000A8ED32322460AE423AD15B974A70AE423AD15B974A102700000000000004435552000000000336333800000000000000000000000000000000000001003A9A916697007D4497F20000000001001F4004F8721B09D9FAAC778F7F4DDB0BDC5069464A249DA53A23971CE0A4D0D066044B4A5E2170B11969FF76091505C243A9C268A5C324F407663E624AED17659B000088DD8AA857163DE549E15434BC6CEFB8F54228B0DE36310C95D6BFAF585A6018008E000000180000000000000000C0000000000000000002010100010000408C7A02EA305570261CD6366651D452F8CFFDD0DCF3D4651FECF4CDD8B57599AC4F770C925E410308000000000000760300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A313A3838333A373339303834323338363532393634393038333A31373230383134363334303135393636010000000000000000000000000001020001000000000000EA3055BBCA92A1AC761AB4F8D4154DE087EEC0FC1FF87A3899FC14CB0ECD779DF33507040800000000000081040000000000000160AE423AD15B974ABA0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0160AE423AD15B974A00000000A8ED32322460AE423AD15B974A70AE423AD15B974A102700000000000004435552000000000336333900000000000000000000000000000000000001003A9A916697007D4497F20000000001002002B0EAE9FA543759515999D6D7DAED00006E79187D51141145D558C2C275D7611FC38B22FFBA8322C7D41915833420F6F6AF8BF43F42331C6468980A0014E98E000066DC07AD40935BB8F346711B3BCC3D75DD782ADFC228F4825D48C2802842B5C40092000000180000000000000000C0000000000000000002010100010000408C7A02EA3055C5AB7D655800C50025BDA1ECE533BA88579247BC7CD0BCDDEDABADF4CCD943350508000000000000770300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A303A3838353A373339303834323338363532393634393038343A31373230383134363334313131353635010000000000000000000000000001020001000000000000EA305541C43F05D465C786EB868DF7E2EB95802DE78F266536821F49AF72D638476B2D060800000000000082040000000000000170AE423AD15B974AB90100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0170AE423AD15B974A00000000A8ED32322470AE423AD15B974A60AE423AD15B974A102700000000000004435552000000000336333800000000000000000000000000000000000001003A9A916697007D4497F20000000001002009E0C06DBB72C4E2FE2916FB4600426890233BBB7C208A7505A1F9F73D28D5D128BCE9EDFDF269BA1DA4F247553A1D5088BB317F47D45D1F9123D5D445ADF15800008DA548D88438331ED51D83839062074493EDBFD2FAABE2B42A7B72FC08601F5E0064000000180000000000000000C0000000000000000002010100010000408C7A02EA30552CA15FE34FD36DB6B5D17E5AD23A9F8B36B3BEF5865C78F4F29633F9C8F42B170708000000000000780300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A313A3838353A373339303834323338363532393634393038343A31373230383134363334313132393632010000000000000000000000000001020001000000000000EA30554DD4557F3B4CCB4986D8C8972419EB38CDE8CE408BCF384D1801D125C38062B2080800000000000083040000000000000170AE423AD15B974ABA0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0170AE423AD15B974A00000000A8ED32322470AE423AD15B974A60AE423AD15B974A102700000000000004435552000000000336333900000000000000000000000000000000000001003A9A916697007D4497F2000000000100202D05D47962A7D38F3D26C8FE49FE39653936604D4E6C24EF319A3EFD623864083EE137F66A6D41AF1F50923A9E6DB8FA7BFDE6AF09068DEF465AC31D2B72FCB10000B92C33B69CACD1B51E8A34D1F10A97AE541AF764E49C29783A43FB18E020B0450085000000180000000000000000C0000000000000000002010100010000408C7A02EA3055B6F1E8B1B7F27A9A4836A97C815E59D20E299BC7C9FF6704A7C17D51FC760CAE0908000000000000790300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A303A3838373A373339303834323338363532393634393038353A31373230383134363334323131353639010000000000000000000000000001020001000000000000EA305572C149E32751024CF2282F294E34606DF30A1100156F8AAE90489E10EBF1358F0A0800000000000084040000000000000160AE423AD15B974ABB0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0160AE423AD15B974A00000000A8ED32322460AE423AD15B974A70AE423AD15B974A102700000000000004435552000000000336333800000000000000000000000000000000000001003A9A916697007D4497F20000000001001F3C71B5B5A3D18D6A3E15085FDC425054B7EBD4E0CD2138809962917803ABCC6E40F2BFCCBDDCD3D2D21845E078B35C3F5C2B1F4197B731D14DB3FE2A98CDE26C00004D8D9BEAF193A4F443B31F2639C11463156710EE0A572388F8A87F61136365AE0064000000180000000000000000C0000000000000000002010100010000408C7A02EA3055A8B7CA8D8AB56E7E6393B70A35EEE67E697DB885B2709E3EB438C489305462A80B080000000000007A0300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A313A3838373A373339303834323338363532393634393038353A31373230383134363334323132393733010000000000000000000000000001020001000000000000EA3055BBCA92A1AC761AB4F8D4154DE087EEC0FC1FF87A3899FC14CB0ECD779DF335070C0800000000000085040000000000000160AE423AD15B974ABC0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0160AE423AD15B974A00000000A8ED32322460AE423AD15B974A70AE423AD15B974A102700000000000004435552000000000336333900000000000000000000000000000000000001003A9A916697007D4497F2000000000100203B577FEB87E9113AAF7B0251E6A0FC168E231F5BC13A8132C242C14094F1ECB24C1E3D1ACC9181BD0F7E545FB4B755AC081139DEED1CEC48D4F1FDD227EAE5FC00003B146DC2F8348E28789AD187B026AA1EAD9880163D16E5E8EA70331AC5666ACC00AA000000180000000000000000C0000000000000000002010100010000408C7A02EA3055EEEA5629E0C698CB1DEDCD73166E5D8F5E01E401657E54C92CA5F60A71099EB40D080000000000007B0300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A303A3838393A373339303834323338363532393634393038363A31373230383134363334333131353537010000000000000000000000000001020001000000000000EA305541C43F05D465C786EB868DF7E2EB95802DE78F266536821F49AF72D638476B2D0E0800000000000086040000000000000170AE423AD15B974ABB0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0170AE423AD15B974A00000000A8ED32322470AE423AD15B974A60AE423AD15B974A102700000000000004435552000000000336333800000000000000000000000000000000000001003A9A916697007D4497F20000000001001F24D6064C13528C4C275563435CB2A629F06CE3FD7AEC08DF9953520A4EECA0543F39AF2BE3EA5AE1A93EE85C53FDB3FF0F8625FFC5277AEED428D3C1A5CB97A10000967F3239C3FBC11DFF6508FE6480B14EB398C1D340E9236846370065B85BFE360064000000180000000000000000C0000000000000000002010100010000408C7A02EA3055A1870F80574C8674EA4DB0BF23FD4A7623897807A34409A6D52E6589E4881F480F080000000000007C0300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A313A3838393A373339303834323338363532393634393038363A31373230383134363334333132393633010000000000000000000000000001020001000000000000EA30554DD4557F3B4CCB4986D8C8972419EB38CDE8CE408BCF384D1801D125C38062B2100800000000000087040000000000000170AE423AD15B974ABC0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0170AE423AD15B974A00000000A8ED32322470AE423AD15B974A60AE423AD15B974A102700000000000004435552000000000336333900000000000000000000000000000000000001003A9A916697007D4497F20000000001001F3509B57E0DC5FF13180D5F97DF5174556FA71E7EEE4844F5085EC9A9A22C48E2562C85897BACD2C18106B6B1BBC45C9C1D50A63BA6C264C084C156291F5B2CA20000BCB534E2478C6B99A29FDE7E93A3D29AF87C596A330C22B021B8C41FD9E5D3BB00B3000000180000000000000000C0000000000000000002010100010000408C7A02EA305516B6E03E79C9BF7E09B50171EA084DAC194C536D104A6E6621FC7D094A4858E511080000000000007D0300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A303A3839313A373339303834323338363532393634393038373A31373230383134363334343131353533010000000000000000000000000001020001000000000000EA305572C149E32751024CF2282F294E34606DF30A1100156F8AAE90489E10EBF1358F120800000000000088040000000000000160AE423AD15B974ABD0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0160AE423AD15B974A00000000A8ED32322460AE423AD15B974A70AE423AD15B974A102700000000000004435552000000000336333800000000000000000000000000000000000001003A9A916697007D4497F20000000001001F19156A0561B9DFB70B4B86D9494D8FEBD8AFF531473A0DE9C5B5FF54245D3509139AB440594AF8CF0AFE7E240ABAE5C6B3E99C4F1B4B23BBB1A5B0E02A52F26D0000F1793C9D909C4AF0184AE16F6B567785B2DED6AEF21B06E340A4C88B0752A71F006C000000180000000000000000C0000000000000000002010100010000408C7A02EA30554617B5AC6B2345D65E4421451CF9B02488228D1C7AC6C97F880BEAEBEA698F6113080000000000007E0300000000000000000000408C7A02EA305500408C7A02EA3055000000000085269D002B2A313A3839313A373339303834323338363532393634393038373A31373230383134363334343132393730010000000000000000000000000001020001000000000000EA3055BBCA92A1AC761AB4F8D4154DE087EEC0FC1FF87A3899FC14CB0ECD779DF33507140800000000000089040000000000000160AE423AD15B974ABE0100000000000002020000000000EA30550000000000EA3055000000572D3CCDCD0160AE423AD15B974A00000000A8ED32322460AE423AD15B974A70AE423AD15B974A102700000000000004435552000000000336333900000000000000000000000000000000000001003A9A916697007D4497F20000000001001F22E70A4DCAAF4A82F3F88F3C4E49AC4F73E22DDACBB260D1254BA8CD2C8C630B0EA6AD59E2E6F5C8B6F04BC79827E1A778F78B0CBEA92EC04265FA9F204839DD00",
    "deltas": "02000E7265736F757263655F757361676503013B0070AE423AD15B974A005591485C0C880700000000000103000000000000005591485CC9940500000000000502000000000000A40A000000000000013B0060AE423AD15B974A005591485CBB900700000000008104000000000000005591485C305C0500000000005103000000000000A40A000000000000013B000000000000EA3055005591485C86930200000000000100000000000000005591485C13760400000000006500000000000000D7FC06000000000000157265736F757263655F6C696D6974735F73746174650101530000DC00000038AE0E4C000000006D0C00000000000000DC00000023ABAA48000000006F0A000000000000000000000000000000000000000000000000000000000000E8EA130000000000B17C090000000000",
    "finality_data": null
}
```

heifner:
`ship_streamer_if_test` failed because it expects a serialized block for each `get_blocks_result_v1`. A `null` block was returned because the request was processed during a fork switch which meant the block by number on best branch in the fork database was not available. Modify SHiP to pull by the `this_block::block_id` so that it will always find the block in the fork database.

Includes clean up of unused methods and providing a `chain_block_id_for_num` and `fork_block_id_for_num` as an alternative to the prior `get_block_id_for_num`.

Resolves #363


spoonincode:
>  A `null` block was returned because the request was processed during a fork switch which meant the block by number on best branch in the fork database was not available.

Can you elaborate a little more what conditions cause this? Maybe an illustrative example? The "during a fork switch" statement is the confusing part to me: the modified code is run on the main thread so it can't run "during" a switch; it'd either have to be before or after it. So I guess I'm not sure under what condition `fetch_block_by_number(block_num)` would be called with a `block_num` not on the best branch

heifner:
`get_block_id` looks in `trace_log`, `chain_state_log`, `finality_data_log`, and then in the fork database. After that call, the code linked below can add to the fork database which can change the best branch head. We no longer track a `head` in the fork database, so the best `head()` can change at any moment.
https://github.com/AntelopeIO/spring/blob/bbc005666dac08e3373ed5ce3295f34a9af6341b/libraries/chain/controller.cpp#L3867-L3867
This code is called from net threads as blocks come in. So the best branch of the fork database can change at any time as new block come in and are added to the fork database. 

Should probably look around at other uses of controller `fetch_block_by_number` to see if anything else might be affected by best head being able to change outside of the main thread.

spoonincode:
> This code is called from net threads as blocks come in. So the best branch of the fork database can change at any time as new block come in and are added to the fork database.

ah okay it can change in net threads... that's what I'm missing

spoonincode:
> Should probably look around at other uses of controller `fetch_block_by_number` to see if anything else might be affected by best head being able to change outside of the main thread.

Yeah.... it feels precarious that some of `controller`'s accessors return results always based on the applied fork while others return based on the best fork that might not have been applied yet and can change at any moment from another thread. For example it seems this innocuous code is now always racy, even if on main thread, because of that discrepancy?

```c++
const uint32_t bnum  = controller.head_block_num();
signed_block_ptr sbp = controller.fetch_block_by_number(bnum);
```

ship is written with the assumption that asking for a blockid of a given blocknum will give the blockid on the currently applied fork. If that's no longer true there may be other things racy in here, I will have to think it through more. I think you're right that since `get_block_id()` first inspects the ship logs (which represent applied blocks) nominally it will find applied blockids like it needs to. However it's possible to run ship without any ship logs and it still ought to stream blocks, and that configuration may be at risk of getting mixed up results currently because `get_block_id()` will always query controller's racy accessor.

Would it make sense for controller to expose a `fetch_block_by_number_on_applied_fork()`? And/Or possibly the existing function should be renamed `fetch_block_by_number_from_best_fork()` (if its existing behavior really even needs to be kept, even)?

heifner:
> Would it make sense for controller to expose a `fetch_block_by_number_on_applied_fork()`? And/Or possibly the existing function should be renamed `fetch_block_by_number_from_best_fork()` (if its existing behavior really even needs to be kept, even)?

Yes, I think it would make sense to provide this functionality. I'll add it.

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Update SHIP to pull by block id, resolves an issue where null blocks were pulled on a fork switch.
Note:end

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -250,16 +250,16 @@ class session final : public session_base {
                      block_to_send->blocks_result_base.this_block  = {self.current_blocks_request.start_block_num, *this_block_id};
```
linh2931, line 250: Should we break in case this_block_id has no value?

heifner: I don't understand what you are asking. This whole block is in the if for if `this_block_id` has value.

linh2931: I could not place the comment right below the `if`. I was asking do we need an `else` to break out the outer `while` in case `this_block_id` has no value.

heifner: Not really sure what we should do in this case. If running without ship logs and without a block log it is possible that we can't find the requested block. In that case we, maybe we should log an error and disconnect? @spoonincode

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -250,16 +250,16 @@ class session final : public session_base {
...
-                     if(chain::signed_block_ptr sbp = get_block(self.next_block_cursor); sbp && self.current_blocks_request.fetch_block)
+                     if(chain::signed_block_ptr sbp = get_block(*this_block_id); sbp && self.current_blocks_request.fetch_block)
                          block_to_send->blocks_result_base.block = fc::raw::pack(*sbp);
                      if(self.current_blocks_request.fetch_traces && self.trace_log)
                         block_to_send->trace_entry = self.trace_log->get_entry(self.next_block_cursor);
                      if(self.current_blocks_request.fetch_deltas && self.chain_state_log)
                         block_to_send->state_entry = self.chain_state_log->get_entry(self.next_block_cursor);
                      if(block_to_send->is_v1_request && *self.current_blocks_request_v1_finality && self.finality_data_log)
                         block_to_send->finality_entry = self.finality_data_log->get_entry(self.next_block_cursor);
+                     ++self.next_block_cursor;
```
spoonincode, line 261: It seems like this change may cause ship to enter a tight busy loop, of sorts, and spam the client forever if we ever enter this failure condition. Because `next_block_cursor` will be stuck forever at a "bad block" (one we can't look up the id to) and after sending this nearly empty block result to the client we'll just reconsider the "bad block" once again immediately once the async_send compeltes. Repeat.

Possibly the discussion over in https://github.com/AntelopeIO/spring/pull/365/files#r1679469382 is right that we could consider asserting `get_block_id(self.next_block_cursor)` succeeds and kill off the connection when it fails. I am not entirely sure off hand what user visible changes this will cause though. It might mean asking for a blocknum before knowledge of all logs (like if you had started from snapshot w/o any previous logs) will now kill the connection instead of streaming "empty" entries until the first populated entry.

Maybe you can say that sort of request is an erroneous request though (i.e. should have done a get_status_request first to see the valid ranges that ship knows about; though that status request doesn't have the block log range for that corner case of no ship logs enabled)

heifner: Seems we should continue to allow requests outside of valid range (move the `++self.next_block_cursor` outside the `if`). Seems like the best approach is to stream "empty" entries until the first populated entry.

linh2931: I was worry about an infinite loop on a bad block ID.

linh2931: Add a comment in the code about the empty entries.

linh2931 (approved):

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
spoonincode: The main problem with those empty entries is they provide no information to the client what block number they correspond to. To an unaware client things might just look broken. Potentially we could change this with the v1 blocks reply and include some always-available bnum; though I'm not convinced it's worth it. (Even though it's racy if pruning/splitting is enabled) well behaved clients that can't assume log availability from genesis really ought to query status first for the minimum starting block to stream from (which, granted, is only available for trace, state, and finality -- not blog -- that too could potentially be changed with v1 status reply but again not sure it's worth it)

heifner: I think that is something to consider for a future PR. In practice what we have now should suffice I think.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -250,15 +250,24 @@ class session final : public session_base {
...
+                        }
+                     }
                      if(self.current_blocks_request.fetch_traces && self.trace_log)
                         block_to_send->trace_entry = self.trace_log->get_entry(self.next_block_cursor);
                      if(self.current_blocks_request.fetch_deltas && self.chain_state_log)
                         block_to_send->state_entry = self.chain_state_log->get_entry(self.next_block_cursor);
                      if(block_to_send->is_v1_request && *self.current_blocks_request_v1_finality && self.finality_data_log)
                         block_to_send->finality_entry = self.finality_data_log->get_entry(self.next_block_cursor);
+                  } else {
+                     fc_wlog(logger, "Unable to retrieve block id for block ${bn}", ("bn", self.next_block_cursor));
```
spoonincode, line 268: This log is certainly more reasonable than the one above since in this case the empty messages that will be sent to the client are kinda mysterious since everything but head & lib is empty, so the log helps out in adding context.

Since I believe the only way we realistically expect to encounter this situation is when a client requests blocks below the `min(blog, trace, state, finality)`, we could consider,

* easy: adding  `remote_endpoint_string` to the message; this alone will help convey the warning is the result of a client's request and not something internally wrong
* medium: possibly only printing the log once per a request to avoid massive log spam if someone mistakenly requests a large range they don't have; like when starting fresh from snapshot but requesting from block 1
* hard: possibly printing a more helpful log such as "client ${remote_endpoint_string} requested blocks from ${bn} but minimum block with any results is ${min_block}; empty results will be streamed until minimum block"; of course `min_block` will require a new function to compute `min(blog, trace, state, finality)`

imo we don't need to do any of these for now; but the 'easy' one might be nice

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -250,15 +250,24 @@ class session final : public session_base {
                      block_to_send->blocks_result_base.this_block  = {self.current_blocks_request.start_block_num, *this_block_id};
                      if(const std::optional<chain::block_id_type> last_block_id = self.get_block_id(self.next_block_cursor - 1))
                         block_to_send->blocks_result_base.prev_block = {self.next_block_cursor - 1, *last_block_id};
-                     if(chain::signed_block_ptr sbp = get_block(self.next_block_cursor); sbp && self.current_blocks_request.fetch_block)
-                         block_to_send->blocks_result_base.block = fc::raw::pack(*sbp);
+                     if (self.current_blocks_request.fetch_block) {
+                        if (chain::signed_block_ptr sbp = get_block(*this_block_id)) {
+                           block_to_send->blocks_result_base.block = fc::raw::pack(*sbp);
+                        } else {
+                           fc_wlog(logger, "Unable to retrieve block ${bn} : ${id}",
```
spoonincode, line 257: This treats the blog specially, and I can't really think of a reason to do so. Either all 4 logs should warn in this case, or none should, to maintain consistency across the 4.

But a log message here just seems ripe for a log flood. For example if someone has `block-log-retain-blocks` set low, a trace log from genesis, and requests block+trace from genesis, I don't think we need to print millions of `wlog`s when the missing block _is already conveyed_ in the still usable response (the block's `optional<bytes>` is unset)

In other words the client already has an explicit visible indicator of this error (on any of the 4 logs), so unless we have a good strong reason to add the log I'd rather leave it without a log and avoid bloating this out more.

spoonincode (changes requested):

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
heifner: I created the following, but realized that all this (except `earliest_available_block_num()`) is available via the SHiP status request. I think we should just remove the log output and let the user use status request if they do not get what they anticipate. 

image

spoonincode (approved):

PR #373, opened 2024-07-17
Write chain_head after vote threads have stopped

### LINKED ISSUE #369: test failure: `nodeos_retry_transaction_if_lr_test`
See [here](https://github.com/AntelopeIO/spring/actions/runs/9964086157/job/27532098443)

```
Total Test time (real) = 224.66 sec
1/1 Test #2162: nodeos_retry_transaction_if_lr_test ...***Failed  226.66 sec
```

heifner:
Make sure vote_processor threads have stopped before writing out the `chain_head` to avoid modifications to the `block_state` while writing. Note `fork_db` was already being written out after the vote processor threads were shutdown.

Resolves #369

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -946,6 +946,8 @@ struct controller_impl {
    authorization_manager           authorization;
    protocol_feature_manager        protocol_features;
    controller::config              conf;
+   // write out chain_head after vote_processor has stopped to avoid concurrent access while writing
+   fc::scoped_exit<std::function<void()>> write_chain_head = [&]() { chain_head.write(conf.state_dir / config::chain_head_filename); };
    const chain_id_type             chain_id; // read by thread_pool threads, value will not be changed
    bool                            replaying = false;
    bool                            is_producer_node = false; // true if node is configured as a block producer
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -946,6 +946,8 @@ struct controller_impl {
    authorization_manager           authorization;
    protocol_feature_manager        protocol_features;
    controller::config              conf;
+   // write out chain_head after vote_processor has stopped to avoid concurrent access while writing
+   fc::scoped_exit<std::function<void()>> write_chain_head = [&]() { chain_head.write(conf.state_dir / config::chain_head_filename); };
```
greg7mdp, line 950: Maybe add a comment stating that this member has to come after `chain_head`, `fork_db` and `conf`.

greg7mdp (approved):

libraries/chain/controller.cpp
greg7mdp: doesn't it depend on fork_db as well?

heifner: In what way does it depend on forkdb?

greg7mdp: I thought [this](https://github.com/AntelopeIO/spring/blob/c2cffd8d7c255e997a0565fbebf3596298f16c5b/libraries/chain/include/eosio/chain/block_handle.hpp#L44) would write the `block_state` which is stored in fork_db?

heifner: The block_state is stored in a shared_ptr that is referenced in forkdb. But the forkdb is not modifying the block state.

greg7mdp: I don't know what I was thinking. Thanks!

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Write out chain_head after vote_processor threads have stopped.
Note:end

Issue #375, opened 2024-07-17
Revert liveness rule


arhag:
Revert the changes of https://github.com/AntelopeIO/spring/issues/293.

Turns out the change to the liveness rule creates scenarios in which liveness is lost/reduced. And it isn't a sufficient solution to solve the safety issues it was intended to solve. The proper safety solution will come in other issues. For this issue we just want to revert the changes of #293.

PR #378, opened 2024-07-18
Revert liveness rule only valid for the same finalizer generation changes

linh2931:
The change to the liveness rule (https://github.com/AntelopeIO/spring/issues/293) create scenarios in which liveness is lost/reduced. Revert the changes. The proper safety solution will be handled by other upcoming PRs.

Resolves https://github.com/AntelopeIO/spring/issues/375


heifner (approved):

libraries/chain/finality/finalizer.cpp
```diff
@@ -99,7 +98,7 @@ bool finalizer::maybe_update_fsi(const block_state_ptr& bsp) {
    auto& final_on_strong_qc_block_ref = bsp->core.get_block_reference(bsp->core.final_on_strong_qc_block_num);
    if (final_on_strong_qc_block_ref.timestamp > fsi.lock.timestamp && bsp->timestamp() > fsi.last_vote.timestamp) {
       fsi.lock = final_on_strong_qc_block_ref;
```
greg7mdp, line 100: I believe we should set the timestamp on `fsi.lock`. Also on line 83.

greg7mdp (changes requested):

libraries/chain/finality/finalizer.cpp
linh2931: `fsi.lock` is from `final_on_strong_qc_block_ref` which already contains the timestamp. What other timestamp to set?

greg7mdp: Oh, my bad, sorry!

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Revert previous changes to liveness rule. After evaluation the previous changes reduced liveness, and engineering identified a better solution.
Note:end

PR #383, opened 2024-07-19
Fix sign-compare warnings in forked_tests_savanna.cpp

linh2931:
Fix a couple of sign compare warnings like

```
unittests/forked_tests_savanna.cpp:232:4:   required from here
libraries/boost/libs/test/include/boost/test/tools/old/impl.hpp:107:17:  warning: comparison of integer expressions of different signedness: ‘const long unsigned int’ and      ‘const int’ [-Wsign-compare]
  107 |     return left == right;
```


greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Fix sign-compare warnings in tests.
Note:end

PR #391, opened 2024-07-23
Finalizer transition requires QCs on both finalizer policy sets

### LINKED ISSUE #376: Policy transitions should use joint policies
When transitioning from one finalizer policy to another there is a stage in which the next policy is in the pending stage before becoming active. During this time, the code as it currently exists requires only QCs from the active finalizer policy to make progress with finality. However, for safety reasons, we require QCs from both the active and pending finalizer policies.

The rules for when the pending policy gets promoted to active still remain the same.

The QC claims also remain the same. There is still only one claim on a prior block (by block number) with an additional flag to indicate whether that QC was weak or strong. There are no changes to the `block_header`. This remains true even if that claimed block required a joint policy (i.e. had both an active policy and a pending policy) and therefore two different QCs are required to be attached to the block to validate its claim.

When a new QC claim is made against a block with a joint policy, the following needs to be true of `quorum_certificate` included in the block extension:
+ two QCs are needed, one for the active policy and one for the pending policy;
+ both QCs need to be valid, however they can independently be weak or strong QCs.

If the QC claim is that it is strong, then both the QCs attached in the block extension need to be strong QCs. If the QC claim is that it is weak, then there are no additional requirements on the two QCs other than they are valid QCs (weak or strong) that meet their respective thresholds.

If a new QC claim is made against a block with a single policy, then the `quorum_certificate` included in the block extension should only have a QC for the active policy.

We still only expect (and in fact require) at most one vote message per block from each finalizer even if the block has a joint policy in which the finalizer exists in both the active and pending policy. But when integrating the signatures from the validated vote message into pending QCs, they must be integrated appropriately into the pending QC for both the active policy and the pending policy. For vote messages on a block with a joint policy that are from finalizers that are only in one of the two policies, the signature should only be integrated into the appropriate pending QC (either active or pending).

When producing a block and searching for the best QC in its branch to include in the new block, the search must respect the rules for joint policies. If a block has a joint policy and only one of its pending QCs are valid, then it cannot be considered for inclusion as the claim for the new block.

Add a new field `last_pending_finalizer_policy_start_num` at the same level as `last_pending_finalizer_policy_digest` which tracks the block number at which the last pending finalizer policy first became pending. This is the second level which should be a struct with the name `level_2_commitments`. For now there are only two levels and so the `base_digest` will also be in this second level. Later we will add a third-level.

heifner:
Require a QC (quorum certificate) on both finalizer policy sets (active and pending) before considering a QC to have reached quorum on a block with a pending finalizer policy change.

Includes refactoring:
- `quorum_certificate.?pp` => `qc.?pp`
- `quorum_certificate` => `qc_t`
- `valid_quorum_certificate` => `qc_sig_t`
- `vote_status` => `vote_result_t` to differentiate from `vote_status_t`
- `pending_quorum_certificate` => `aggregating_qc_sig_t` Use `aggregating` instead of `pending` to avoid confusion with pending finalizer policy
- `aggregating_qc_t` manages the two `aggregating_qc_sig_t`, one for active policy sig and optionally one for pending policy sig.
- Move QC processing out of `block_state` into QC types

Adds `last_pending_finalizer_policy_start_num` for tracking when pending finalizer policy first became pending.

Resolves #376


libraries/chain/include/eosio/chain/finality/qc.hpp
```diff
@@ -0,0 +1,261 @@
...
+      bool received_qc_is_strong() const;
+      vote_status aggregate_vote(uint32_t connection_id, const vote_message& vote,
+                                 block_num_type block_num, std::span<const uint8_t> finalizer_digest);
+      has_vote_status_t has_voted(const bls_public_key& key) const;
+      bool is_quorum_met() const;
+
+   private:
+      friend struct fc::reflector<open_qc_t>;
+      finalizer_policy_ptr         active_finalizer_policy;  // not modified after construction
+      finalizer_policy_ptr         pending_finalizer_policy; // not modified after construction
```
greg7mdp, line 247: why not make them const if they are not modified after construction

libraries/chain/include/eosio/chain/finality/qc.hpp
```diff
@@ -0,0 +1,261 @@
...
+            for (size_t i = 0; i < bitset.size(); ++i) {
+               if (bitset[i])
+                  cb(i);
+            }
+         }
+      };
+
+      open_qc_sig_t();
+
+      explicit open_qc_sig_t(size_t num_finalizers, uint64_t quorum, uint64_t max_weak_sum_before_weak_final);
```
greg7mdp, line 145: Why not just have a constructor taking a `const finalizer_policy_ptr&` to avoid duplicated code at lines 218-226

libraries/chain/include/eosio/chain/finality/qc.hpp
heifner: That was my first thought as well. Unfortunately, it is used in `FC_REFLECT`
```c++
/home/heifner/ext/leap/libraries/libfc/include/fc/io/raw.hpp:250:8: error: static assertion failed due to requirement '!std::is_same_v<const std::shared_ptr<eosio::chain::finalizer_policy>, const std::shared_ptr<eosio::chain::finalizer_policy>>': can't unpack const type
       static_assert(not std::is_same_v<const T, const T>, "can't unpack const type");
```

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,505 @@
...
+         return std::optional{qc_t{ block_num, std::move(*active_best_qc), {} }};
+      }
+      return {};
+   }
+   // consider both
+   std::optional<qc_sig_t> pending_best_qc = pending_policy_sig->get_best_qc();
+   if (pending_best_qc) {
+      return std::optional{qc_t{ block_num, std::move(*active_best_qc), std::move(*pending_best_qc) }};
+   }
+   return {};
```
greg7mdp, line 310: I think it could be simplified to:

```suggestion
   std::optional<qc_sig_t> active_best_qc = active_policy_sig.get_best_qc();
   if (!active_best_qc)
      return {};

   if (pending_policy_sig) 
      return std::optional{qc_t{ block_num, std::move(*active_best_qc), pending_policy_sig->get_best_qc() });

   return std::optional{qc_t{ block_num, std::move(*active_best_qc), {} }};
```

heifner: If there is a `pending_policy_sig` then it must return a best_qc or the pair does not have a best_qc. But it can be simpler in the same spirit of what you have.

greg7mdp: If there is no `active_best_qc`, regardless of whether there is a `pending_policy_sig` or not, a qc will not be achieved, right?

heifner: Correct.

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,503 @@
...
+   qc_sig_t qc_sig_from_open = extract_qc_sig_from_open();
+
+   // if received_qc_sig does not have value, consider valid_qc_sig_from_open only
+   if( !received_qc_sig ) {
+      return std::optional{std::move(qc_sig_from_open)};
+   }
+
+   // Both received_qc_sig and qc_sig_from_open have value. Compare them and select a better one.
+   // Strong beats weak. Tie-break by received_qc_sig, strong beats weak
+   bool use_received_qc_sig = received_qc_sig->is_strong() || (received_qc_sig->is_weak() && qc_sig_from_open.is_weak());
```
greg7mdp, line 274: If it is not strong, it has to be weak :-)

```suggestion
   bool use_received_qc_sig = received_qc_sig->is_strong() || qc_sig_from_open.is_weak();
```

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,502 @@
...
+   // Strong beats weak. Tie-break by received_qc_sig, strong beats weak
+   if (received_qc_sig->is_strong() || qc_sig_from_open.is_weak()) {
+      return std::optional{qc_sig_t{ *received_qc_sig }};
+   }
+   return std::optional{qc_sig_t{ std::move(qc_sig_from_open) }};
+}
+
+void open_qc_sig_t::set_received_qc_sig(const qc_sig_t& qc) {
+   std::lock_guard g(*_mtx);
+   received_qc_sig = qc;
```
greg7mdp, line 282: I believe we may already hold a `received_qc_sig`. If so, we should check which one is the strongest one.

libraries/chain/include/eosio/chain/finality/qc.hpp
```diff
@@ -0,0 +1,256 @@
...
+      qc_sig_t                  active_policy_sig;
+      // signatures corresponding to the pending finalizer policy if there is one
+      std::optional<qc_sig_t>   pending_policy_sig;
+
+      bool is_strong() const {
+         return active_policy_sig.is_strong() && (!pending_policy_sig || pending_policy_sig->is_strong());
+      }
+
+      bool is_weak() const {
+         return active_policy_sig.is_weak() || (pending_policy_sig && pending_policy_sig->is_weak());
```
greg7mdp, line 77:
```suggestion
         return !is_strong();
```

libraries/chain/finality/qc.cpp
heifner: It does seem like we should save the best qc.

libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -111,13 +108,25 @@ namespace eosio::chain {
...
+               if (itr != bsp->pending_finalizer_policy->second->finalizers.end()) {
+                  if (std::ranges::find_if(votes, [&](const vote_message_ptr& vote_msg) { return vote_msg->finalizer_key == f.first; }) == votes.end()) {
+                     vote_message_ptr vote_msg = f.second.maybe_vote(f.first, bsp, bsp->strong_digest);
+                     if (vote_msg)
+                        votes.push_back(std::move(vote_msg));
+                  }
+               }
             }
          }
+
```
greg7mdp, line 129: I think we can make it simpler by voting iff the local finalizer is in active or pending policy.

```suggestion
         auto in_policy = [](const finalizer_policy_ptr& policy, const bls_public_key& key ){
            return std::ranges::any_of(policy->finalizers, [&](const auto& fin) { return fin.public_key == key; });
         };

         for (auto& f : finalizers) {
            if (in_policy(bsp->active_finalizer_policy, f.first) ||
                (bsp->pending_finalizer_policy && in_policy(bsp->pending_finalizer_policy->second, f.first))) {
               vote_message_ptr vote_msg = f.second.maybe_vote(f.first, bsp, bsp->strong_digest);
               if (vote_msg)
                  votes.push_back(std::move(vote_msg));
            }
         }
```

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+qc_vote_metrics_t open_qc_t::vote_metrics(const qc_t& qc) const {
+   qc_vote_metrics_t result;
+
+   auto add_votes = [&](const finalizer_policy_ptr& finalizer_policy, const auto& votes, std::set<finalizer_authority_ptr>& results) {
+      assert(votes.size() == finalizer_policy->finalizers.size());
+      size_t added = 0;
+      for (size_t i = 0; i < votes.size(); ++i) {
+         if (votes[i]) {
+            results.insert(finalizer_authority_ptr{finalizer_policy, &finalizer_policy->finalizers[i]}); // use aliasing shared_ptr constructor
```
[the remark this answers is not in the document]
heifner: It is guaranteed in this case, finalizers are owned by the finalizer_policy. " typical use cases where ptr is a member of the object managed by r" which is exactly the case here.

libraries/chain/include/eosio/chain/finality/qc.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+      bool is_quorum_met_no_lock() const;
+      qc_sig_t extract_qc_sig_from_open() const;
+   };
+
+   // finalizer authority of strong, weak, or missing votes
+   struct qc_vote_metrics_t {
+      std::set<finalizer_authority_ptr> strong_votes;
+      std::set<finalizer_authority_ptr> weak_votes;
+      std::set<finalizer_authority_ptr> missing_votes;
```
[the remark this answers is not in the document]
heifner: I need to provide a less operator.

libraries/chain/finality/qc.cpp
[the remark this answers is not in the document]
greg7mdp: Oh yes, great use case, thanks!

libraries/chain/include/eosio/chain/finality/qc.hpp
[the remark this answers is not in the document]
greg7mdp: Why use `std::set` and not `std::vector`?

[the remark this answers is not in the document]
heifner: Because there will almost always be duplicates from pending when pending is present.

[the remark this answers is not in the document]
greg7mdp: Will the `shared_ptr` created from the active and pending policies be the same?

libraries/chain/include/eosio/chain/finality/finalizer.hpp
heifner: Very nice.

unittests/svnn_ibc_tests.cpp
```diff
@@ -574,12 +588,18 @@ BOOST_AUTO_TEST_SUITE(svnn_ibc)
       // The QC provided to prove this also proves a commitment from finalizers to this policy, so the smart contract can accept it.
       action_trace check_heavy_proof_3_trace = cluster.node0.push_action("ibc"_n, "checkproof"_n, "ibc"_n, heavy_proof_3)->action_traces[0];
 
+      BOOST_CHECK(true);
```
greg7mdp, line 591: ???

libraries/chain/include/eosio/chain/finality/qc.hpp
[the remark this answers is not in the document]
heifner: No, which is why I needed to add a less compare to the set. I forgot to do that originally.

unittests/svnn_ibc_tests.cpp
```diff
@@ -574,12 +588,18 @@ BOOST_AUTO_TEST_SUITE(svnn_ibc)
       // The QC provided to prove this also proves a commitment from finalizers to this policy, so the smart contract can accept it.
       action_trace check_heavy_proof_3_trace = cluster.node0.push_action("ibc"_n, "checkproof"_n, "ibc"_n, heavy_proof_3)->action_traces[0];
 
+      BOOST_CHECK(true);
+
       // now that we have successfully proven finalizer policy generation #2, the contract has it, and we can prove heavy_proof_4
       action_trace check_heavy_proof_4_trace = cluster.node0.push_action("ibc"_n, "checkproof"_n, "ibc"_n, heavy_proof_4)->action_traces[0];
 
+      BOOST_CHECK(true);
```
greg7mdp, line 596: ???

unittests/svnn_ibc_tests.cpp
```diff
@@ -574,12 +588,18 @@ BOOST_AUTO_TEST_SUITE(svnn_ibc)
...
       // now that we have successfully proven finalizer policy generation #2, the contract has it, and we can prove heavy_proof_4
       action_trace check_heavy_proof_4_trace = cluster.node0.push_action("ibc"_n, "checkproof"_n, "ibc"_n, heavy_proof_4)->action_traces[0];
 
+      BOOST_CHECK(true);
+
       // we now test light proof we should still be able to verify a proof of finality for block #2 without finality proof,
       // since the previous root is still cached
       cluster.node0.push_action("ibc"_n, "checkproof"_n, "ibc"_n, light_proof_1);
+      
+      BOOST_CHECK(true);
```
greg7mdp, line 602: ???

unittests/finality_proof.hpp
```diff
@@ -247,7 +255,7 @@ namespace finality_proof {
          qc_data_t qc_data = extract_qc_data(block);
 
          // return relevant IBC information
-         return {block, qc_data, onblock_trace, finality_data, active_finalizer_policy.generation, last_pending_finalizer_policy.generation, last_proposed_finalizer_policy.generation, action_mroot, base_digest, active_finalizer_policy_digest, last_pending_finalizer_policy_digest, last_proposed_finalizer_policy_digest, finality_digest, level_3_commitments_digest, level_2_commitments_digest, finality_leaf, finality_root };
```
greg7mdp, line 250: to make it more error-prone, can we use the syntax with the field names (because many members are the same type `digest_type`, if we add a digest_type in the middle the return would still work but be incorrect.

unittests/svnn_ibc_tests.cpp
systemzax: Sorry for that. I added these checkpoints to verify which push_action calls were failing (ctest lets you know which was the last valid checkpoint), and forgot to remove them. I removed them now

unittests/finality_proof.hpp
systemzax: Fixed.

greg7mdp: Nice!

libraries/chain/include/eosio/chain/finality/qc.hpp
linh2931: Maybe add a comment saying why `const` is not used.

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,508 @@
...
+   if (pending_finalizer_policy) {
+      qc.pending_policy_sig->verify(pending_finalizer_policy, strong_digest, weak_digest);
+   }
+}
+
+bool open_qc_t::set_received_qc(const qc_t& qc) {
+   bool active_better = active_policy_sig.set_received_qc_sig(qc.active_policy_sig);
+   bool pending_better = false;
+   if (qc.pending_policy_sig) {
+      assert(pending_policy_sig);
```
greg7mdp, line 333: Probably should not be `assert`, maybe `FC_ASSERT`?

libraries/chain/controller.cpp
```diff
@@ -3958,34 +3955,32 @@ struct controller_impl {
...
-      // claimed->valid_qc_is_strong() acquires a mutex.
-      if (received_qc.is_weak() || claimed->valid_qc_is_strong()) {
-         dlog("qc not better, claimed->valid: ${qbn} ${qid}, strong=${s}, received: ${rqc}, for block ${bn} ${id}",
-              ("qbn", claimed->block_num())("qid", claimed->id())("s", !received_qc.is_weak()) // use is_weak() to avoid mutex on valid_qc_is_strong()
+      // Don't save the QC from block extension if the claimed block has a better or same received_qc
+      if (!claimed->set_received_qc(received_qc)) {
+         dlog("qc not better, claimed->received: ${qbn} ${qid}, strong=${s}, received: ${rqc}, for block ${bn} ${id}",
+              ("qbn", claimed->block_num())("qid", claimed->id())("s", !received_qc.is_weak()) // use is_weak() to avoid mutex on received_qc_is_strong()
               ("rqc", qc_ext.qc.to_qc_claim())("bn", bsp_in->block_num())("id", bsp_in->id()));
          return;
```
greg7mdp, line 3980: We should probably not `return` here as it makes sense to still do the `maybe_update_fsi()` below unless the incoming qc is exactlt the same we already had.

libraries/chain/finality/finalizer.cpp
```diff
@@ -154,8 +151,13 @@ void my_finalizers_t::maybe_update_fsi(const block_state_ptr& bsp, const valid_q
 
    bool updated = false;
    for (auto& f : finalizers) {
-      if (has_voted_strong(bsp->active_finalizer_policy->finalizers, received_qc, f.first)) {
+      if (has_voted_strong(bsp->active_finalizer_policy->finalizers, received_qc.active_policy_sig, f.first)) {
          updated |= f.second.maybe_update_fsi(bsp);
+      } else if (bsp->pending_finalizer_policy) {
+         assert(received_qc.pending_policy_sig);
```
greg7mdp, line 157: I don't think it should be an `assert` here, maybe:

```suggestion
      } else if (bsp->pending_finalizer_policy && received_qc.pending_policy_sig) {
```

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,508 @@
...
+      return vote_status::success;
+   };
+
+   auto add_vote = [&](const finalizer_policy_ptr& finalizer_policy, open_qc_sig_t& open_qc_sig) -> vote_status {
+      const auto& finalizers = finalizer_policy->finalizers;
+      auto itr = std::ranges::find_if(finalizers, [&](const auto& finalizer) { return finalizer.public_key == vote.finalizer_key; });
+      vote_status s = vote_status::unknown_public_key;
+      if (itr != finalizers.end()) {
+         auto index = std::distance(finalizers.begin(), itr);
+         if (open_qc_sig.has_voted(vote.strong, index)) {
```
greg7mdp, line 366: We should use the `has_voted` version which does not take a `bool strong` first param. After all we should not accept both a weak and strong vote from the same finalizer.

greg7mdp (changes requested):

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -110,6 +111,9 @@ struct block_header_state {
    // in the history of the blockchain so far that is not in proposed state (so either pending or active state)
    digest_type                         last_pending_finalizer_policy_digest;
 
+   // tracks the block number at which the last pending finalizer policy first became pending
```
linh2931, line 114: If no pending policy, does this use active policy's block number (to match `last_pending_finalizer_policy_digest` semantics) or the name should be changed.

If for real pending, we don't even need it as a separate field, you can find it from `pending_finalizer_policy`'s first member.

libraries/chain/include/eosio/chain/finality/qc.hpp
```diff
@@ -0,0 +1,262 @@
...
+                               // disaster recovery), we may not have a qc_t so we use the `lib` block_num
+                               // and specify `weak`.
+   };
+
+   /**
+    * All public methods are thread-safe.
+    * Used for incorporating votes into a qc signature.
+    * "open" in that it allows new votes to be added at any time.
+    */
+   class open_qc_sig_t {
```
linh2931, line 95: I feel `pending_qc_sig_t` and `pending_qc` are more appropriate. `pending` means something in progress; that's what they are. No need to worry about it might be confused with pending finalizer policy. We use pending in other cases in the code base too.

linh2931 (commented):
Will add more comments while reviewing.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Require a QC (quorum certificate) on both finalizer policy sets (active and pending) before considering a QC to have reached quorum on a block with a pending finalizer policy change.
Note:end

libraries/chain/controller.cpp
heifner: Excellent point. That is a really good improvement over what we had before.

libraries/chain/finality/finalizer.cpp
heifner: As in the above case, the received qc should be as expected and should have already been validated. But I'll add an `EOS_ASSERT` to be paranoid.

libraries/chain/include/eosio/chain/block_header_state.hpp
heifner: From Areg:

> It is always the block number at which the last pending policy first was promoted to pending.
> 
> If the last pending policy is the current active policy, then it is the block number at which that active policy first was promoted to pending.
> 
> If the last pending policy is the genesis policy (and therefore it went straight to active without a pending stage), then it is the genesis block number.

libraries/chain/include/eosio/chain/finality/qc.hpp
heifner: Renamed to `in_progress_qc_sig_t` and `in_progress_qc_t`.

libraries/chain/include/eosio/chain/block_header_state.hpp
greg7mdp: Let's make sure we add this in the comment.

libraries/chain/include/eosio/chain/finality/qc.hpp
heifner: Renamed to `aggregating_qc_sig_t` and `aggregating_qc_t`.

greg7mdp (approved):

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,507 @@
...
+std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) const {
+   std::optional<qc_sig_t> active_best_qc = active_policy_sig.get_best_qc();
+   if (!active_best_qc) // active is always required
+      return {};
+
+   if (pending_policy_sig) {
+      std::optional<qc_sig_t> pending_best_qc = pending_policy_sig->get_best_qc();
+      if (pending_best_qc)
+         return std::optional<qc_t>{qc_t{block_num, std::move(*active_best_qc), std::move(pending_best_qc)}};
+      return {}; // no quorum on pending_policy_sig so no qc for this block
```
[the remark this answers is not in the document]
heifner: Because that is not a valid QC if there is a pending finalizer policy, we must reach quorum on pending as well as on the active finalizer policy otherwise we do not have a valid QC.

libraries/chain/finality/qc.cpp
```diff
@@ -0,0 +1,507 @@
...
+      }
+      return vote_status_t::irrelevant_finalizer;
+   };
+
+   vote_status_t active_status = finalizer_has_voted(active_finalizer_policy, active_policy_sig, key);
+   if (!pending_finalizer_policy || active_status == vote_status_t::not_voted) {
+      return active_status;
+   }
+
+   EOS_ASSERT(pending_policy_sig, invalid_qc_claim,
```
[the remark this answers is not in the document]
heifner: If a pending finalizer policy does not exist then the function is exited at the return on line 423.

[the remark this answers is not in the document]
linh2931: Oh my bad. You are correct.

linh2931 (approved):

PR #392, opened 2024-07-23
Add `timestamp` and `parent_timestamp` to finality leaf nodes.

greg7mdp:
Each finality leaf node commits to that block's timestamp (called `timestamp`), and also the timestamp of the parent block as well (called `parent_timestamp`). This data is always accessible from the `finality_core` (except for the genesis block). For the genesis block, the `parent_timestamp` is simply be the the earliest representable timestamp.

IBC unit tests have been updated to that this addition into account (thanks Guillaume!).


libraries/chain/block_state.cpp
```diff
@@ -103,10 +103,13 @@ block_state_ptr block_state::create_if_genesis_block(const block_state_legacy& b
 
    // build leaf_node and validation_tree
    valid_t::finality_leaf_node_t leaf_node {
-      .block_num       = bsp.block_num(),
-      .finality_digest = result.strong_digest,
-      .action_mroot    = *bsp.action_mroot_savanna
+      .block_num        = bsp.block_num(),
+      .timestamp        = bsp.timestamp().to_time_point(),
+      .parent_timestamp = block_timestamp_type().to_time_point(), // for the genesis block, the parent_timestamp is the the earliest representable timestamp.
```
heifner, line 108: I think both of these should be `block_timestamp_type` instead of `time_point`. `fc::time_point` is a `int64_t` where `block_timestamp_type` is a `uint32_t`. Also seems more reflective of what this is. It is a block time stamp not a time point.

heifner (changes requested):

libraries/chain/block_state.cpp
greg7mdp: image

heifner: @systemzax I believe this should be block_timestamp_type, what serialization issues does it cause?

systemzax: Originally, @greg7mdp had them as block_timestamp_type, but we don't really have an equivalent in CDT, do we?

If I'm going to work with time_point on the CDT side, it seems easier to serialize them this way already, rather than have to create a custom serializer on the contract side?

heifner: There is `eosio::block_timestamp` in `<eosio/time.hpp>`

systemzax: Oh somehow I missed that! Thanks, will use this instead

systemzax: Addressed here : https://github.com/AntelopeIO/spring/pull/396/commits/db6248f1e170314b2fd95f98fd906a66d05ecf50

greg7mdp: @heifner I reverted to timestamps in the PR and disabled the `svnn_ibc` tests. I think the PR is in a state where it can be merged.

unittests/svnn_ibc_tests.cpp
```diff
@@ -18,6 +18,13 @@ using namespace eosio::testing;
 
 using mvo = mutable_variant_object;
 
+#if 1
+
+BOOST_AUTO_TEST_SUITE(svnn_ibc)
```
heifner, line 23: This doesn't work because of the way cmake greps the tests.

heifner (approved):

PR #393, opened 2024-07-23
don't use a VLA to avoid clang warning

spoonincode:
clang18 flags this with `warning: variable length arrays in C++ are a Clang extension`.


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Do not use variable length arrays to avoid clang warning.
Note:end

PR #395, opened 2024-07-24
add progress logs during snapshot loading

spoonincode:
Adds progress logs during snapshot loading; can look something like
```
info  2024-07-24T03:30:21.700 nodeos    controller.cpp:1831           startup              ] Starting initialization from snapshot and no block log, this may take a significant amount of time
info  2024-07-24T03:30:26.008 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 2% complete
info  2024-07-24T03:30:31.001 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 4% complete
info  2024-07-24T03:30:36.003 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 17% complete
info  2024-07-24T03:30:41.008 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 28% complete
...
info  2024-07-24T03:31:31.029 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 90% complete
info  2024-07-24T03:31:36.009 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 92% complete
info  2024-07-24T03:31:41.019 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 93% complete
info  2024-07-24T03:31:46.035 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 96% complete
info  2024-07-24T03:31:51.017 nodeos    snapshot.hpp:418              progress             ] Snapshot initialization 99% complete
info  2024-07-24T03:31:53.283 nodeos    controller.cpp:1838           startup              ] Snapshot loaded, lib: 310859547
```

Resolves AntelopeIO/leap#1167

### DIFF
--- a/libraries/chain/include/eosio/chain/snapshot.hpp
+++ b/libraries/chain/include/eosio/chain/snapshot.hpp
@@ -277,6 +277,8 @@ namespace eosio { namespace chain {
 
       virtual void return_to_header() = 0;
 
+      virtual size_t total_row_count() = 0;
+
       virtual ~snapshot_reader(){};
 
       protected:
@@ -313,6 +315,7 @@ namespace eosio { namespace chain {
          bool empty ( ) override;
          void clear_section() override;
          void return_to_header() override;
+         size_t total_row_count() override;
 
       private:
          const fc::variant& snapshot;
@@ -364,6 +367,7 @@ namespace eosio { namespace chain {
          bool empty ( ) override;
          void clear_section() override;
          void return_to_header() override;
+         size_t total_row_count() override;
 
       private:
          bool validate_section() const;
@@ -385,6 +389,7 @@ namespace eosio { namespace chain {
          bool empty ( ) override;
          void clear_section() override;
          void return_to_header() override;
+         size_t total_row_count() override;
 
       private:
          bool validate_section() const;
@@ -406,4 +411,16 @@ namespace eosio { namespace chain {
 
    };
 
+   struct snapshot_loaded_row_counter {
+      snapshot_loaded_row_counter(const size_t total) : total(total) {}
+      void progress() {
+         if(++count % 50000 == 0 && time(NULL) - last_print >= 5) {
+            ilog("Snapshot initialization ${pct}% complete", ("pct",(unsigned)(((double)count/total)*100)));
+            last_print = time(NULL);
+         }
+      }
+      size_t count = 0;
+      const size_t total = 0;
+      time_t last_print = time(NULL);
+   };
 }}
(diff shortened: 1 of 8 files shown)


unittests/snapshot_tests.cpp
```diff
@@ -61,6 +61,7 @@ class snapshotted_tester : public base_tester {
       controller::config copied_config = (copy_files_from_config == copy_config_files)
                                          ? copy_config_and_files(config, ordinal) : copy_config(config, ordinal);
 
+      BOOST_CHECK_GT(snapshot->total_row_count(), 0);
```
spoonincode, line 64: This is good enough to make sure all three impls of `total_row_count()` are sane (and indeed I found my rapidjson impl was incorrect initially). I don't think we need to actually make sure the exact count is accurate and precise.

libraries/chain/include/eosio/chain/snapshot.hpp
```diff
@@ -406,4 +411,16 @@ namespace eosio { namespace chain {
...
+      void progress() {
+         if(++count % 50000 == 0 && time(NULL) - last_print >= 5) {
+            ilog("Snapshot initialization ${pct}% complete", ("pct",(unsigned)(((double)count/total)*100)));
+            last_print = time(NULL);
+         }
+      }
+      size_t count = 0;
+      const size_t total = 0;
+      time_t last_print = time(NULL);
+   };
```
spoonincode, line 425: It's a little tempting to move this completely inside of snapshot_reader so that maybe the onus to call `progress()` remains entirely inside of the `snapshot_reaader` & friends (which would probably help ensuring there isn't missed calls). But `snapshot_reader`'s interface technically allows you to read the same section multiple times (even though nobody does that..) so, eh.. not a perfect match.

heifner: I could see a desire for a user to provide their own version of `snapshot_loaded_row_counter` say for diff logging purposes. Keeping it external would make that easier. I don't think it worth making it a template or derivable type at this point until that is needed.

heifner (approved):

greg7mdp (approved):

libraries/chain/include/eosio/chain/snapshot.hpp
spoonincode: Good idea buuuutt.. this all probably needs to be refactored again when making loading parallel so that might not be possible. I didn't mention that this may be somewhat temporary since I'm not committing to parallel loading just yet..

ericpassmore:
Note:start
group: LOGGING
category: INTERNALS
summary: Log progress during snapshot loading.
Note: end

PR #336, opened 2024-07-01
Update boost submodule to tag `boost-1.86.0`.

greg7mdp:
This updates spring to use the [latest boost version](https://www.boost.org/users/history/) .

PR #400, opened 2024-07-24
update pinned build to clang 18.1.8

spoonincode:
Update the pinned build to clang 18.1.8. Normally I wouldn't have advocated for moving from 17->18 but I stumbled on a c++20 feature that while it's in libstdc++10, is still missing from libc++17 but just added to libc++18: `jthread`. There seems to be an appetite to take on environment upgrades (e.g. #336) so nice to fill out c++20 support more before 1.0 release.

I measured something around a 0.5% speed improvement for a replay over blocks 381452700 through 381632370 (sadly I didn't think to include #344 as part of that; though at least we know #344 wouldn't have been resolved just by updating to clang18).

The upgrade may also unlock another useful feature: `_LIBCPP_HARDENING_MODE` was added in 18 and is similar to `_GLIBCXX_ASSERTIONS` in enabling stdlib hardening checks intended to be operational in production. To be explored still.

This PR also tweaks the Dockerfile to remove intermediary build files at each step instead of at the end. This reduces the size of Docker's build cache from ~20GB to ~14GB

### DIFF
--- a/tools/reproducible.Dockerfile
+++ b/tools/reproducible.Dockerfile
@@ -56,7 +56,8 @@ RUN tar xf cmake-*.tar.gz && \
     cd cmake*[0-9] && \
     echo 'set(CMAKE_USE_OPENSSL OFF CACHE BOOL "" FORCE)' > spring-init.cmake && \
     ./bootstrap --parallel=$(nproc) --init=spring-init.cmake --generator=Ninja && \
-    ninja install
+    ninja install && \
+    rm -rf cmake*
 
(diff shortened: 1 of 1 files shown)


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: CHORE
summary: Upgrade to clang v18 in Docker repeatable build. Improvements to reduce Docker build cache.
Note:end

Issue #409, opened 2024-07-25
Add pending policy generation number to finality digest


arhag:
Add `last_pending_finalizer_policy_generation` to `finality_digest_data_v1` (which is used in the finality digest calculation). If a pending policy is currently present, then this field will hold the generation number of that pending policy. Otherwise, the field will hold the same number as `active_finalizer_policy_generation`.

`last_pending_finalizer_policy_generation` should also be added into `finality_data_t` which is sent via SHiP.

PR #410, opened 2024-07-25
Add pending policy generation number to finality digest

linh2931:
* Add pending_finalizer_policy_generation to finality_digest_data_v1. If a pending policy is currently present, this field holds the generation number of that pending policy. Otherwise, the field holds the same number as active_finalizer_policy_generation.
* Add pending_finalizer_policy_generation SHiP finality_data_t .
* Remove timestamp from finality_data_t as it can be retrieved from block.

Resolve https://github.com/AntelopeIO/spring/issues/409


heifner (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -63,6 +63,7 @@ digest_type block_header_state::compute_finality_digest() const {
    assert(active_finalizer_policy);
    finality_digest_data_v1 finality_digest_data {
       .active_finalizer_policy_generation  = active_finalizer_policy->generation,
+      .pending_finalizer_policy_generation = get_last_pending_finalizer_policy().generation,
```
greg7mdp, line 66: I think this member should be called `last_pending_finalizer_policy_generation`

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -41,7 +41,12 @@ struct level_2_commitments_t {
 struct finality_digest_data_v1 {
    uint32_t    major_version{light_header_protocol_version_major};
    uint32_t    minor_version{light_header_protocol_version_minor};
-   uint32_t    active_finalizer_policy_generation {0};
+   uint32_t    active_finalizer_policy_generation{0};
+
+   // Hold the generation number of pending_finalizer_policy if it exists,
+   // otherwise hold the same value as active_finalizer_policy_generation
+   uint32_t    pending_finalizer_policy_generation{0};
```
greg7mdp, line 48: If called `last_pending_finalizer_policy_generation` we can just have a short comment after the semicolon.

greg7mdp (approved):

systemzax (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Add last_pending_finalizer_policy_generation into the finality digest calculation.
Note:end

PR #414, opened 2024-07-26
Use last_pending_finalizer_policy_start_timestamp instead of last_pending_finalizer_policy_start_num

### LINKED ISSUE #412: Use last_pending_finalizer_policy_start_timestamp instead of last_pending_finalizer_policy_start_num
Instead of `last_pending_finalizer_policy_start_num` we should use `last_pending_finalizer_policy_start_timestamp`. 

Then if `last_pending_finalizer_policy_start_timestamp <= latest_qc_claim_timestamp` and `active_finalizer_policy_generation != last_pending_finalizer_policy_generation` for a block (call it block A) that has a strong QC on it, we know that this strong QC, if claimed in a block (call it block B), would prove that the last pending finalizer policy referenced by the `last_pending_finalizer_policy_generation` in block A would become the active finalizer policy in block B.

With this change `latest_qc_claim_block_num` might not even be necessary in `level_3_commitments_t`, but it may be best to not get rid of it yet.

There is no harm in using the timestamp version rather than the number version and it is more consistent with other places in our implementation where we prefer timestamp over block numbers. 

Furthermore, committing to the `last_pending_finalizer_policy_start_timestamp` in the finalizer digest may be useful in proving finality violations in certain cases; however it may not be sufficient (we might require a `last_pending_finalizer_policy_finality_digest` as well).

greg7mdp:
Resolves #412 

See issue for background explanation for this change.


libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -115,10 +115,10 @@ struct block_header_state {
    // in the history of the blockchain so far that is not in proposed state (so either pending or active state)
    digest_type                         last_pending_finalizer_policy_digest;
 
-   // Block number at which the last pending finalizer policy first was promoted to pending.
-   // If the last pending finalizer policy is the current active finalizer policy, then it is the block number at which
+   // Timestamp of the block at which the last pending finalizer policy first was promoted to pending.
+   // If the last pending finalizer policy is the current active finalizer policy, then it is the timestamp of the block at which
    // that active finalizer policy first was promoted to pending. Savanna genesis block it is the genesis block number.
-   block_num_type                      last_pending_finalizer_policy_start_num {0};
+   block_timestamp_type                      last_pending_finalizer_policy_start_timestamp;
```
heifner, line 121: Minor: indentation.

heifner (approved):

linh2931 (approved):

ericpassmore:
last_pending_finalizer_policy_start_timestamp
Note:start
group: STABILITY
category: INTERNALS
summary: Switch to timestamp for better detection of most recent policy change.
Note:end

PR #420, opened 2024-07-26
Only vote on recent blocks

### LINKED ISSUE #418: Do not vote on old blocks
Relevant checks we already have:
https://github.com/AntelopeIO/spring/blob/e2f0e9000f21edf3b65197914ac0938a048a301a/libraries/chain/controller.cpp#L3299
https://github.com/AntelopeIO/spring/blob/e2f0e9000f21edf3b65197914ac0938a048a301a/libraries/chain/controller.cpp#L4005
https://github.com/AntelopeIO/spring/blob/e2f0e9000f21edf3b65197914ac0938a048a301a/libraries/chain/include/eosio/chain/finality/finalizer.hpp#L96-L99

What we want is to not ever vote on old blocks (more than 30 seconds old relative to the wall clock time) regardless of the status of `enable_voting` and where the finalizers are active.

heifner:
Simplify vote restriction by only looking at block timestamp vs wall clock instead of flipping an `enable_voting` flag.
The `testing_allow_voting` change was needed because the tester creates blocks starting with `2020-01-01`.

Resolves #418

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -960,7 +960,7 @@ struct controller_impl {
    named_thread_pool<chain>        thread_pool;
    deep_mind_handler*              deep_mind_logger = nullptr;
    bool                            okay_to_print_integrity_hash_on_stop = false;
-   bool                            testing_allow_voting = true; // used in unit tests to create long forks or simulate not getting votes
+   bool                            testing_allow_voting = false; // used in unit tests to create long forks or simulate not getting votes
    async_t                         async_voting = async_t::yes;  // by default we post `create_and_send_vote_msg()` calls, used in tester
    async_t                         async_aggregation = async_t::yes; // by default we process incoming votes asynchronously
    my_finalizers_t                 my_finalizers;
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3296,7 +3296,7 @@ struct controller_impl {
 
             if (!my_finalizers.empty()) {
                block_handle_accessor::apply_s<void>(chain_head, [&](const auto& head) {
-                  if (head->is_recent() || my_finalizers.is_active()) {
+                  if (head->is_recent()) {
                      if (async_voting == async_t::no)
                         create_and_send_vote_msg(head);
                      else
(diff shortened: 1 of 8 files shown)


linh2931 (dismissed):

libraries/chain/controller.cpp
```diff
@@ -3296,7 +3296,7 @@ struct controller_impl {
 
             if (!my_finalizers.empty()) {
                block_handle_accessor::apply_s<void>(chain_head, [&](const auto& head) {
-                  if (head->is_recent() || my_finalizers.is_active()) {
+                  if (head->is_recent() || testing_allow_voting) {
```
greg7mdp, line 3299: This doesn't seem right. By default `testing_allow_voting` is true, which means that by default we allow voting on old blocks.

libraries/chain/controller.cpp
```diff
@@ -960,7 +960,7 @@ struct controller_impl {
    named_thread_pool<chain>        thread_pool;
    deep_mind_handler*              deep_mind_logger = nullptr;
    bool                            okay_to_print_integrity_hash_on_stop = false;
-   bool                            testing_allow_voting = false; // used in unit tests to create long forks or simulate not getting votes
+   bool                            testing_allow_voting = true; // used in unit tests to create long forks or simulate not getting votes
```
linh2931, line 963: Is this correct to set default to true?

heifner: Nope, quick response to a failed test. Addressing now. Thanks @greg7mdp & @linh2931

linh2931: Later we should have a test to verify old blocks are not voted.

linh2931 (approved):

greg7mdp:
Approving, but I'm a little uncomfortable that `testing_allow_voting`  is used for two completely separate purposes:

1. when producing a series of blocks in testing, vote on only the last 4 to save test time
2. to disable the controller feature of not voting on old blocks, because tests create blocks with old timestamps

I think it would be cleaner to have two separate states, or even better removing the test optimization (first_usage), in which case `testing_allow_voting` could be renamed `allow_voting_on_old_blocks`.

greg7mdp (approved):

heifner:
> 1. when producing a series of blocks in testing, vote on only the last 4 to save test time

If I remember correctly this made a big difference.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Only vote on recent blocks and simply voting by comparing time.
Note:end

PR #421, opened 2024-07-26
Fix a sign-compare warning in snapshot_tests.cpp

linh2931:
```
unittests/snapshot_tests.cpp:64:7:   required from here
libraries/boost/libs/test/include/boost/test/tools/old/impl.hpp:185:21:    warning: comparison of integer expressions of different signedness: ‘const long unsigned int’ and  ‘const int’ [-Wsign-compare]
  185 |         return left > right;
```


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Fix sign-compare warning in snapshot tests.
Note:end

PR #434, opened 2024-07-29
[1.0-beta4] Log when block has no QCs

heifner:
Helpful to distinguish between only missing one vote or no votes making into a block.
Add log of all recevied blocks with qc claim, previous, and timestamp.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3968,6 +3968,10 @@ struct controller_impl {
       auto qc_ext = bsp_in->block->extract_extension<quorum_certificate_extension>();
       const qc_t& received_qc = qc_ext.qc;
 
+      dlog("received block: #${bn} ${t} ${prod} ${id}, qc claim: ${qc}, previous: ${p}",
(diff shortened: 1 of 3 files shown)


linh2931 (approved):

ericpassmore:
Note:start
group: LOGGING
category: INTERNALS
summary: Add logging when block has no QCs.
Note:end

greg7mdp (approved):

libraries/chain/controller.cpp
```diff
@@ -3968,6 +3968,10 @@ struct controller_impl {
       auto qc_ext = bsp_in->block->extract_extension<quorum_certificate_extension>();
       const qc_t& received_qc = qc_ext.qc;
 
+      dlog("received block: #${bn} ${t} ${prod} ${id}, qc claim: ${qc}, previous: ${p}",
```
linh2931, line 3971: You can also log after line 3965 if the block does not contain a QC.

linh2931 (approved):

greg7mdp (approved):

PR #439, opened 2024-07-30
guard against throwing in `random_access_file`'s `write_datastream` dtor

spoonincode:
I encountered an unfortunate design deficiency in `random_access_file`'s `write_datastream`. Upon destruction `write_datastream` will flush its buffer via `do_write()` but that call can throw. When this occurs obviously it's Game Over.

This tweaks the interface to `write_datastream`: exceptions during dtor are swallowed (with a log warning); if a user of the class wants errors to be reported they must call a new `flush()` which still throws if writing the buffer fails.

I don't really like that and I'm open to other potential tweaks, but this pattern does have some precedent in `ofstream` (with `basic_filebuf`).

### DIFF
--- a/libraries/libfc/include/fc/io/random_access_file.hpp
+++ b/libraries/libfc/include/fc/io/random_access_file.hpp
@@ -30,7 +30,8 @@ read_ds() and write_ds() may be called from multiple threads simultaneously. The
  fc's pack/unpack. Multiple read_datastream and write_datastream created from the same random_access_file can be used simultaneously from different
  threads, but an individual read/write_datastream must only be used by a single thread at a time. unpack_from() and pack_to() are just
  simple helper functions based on these read/write_datastreams. Be aware that read/write_datastreams are buffered. Write buffers are flushed
- as needed, both upon destruction of the datastream and also when an internal buffer becomes full.
+ on an internal threshold, upon a call to flush(), and upon destruction of the datastream. If a buffered write fails during destruction an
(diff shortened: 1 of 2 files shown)


libraries/libfc/include/fc/io/random_access_file.hpp
```diff
@@ -419,12 +433,14 @@ class random_access_file {
    void pack_to(const T& v, const ssize_t offset) {
       write_datastream ds(ctx, offset);
       fc::raw::pack(ds, v);
+      ds.flush();
```
greg7mdp, line 436: Could the `ctx->write_to()` on each `pack_to()` call hurt performance?

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Guard against throwing an error write_deatastream's destructor. 
Note:end

libraries/libfc/include/fc/io/random_access_file.hpp
spoonincode: Not in this case: the buffer is/was flushed upon destruction of `write_datastream` so the buffer was already being flushed at the end of `pack_to()`.

spoonincode:
An alternative might be to actually allow `write_datastream`'s dtor to throw. I hadn't really considered that an option previously since that's considered a highly unusual (and discouraged) pattern, but it might be workable in this situation.

heifner (approved):

greg7mdp (approved):

PR #440, opened 2024-07-30
multi threaded snapshot loading

### LINKED ISSUE #437: Make snapshot process multi-threaded
The snapshot process seems to be single-threaded according to looking at "top". Perhaps there is opportunity to speed up the snapshot process by making it multi-threaded.

Originally: https://github.com/eosnetworkfoundation/mandel/issues/77

spoonincode:
Add multi threaded snapshot loading.

Different environments will have different performance uplifts, but in my environment with a recent EOS snapshot, loading time decreased from 90 seconds to 60 seconds.

This implements a very simple parallelization scheme where each snapshot section (which roughly corresponds to each chainbase table) is loaded in parallel. To get better parallelization we will need to load a single chainbase table in parallel since the contract table far outweighs the other tables. That is certainly doable, but far more complex.

Resolves #437


libraries/chain/controller.cpp
```diff
@@ -2300,21 +2303,53 @@ struct controller_impl {
...
 
-      read_contract_tables_from_snapshot(snapshot, row_counter);
+      snapshot_load_ctx.post([this,&snapshot,&rows_loaded]() {
+         read_contract_tables_from_snapshot(snapshot, rows_loaded);
+      });
+
+      authorization.read_from_snapshot(snapshot, rows_loaded, snapshot_load_ctx);
+      resource_limits.read_from_snapshot(snapshot, rows_loaded, snapshot_load_ctx);
+
+      constexpr unsigned max_snapshot_load_threads = 4;
```
spoonincode, line 2324: There is diminishing returns beyond this. Really even 2 or 3 is probably similar performance to 4 with current EOS data set.

libraries/chain/controller.cpp
```diff
@@ -2300,21 +2303,53 @@ struct controller_impl {
...
+      const unsigned snapshot_load_threads = snapshot->supports_threading() ? max_snapshot_load_threads : 1;
+
+      std::vector<std::promise<void>> thread_promises(snapshot_load_threads);
+      std::list<std::thread> threads;
+      //this scoped_exit can go away with jthread; but still marked experimental in libc++18
```
spoonincode, line 2329: I guess I goofed this up before making #400. I thought I tested locally with libc++18 I could use `jthread` but clearly that isn't the case.

ericpassmore:
Note:start
group: STABILITY
category: PERFORMANCE
summary: Reduce snapshot loading time by utilizing multiple threads. 
Note:end

libraries/chain/controller.cpp
```diff
@@ -2300,21 +2303,53 @@ struct controller_impl {
...
 
-      authorization.read_from_snapshot(snapshot, row_counter);
-      resource_limits.read_from_snapshot(snapshot, row_counter);
+      for(std::promise<void>& p : thread_promises) {
+         std::future<void> f = p.get_future();
+         while(f.wait_for(std::chrono::seconds(5)) != std::future_status::ready) {
+            ilog("Snapshot initialization ${pct}% complete", ("pct",(unsigned)(((double)rows_loaded/total_snapshot_rows)*100)));
+         }
+         f.get();
+      }
```
heifner, line 2352: Would be nice to package the thread pool run to completion into a class in thead_utils.hpp. Either enhance `named_thread_pool` to have this additional capability (provided ctx and set promise when run complete), or add a new class that just does this functionality. Bonus points for setting the thread name if you add a new class.

heifner (changes requested):

libraries/chain/controller.cpp
spoonincode: def thinks this should be separate from `named_thread_pool` instead of trying to kludge in two modes of operation. I'll see what I can do

heifner (approved):

PR #448, opened 2024-07-30
[1.0-beta4] Terminate on large reversible size

### LINKED ISSUE #447: Terminate nodeos when reversible blocks exceeds irreversible blocks by excessive amount

heifner:
Terminate nodeos when HEAD is greater than LIB + 1000. Puts a limit on how many reversible blocks on a best HEAD can be created before nodeos gives up on LIB moving and exits.

- Adds missing mutex lock in fork database.
- Refactors terminate-at-block handling into one function that checks both conditions for terminate.

Resolves #447

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3523,8 +3523,7 @@ struct controller_impl {
                      const trx_meta_cache_lookup& trx_lookup ) {
       try {
          try {
-            if( conf.terminate_at_block > 0 && conf.terminate_at_block < bsp->block_num() ) { // < since checked before apply
-               ilog("Block ${n} reached configured maximum block ${num}; terminating", ("n", bsp->block_num()-1)("num", conf.terminate_at_block) );
+            if (should_terminate()) {
                shutdown();
                return false;
             }
@@ -3898,6 +3897,8 @@ struct controller_impl {
          consider_voting(bsp, use_thread_pool_t::no);
       }
 
+      // should_terminate() is not thread-safe requires chain_head.block_num()
+      // also when syncing allow many blocks in forkdb queued to be applied
       if (conf.terminate_at_block == 0 || bsp->block_num() <= conf.terminate_at_block) {
          forkdb.add(bsp, ignore_duplicate_t::yes);
          if constexpr (savanna_mode)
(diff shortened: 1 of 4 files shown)


greg7mdp:
Does this mean that if 1/3 of finalizer are off-line for 10 minutes, all nodes will shutdown?

heifner:
> Does this mean that if 1/3 of finalizer are off-line for 10 minutes, all nodes will shutdown?

Closer to 8.3 minutes, yes.

greg7mdp:
> > Does this mean that if 1/3 of finalizer are off-line for 10 minutes, all nodes will shutdown?
> 
> Closer to 8.3 minutes, yes.

This doesn't seem desirable though. Maybe it would be better to just skip more and more block producing slots until we start getting QCs again? For example every 10 minutes, skip twice as many producing slots.

libraries/chain/controller.cpp
```diff
@@ -4701,6 +4702,25 @@ struct controller_impl {
       return conf.block_validation_mode == validation_mode::LIGHT || conf.trusted_producers.count(producer);
    }
 
+   bool should_terminate() const {
+      constexpr block_num_type max_reversible_blocks = 1000;
```
heifner, line 4706: Temporary solution until we can get the consensus level change that will restrict producing higher block heights until finalizers come back online.

heifner:
> This doesn't seem desirable though. Maybe it would be better to just skip more and more block producing slots until we start getting QCs again? For example every 10 minutes, skip twice as many producing slots.

see https://github.com/AntelopeIO/spring/pull/448/files#r1697549049

libraries/chain/controller.cpp
heifner: We have some libtester tests that failed because they produce more than 1000 blocks (more than 2300 if I counted correctly). How about we go with 30 minutes; 3600 blocks.

linh2931: As long as we have some protection against OOM, 30 minutes is fine for now.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Duplicate auto-merged by github.
Note:end

PR #451, opened 2024-07-31
[1.0-beta4] Some misc improvements

heifner:
- Add missing mutex to `fork_database_t::has_root()`.
- When a block fails to produce, do not attempt to produce until the next block time stamp. This prevents a tight spin on failures. We had an issue somewhere for this, but can't find it. I found the problem when testing a possible alternative to #448.
- Refactors terminate-at-block handling into a function.
- Additional `std::move`  in assemble block.
- Logging improvements
  - Do not log expired transaction count during replay.
  - Do not log expired transaction count if there are none.
  - Remove confusing "maybe switch forks" log statement.
  - Log block timestamp instead of wall clock time (which is always logged anyway) during start block.

I could separate some/all of these into different PRs if desired.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3523,8 +3523,7 @@ struct controller_impl {
                      const trx_meta_cache_lookup& trx_lookup ) {
       try {
          try {
-            if( conf.terminate_at_block > 0 && conf.terminate_at_block < bsp->block_num() ) { // < since checked before apply
-               ilog("Block ${n} reached configured maximum block ${num}; terminating", ("n", bsp->block_num()-1)("num", conf.terminate_at_block) );
+            if (should_terminate()) {
                shutdown();
                return false;
             }
@@ -3898,6 +3897,8 @@ struct controller_impl {
          consider_voting(bsp, use_thread_pool_t::no);
       }
 
+      // should_terminate() is not thread-safe requires chain_head.block_num()
+      // also when syncing allow many blocks in forkdb queued to be applied
       if (conf.terminate_at_block == 0 || bsp->block_num() <= conf.terminate_at_block) {
          forkdb.add(bsp, ignore_duplicate_t::yes);
          if constexpr (savanna_mode)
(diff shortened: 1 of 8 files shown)


libraries/chain/controller.cpp
```diff
@@ -4367,8 +4366,10 @@ struct controller_impl {
             break;
          }
       }
-      dlog("removed ${n} expired transactions of the ${t} input dedup list, pending block time ${pt}",
-           ("n", num_removed)("t", total)("pt", now));
+      if (!replaying && total > 0) {
```
linh2931, line 4369: I like this removal. Was thinking about removing this spamming `removed 0 expired transactions...` too.

linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -3898,6 +3897,8 @@ struct controller_impl {
          consider_voting(bsp, use_thread_pool_t::no);
       }
 
+      // should_terminate() is not thread-safe requires chain_head.block_num()
+      // also when syncing allow many blocks in forkdb queued to be applied
```
greg7mdp, line 3901: What about having both:

```
// thread-safe
bool should_terminate(block_num_type head_block_num) const {
   if (conf.terminate_at_block > 0 && conf.terminate_at_block <= head_block_num) {
      ilog("Block ${n} reached configured maximum block ${num}; terminating",
           ("n", head_block_num)("num", conf.terminate_at_block) );
      return true;
   }
   return false;
}

// not thread-safe because of chain_head access
bool should_terminate() const {
   return should_terminate(chain_head.block_num());
}
```

greg7mdp (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Improvements to Finalizer, ForkDB, and Controller. Refactors code for readability, adds logging, and improves concurrency with additional mutex.
Note:end

PR #452, opened 2024-07-31
[1.0-beta4 -> main] Some misc improvements

heifner:
- Add missing mutex to `fork_database_t::has_root()`.
- When a block fails to produce, do not attempt to produce until the next block time stamp. This prevents a tight spin on failures. We had an issue somewhere for this, but can't find it. I found the problem when testing a possible alternative to #448.
- Refactors terminate-at-block handling into a function.
- Additional `std::move`  in assemble block.
- Logging improvements
  - Do not log expired transaction count during replay.
  - Do not log expired transaction count if there are none.
  - Remove confusing "maybe switch forks" log statement.
  - Log block timestamp instead of wall clock time (which is always logged anyway) during start block.

Merges `release/1.0-beta4` into `main` including #451

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -812,11 +812,11 @@ struct building_block {
 
                assembled_block::assembled_block_if ab{
                   bb.active_producer_authority,
-                  bhs,
+                  std::move(bhs),
                   std::move(bb.pending_trx_metas),
                   std::move(bb.pending_trx_receipts),
-                  valid,
-                  qc_data.qc,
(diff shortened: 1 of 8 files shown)


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Improvements to Finalizer, ForkDB, and Controller. Refactors code for readability, adds logging, and improves concurrency with additional mutex.
Note:end

Issue #456, opened 2024-08-01
beta4 testnet - finalizer not voting


heifner:
- Started without a safety.dat file
```
error 2024-07-31T15:33:51.862 nodeos    finalizer.cpp:218             load_finalizer_safet ] unable to open finalizer safety persistence file /opt/SLOWtestNet/finalizers/safety.dat, file doesn't exist
```

- Synced through transition
```
info  2024-07-31T15:33:55.216 net-3     net_plugin.cpp:2183           operator()           ] ["slownet.eosusa.io:19883 - 42b454f" - 2 10.10.10.116:19883] requesting range 30602 to 30801, fhead 30434, lib 30090
info  2024-07-31T15:33:55.233 nodeos    controller.cpp:1360           transition_to_savann ] Transitioning to savanna, IF Genesis Block 30275, IF Critical Block 30607
info  2024-07-31T15:33:55.305 nodeos    controller.cpp:1405           operator()           ] Transition to instant finality happening after block 30607, First IF Proper Block 30608
info  2024-07-31T15:33:55.306 nodeos    producer_plugin.cpp:688       on_block             ] Block 19068e4ffe4f3b17... #30608 @ 2024-07-29T15:22:24.500 produced by pythoncolors, latency: 173490806ms has no votes
```

- Producer/finalizer became active
```
[...]

PR #458, opened 2024-08-01
[1.0-beta4] Log liveness/safety check failures info/warn log level

heifner:
- Use `vote` logger for logging of decide vote conditions
- Use `info` level instead of `debug` for liveness check failure
- Use `warn` level instead of `debug` for safety check failure
- Add some static asserts on size of safety.dat file types

Better logging for issue: #456

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -669,7 +669,8 @@ public:
       auto& chain  = chain_plug->chain();
       auto  before = _unapplied_transactions.size();
       _unapplied_transactions.clear_applied(block);
-      chain.get_mutable_subjective_billing().on_block(_log, block, fc::time_point::now());
+      auto now = fc::time_point::now();
+      chain.get_mutable_subjective_billing().on_block(_log, block, now);
       if (before > 0) {
          fc_dlog(_log, "Removed applied transactions before: ${before}, after: ${after}", ("before", before)("after", _unapplied_transactions.size()));
       }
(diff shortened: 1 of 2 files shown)


greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -684,7 +685,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                auto missing = chain.missing_votes(id, qc_ext.qc);
                log_missing_votes(block, id, missing);
             }
-         } else if (block->is_proper_svnn_block()) {
+         } else if (block->is_proper_svnn_block() && (now - block->timestamp < fc::minutes(5) || (block->block_num() % 1000 == 0)) ) {
```
linh2931, line 688: What's purpose of the timestamp check here? Should `<` be `>=`?

heifner: The check is so you don't spam the log while syncing.

linh2931: So why not log when the elapse time (now - block->timestamp) is greater than 5 minutes? With existing `now - block->timestamp < fc::minutes(5)`, you log every time when the difference is less than 5 minutes.

heifner: now  - old block < 5 minutes
10:10:10.340 - 05:05:05.000 < 5 minutes
Only logs if the delta is small.

linh2931 (approved):

ericpassmore:
Note:start
group: LOGGING
category: INTERNALS
summary: More appropriate logging levels for liveness/safety check failures.
Note: end

Issue #467, opened 2024-08-02
Verify dual finalizers in active and pending finalizer policies vote in the same way in received QCs


linh2931:
For a received QC, if a finalizer is in both active finalizer policy and pending finalizer policy (_**dual finalizer**_), verify the finalizer votes in the same way on both policies or not vote at all.

greg7mdp:
Is this needed? The finalizer votes only once on each block, and the vote is added to either one or two qcs depending on whether a pending policy is active. So unless the nodeos code is maliciously modified the issue would not occur.

Also, what should we do if that were to happen? Disregard the QC entirely? This would allow a single faulty finalizer to prevent the chain from finalizing.

linh2931:
Yes, this is to protect against malicious actors (who can even modify nodes code) in the same way we verify other parts of incoming QCs. We talked about this in the final hour of this Wednesday meeting.

@heifner @arhag

heifner:
Yes, we should disregard the QC entirely and also the block it came in on. This can only happen by a misbehaving producer. We should not validate that block or the QC from that producer.

PR #475, opened 2024-08-05
Add more validations on QC's votes

linh2931:
Add more validations on QC's votes:
* Verify a finalizer does not vote both strong and weak for the same block
* Verify a dual finalizer (which votes on both active and pending finalizer policies) votes the same way
* Verify at least one of strong and weak bitsets present in QC signature
* Add tests for dual finalizer validation

Resolves https://github.com/AntelopeIO/spring/issues/467


libraries/chain/finality/qc.cpp
```diff
@@ -309,10 +375,54 @@ std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) cons
...
+}
+
 void aggregating_qc_t::verify_qc(const qc_t& qc, const digest_type& strong_digest, const weak_digest_t& weak_digest) const {
+   qc.active_policy_sig.verify_vote_format(active_finalizer_policy);
+
    if (qc.pending_policy_sig) {
       EOS_ASSERT(pending_finalizer_policy, invalid_qc_claim,
                  "qc ${bn} contains pending policy signature for nonexistent pending finalizer policy", ("bn", qc.block_num));
+
+      qc.pending_policy_sig->verify_vote_format(active_finalizer_policy);
```
heifner, line 424: Should be `pending_finalizer_policy`

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,77 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+                                         pending_vote_index);
+}
+
+// returns true if vote indicated by my_vote_index in strong_votes or
+// weak_votes are is the same as the one indicated by other_vote_index
+// in `other` signature's strong_votes or weak_votes
+bool qc_sig_t::vote_same_at(const qc_sig_t& other, uint32_t my_vote_index, uint32_t other_vote_index) const {
+   auto is_vote_same = [&] (std::optional<vote_bitset_t> my_votes,
+                            uint32_t my_vote_index,
+                            std::optional<vote_bitset_t> other_votes,
```
heifner, line 39: Both these `std::optional<vote_bitset_t>` should be `const &`

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,77 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+   }
+   if (weak_votes) {
+      EOS_ASSERT( num_finalizers == weak_votes->size(), invalid_qc_claim,
+                  "vote bitset size is not the same as the number of finalizers for the policy it refers to, "
+                  "vote bitset size: ${s}, num of finalizers for the policy: ${n}",
+                  ("s", weak_votes->size())("n", num_finalizers) );
+   }
+
+   // verify a finalizer cannot vote both strong and weak
+   if (strong_votes && weak_votes) {
```
heifner, line 84: Need an `EOS_ASSERT` that either `strong_votes` or `weak_votes` is present.

libraries/chain/finality/qc.cpp
```diff
@@ -309,10 +375,54 @@ std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) cons
    return std::optional<qc_t>{qc_t{block_num, std::move(*active_best_qc), {}}};
 }
 
+// A dual finalizer votes on both active and pending finalizer policies
+void aggregating_qc_t::verify_dual_finalizers_votes(const qc_t& qc) const {
+   auto build_vote_index = [&](const std::vector<finalizer_authority> finalizers, std::set<bls_public_key>& keys, std::map<bls_public_key, uint32_t>& index) {
```
heifner, line 380: `finalizers` should be `const&`

libraries/chain/finality/qc.cpp
```diff
@@ -309,10 +375,54 @@ std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) cons
    return std::optional<qc_t>{qc_t{block_num, std::move(*active_best_qc), {}}};
 }
 
+// A dual finalizer votes on both active and pending finalizer policies
+void aggregating_qc_t::verify_dual_finalizers_votes(const qc_t& qc) const {
```
heifner, line 379: This function seems rather inefficient. I think it would be better if it did sequential scans of the vector instead of creating the sets and maps with all the copies. Alternatively, could sort the finalizers. Seems like we could sort the finalizers when we set them. Going to want them sorted for the diff algorithm also.

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,77 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+      assert(my_votes.has_value() == other_votes.has_value());
+      // validate indexes
+      assert(my_vote_index < my_votes->size() && other_vote_index < other_votes->size());
+
+      return (*my_votes)[my_vote_index] == (*other_votes)[other_vote_index];
+   };
+
+   // if strong_votes exist, both votes the same
+   if (strong_votes) {
+      return is_vote_same(strong_votes, my_vote_index, other.strong_votes, other_vote_index);
```
heifner, line 51: I've confused by this, just because `strong_votes` exist it doesn't mean the `my_vote_index` is for the `strong_votes`.

heifner (changes requested):

libraries/chain/finality/qc.cpp
linh2931: Thanks! (cut and paste...). Fixed by https://github.com/AntelopeIO/spring/pull/475/commits/2841dd00717e70d2ae545cb8cb73c7dee7c356e0

libraries/chain/finality/qc.cpp
linh2931: Initially I was thinking about doing it manually, but ended up with STD containers for simplicity.  Will think about more.

libraries/chain/finality/qc.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/475/commits/58edf7588ed6bafa7a709d0e8c866d6d33f803e6

libraries/chain/finality/qc.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/475/commits/58edf7588ed6bafa7a709d0e8c866d6d33f803e6

libraries/chain/finality/qc.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/475/commits/58edf7588ed6bafa7a709d0e8c866d6d33f803e6

libraries/chain/finality/qc.cpp
linh2931: Thanks. Fixed it and made it simpler. https://github.com/AntelopeIO/spring/pull/475/commits/d59b1ec7c95c94c44f3d5944eb85a362154f535b

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,80 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+// returns true if ithe other and I vote iin the same way on my_vote_index
+// and other's index.
+bool qc_sig_t::vote_same_at(const qc_sig_t& other, uint32_t my_vote_index, uint32_t other_vote_index) const {
+   assert(!strong_votes || my_vote_index < strong_votes->size());
+   assert(!weak_votes || my_vote_index < weak_votes->size());
+
+   // I vote strongly on my_vote_index, the other must vote strongly
+   // on other_vote_index too
+   if (strong_votes && (*strong_votes)[my_vote_index]) {
+      return other.strong_votes && (*other.strong_votes)[other_vote_index];
```
heifner, line 42: Don't we also need to verify other didn't also vote weak at the same index?

linh2931: For a particular finalizer (identified by index), it can only vote strong on both active (I) and pending (other), or vote weak on both active and pending. When we know I votes strong, we only need to check if the other votes strong or not; if the other votes strong then returns true; and if the other does not vote on strong then returns false (not matter if the other votes on weak or not). Will update the comments if this makes sense.

heifner: But if someone crafts an invalid QC they could create one that has both a strong and vote index marked as true.

greg7mdp: > For a particular finalizer (identified by index), it can only vote strong on both active (I) and pending (other), or vote weak on both active and pending. 

I'm not sure I follow. For a given index, the active and pending policy may have a different finalizer. There is no guarantee that any finilizer votes on both policies.

linh2931: We have already checked `qc_sig_t` that strong and weak cannot be set at the same index.

linh2931: > > For a particular finalizer (identified by index), it can only vote strong on both active (I) and pending (other), or vote weak on both active and pending.
> 
> I'm not sure I follow. For a given index, the active and pending policy may have a different finalizer. There is no guarantee that any finalizer votes on both policies.

Yes, active and pending policies can have different indexes for a same finalizer.

linh2931: Oh I see, we should continue to check the other should not vote on weak if I vote on strong -- we should eosio_assert in this case.

libraries/chain/finality/qc.cpp
linh2931: Changed to use sorted pending policy keys directly https://github.com/AntelopeIO/spring/pull/475/commits/90643d0213721fa5e1f0cac1b7f1ffe368d5fb0e

libraries/chain/finality/qc.cpp
```diff
@@ -309,10 +365,55 @@ std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) cons
...
+      pending_finalizer_keys_sorted.emplace_back(f.public_key);
+   }
+   std::sort(pending_finalizer_keys_sorted.begin(), pending_finalizer_keys_sorted.end());
+
+   // Find dual finalizers (which vote on both active and pending policies)
+   // and verify each dual finalizer votes in the same way.
+   for (const auto& f: active_finalizer_policy->finalizers) {
+      bool found = std::binary_search(pending_finalizer_keys_sorted.begin(),
+                                      pending_finalizer_keys_sorted.end(),
+                                      f.public_key);
```
heifner, line 397: Instead of sorting (which requires copying all the `bls_public_key`s), I think it would be better to just sequential scan the entire vector each time. `bls_public_key` is rather large. I wonder if we would be better off using `shared_ptr<bls_public_key>` everywhere instead.

libraries/chain/finality/qc.cpp
```diff
@@ -309,10 +365,55 @@ std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) cons
...
+      if (found) {
+         auto active_vote_index  = active_finalizer_vote_index[f.public_key];
+         auto pending_vote_index = pending_finalizer_vote_index[f.public_key];
+         EOS_ASSERT(qc.vote_same_at(active_vote_index, pending_vote_index),
+                    invalid_qc_claim,
+                    "qc ${bn} contains a dual finalizer ${k} which does not vote the same on active and pending policies", ("bn", qc.block_num)("k", f.public_key));
```
heifner, line 403: Can you fix code wrapper here.

heifner (changes requested):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Additional validations on QC's votes.
Note:end

libraries/chain/include/eosio/chain/finality/qc.hpp
```diff
@@ -234,6 +244,8 @@ namespace eosio::chain {
 
       std::optional<qc_t> get_best_qc(block_num_type block_num) const;
       // verify qc against active and pending policy
+      void verify_dual_finalizers_votes(const qc_t& qc) const;
```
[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/475/commits/f1f3ac2befc80c8ffd3faa0916970f792b09b4ec

libraries/chain/finality/qc.cpp
```diff
@@ -309,10 +365,55 @@ std::optional<qc_t> aggregating_qc_t::get_best_qc(block_num_type block_num) cons
...
+   auto build_vote_index = [&](const std::vector<finalizer_authority>& finalizers,
+                               std::map<bls_public_key, uint32_t>& index) {
+      uint32_t i = 0;
+      for (const auto& f: finalizers) {
+         index[f.public_key] = i;
+         ++i;
+      }
+   };
+
+   std::map<bls_public_key, uint32_t> active_finalizer_vote_index; // from public key to vote bit index
```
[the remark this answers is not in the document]
linh2931: No longer needed in new implementation. https://github.com/AntelopeIO/spring/pull/475/commits/f1f3ac2befc80c8ffd3faa0916970f792b09b4ec

libraries/chain/finality/qc.cpp
linh2931: It is now much simpler. Thanks, https://github.com/AntelopeIO/spring/pull/475/commits/f1f3ac2befc80c8ffd3faa0916970f792b09b4ec

heifner (approved):

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,67 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+// returns true if vote indicated by active_vote_index in active_policy
+// is the same as vote indicated by pending_vote_index in pending_policy
+bool qc_t::vote_same_at(uint32_t active_vote_index, uint32_t pending_vote_index) const {
+   assert(pending_policy_sig);
+   return active_policy_sig.vote_same_at(*pending_policy_sig,
+                                         active_vote_index,
+                                         pending_vote_index);
+}
+
+// returns true if ithe other and I vote iin the same way on my_vote_index
```
greg7mdp, line 33:
```suggestion
// returns true iff the other and I voted the same way
```

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,67 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+                  "vote bitset size: ${s}, num of finalizers for the policy: ${n}",
+                  ("s", weak_votes->size())("n", num_finalizers) );
+   }
+
+   // verify a finalizer cannot vote both strong and weak
+   if (strong_votes && weak_votes) {
+      for (size_t i=0; i<strong_votes->size(); ++i) {
+         // at most one is true
+         EOS_ASSERT( !((*strong_votes)[i] && (*weak_votes)[i]), invalid_qc_claim,
+                     "finalizer (bit index ${i}) votes both strong and weak",
```
greg7mdp, line 78:
```suggestion
                     "finalizer (bit index ${i}) voted both strong and weak",
```

libraries/chain/finality/qc.cpp
```diff
@@ -21,6 +21,67 @@ inline std::vector<uint32_t> bitset_to_vector(const vote_bitset_t& bs) {
...
+}
+
+// returns true if ithe other and I vote iin the same way on my_vote_index
+// and other's index.
+bool qc_sig_t::vote_same_at(const qc_sig_t& other, uint32_t my_vote_index, uint32_t other_vote_index) const {
+   assert(!strong_votes || my_vote_index < strong_votes->size());
+   assert(!weak_votes || my_vote_index < weak_votes->size());
+
+   // We have already verified the same index is not voted on both strong
+   // and weak for a given qc_sig_t (I or other).
```
greg7mdp, line 40:
```suggestion
   // We have already verified the same index has not voted both strong
   // and weak for a given qc_sig_t (I or other).
```

greg7mdp (approved):

libraries/chain/finality/qc.cpp
linh2931: Thanks.

PR #142, opened 2024-05-15
upgrade to libsecp256k1 v0.5.0; improves k1 signing performance ~10%

spoonincode:
Time for another awesome libsecp256k1 update. v0.5.0 improves signing performance by about ~10% in my tests. The improvement is via `https://github.com/bitcoin-core/secp256k1/pull/1058`

PR #481, opened 2024-08-06
update to libsecp256k1 0.5.1 for small signing performance improvement

spoonincode:
Bump libsecp256k1 to v0.5.1.

v0.5.1 changes the default pre computed table size used for signing. The default was changed to match what Bitcoin Core is using (the highest configuration; i.e. largest table). So, change our configuration to match the new default. It seems entirely reasonable to use the same configuration Bitcoin Core does.

On a i7-12700K this improves signing time from about 17.4µs to 16.8µs on a gcc14 build. A 3% improvement at the cost of the secp256k1 context growing by 64KB (inconsequential growth).

### DIFF
--- a/libraries/libfc/secp256k1/CMakeLists.txt
+++ b/libraries/libfc/secp256k1/CMakeLists.txt
@@ -5,8 +5,7 @@ add_library(secp256k1-internal INTERFACE)
 target_include_directories(secp256k1-internal INTERFACE secp256k1/src)
 
 target_compile_definitions(secp256k1-internal INTERFACE ENABLE_MODULE_RECOVERY=1
-                                                        ENABLE_MODULE_EXTRAKEYS=1
-                                                        COMB_BLOCKS=11
+                                                        COMB_BLOCKS=43
                                                         COMB_TEETH=6
                                                         ECMULT_WINDOW_SIZE=15
                                                         SECP256K1_STATIC=1)


libraries/libfc/secp256k1/CMakeLists.txt
```diff
@@ -5,8 +5,7 @@ add_library(secp256k1-internal INTERFACE)
 target_include_directories(secp256k1-internal INTERFACE secp256k1/src)
 
 target_compile_definitions(secp256k1-internal INTERFACE ENABLE_MODULE_RECOVERY=1
-                                                        ENABLE_MODULE_EXTRAKEYS=1
-                                                        COMB_BLOCKS=11
+                                                        COMB_BLOCKS=43
```
spoonincode, line 8: These `COMB` settings come from
https://github.com/bitcoin-core/secp256k1/blob/3fdf146bad042a17f6b2f490ef8bd9d8e774cdbd/CMakeLists.txt#L93-L105
notice the default setting is now 86 and the `COMB` defines match what is here.

libraries/libfc/secp256k1/CMakeLists.txt
```diff
@@ -5,8 +5,7 @@ add_library(secp256k1-internal INTERFACE)
 target_include_directories(secp256k1-internal INTERFACE secp256k1/src)
 
 target_compile_definitions(secp256k1-internal INTERFACE ENABLE_MODULE_RECOVERY=1
-                                                        ENABLE_MODULE_EXTRAKEYS=1
```
spoonincode, line 8: `EXTRAKEYS` was added in https://github.com/AntelopeIO/spring/pull/142#discussion_r1602040845 and as I suspected at that time, this was indeed a dependency oversight and has been fixed in 0.5.1.

heifner (approved):

linh2931 (approved):

Issue #454, opened 2024-07-31
Fix proposer policy transition rules


arhag:
Consider the following case:
1. In round 1, a block proposes a proposer policy.
2. In round 2, the block in the 12th time slot of the round is not present.
3. In round 3, there exists at least one block.

In the current implementation, the first block in round 3 will not use the proposer policy that was proposed in round 1 despite that being our intention. That would however be the case if there was a block present in the 12th time slot of round 2. This is a bug.

We want to always promote a proposer policy to active on the first block of a round. That proposer policy should never be the one that was proposed in the prior round. It should have been proposed at least two rounds back.

Fixing the bug means that at the time of building a new block that is the first block of a round, it will always choose as the active proposal policy the last proposed proposer policy in the blockchain that was not proposed from a block in the current round or the one prior.

This means we do not actually need a `map` to track all the proposed proposal policies in the `block_header_state`. It is sufficient to have only two `optional` fields: `latest_pending_proposer_policy` and `latest_proposed_proposer_policy`. 

Additionally, instead of tracking an `active_time` with each proposed policy, we should just track the timestamp in which it was proposed: `proposal_time`.

When creating a `block_header_state` for the first block of a round (which can be determined by comparing its timestamp with the timestamp of its parent block), some changes to these fields may need to occur:
1. If `latest_proposed_proposer_policy` is not `nullopt` and its `proposal_time` indicates that it was proposed in a round before the prior round, then maybe promote `latest_proposed_proposer_policy` to `active_proposer_policy` (maybe because other conditions may also need to apply, to be discussed in a bit). If it is promoted, then `latest_pending_proposer_policy` and `latest_proposed_proposer_policy` should be set to `nullopt` at this point in time (`latest_proposed_proposer_policy` may still be changed in later steps).
2. Otherwise, if `latest_pending_proposer_policy` is not `nullopt`, maybe promote `latest_pending_proposer_policy` to `active_proposer_policy` (maybe because other conditions may also need to apply, to be discussed in a bit). If it is promoted, then `latest_pending_proposer_policy` should be set to `nullopt` at this point in time (may still be changed in later steps).
3. If `latest_proposed_proposer_policy` is not `nullopt` and `latest_pending_proposer_policy` is now `nullopt`, `latest_proposed_proposer_policy` is moved into `latest_pending_proposer_policy`. If it is moved, then `latest_proposed_proposer_policy` should be set to `nullopt` at this point in time (may still be changed in later steps).

[...]

PR #483, opened 2024-08-06
Fix proposal policy transition rules

linh2931:
Consider the following case:
1. In round 1, a block proposes a proposer policy.
2. In round 2, the block in the 12th time slot of the round is not present.
3. In round 3, there exists at least one block.

In the current implementation, the first block in round 3 will not use the proposer policy that was proposed in round 1 despite that being our intention. That would however be the case if there was a block present in the 12th time slot of round 2. 

Fixing the bug means that at the time of building a new block that is the first block of a round, it will always choose as the active proposal policy the last proposed proposer policy in the blockchain that was not proposed from a block in the current round or the one prior.

Instead of a `map` to track all the proposed proposal policies in the `block_header_state`, only two `optional` fields: `latest_pending_proposer_policy` and `latest_proposed_proposer_policy` are needed. Additionally, instead of tracking an `active_time` with each proposed policy, we just track the timestamp in which it was proposed: `proposal_time`.

When creating a `block_header_state` for the first block of a round, those fields are computed by:
1. If `latest_proposed_proposer_policy` is not `nullopt` and its `proposal_time` indicates that it was proposed in a round before the prior round, then promote `latest_proposed_proposer_policy` to `active_proposer_policy`.  `latest_pending_proposer_policy` and `latest_proposed_proposer_policy` should be set to `nullopt` at this point in time (`latest_proposed_proposer_policy` may still be changed in later steps).
2. Otherwise, if `latest_pending_proposer_policy` is not `nullopt`, promote `latest_pending_proposer_policy` to `active_proposer_policy`. `latest_pending_proposer_policy` should be set to `nullopt` at this point in time (may still be changed in later steps).
3. If `latest_proposed_proposer_policy` is not `nullopt` and `latest_pending_proposer_policy` is now `nullopt`, `latest_proposed_proposer_policy` is moved into `latest_pending_proposer_policy`. If it is moved, then `latest_proposed_proposer_policy` is set to `nullopt` at this point in time (may still be changed in later steps).

The final related step should occur for any block (whether it is the first block of the round or not) and can only be computed after the block has been assembled:

4. If this block proposes a new proposer policy, that policy should be set in `latest_proposed_proposer_policy` (replacing whatever happened to be there).

Resolves https://github.com/AntelopeIO/spring/issues/454


unittests/producer_schedule_savanna_tests.cpp
```diff
@@ -434,4 +434,36 @@ BOOST_AUTO_TEST_CASE( switch_producers_test ) try {
...
+   // In round 1, a block proposes a proposer policy.
+   create_accounts( {"alice"_n} );
+   set_producers( {"alice"_n} );
+   auto b = produce_block();
+   auto index = b->timestamp.slot % config::producer_repetitions; // current index in current round
+   produce_blocks(config::producer_repetitions - index - 1); // to end of the round
+
+   // In round 2, the block in the last time slot of the round is not present.
+   produce_blocks(config::producer_repetitions - 1);
+   const auto time_to_skip = fc::milliseconds(config::block_interval_ms);
```
heifner, line 451: Default `time_to_skip` is `fc::microseconds default_skip_time = fc::milliseconds(config::block_interval_ms)` so this is just doing the default produce one block. I think you need `2 * config::block_interval_ms`.

linh2931: Tried `2 * config::block_interval_ms`. Test still passed. Will dig into it tomorrow.

linh2931:
@heifner The commit https://github.com/AntelopeIO/spring/pull/483/commits/59e716e408b9f3177ef8a74db9102d5d3d9fe52d passed most tests, but failed a few. Can you see anything obviously wrong? Thanks.

```
unittests/unit_test -t producer_schedule_savanna_tests
Random number generator seeded to 1723148650
Running 5 test cases...
unittests/producer_schedule_savanna_tests.cpp(35): error: in "producer_schedule_savanna_tests/verify_producer_schedule_after_savanna_activation": check head().block_num() == expected_block_num has failed [25 != 24]
unittests/producer_schedule_savanna_tests.cpp(35): error: in "producer_schedule_savanna_tests/verify_producer_schedule_after_savanna_activation": check head().block_num() == expected_block_num has failed [49 != 48]
unittests/producer_schedule_savanna_tests.cpp(364): error: in "producer_schedule_savanna_tests/proposer_policy_progression_test": check 17u == control->active_producers().version has failed [17 != 12]
unittests/producer_schedule_savanna_tests.cpp(367): error: in "producer_schedule_savanna_tests/proposer_policy_progression_test": check 22u == control->active_producers().version has failed [22 != 17]
nittests/producer_schedule_savanna_tests.cpp(468): error: in "producer_schedule_savanna_tests/policy_switching_corner_case_test": check b->producer == "alice"_n has failed [eosio != alice]
```

linh2931:
@arhag The current PR has not implemented the enhancement described by https://github.com/AntelopeIO/spring/issues/454:

> not promote the policy if the `proposal_time` of the policy is greater than the `last_final_block_timestamp` (computed after considering the claim of the current block).

This is due to the fact the `producer_plugin` needs to know the producer for the next block https://github.com/AntelopeIO/spring/blob/ea5c02580b75749202d77d5c3331f20361f74c28/plugins/producer_plugin/producer_plugin.cpp#L1795
But at that time we don't know the QC claim of the next block and cannot calculate the `last_final_block_timestamp`.

But we know the `last_final_block_timestamp` of the head block. We could check against it. This will miss one corner case that if the next block advances the LIB.

libraries/chain/controller.cpp
```diff
@@ -4802,6 +4808,13 @@ struct controller_impl {
       return pending->active_producers();
    }
 
+   const producer_authority_schedule& scheduled_active_producers_at(block_timestamp_type t)const {
+      if( !(pending) )
+         return head_compute_active_schedule_auth(t);
+
+      return pending->active_producers();
```
heifner, line 4815: This doesn't seem right, why is `get_scheduled_active_proposer_policy_at` not used?

libraries/chain/include/eosio/chain/block_header_state_utils.hpp
```diff
@@ -20,6 +20,26 @@ namespace eosio::chain::detail {
       return block_timestamp_type{t.slot + (config::producer_repetitions - index) + config::producer_repetitions};
    }
```
heifner, line 21: Remove unused `get_next_next_round_block_time`. The log statement using should not report when it will change because it can't know when that will be.

heifner (changes requested):

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -268,6 +268,7 @@ namespace eosio::chain {
          uint32_t                       pending_block_num()const;
 
          const producer_authority_schedule&         active_producers()const;
+         const producer_authority_schedule&         scheduled_active_producers_at(block_timestamp_type t)const;
```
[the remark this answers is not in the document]
linh2931: If `pending` does not exist, as in the use case of producer_plugin, we need to provide the next block time. That's why `scheduled_active_producers_at` is added.

libraries/chain/controller.cpp
linh2931: If `pending` does not exist, we call `head_compute_active_schedule_auth` which in turns calls `get_scheduled_active_proposer_policy_at`. (I should have changed `head_compute_active_schedule_auth` to `head_scheduled_active_schedule_auth_at`).

If `pending` exists, we call `pending`'s `active_producers()`, which uses actual computed active policy.

libraries/chain/include/eosio/chain/block_header_state_utils.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/483/commits/fa43246df35fe1e08d47f782bb27e66d86f950ac

libraries/chain/include/eosio/chain/controller.hpp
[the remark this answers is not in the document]
heifner: `scheduled_active_producers_at` Does not provide what it advertises. For example, I can't give it a blocktime of 3 days ago or 3 days in the future and get the active schedule.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Improve finalizer policy transition by always promoting a proposer policy to active on the first block of a round. That proposer policy should never be the one that was proposed in the prior round. It should have been proposed at least two rounds back.
Note:end

heifner (changes requested):

libraries/chain/block_header_state.cpp
```diff
@@ -191,6 +233,33 @@ void evaluate_finalizer_policies_for_promotion(const block_header_state& prev,
...
+   } else if (curr.latest_pending_proposer_policy &&
+                ((*curr.latest_pending_proposer_policy)->proposal_time <= prev.core.last_final_block_timestamp())) {
+      curr.active_proposer_policy = *curr.latest_pending_proposer_policy;
+      curr.latest_pending_proposer_policy = std::nullopt;
+   }
+
+   if (curr.latest_proposed_proposer_policy && !curr.latest_pending_proposer_policy) {
+      curr.latest_pending_proposer_policy = curr.latest_proposed_proposer_policy;
+      curr.latest_proposed_proposer_policy = std::nullopt;
+   }
```
greg7mdp, line 260: It doesn't seem a good idea to duplicate the promotion logic in this function and `get_active_proposer_policy_for_block_at()`.

I propose the following:
```suggestion
   auto& new_policy = prev.get_active_proposer_policy_for_block_at(curr.timestamp());
   if (new_policy != curr.active_proposer_policy) {
      curr.active_proposer_policy = new_policy;

      if (curr.latest_proposed_proposer_policy && new_policy == *curr.latest_proposed_proposer_policy) {
         curr.latest_proposed_proposer_policy = std::nullopt;
         curr.latest_pending_proposer_policy = std::nullopt;
      } else if (curr.latest_pending_proposer_policy && new_policy == *curr.latest_pending_proposer_policy)
         curr.latest_pending_proposer_policy = std::nullopt;
   }
   if (detail::first_block_of_round(curr.timestamp(), prev.timestamp()) &&
       curr.latest_proposed_proposer_policy && !curr.latest_pending_proposer_policy) {
      curr.latest_pending_proposer_policy = curr.latest_proposed_proposer_policy;
      curr.latest_proposed_proposer_policy = std::nullopt;
   }
```

greg7mdp (changes requested):

libraries/chain/block_header_state.cpp
linh2931: Thanks. I was thinking about doing the same but did not proceed as I needed the same number of `if` statements. But those new `if`s are simpler. https://github.com/AntelopeIO/spring/pull/483/commits/b55525d46d8ca3ca3ec5c28bbe4884dddcd9775d

libraries/testing/tester.cpp
```diff
@@ -561,8 +561,12 @@ namespace eosio::testing {
          // otherwise the testing_allow_voting would be set to true when `i >= 4` even though the user of
          // `produce_blocks` wants it to be true.
          // This is 4 instead of 3 because the extra block has to be produced to log_irreversible
-         if (allow_voting_originally && n > 4)
-            control->testing_allow_voting(i >= n - 4);
+         if (allow_voting_originally && n > 4) {
```
greg7mdp, line 564: Since we vote on the first two and the last four.
```suggestion
         if (allow_voting_originally && n > 6) {
```

libraries/testing/tester.cpp
```diff
@@ -561,8 +561,12 @@ namespace eosio::testing {
...
+            // Some tests like the ones for proposer policy transition rely on LIB advance
+            // at least every round (12 blocks). Vote first 2 out of every 12 blocks.
+            auto j = i%12;
+            control->testing_allow_voting( j == 0 || j == 1 || i >= n - 4);
```
greg7mdp, line 568: no big deal, but you could use `j < 2`

libraries/chain/block_header_state.cpp
```diff
@@ -216,26 +277,18 @@ void finish_next(const block_header_state& prev,
...
    std::optional<proposer_policy> new_proposer_policy;
    if (f_ext.new_proposer_policy_diff) {
       new_proposer_policy = prev.get_last_proposed_proposer_policy().apply_diff(*f_ext.new_proposer_policy_diff);
    }
    if (new_proposer_policy) {
       // called when assembling the block
-      next_header_state.proposer_policies[new_proposer_policy->active_time] =
-         std::make_shared<proposer_policy>(std::move(*new_proposer_policy));
+      next_header_state.latest_proposed_proposer_policy = std::make_shared<proposer_policy>(std::move(*new_proposer_policy));
    }
```
greg7mdp, line 292: I don't believe `apply_diff` returns a `std::optional`.
So the `if` at line 289 can be removed and the whole thing simplified to:

```
   if (f_ext.new_proposer_policy_diff) {
      next_header_state.latest_proposed_proposer_policy =
         std::make_shared<proposer_policy>(prev.get_last_proposed_proposer_policy().apply_diff(*f_ext.new_proposer_policy_diff));
   }
```

libraries/chain/controller.cpp
```diff
@@ -3245,11 +3251,11 @@ struct controller_impl {
...
                      if (version) {
                         new_proposer_policy.emplace();
-                        new_proposer_policy->active_time       = detail::get_next_next_round_block_time(bb.timestamp);
+                        new_proposer_policy->proposal_time     = bb.timestamp;
                         new_proposer_policy->proposer_schedule = std::move(bb.trx_blk_context.proposed_schedule);
                         new_proposer_policy->proposer_schedule.version = *version;
-                        ilog("Scheduling proposer schedule change at ${t}: ${s}",
-                             ("t", new_proposer_policy->active_time)("s", new_proposer_policy->proposer_schedule));
+                        ilog("Scheduling proposer schedule ${p}, proposed at: ${s}",
+                             ("p", new_proposer_policy->proposal_time)("s", new_proposer_policy->proposer_schedule));
```
greg7mdp, line 3258: `p` and `s` are be swapped.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -89,13 +89,8 @@ struct block_header_state {
...
 
-   // block time when proposer_policy will become active
-   // current algorithm only two entries possible, for the next,next round and one for block round after that
-   // The active time is the next,next producer round. For example,
-   //   round A [1,2,..12], next_round B [1,2,..12], next_next_round C [1,2,..12], D [1,2,..12]
-   //   If proposed in A1, A2, .. A12 becomes active in C1
-   //   If proposed in B1, B2, .. B12 becomes active in D1
-   flat_map<block_timestamp_type, proposer_policy_ptr>     proposer_policies;
+   std::optional<proposer_policy_ptr> latest_proposed_proposer_policy;
+   std::optional<proposer_policy_ptr> latest_pending_proposer_policy;
```
greg7mdp, line 93: Can you add the rules 1 to 4 from the issue here as comments, so that we don't have to lookup the issue to understand what the rules are.

libraries/chain/controller.cpp
linh2931: Thanks!

libraries/chain/block_header_state.cpp
```diff
@@ -71,10 +68,54 @@ digest_type block_header_state::compute_finality_digest() const {
...
 const producer_authority& block_header_state::get_scheduled_producer(block_timestamp_type t) const {
    return detail::get_scheduled_producer(active_proposer_policy->proposer_schedule.producers, t);
 }
 
+// returns producer using the proposer policy calculated by time `t`
+const producer_authority& block_header_state::get_producer_for_block_at(block_timestamp_type t) const {
+   return detail::get_scheduled_producer(get_active_proposer_policy_for_block_at(t)->proposer_schedule.producers, t);
+}
+
+const producer_authority_schedule* block_header_state::get_next_producer_schedule() const {
```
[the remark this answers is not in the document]
linh2931: This is changed to `pending_producers`.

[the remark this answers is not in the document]
greg7mdp: This still seems like an odd API to me. What do we need it for?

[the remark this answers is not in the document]
linh2931: By net_plugin and chain_plugin. Please see the discussion in telegram channel.

heifner (approved):

greg7mdp (approved):

PR #492, opened 2024-08-07
Add support for changing the content of the finalizer safety info (fsi) file and loading older versions

### LINKED ISSUE #463: Add support for changing the content of the finalizer safety info file (fsi) and loading older version
In order to support future changes to the content of the finalizer safety info file (fsi), we should probably add a `version` number in addition to the `magic` number.

greg7mdp:
Resolves #463.

This adds a version number to the fsi file, while still allowing to load the old file format.
It also adds a testing infrastructure which verifies that all older file versions (currently there is only one) can successfuly be loaded.


libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -36,7 +36,8 @@ namespace eosio::chain {
       block_ref            last_vote;
       block_ref            lock;
 
-      static constexpr uint64_t magic = 0x5AFE11115AFE1111ull;
+      static constexpr uint64_t magic_unversioned = 0x5AFE11115AFE1111ull;
```
heifner, line 39: I don't see any need to have the extra complexity of supporting the un-versioned file.

heifner (changes requested):

libraries/chain/include/eosio/chain/finality/finalizer.hpp
greg7mdp: Why not since people already have unversioned files? And it just a couple lines.

heifner (approved):

libraries/chain/finality/finalizer.cpp
```diff
@@ -305,4 +323,16 @@ void my_finalizers_t::set_default_safety_information(const fsi_t& fsi) {
    default_fsi = fsi;
 }
 
+bool  my_finalizers_t::are_equivalent(uint32_t version, const fsi_map& old, const fsi_map& current) {
+   assert(version < current_safety_file_version); // this function compares an older version with the current one
+   switch(version) {
```
linh2931, line 328: Why do we need `version`, as it is hardcoded to `0` anyway?

greg7mdp: > Why do we need `version`, as it is hardcoded to `0` anyway?

Where is it hardcoded to zero? It can take all values less than `current_safety_file_version`

linh2931: I meant the `case 0:`.  OK, currently `current_safety_file_version` is 1.

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Versioning for Finalizer Safety Information file.
Note:end

PR #493, opened 2024-08-07
[1.0-beta4] Add max-reversible-blocks option

### LINKED ISSUE #447: Terminate nodeos when reversible blocks exceeds irreversible blocks by excessive amount

heifner:
New `chain_plugin` option `max-reversible-blocks`. Specifies the maximum number of reversible blocks beyond current LIB to allow before the node is gracefully shutdown. The graceful shutdown is provided as a way to prevent out of memory kill by OS. Once consensus is working again the node can be started with a larger `max-reversible-blocks` value to continue.

```
  --max-reversible-blocks arg (=3600)   Approximate maximum allowed reversible
                                        blocks before shutdown. Will shut down
                                        if limit reached. Specify 0 to disable.
```

Resolves #447


linh2931 (approved):

greg7mdp:
I think it might be a good idea to allow the user to disable the check with  `--max-reversible-blocks 0`.

greg7mdp:
The name is a bit misleading. It is not really `max_reversible_blocks`, but `longuest_reversible_branch_allowed`.

heifner:
> The name is a bit misleading. It is not really `max_reversible_blocks`, but `longuest_reversible_branch_allowed`.

It is not that either. It is `max-reversible-blocks-allowed-on-best-branch` or `max-reversible-blocks` for short.

libraries/chain/controller.cpp
```diff
@@ -5503,8 +5515,8 @@ bool controller::should_terminate() const {
    return my->should_terminate();
 }
 
-bool controller::should_terminate(block_num_type head_block_num) const {
-   return my->should_terminate(head_block_num);
+bool controller::should_terminate(block_num_type head_block_num, block_num_type lib) const {
```
greg7mdp, line 5518: why expose this API in the controller interface.

libraries/chain/controller.cpp
```diff
@@ -4707,17 +4709,27 @@ struct controller_impl {
       return conf.block_validation_mode == validation_mode::LIGHT || conf.trusted_producers.count(producer);
    }
 
-   bool should_terminate(block_num_type head_block_num) const {
+   bool should_terminate(block_num_type head_block_num, block_num_type lib) const {
```
greg7mdp, line 4712: parameter name `head_block_num` is misleading. It is not necessarily the head since it is called in `create_block_state_i`.

I think it would be nicer for this to be a `template<class BS>` function, taking either a `block_state` or `block_state_legacy` as parameter.

greg7mdp:
> > The name is a bit misleading. It is not really `max_reversible_blocks`, but `longuest_reversible_branch_allowed`.
> 
> It is not that either. It is `max-reversible-blocks-allowed-on-best-branch` or `max-reversible-blocks` for short.

Not quite, it is used in `create_block_state_i` on a block pushed on a branch which may not be the best branch, right?

heifner:
> Not quite, it is used in `create_block_state_i` on a block pushed on a branch which may not be the best branch, right?

Correct, in that case it could be any branch.

libraries/chain/controller.cpp
heifner: Removed

greg7mdp:
> > Not quite, it is used in `create_block_state_i` on a block pushed on a branch which may not be the best branch, right?
> 
> Correct, in that case it could be any branch.

So really we cannot create any branch longer than `max_reversible_blocks`, right?

libraries/chain/controller.cpp
heifner: Renamed to `reversible_block_num`.

heifner:
> So really we cannot create any branch longer than `max_reversible_blocks`, right? If so, it really is `longuest_reversible_branch_allowed`.
> 
> I feel that `max_reversible_blocks` implies the total number of blocks in `fork_db`, which this definitely isn't.

The description is `Approximate maximum allowed reversible blocks`. I didn't want to tie down the implementation in case we have reason to change it. Would like to allow this config option to work even if we change to `forkdb.size()` or `fork_db.root()->block_num() + {head or reversible_block_num}`

greg7mdp:
>  I didn't want to tie down the implementation in case we have reason to change it...

Sounds good, thanks!

heifner:
> > I didn't want to tie down the implementation in case we have reason to change it...

Discussed with team and decided to tie the implementation to `forkdb.size()`.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -360,7 +360,7 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
 #endif
          ("enable-account-queries", bpo::value<bool>()->default_value(false), "enable queries to find accounts by various metadata.")
          ("max-reversible-blocks", bpo::value<uint32_t>()->default_value(config::default_max_reversible_blocks),
-          "Approximate maximum allowed reversible blocks before shutdown. Will shut down if limit reached.")
+          "Approximate maximum allowed reversible blocks before shutdown. Will shut down if limit reached. Specify 0 to disable.")
```
linh2931, line 363: I don't see where you set it to an internal big number to disable if the user configures it to 0.

heifner: https://github.com/AntelopeIO/spring/pull/493/files#diff-42e9f97eed543dd784b1b77a536c50ff8e2403d5bbd3b1fb7f5dd90201391061R4717

linh2931 (approved):

plugins/chain_plugin/chain_plugin.cpp
heifner: Now that I'm not adding the value the `reversible_block_num`, might be nicer to just set it to `std::numeric_limits<uint32_t>::max()` if `0`.

heifner: Ah, but the `if > 0` does avoid the mutex if that has any worth. I guess I'll leave it as is.

greg7mdp (approved):

ericpassmore:
Note:start
group: PROTOCOL
category: HTTP
summary: Adds new chain_plugin option max-reversible-blocks. Specifies the maximum number of reversible blocks beyond current LIB to allow before the node is gracefully shutdown.
Note:end

PR #507, opened 2024-08-08
treat rewriting identical genesis block as no-op in state history log

### LINKED ISSUE #413: SHiP write error when using --replay-blockchain
Had a node crash and didnt have snapshots available, so I wanted to just kick off a --replay-blockchain and let it rebuild the state, but when I launched it, I received an ambiguous ship write error that didnt indicate really what the issue was:

```error 2024-07-26T11:46:02.499 nodeos    main.cpp:217                  main                 ] 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:149 on_accepted_block

    {}
    nodeos  controller.cpp:3662 apply_block
rethrow
    {}
    nodeos  controller.cpp:4111 replay_push_block

    {}
    nodeos  chain_plugin.cpp:1152 plugin_startup
   ``` 
    
I understand this is to prevent accidentally wiping a full SHiP log by accidentally replaying, but the error should be more apparent that you need to manually remove the existing SHiP logs in order to complete the replay from genesis.

### PR DESCRIPTION
**Pull request by spoonincode**
Prior to leap 3.2, when writing a state history log entry with a block number less than the log's head block number, the log would be truncated to the new lower block number (truncated as in the file's contents are removed entirely).

This behavior allowed an accidental erasure of the entire state history log (see AntelopeIO/leap#659), so in leap 3.2 (AntelopeIO/leap#737) the behavior was modified: writing the genesis block in to a non empty state history log was immediately assert failure, and writing any other blocks in to the log would only truncate the log if the blockid is different; otherwise the logs were left unaltered.

But that change has a side effect that makes it impossible to replay a chain from genesis while keeping the state history logs intact (see #413). So instead, modify behavior such that writing any existing block -- even the genesis block -- with the same blockid is treated as a "no-op" (logs left unaltered), and disallowing rewriting the genesis block with a different blockid.

Resolves #413

### DIFF
--- a/tests/ship_log.cpp
+++ b/tests/ship_log.cpp
@@ -265,12 +265,6 @@ BOOST_DATA_TEST_CASE(basic_prune_test, bdata::xrange(2) * bdata::xrange(2) * bda
       return e.to_detail_string().find("missed a fork change") != std::string::npos;
    });
 
-   //start from genesis not allowed
-   BOOST_REQUIRE_EXCEPTION(t.add(2, payload_size, 'A', 'A');, eosio::chain::plugin_exception, [](const eosio::chain::plugin_exception& e) {
-      std::string err = e.to_detail_string();
-      return err.find("existing ship log") != std::string::npos && err.find("when starting from genesis block") != std::string::npos;
-   });
-
 } FC_LOG_AND_RETHROW() }
 
 BOOST_DATA_TEST_CASE(basic_test, bdata::xrange(2) * bdata::xrange(2) * bdata::xrange(2), enable_read, reopen_on_mark, remove_index_on_reopen)  { try {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/ship_log.cpp [hunk 1]
```diff
@@ -265,12 +265,6 @@ BOOST_DATA_TEST_CASE(basic_prune_test, bdata::xrange(2) * bdata::xrange(2) * bda
       return e.to_detail_string().find("missed a fork change") != std::string::npos;
    });
 
-   //start from genesis not allowed
-   BOOST_REQUIRE_EXCEPTION(t.add(2, payload_size, 'A', 'A');, eosio::chain::plugin_exception, [](const eosio::chain::plugin_exception& e) {
-      std::string err = e.to_detail_string();
-      return err.find("existing ship log") != std::string::npos && err.find("when starting from genesis block") != std::string::npos;
-   });
-
```
spoonincode, line 273: I think it's okay to remove this particular check: the test case is actually for pruned logs and pruned logs for this test case won't have the genesis block anyways. The test worked previously because the `blocknum==2` check was so early in the logic.

The test case further down exercises the genesis block behavior.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Allow replaying a chain from genesis while keeping the state history logs intact.
Note:end

heifner (approved):

greg7mdp (approved):

PR #508, opened 2024-08-08
Nodeos should refuse starting from a snapshot if there is no blocks log, but a fork database is present

### LINKED ISSUE #482: Nodeos should refuse starting from a snapshot if there is no blocks log, but a fork database is present

greg7mdp:
Resolves #482.


linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -1820,6 +1820,8 @@ struct controller_impl {
                   ("b", blog.first_block_num())("e", blog_head->block_num()) );
             block_states = read_from_snapshot( snapshot, blog.first_block_num(), blog_head->block_num() );
          } else {
+            EOS_ASSERT( !fork_db.file_exists(), snapshot_exception,
```
heifner, line 1823: Nothing wrong with the snapshot. I think this should be a `fork_database_exception`.

unittests/savanna_misc_tests.cpp
```diff
@@ -0,0 +1,43 @@
...
+   auto snapshot = B.snapshot();
+   A.produce_blocks(3);
+
+   B.close();
+   B.remove_blocks_log(); // remove blocks log, but *not* fork database
+   B.remove_state();
+   BOOST_CHECK_EXCEPTION(B.open_from_snapshot(snapshot), snapshot_exception,
+                         fc_exception_message_is("When starting from a snapshot with no block log, we shouldn't have a fork database either"));
+
+} FC_LOG_AND_RETHROW()
```
heifner, line 39: Nice test.

heifner (changes requested):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Nodes will not refuse to start from a snapshot if there is no blocks log but there is a fork database.
Note:end

libraries/chain/controller.cpp
greg7mdp: Thanks, done.

heifner (approved):

PR #500, opened 2024-08-08
Fix v1/chain/get_finalizer_info to update votes

heifner:
Fixed issue with `tracked_votes` not updating the vote info after the first insert.
Moved vote logging and prometheus logging into `tracked_votes`. This removes an extra call to `controller.vote_metrics()` per block.

Resolves #479

Issue #479, opened 2024-08-06
v1/chain/get_finalizer_info not updating vote info


heifner:
As can be seen by: http://3.227.1.63:8888/v1/chain/get_finalizer_info 
The votes are not updated after initial block timestamp of first vote.

heifner:
```
                  last_votes.emplace(f.fin_auth->public_key, std::move(v_info)); // track the voting information for the finalizer
```
Does not update as `emplace` does not update if already exists. 

- Should consolidate this and the existing Prometheus and text logging in producer plugin so that `controller.vote_metrics` is only called once.
- Should not create `tracked_votes` if `chain_api_plugin` not enabled.

heifner:
```
diff --git a/plugins/chain_plugin/tracked_votes.cpp b/plugins/chain_plugin/tracked_votes.cpp
index 5e4f39824e..faf4aa7a61 100644
--- a/plugins/chain_plugin/tracked_votes.cpp
+++ b/plugins/chain_plugin/tracked_votes.cpp
@@ -47,7 +47,7 @@ namespace eosio::chain_apis {
                      .voted_for_block_timestamp    = vm.voted_for_block_timestamp
                   };
 
-                  last_votes.emplace(f.fin_auth->public_key, std::move(v_info)); // track the voting information for the finalizer
+                  last_votes[f.fin_auth->public_key] = std::move(v_info); // track the voting information for the finalizer
                }
             };
 
diff --git a/tests/plugin_http_api_test_savanna.py b/tests/plugin_http_api_test_savanna.py
index dc2e9c8715..a0e5ed5b5b 100755
--- a/tests/plugin_http_api_test_savanna.py
+++ b/tests/plugin_http_api_test_savanna.py
@@ -61,8 +61,11 @@ try:
     ret_json = node.processUrllibRequest(resource, command, payload)
     assert type(ret_json["payload"]["active_finalizer_policy"]) == dict
     assert type(ret_json["payload"]["last_tracked_votes"]) == list
+    node.waitForLibToAdvance()
+    org_ret_json = ret_json
     # get_finalizer_info with empty content parameter
     ret_json = node.processUrllibRequest(resource, command, empty_content_dict, payload)
+    assert org_ret_json != ret_json, f"{ret_json}"
     assert type(ret_json["payload"]["active_finalizer_policy"]) == dict
     assert type(ret_json["payload"]["last_tracked_votes"]) == list
     # get_finalizer_info with invalid parameter
```

PR #512, opened 2024-08-08
Only track/log votes on recent blocks

heifner:
Follow on to #500. Only track / log votes for recent blocks to avoid overhead during syncing and replaying.

See #479

### DIFF
--- a/plugins/chain_plugin/tracked_votes.cpp
+++ b/plugins/chain_plugin/tracked_votes.cpp
@@ -34,10 +34,15 @@ namespace eosio::chain_apis {
             if (!_update_vote_block_metrics && !tracking_enabled && !chain::vote_logger.is_enabled(fc::log_level::info))
                return;
 
+            // do not bother tracking/logging when syncing or replaying
+            auto now = fc::time_point::now();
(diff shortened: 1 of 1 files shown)


plugins/chain_plugin/tracked_votes.cpp
```diff
@@ -34,10 +34,15 @@ namespace eosio::chain_apis {
             if (!_update_vote_block_metrics && !tracking_enabled && !chain::vote_logger.is_enabled(fc::log_level::info))
                return;
 
+            // do not bother tracking/logging when syncing or replaying
+            auto now = fc::time_point::now();
+            if (now - block->timestamp > fc::minutes(5) && (block->block_num() % 1000 != 0))
```
greg7mdp, line 39: Do we still want to do it every thousand blocks?

greg7mdp (approved):

plugins/chain_plugin/tracked_votes.cpp
heifner: I think so, gives a bit of status as it syncs.

linh2931 (approved):

ericpassmore:
Note:start
group: LOGGING
category: INTERNALS
summary: More efficient logging track votes only for recent blocks.
Note: end

PR #514, opened 2024-08-09
split snapshot contract tables in to separate snapshot sections for faster parallel loading

spoonincode:
Snapshots have all chainbase contract tables (table id, primary key table, secondary key tables) commingled together as a single `contract_tables` snapshot section. This is at odds with the new parallelization scheme where each snapshot section is loaded in parallel since each snapshot section (otherwise) refers to a single chainbase table.

Change contract tables to be stored in separate snapshot sections so each table (table id, primary key table, secondary key tables) can be loaded in parallel. For a recent EOS snapshot I've been consistently using for performance testing on recent PRs, snapshot load time is reduced from 60 seconds to 32 seconds with this change.

The new snapshot format is about 0.5% larger than before, for this recent EOS block.


spoonincode:
hm I will look in to failures; thought I ran that test locally

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: PERFORMANCE
summary: Modify snapshot sections to allow parallel loading, significant decrease in snapshot loading time. 
Note:end

greg7mdp (approved):

PR #520, opened 2024-08-12
Pause production when not receiving votes

### LINKED ISSUE #495: Pause production when not receiving votes
## Background

This behavior is intended as a medium-term restriction in nodeos to help block producers be aware of common misconfigurations of their nodes that would hinder vote propagation and advancement of finality. It makes an assumption that the same entities who act as proposers are also acting as finalizers. It also assumes that the system contract that proposes proposer and finality policies couples the finalizer keys to producer accounts via the finalizer key description. These assumptions are reasonable for how Spring is intended to be used in the near future.

This behavior may be reverted in the future when better tooling and/or practices are formed to give the network confidence that finality advancement performs well even without these restrictions. It will especially be necessary to revert this behavior if in the future we wish to fully decouple proposer and finalizer roles.

## Changes

There are two conditions that each independently can trigger a implicit pause in block production. An implicit pause in block production can be undone once the conditions that triggered them are no longer active. If block production was explicitly paused via the RPC, then block production remains paused even if the implicit pause is removed. If block production was explicitly unpaused via the RPC, then the internal pause should also be reset. Finally, the `paused` query should always reflect whether block production is paused or not regardless of whether it is due to an implicit pause or explicit pause or both.

In the conditions below, the description gives a period of 6 seconds to receive expected votes before the conditions is triggered. To explain expected behavior, consider the following analogy. A 6-second timer (for each condition) is started when a block is accepted. The timer is disabled if an expected vote (related to the specific condition) for the block is received or if a timer (for that same condition type) for any other block is disabled. Otherwise, if the timer goes off after 6 seconds, then the condition is triggered and remains triggered unless the timer gets disabled. This is only used as an analogy to describe what correct timing behavior is expected. The actual implementation does not need to use explicit timers.

### Condition 1: Do not receive other votes

After Savanna transition, if nodeos does not receive any expected votes from others for a period of time exceeding 6 seconds, then pause production. Expected votes from others are votes from finalizers in the active (or pending) policy that have a description consisting of a producer account name that is not locally configured in nodeos as a producer.

This condition is typically triggered because a canary nodeos guarding the producer node is misconfigured to not process any votes (`vote-threads` is 0).

### Condition 2: Do not receive your own votes

After Savanna transition, if nodeos does not receive any expected votes from self (including internal delivery within controller) for a period of time exceeding 6 seconds, then pause production. Expected votes from self are votes from finalizers in the active (or pending) policy that have a description consisting of a producer account name that is locally configured in nodeos as a producer.

This condition is typically triggered because the producer nodeos has not configured the necessary BLS finalizer key.

heifner:
- Updates integration tests to use producer names for finalizer descriptions to match what core system contract does and what `producer_plugin` now expects.
- Updated vote signal to include the active finalizer authority and/or the pending finalizer authority
  - This is not strictly necessary as the `producer_plugin` could track the BLS public keys of the active and pending finalizer policies and see if the those vote BLS public keys match any of the producer account names of the finalizers with the same name <=> description. I thought it was nice to signal the finalizer authority. For one thing, we can now log the finalizer description of the vote instead of just the BLS public key. However, if we think the overhead of this is too much I could pull it out and go with the producer plugin doing all the heavy lifting.

- New config option `production-pause-vote-timeout-ms`
```
  --production-pause-vote-timeout-ms arg (=6000)
                                        Received vote timeout. If no vote from
                                        producer-name finalizers or other
                                        finalizers then pauses block
                                        production. 0 disables.
```

Resolves #495


ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Pause block production when not receiving votes. This behavior is intended as a medium-term restriction in nodeos to help block producers be aware of common misconfigurations of their nodes that would hinder vote propagation and advancement of finality.
Note:end

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3959,10 +3962,22 @@ namespace eosio {
...
+                                                const vote_message_ptr& msg,
+                                                const finalizer_authority_ptr& active_auth,
+                                                const finalizer_authority_ptr& pending_auth) {
+      auto get_desc = [&]() -> std::string {
+         if (active_auth)
+            return active_auth->description;
+         if (pending_auth)
+            return pending_auth->description;
+         return "unknown";
+      };
```
greg7mdp, line 3975: avoiding string copies
```suggestion
       auto get_desc = [&]() -> const std::string& {
         static constexpr std::string unknown {"unknown"};
         if (active_auth)
            return active_auth->description;
         if (pending_auth)
            return pending_auth->description;
         return unknown;
      };
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1034,7 +1158,10 @@ void producer_plugin::set_program_options(
    producer_options.add_options()
       ("enable-stale-production,e", boost::program_options::bool_switch()->notifier([this](bool e){my->_production_enabled = e;}),
        "Enable block production, even if the chain is stale.")
-         ("pause-on-startup,x", boost::program_options::bool_switch()->notifier([this](bool p){my->_pause_production = p;}), "Start this node in a state where production is paused")
+         ("pause-on-startup,x", boost::program_options::bool_switch()->notifier([this](bool p){my->_pause_production = p;}),
+          "Start this node in a state where production is paused")
+         ("production-pause-vote-timeout-ms", bpo::value<uint32_t>()->default_value(config::default_production_pause_vote_timeout_ms),
```
linh2931, line 1163: I am debating what's a best name for this. It sounds like `pause vote` not `pause production`. Maybe just `vote-timeout-ms` as the name, and the description describe what option does in details?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -499,8 +499,14 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
    pending_block_mode                                _pending_block_mode = pending_block_mode::speculating;
    unapplied_transaction_queue                       _unapplied_transactions;
+   alignas(hardware_destructive_interference_sz)
    std::atomic<int32_t>                              _max_transaction_time_ms; // modified by app thread, read by net_plugin thread pool
+   alignas(hardware_destructive_interference_sz)
    std::atomic<uint32_t>                             _received_block{0};       // modified by net_plugin thread pool
+   alignas(hardware_destructive_interference_sz)
+   std::atomic<fc::time_point>                       _last_other_vote_recevied;
+   std::atomic<fc::time_point>                       _last_producer_vote_received;
+   fc::microseconds                                  _production_pause_vote_timeout;
```
linh2931, line 509: maybe add a `_us` postfix.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -510,6 +516,10 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    bool                                              _disable_subjective_p2p_billing              = true;
    bool                                              _disable_subjective_api_billing              = true;
    fc::time_point                                    _irreversible_block_time;
+   fc::time_point                                    _accepted_block_time;
+   bool                                              _is_savanna_active                           = false;
+   bool                                              _is_producer_active_finalizer                = false;
+   bool                                              _other_active_finalizers                     = false;
```
linh2931, line 522: What does this boolean mean?

linh2931 (commented):
Not done yet. Just post comments so far.

plugins/net_plugin/net_plugin.cpp
heifner: I don't think this avoids any string copies over the existing. Both will create a string in the `mutable_variant_object` `operator()( std::string key, T&& var )` `variant( std::string val )` constructor via RVO once I add `return std::string{"unknown");`

plugins/producer_plugin/producer_plugin.cpp
heifner: This name was decided in our meeting. We wanted `production-pause-*` as we might add new `production-pause-*` options in the future.

plugins/producer_plugin/producer_plugin.cpp
heifner: I generally go with `_us` and `_ms` for int types. `fc::microseconds` prevents misuse.

plugins/producer_plugin/producer_plugin.cpp
heifner: Are there any active finalizers that are not part of the producers of this nodeos. I'll add a comment.

plugins/net_plugin/net_plugin.cpp
greg7mdp: right, my bad!

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -649,6 +694,81 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+      }
+
+      if (_producers.contains(auth_desc)) { // _producers not modified, thread safe
+         _last_producer_vote_received.store(now, std::memory_order_relaxed);
+      } else {
+         _last_other_vote_recevied.store(now, std::memory_order_relaxed);
+      }
+   }
+
+   // true if paused do to not receiving votes
```
greg7mdp, line 756:
```suggestion
   // true if paused due to not receiving votes
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -649,6 +694,81 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+   bool is_implicitly_paused() const {
+      if (_producers.empty() || _production_pause_vote_timeout.count() == 0)
+         return false;
+      if (!_is_savanna_active)
+         return false; // no implicit pause in legacy
+      if (!_is_producer_active_finalizer)
+         return false;
```
greg7mdp, line 763: This seems wrong. Couldn't we be paused because of  not receiving votes from `_other_active_finalizers`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -499,8 +499,14 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    block_timing_util::producer_watermarks            _producer_watermarks;
    pending_block_mode                                _pending_block_mode = pending_block_mode::speculating;
    unapplied_transaction_queue                       _unapplied_transactions;
+   alignas(hardware_destructive_interference_sz)
    std::atomic<int32_t>                              _max_transaction_time_ms; // modified by app thread, read by net_plugin thread pool
+   alignas(hardware_destructive_interference_sz)
    std::atomic<uint32_t>                             _received_block{0};       // modified by net_plugin thread pool
+   alignas(hardware_destructive_interference_sz)
+   std::atomic<fc::time_point>                       _last_other_vote_recevied;
```
greg7mdp, line 507:
```suggestion
   std::atomic<fc::time_point>                       _last_other_vote_received;
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -626,12 +641,42 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               }
+               if (!_other_active_finalizers) {
+                  _other_active_finalizers = std::ranges::any_of(fin_policy->finalizers, [&](const auto& f) {
+                     return !_producers.contains(to_account_name_safe(f.description));
+                  });
+               }
+            }
+         }
+      }
+   }
```
greg7mdp, line 670: Unless I'm mistaken, you can simplify like this:
```suggestion
   void update_active_finalizers() {
      if (_is_savanna_active && !_producers.empty()) {
         auto& chain  = chain_plug->chain();
         finalizer_policy_ptr fin_policy = chain.head_active_finalizer_policy();
         assert(fin_policy);
         auto is_producers_in = [&](const auto& f) -> bool {
            return _producers.contains(to_account_name_safe(f.description);
         };

         _is_producer_active_finalizer = std::ranges::any_of(fin_policy->finalizers, is_producers_in);
         _other_active_finalizers = !std::ranges::all_of(fin_policy->finalizers, is_producers_in);

         if (!_is_producer_active_finalizer || !_other_active_finalizers) {
            if (fin_policy = chain.head_pending_finalizer_policy(); fin_policy) {
               if (!_is_producer_active_finalizer)
                  _is_producer_active_finalizer = std::ranges::any_of(fin_policy->finalizers, is_producers_in);
               if (!_other_active_finalizers)
                  _other_active_finalizers = !std::ranges::all_of(fin_policy->finalizers, is_producers_in);
            }
         }
      }
   }
```

plugins/producer_plugin/producer_plugin.cpp
heifner: Oops, meant to remove that.

plugins/producer_plugin/producer_plugin.cpp
heifner: Looks correct to me, thanks.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -510,6 +516,11 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    bool                                              _disable_subjective_p2p_billing              = true;
    bool                                              _disable_subjective_api_billing              = true;
    fc::time_point                                    _irreversible_block_time;
+   fc::time_point                                    _accepted_block_time;
```
greg7mdp, line 519: Can we rename this `_last_accepted_block_time`?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1363,6 +1485,17 @@ void producer_plugin_impl::plugin_startup() {
             }
          }));
 
+         if (!_producers.empty()) { // track votes if producer to verify votes are being processed
+            auto on_vote_signal = [this]( const vote_signal_params& vote_signal ) {
+               try {
+                  const auto& [connection_id, status, msg, active_auth, pending_auth] = vote_signal;
+                  on_vote(connection_id, status, msg, active_auth, pending_auth);
```
greg7mdp, line 1492: No big deal, but why not have `on_vote` take `const vote_signal_params& ` and unbind in `on_vote`?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1813,16 +1947,25 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
+   } else if (is_implicitly_paused()) {
+      fc_elog(_log, "Not producing block because no recent votes, last producer vote ${pv}, other votes ${ov}, last block time ${bt}",
+              ("pv", _last_producer_vote_received.load(std::memory_order_relaxed))
+              ("ov", _last_other_vote_received.load(std::memory_order_relaxed))("bt", _accepted_block_time));
+      _pending_block_mode = pending_block_mode::speculating;
+      not_producing_when_time = true;
    }
 
-   if (in_speculating_mode()) {
+   if (in_speculating_mode() && !not_producing_when_time) {
```
greg7mdp, line 1968: why ` && !not_producing_when_time`?

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: It separates concerns of the signaled type/info from what `on_vote` needs.

plugins/producer_plugin/producer_plugin.cpp
heifner: This matches `_irreversible_block_time` in naming and I would prefer not to rename `_irreversible_block_time`.

plugins/producer_plugin/producer_plugin.cpp
heifner: `!not_producing_when_time to avoid tight spin because of error or paused production` I'll add a comment

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: > Not correct. Any of not != not all.

I don't follow. it is literally the same.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -626,12 +641,43 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+      assert(fin_policy);
+      _is_producer_active_finalizer = std::ranges::any_of(fin_policy->finalizers, [&](const auto& f) {
+         return _producers.contains(to_account_name_safe(f.description));
+      });
+      _other_active_finalizers = std::ranges::any_of(fin_policy->finalizers, [&](const auto& f) {
+         return !_producers.contains(to_account_name_safe(f.description));
+      });
+      if (!_is_producer_active_finalizer || !_other_active_finalizers) {
+         if (fin_policy = chain.head_pending_finalizer_policy(); fin_policy) {
+            if (!_is_producer_active_finalizer) {
```
linh2931, line 659: This section of code is the same as starting from line 651.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -649,6 +695,79 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+   void on_vote(uint32_t connection_id, vote_result_t status, const vote_message_ptr& msg,
+                const finalizer_authority_ptr& active_auth, const finalizer_authority_ptr& pending_auth) {
+      switch( status ) {
+      case vote_result_t::success:
+         break;
+      case vote_result_t::unknown_public_key:
+      case vote_result_t::invalid_signature:
+      case vote_result_t::max_exceeded:
+      case vote_result_t::unknown_block:
+      case vote_result_t::duplicate:
```
linh2931, line 726: Maybe `duplicate` can be counted as success so the timer has a new value.

plugins/producer_plugin/producer_plugin.cpp
heifner: Not clear how to make it much simplier.

plugins/producer_plugin/producer_plugin.cpp
heifner: Hmm, yeah, that makes sense.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -649,6 +695,79 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         return false;
+      if (!_is_savanna_active)
+         return false; // no implicit pause in legacy
+      if (_producers.contains(eosio::chain::config::system_account_name)) // disable implicit pause for eosio
+         return false;
+
+      // need a vote within timeout of last accepted block
+      auto time_limit = _accepted_block_time - _production_pause_vote_timeout;
+
+      return (_is_producer_active_finalizer && _last_producer_vote_received.load(std::memory_order_relaxed) < time_limit)
```
linh2931, line 769: Maybe `_accepted_block_time- _last_producer_vote_received.load(std::memory_order_relaxed) >  _production_pause_vote_timeout` is more clearer.

plugins/producer_plugin/producer_plugin.cpp
linh2931: OK.

plugins/producer_plugin/producer_plugin.cpp
heifner: Thanks! I do think this makes it easier to read.

linh2931 (approved):

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -649,6 +695,82 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         return false;
+
+      auto last_producer_vote_received = _last_producer_vote_received.load(std::memory_order_relaxed);
+      auto last_other_vote_received    = _last_other_vote_received.load(std::memory_order_relaxed);
+
+      // need a vote within timeout of last accepted block
+      return (_is_producer_active_finalizer &&
+              _accepted_block_time - last_producer_vote_received > _production_pause_vote_timeout)
+          || (_other_active_finalizers &&
+              _accepted_block_time - last_other_vote_received    > _production_pause_vote_timeout);
```
greg7mdp, line 773: Not an issue, but I liked the previous version better. Here we can have negative time values and I think it is less clear.

greg7mdp (approved):

PR #521, opened 2024-08-12
Fix some incorrect extension lookups using `lower_bound` 

greg7mdp:
We typically use the map `lower_bound()` api to retrieve an iterator to an extension we are interested in. However this api returns the first entry in the map that is *not less* than the key we are looking for. If the key we are looking for is not present, it could return a different, greater key.

Foe example, see:

https://github.com/AntelopeIO/spring/blob/b58cbbcdc532f430e87a430ed1209bc9c9fce3e2/libraries/chain/block_state.cpp#L197-L201

This doesn't cause an issue right now because `finality_extension` is the last of the header extensions, and `quorum_certificate_extension` is the last of the block extensions (so they have the highest ids), but it is quite fragile.

A common pattern is to do a `count()` to check if the extension is present, followed by a `lower_bound()` to retrieve an iterator to it. I guess it could make sense in cases where we can have multiple extensions of a given type, but for *unique* extensions, I feel that doing a single lookup using `find()` instead of two lookups is better.

This PR updates the extension lookups to use `find()` instead of `lower_bound()`

### DIFF
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -369,8 +369,7 @@ block_header_state block_header_state::next(const signed_block_header& h, valida
    // retrieve protocol_feature_activation from incoming block header extension
    // -------------------------------------------------------------------------
    vector<digest_type> new_protocol_feature_activations;
-   if( exts.count(protocol_feature_activation::extension_id() > 0) ) {
-      auto  pfa_entry = exts.lower_bound(protocol_feature_activation::extension_id());
+   if (auto  pfa_entry = exts.find(protocol_feature_activation::extension_id()); pfa_entry != exts.end()) {
       auto& pfa_ext   = std::get<protocol_feature_activation>(pfa_entry->second);
       new_protocol_feature_activations = pfa_ext.protocol_features;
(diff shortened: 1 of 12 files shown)


heifner (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -369,18 +369,17 @@ block_header_state block_header_state::next(const signed_block_header& h, valida
    // retrieve protocol_feature_activation from incoming block header extension
    // -------------------------------------------------------------------------
    vector<digest_type> new_protocol_feature_activations;
-   if( exts.count(protocol_feature_activation::extension_id() > 0) ) {
-      auto  pfa_entry = exts.lower_bound(protocol_feature_activation::extension_id());
+   if (auto  pfa_entry = exts.find(protocol_feature_activation::extension_id()); pfa_entry != exts.end()) {
```
spoonincode, line 372: The behavior of `find()` is different then `lower_bound()` when the mutlimap contains identical keys. Fortunately it seems all these extensions currently `enforce_unique = true` so there appears to be no way to concoct multiple extensions of a single id in the header. (if we did allow that, I think this would introduce a determinism problem here)

spoonincode (approved):

PR #524, opened 2024-08-12
Provide 500ms for speculative blocks

heifner:
Since https://github.com/AntelopeIO/leap/pull/1481 speculative blocks honor block deadlines so that "Start speculative blocks every block interval but not more often". However, since 50ms threshold was use it could set a deadline that would be hit before another block came causing a needless restart of a transaction and restart of a block.

For example:
```
info  2024-08-12T15:18:22.511 nodeos    controller.cpp:3525           log_applied          ] Received block ebede3ba155c0f61... #2154995 @ 2024-08-12T15:18:22.500 signed by lioninjungle [trxs: 2, lib: 2154993, net: 432, cpu: 351, elapsed: 619, time: 1966, latency: 11 ms]
debug 2024-08-12T15:18:22.511 nodeos    producer_plugin.cpp:2005      start_block          ] Starting block #2154996 2024-08-12T15:18:23.000 producer lioninjungle
debug 2024-08-12T15:18:22.511 nodeos    producer_plugin.cpp:2162      remove_expired_trxs  ] Processed 0 expired transactions of the 1 transactions in the unapplied queue.
debug 2024-08-12T15:18:22.511 nodeos    producer_plugin.cpp:2643      schedule_production_ ] Speculative Block Created; Scheduling Speculative/Production Change
debug 2024-08-12T15:18:22.511 nodeos    producer_plugin.cpp:2699      schedule_delayed_pro ] Scheduling Speculative/Production Change at 2024-08-12T15:18:47.500
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:869       restart_speculative_ ] Restarting exhausted speculative block #2154996
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:309       report               ] Block #2154996 lioninjungle trx idle: 173057us out of 173987us, success: 0, 0us, fail: 0, 0us, transient: 0, 0us, other: 930us
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:2005      start_block          ] Starting block #2154996 2024-08-12T15:18:23.000 producer lioninjungle
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:2162      remove_expired_trxs  ] Processed 0 expired transactions of the 2 transactions in the unapplied queue.
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:2553      process_incoming_trx ] Processing 1 pending transactions
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:2579      process_incoming_trx ] Processed 1 pending transactions, 0 left
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:2643      schedule_production_ ] Speculative Block Created; Scheduling Speculative/Production Change
debug 2024-08-12T15:18:22.685 nodeos    producer_plugin.cpp:2699      schedule_delayed_pro ] Scheduling Speculative/Production Change at 2024-08-12T15:18:47.500
debug 2024-08-12T15:18:22.971 nodeos    producer_plugin.cpp:783       on_incoming_block    ] received incoming block 2154996 0020e1f42cb0c29fb64575c9b3ed4a4135a43180eab3a0a950aa15c3bf189567
debug 2024-08-12T15:18:22.971 nodeos    producer_plugin.cpp:309       report               ] Block #2154996 lioninjungle trx idle: 284523us out of 286360us, success: 3, 1161us, fail: 0, 0us, transient: 0, 0us, other: 676us
debug 2024-08-12T15:18:22.973 nodeos    producer_plugin.cpp:666       on_block             ] Removed applied transactions before: 3, after: 1
info  2024-08-12T15:18:22.973 nodeos    controller.cpp:3525           log_applied          ] Received block 2cb0c29fb64575c9... #2154996 @ 2024-08-12T15:18:23.000 signed by lioninjungle [trxs: 2, lib: 2154994, net: 432, cpu: 251, elapsed: 641, time: 2026, latency: -26 ms]
```

Instead give a speculative block a full block interval (500ms) before considering it exhausted.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1868,14 +1868,15 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
    if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has passed
       // speculative block, no reason to start a block that will immediately be re-started, set deadline in the future
-      // a block should come in during this time, if not then just keep creating the block every produce_block_cpu_effort
-      if (now + fc::milliseconds(config::block_interval_ms/10) > _pending_block_deadline) {
-         _pending_block_deadline = now + _produce_block_cpu_effort;
+      // a block should come in during this time, if not then just keep creating the block every block_interval_ms
+      if (now + fc::milliseconds(config::block_interval_ms) > _pending_block_deadline) {
+         _pending_block_deadline = now + fc::milliseconds(config::block_interval_ms);
       }
    }
    const auto& preprocess_deadline = _pending_block_deadline;
 
-   fc_dlog(_log, "Starting block #${n} ${bt} producer ${p}", ("n", pending_block_num)("bt", block_time)("p", scheduled_producer.producer_name));
+   fc_dlog(_log, "Starting block #${n} ${bt} producer ${p}, deadline ${d}",
+           ("n", pending_block_num)("bt", block_time)("p", scheduled_producer.producer_name)("d", _pending_block_deadline));
 
    try {
 


ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Provide 500ms for speculative blocks.
Note:end

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1868,14 +1868,15 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
    if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has passed
       // speculative block, no reason to start a block that will immediately be re-started, set deadline in the future
-      // a block should come in during this time, if not then just keep creating the block every produce_block_cpu_effort
-      if (now + fc::milliseconds(config::block_interval_ms/10) > _pending_block_deadline) {
-         _pending_block_deadline = now + _produce_block_cpu_effort;
+      // a block should come in during this time, if not then just keep creating the block every block_interval_ms
```
greg7mdp, line 1871: that three line comment doesn't seem very clear to me.

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: I re-read the comments. Not sure what I can change/add there. Any ideas?

greg7mdp: I guess I don't understand why `block_interval_ms` is the right deadline, why we thought that 50ms was better before, etc...
I know we talked about it during the meeting but I don't really understand the reasoning behind it. Maybe you can explain it to me in a conversation and then I can suggest a new comment.

heifner: This is for speculative block production. Before we used `_produce_block_cpu_effort` which is `block_interval_ms - block-offset`. The block offset is just a configurable amount that provides time for the block to reach the next block producer. I don't really see a point of factoring that into this. The 50ms before was just a threshold for if the `_produce_block_cpu_effort` should be added on to now. We used to wait indefinitely for a block to come in when creating a speculative block. We decided it was good to re-start a speculative block every block interval so that any speculative transactions would be able to run using the current block time. This change simplified the logic and just uses `block_interval_ms` so we get a new speculative block every 500ms. In general we expect a new block to be received in that time frame which will abort the speculative block, apply the received block, and start a new speculative block.

greg7mdp: Wow, this is a great explanation. Thank you Kevin.

I tried to translate my understanding into a comment:

```
// in speculating mode, we want to start creating blocks on a regular basis so that we can
// execute speculative transaction with the current block time.
// If a block is received, we will be able to abort the speculative block when the current
// transaction completes (a maximum of 150ms) - this because we currently don't interrupt transactions.
// We could set a `_pending_block_deadline` less than 150ms in order to minimize the delay before
// an incoming block can be processed. However the drawback is that the speculative block could
// potentially be restarted multiple times before incoming blocks.
```

PS: Why is it a problem if the speculative  block is restarted?

heifner: There is no problem to restart the speculative block, just wasteful. Wasteful in the time it takes to abort the current block and start a new one; but mostly wasteful in that you abort the trx at the deadline and it has to be restarted on the next speculative block.

greg7mdp: > There is no problem to restart the speculative block, just wasteful. Wasteful in the time it takes to abort the current block and start a new one; but mostly wasteful in that you abort the trx at the deadline and it has to be restarted on the next speculative block.

Do we care if it is wasteful? Since we didn't get a block we don't have much to do anyways, right? And maybe that makes nodeos more responsive to things like `get_info`?

heifner: There could be thousands of trxs queued up ready to execute. I think that is the important thing to optimize in this case.

This is one case where knowing if the node is a relay node vs an API node would allow us to make different choices.

greg7mdp: Thanks for the comment update!

greg7mdp (approved):

PR #519, opened 2024-08-12
Disable `NEON` on arm in boost `unordered`.

greg7mdp:
Disable `NEON` on arm in boost `unordered` because of conflicting declaration for `float16_t` in softfloat library.

Resolves #517.

Issue #517, opened 2024-08-12
Cannot build unit_tests on ARM: conflicting declaration for float types


yarkineos:
Got things like
```
In file included from /home/parallels/code/antelope_repos/spring/libraries/boost/libs/unordered/include/boost/unordered/detail/foa/core.hpp:62,
                 from /home/parallels/code/antelope_repos/spring/libraries/boost/libs/unordered/include/boost/unordered/detail/foa/table.hpp:18,
                 from /home/parallels/code/antelope_repos/spring/libraries/boost/libs/unordered/include/boost/unordered/unordered_flat_map.hpp:14,
                 from /home/parallels/code/antelope_repos/spring/unittests/savanna_cluster.hpp:8,
                 from /home/parallels/code/antelope_repos/spring/unittests/forked_tests_savanna.cpp:1:
/usr/lib/gcc/aarch64-linux-gnu/11/include/arm_neon.h:72:16: error: conflicting declaration ‘typedef __fp16 float16_t’
   72 | typedef __fp16 float16_t;
      |                ^~~~~~~~~
In file included from /home/parallels/code/antelope_repos/spring/libraries/softfloat/source/include/softfloat.hpp:3,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/database_utils.hpp:5,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/code_object.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/wasm_interface.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/config.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/block_timestamp.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/block_header.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/block_header_state_legacy.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/block_state_legacy.hpp:3,
                 from /home/parallels/code/antelope_repos/spring/libraries/chain/include/eosio/chain/controller.hpp:2,
                 from /home/parallels/code/antelope_repos/spring/libraries/testing/include/eosio/testing/tester.hpp:2,
[...]

PR #532, opened 2024-08-13
also add `BOOST_UNORDERED_DISABLE_NEON` for users of libtester

spoonincode:
Follow up to #519. I think we probably need to do this or run the risk of ODR violations (on ARM) if something else uses `boost::unordered` as part of the external test compilation. It's also possible it's already a ODR violation as-is since `apply_context.hpp` -- exposed as public libtester interface -- does a `#include <boost/unordered/unordered_flat_map.hpp>`. (I admit to not looking at `unordered`'s impl yet to know if this ODR violation theory is possible or not, but seems like this is right thing to do -- it's possible original reporter in #517 will need it anyways to compile their external tests)

### DIFF
--- a/CMakeModules/EosioTester.cmake.in
+++ b/CMakeModules/EosioTester.cmake.in
@@ -35,6 +35,8 @@ else ( APPLE )
    set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} ${CMAKE_CXX_FLAGS} -Wall")
 endif ( APPLE )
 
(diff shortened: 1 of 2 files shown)


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Add compile option BOOST_UNORDERED_DISABLE_NEON to better support arm.
Note:end

Issue #535, opened 2024-08-14
Use shared_ptr instead of optional<shared_ptr> for latest_proposed_proposer_policy and latest_pending_proposer_policy


linh2931:
This is to avoid an extra thing to encode in the base digest.

PR #536, opened 2024-08-14
Use shared_ptr instead of std::optional<shared_ptr> for latest_proposed_proposer_policy and latest_pending_proposer_policy

linh2931:
Resolves https://github.com/AntelopeIO/spring/issues/535


libraries/chain/block_header_state.cpp
```diff
@@ -24,8 +24,12 @@ digest_type block_header_state::compute_base_digest() const {
...
    fc::raw::pack( enc, *active_proposer_policy );
 
-   fc::raw::pack( enc, latest_proposed_proposer_policy );
-   fc::raw::pack( enc, latest_pending_proposer_policy );
+   if (latest_proposed_proposer_policy) {
+      fc::raw::pack( enc, *latest_proposed_proposer_policy );
+   }
+   if (latest_pending_proposer_policy) {
+      fc::raw::pack( enc, *latest_pending_proposer_policy );
+   }
```
greg7mdp, line 32: why this change? I believe we should pack the `shared_ptr` itself.

greg7mdp (changes requested):

libraries/chain/block_header_state.cpp
linh2931: We don't pack `activated_protocol_features` (a shared ptr) if it does not exist. We also pack the actual content of `active_proposer_policy` (we assert it is always present).

heifner: For a digest it doesn't matter, but if we were packing for writing we would want the shared_ptr written so that we could unpack it. Otherwise, you would not know that entry from the next. 

Seems we should be consistent with `activated_protocol_features`. Curious if @arhag has thoughts on this.

heifner: I talked to @arhag about this. 

- Even though this is for a digest, it is good practice to keep it consistent with what you would do for a packed that you would write. 
- `activated_protocol_features` should always be present. We can change that to an `assert` and just write it out.
- For things that are optionally present we should always pack the bool indicating if they are there. So that means packing the shared_ptr, not the thing itself.

heifner (approved):

greg7mdp (approved):

arhag (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Use shared pointer to reduce cost of encoding base digest.
Note:end

PR #537, opened 2024-08-14
Fix issue with fsi handling and `block_ref` creation.

### LINKED ISSUE #534: Bugs in how fsi is handled
There are a few problems with the way finalizer safety information (fsi) records are handled in the code.

First, there is an implementation issue in which everywhere in the code that sets a `block_ref` in the fsi is not correctly setting the finality digest (leaving it the default construct empty hash) despite it existing as a record in `block_ref`. This does not break anything since `block_id` is currently being used to check to see if a locked block or last voted block is an ancestor of a block being considered for voting. But the data stored to the fsi file would not be correct since `block_ref` is not constructed properly. To allow the compiler to assist in finding all locations where this happens, a constructor should be added to `block_ref` that takes all three fields so that the compiler guarantees the appropriate data is provided everywhere a `block_ref` is set in the fsi. 

It may be necessary to keep the concept of an `empty` `block_ref` to capture the concept of no `last_vote` in the fsi. One way of achieving this is to add a default constructor for `block_ref` that allows creating this appropriate `empty` representation. A cleaner way (though one that adds an extra byte in the serialization of fsi) is to remove the concept of an `empty` `block_ref` and instead change the type of `last_vote` in the fsi to `std::optional<block_ref>`.

Second, there was a design flaw related to rule 2 (the rule that determines whether to vote strong vs weak) that needs to be corrected. It is possible for the latest (weak) vote to mask that the vote prior (weak) vote for a block not on the same branch. Then if later an attempt is made at a strong vote on a block with a latest QC claim block timestamp less than that of block the prior weak vote was on, `decide_vote` may (incorrectly) conclude it can vote strong because the latest vote was on a block that is ancestor of the block it is considering voting on. In reality, a strong vote would violate rule 2 (see https://github.com/AntelopeIO/spring/issues/92#issuecomment-2287315555) since it would commit to an implied time interval that contains the timestamp of the block the prior weak vote was on despite that block not being an ancestor of the block it is considering voting on. Therefore, `decide_vote` would have to vote weak in that situation.

To correct this masking problem, a slightly stricter check than what is required for rule 2 can be implemented. A new boolean called `votes_forked_since_latest_strong_vote` should be tracked in the fsi. A default fsi can start with `votes_forked_since_latest_strong_vote` set to false. And a strong vote should always set `votes_forked_since_latest_strong_vote` to false. On a weak vote, if `votes_forked_since_latest_strong_vote` was already true, then it should remain true. The only way `votes_forked_since_latest_strong_vote` can change from false to true is on a weak vote for a block `b` where the `last_vote` references a block that is not an ancestor of `b`.

In the `if (can_vote)` block https://github.com/AntelopeIO/spring/blob/7bd7153e4897e333ec1c31b5f62ec3b63c5b9d39/libraries/chain/finality/finalizer.cpp#L67-L70 we still need to decide if the vote will be strong or weak. However, it no longer needs `last_vote_range_start` to determine that; `last_vote_range_start` should be removed from the fsi. Instead, it can first compare `fsi.last_vote.timestamp <= bsp->core.latest_qc_block_timestamp()`. If that condition is true (or if `fsi.last_vote` is empty), then we immediately know that we can vote strong. If it is not true, then we can only vote strong if `votes_forked_since_latest_strong_vote == false` and `bsp->core.extends(fsi.last_vote.block_id)`.

These changes to fsi do not require versioning. We still have the opportunity to retroactively change the fsi definition and we should take advantage of that rather than complicating the logic further by adding a new version. We should not bother supporting the old (incorrect) version that still tracked `last_vote_range_start`.

Finally, there were other small things I noticed that do not necessarily need to be changed as part of this issue, but is worth noting:

1. We only consider strong QCs and thus strong votes as evidence from the blockchain when updating the fsi during syncing (prior to first vote). I believe we decided to do this because we were only concerned about the monotonicity condition and the safety and liveness conditions related to the `lock` in the fsi; we had a stricter protection for monotonicity based on using the nodeos startup time to protect against votes on earlier timestamp, and the `lock` is only updated on strong votes. However, there is useful information from weak votes as well. Weak votes may not update the `lock` in the fsi, but they do update the `last_vote` and now (with the changes in this issue) the `votes_forked_since_latest_strong_vote` boolean as well. Even ignoring the monotonicity concern and assuming we have the most up-to-date lock, a stale value for `last_vote` or `votes_forked_since_latest_strong_vote` may incorrectly force `decide_vote` to allow a strong vote when it should have only allowed a weak vote (a rule 2 violation). We could instead consider all votes found as evidence for updating the fsi (prior to first vote) while syncing. While recognizing that the exact order in which the votes were originally made may not be respected and some subset of votes may be skipped (either because not they were not found or because they were ignored from the monotonicity check due to processing out of order), it is still desirable to update the fsi according to the same logic that is normally followed. That means that if the monotonicity condition is satisfied, the `last_vote` (and possibly `votes_forked_since_latest_strong_vote`) should be updated even if the `lock` is not because the new lock from the vote does not have a better timestamp then the one already in the fsi. 
2. The condition https://github.com/AntelopeIO/spring/blob/7bd7153e4897e333ec1c31b5f62ec3b63c5b9d39/libraries/chain/finality/finalizer.cpp#L16-L18 in `decide_vote` should never be satisfied (and so the log should never be printed) because if the `last_vote` is empty, then the montonicity condition should necessarily be satisfied.
3. The code https://github.com/AntelopeIO/spring/blob/7bd7153e4897e333ec1c31b5f62ec3b63c5b9d39/libraries/chain/finality/finalizer.cpp#L39-L42 is not necessary because if `bsp->core.last_final_block_timestamp() >= fsi.lock.timestamp` then it is necessarily true that `bsp->core.latest_qc_block_timestamp() > fsi.lock.timestamp`.

greg7mdp:
Resolves #534 

1. consistently initialize `block_ref` with the `finality_digest`.
2. update voting rule 2 (determines whether to vote strong vs weak)) and update fsi and file format.
3. update tests.


heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -1416,12 +1416,14 @@ struct controller_impl {
          // information for those finalizers that don't already have one. This typically should be done when
          // we create the non-legacy fork_db, as from this point we may need to cast votes to participate
          // to the IF consensus. See https://github.com/AntelopeIO/leap/issues/2070#issuecomment-1941901836
-         auto start_block = chain_head; // doesn't matter this is not updated for IRREVERSIBLE, can be in irreversible mode and be a finalizer
-         auto lib_block   = chain_head;
+         block_ref ref = block_handle_accessor::apply<block_ref>(chain_head,
+            overloaded{[&](const block_state_legacy_ptr& head) { return block_ref{head->id(), head->timestamp(), {}}; },
```
arhag, line 1420: If we starting before the IF transition, why not use a default safety information that consists of the empty `last_vote` and empty `lock`? That would at least be more consistent then having a malformed block reference with non-empty `block_id` but empty `finality_digest`.

libraries/chain/finality/finalizer.cpp
```diff
@@ -242,15 +248,13 @@ my_finalizers_t::fsi_map my_finalizers_t::load_finalizer_safety_info() {
...
-      EOS_ASSERT(magic == fsi_t::magic_unversioned || magic == fsi_t::magic, finalizer_safety_exception,
+      EOS_ASSERT(magic == fsi_t::magic, finalizer_safety_exception,
                  "bad magic number in finalizer safety persistence file: ${p}", ("p", persist_file_path));
 
-      // If we expect the file to contain a version number, read it. We can load files with older
-      // versions, but it better not be a file with a version higher that the running nodeos understands.
-      // ------------------------------------------------------------------------------------------------
+      // We can load files with older, but not files with a version higher that the running nodeos understands.
+      // -----------------------------------------------------------------------------------------------------
       uint64_t file_version = 0; // default version for file with magic == fsi_t::magic_unversioned
```
arhag, line 256: Change this comment now that `magic_unversioned` has been removed.

unittests/finalizer_tests.cpp
```diff
@@ -281,7 +282,7 @@ BOOST_AUTO_TEST_CASE( finalizer_safety_file_versioning ) try {
    // --------------------------------------------------------------------
    fc::temp_directory tempdir;
 
-   for (size_t i=0; i<current_version; ++i) {
+   for (size_t i=1; i<current_version; ++i) {
```
arhag, line 285: Doesn't this now do nothing since `current_version` is 1?

arhag: Okay. So then should we start the `current_version` at 0 now and change the starting `i` back to 0?

libraries/chain/include/eosio/chain/finality/finality_core.hpp
```diff
@@ -10,6 +10,13 @@ using block_time_type = chain::block_timestamp_type;
 
 struct block_ref
 {
+   block_ref(const block_id_type& block_id, block_time_type timestamp, const digest_type& finality_digest)
+      : block_id(block_id)
+      , timestamp(timestamp)
+      , finality_digest(finality_digest) {}
+
+   block_ref() = default; // creates `empty` representation where `block_id.empty() == true`
```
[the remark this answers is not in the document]
greg7mdp: Tried it, but it causes `finalizer_vote_tests` to fail, most likely an issue with the test, so I'd rather not change this now.

libraries/chain/finality/finalizer.cpp
```diff
@@ -68,23 +64,33 @@ finalizer::vote_result finalizer::decide_vote(const block_state_ptr& bsp) {
...
+      if (voting_strong) {
+         fsi.votes_forked_since_latest_strong_vote = false;             // always reset to false on strong vote
+         if (latest_qc_claim__block_ref.timestamp > fsi.lock.timestamp)
+            fsi.lock = latest_qc_claim__block_ref;
+      } else {
+         // On a weak vote, if `votes_forked_since_latest_strong_vote` was already true, then it should remain true.
+         // The only way `votes_forked_since_latest_strong_vote` can change from false to true is on a weak vote
+         // for a block b where the last_vote references a block that is not an ancestor of b
+         // --------------------------------------------------------------------------------------------------------
+         fsi.votes_forked_since_latest_strong_vote |= !bsp->core.extends(fsi.last_vote.block_id);
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/chain/finality/finalizer.cpp
greg7mdp: done

unittests/finalizer_tests.cpp
greg7mdp: done.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -164,6 +164,8 @@ struct block_header_state {
    digest_type compute_base_digest() const;
    digest_type compute_finality_digest() const;
 
+   block_ref make_block_ref() const { return block_ref{block_id, timestamp(), compute_finality_digest() }; }
```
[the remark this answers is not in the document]
greg7mdp: Indeed, done.

libraries/chain/controller.cpp
greg7mdp: done

libraries/chain/finality/finalizer.cpp
heifner: Did you push it?

arhag: I think you changed another comment I didn't notice earlier in the code, but not yet the one on this line.

greg7mdp: oops, sorry, I just fixed the other one.

arhag (approved):

heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Improve Finality Safety Information, consistently initializing block_ref, and ensuring weak votes aren't masked.
Note:end

PR #538, opened 2024-08-14
Move verify_qc_claim onto the chain thread pool

heifner:
During syncing one `net` thread was at 100% CPU. Moving the `verify_qc_claim` onto the `chain-threads` thread pool gives about 10% CPU on a net thread and 50% on two `chain-threads`. Did not do detailed timing, but sync time improvement on my machine for 1000 block sync from ~ 3.5 seconds to ~3 seconds.


libraries/chain/controller.cpp
```diff
@@ -3961,7 +3964,7 @@ struct controller_impl {
 
       const bool skip_validate_signee = false;
       auto bsp = std::make_shared<BS>(
-            prev,
+            *prev,
             b,
```
arhag, line 3968: Prior to calling this constructor, we need to at least make sure that the QC claim in the block header of block `b` does not go backwards relative to the QC claim in the previous block (`prev`). 

This is needed to satisfy the precondition `this->latest_qc_claim() <= most_recent_ancestor_with_qc` assumed in `finality_core::next` and `finality_core::next_metadata` which is (indirectly) called from this constructor.

Previously that condition was checked in `verify_qc_claim`. If it was not satisfied, it threw an exception to reject the block and exited earlier before getting to this part of the code.

Now that it is posted to another thread and the result (throwing an exception or not) is only guaranteed by the point `verify_future.get()` is called below, we cannot deterministically guarantee that the precondition will be satisfied by the time this constructor is called.

At the very least we would need check the QC claim earlier in this function (outside of the future) to guarantee the precondition, and the remainder of the checks in `verify_qc_claim` can be done later as long as it is prior to calling `integrate_received_qc_to_block`.

arhag (changes requested):

ericpassmore:
Note:start
group: STABILITY
category: PERFORMANCE
summary: Move verify_qc_claim ont the chain thread pool.
Note:end

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Move `verify_qc_claim` onto the chain thread pool .
Note:end

heifner:
Replaced by https://github.com/AntelopeIO/spring/pull/736

PR #554, opened 2024-08-15
[1.0 -> main] Fix issue with last_pending_finalizer_policy_start_timestamp not being set/reflected

### LINKED ISSUE #549: Add restart production integration test - Fix rc1 issue
Start up a network with two nodes. The first node is a producer node, defproducera that has vote-threads enabled. The second node has a producer (defproducerb) and a single finalizer key configured. Use the bios contract to transition to Savanna consensus while keeping the existing producers and using a finalizer policy with the two finalizers.

Once everything has been confirmed to be working correctly and finality is advancing, cleanly shut down the producer defproducera node but keep the finalizer node of defproducerb running. 

Then change the finalizer policy (e.g. switch the order of the two finalizers) to get the nodes into a state where they have a pending finalizer policy. At that point restart the producer node defproducera (with stale production enabled so it produces blocks again).

The correct behavior is for votes from the finalizer node on the newly produced blocks to be accepted by producer node defproducera, QCs to be formed and included in new blocks, and finality to advance.

Due to the bug in pre-1.0.0-rc1, we expect that on restart the producer node will reject the votes received by the finalizer node because the producer node will be computing the wrong finality digest.

arhag:
Forwards #551 to main.

Resolves #549.


heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Fix edge case transitioning to finalizer policy on shutdown by setting the correct timestamp for pending finalizer policy change.
Note:end

PR #557, opened 2024-08-15
Remove unnecessary code that cannot be executed.

greg7mdp:
As described in issue https://github.com/AntelopeIO/spring/issues/534 (3rd `small things` at the end), we have a small test that cannot possibly be true, so this PR removes that dead code.


heifner (changes requested):
Should target `release/1.0`

greg7mdp:
> Should target `release/1.0`

why? It is not critical for sure. But I'm happy to change it.

linh2931 (approved):

greg7mdp:
Will open a new PR with correct target branch

PR #560, opened 2024-08-15
[1.0 -> main] Remove unnecessary code that cannot be executed.Gh 534 2 main

greg7mdp:
As described in issue https://github.com/AntelopeIO/spring/issues/534 (3rd `small things` at the end), we have a small test that cannot possibly be true, so this PR removes that dead code.


linh2931 (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Remove unnecessary code that cannot be executed.
Note:end

Issue #570, opened 2024-08-17
Add pause-at-block option


heifner:
Currently one use-case (possibly the largest use-case) of `terminate-at-block` is to provide the ability to stop a node at a given block height; remove any configured peers, disable incoming peers, and restart the node so that it can be queried at the specified block. 

In Spring, we process blocks into the fork database immediately. This means that on shutdown there can be many un-applied blocks in the fork database that would normally be applied on startup. For Spring 1.0.0, we have preserved the functionality (when configured with no configured peers (`p2p-peer-address`)), to not apply blocks on startup so that if the node was previously stopped with `terminate-at-block` chain head remains at that block on startup. See #565.

Add a new feature, `pause-at-block` that would allow a node operator to stop processing of blocks at a specified block height without shutting down. This feature could be used by itself to pause a node at a specified block or with terminate-at-block followed by pause-at-block to restart a node at a specific block without the node possibly processing any additional received blocks from the fork database or from the network. 

A new endpoint will be needed as a way to unpause, optionally with a new block pause height. Or we could just require a restart of the node after a pause.

PR #545, opened 2024-08-14
[1.0] Limit sync-fetch-span by max-reversible-blocks if needed

heifner:
Example output:
```
warn  2024-08-16T12:56:55.440 net-7     net_plugin.cpp:2021           active_sync_fetch_sp ] sync-fetch-span 5000 restricted to 3270 by max-reversible-blocks 3600, fork_db_size 329
info  2024-08-16T12:56:55.440 net-7     net_plugin.cpp:2179           operator()           ] ["jungle4.cryptolions.io:9870 - 34355da" - 2 95.216.44.76:9870] requesting range 1480342 to 1483611, fhead 1480341, lib 1480012
```

Resolves #528

PR #572, opened 2024-08-17
[1.0] Normally process blocks from the forkdb on startup

### LINKED ISSUE #565: P2P: Sync mode stuck after hitting max-reversible-blocks, after restart
After hitting `max-reversible-blocks` during a `read-mode = irreversible` sync, upon restart the node is stuck in `already syncing, start sync ignored`.

Hitting `max-reversible-blocks` should be avoided after #545. Need to investigate why node is stuck in `already syncing, start sync ignored` and also test #545 with the reported `config.ini`. 

`max-reversible-blocks` may not work correctly with `read-mode = irreversible`.

heifner:
In Spring 1.0.0, unlike Leap, we process blocks into the fork database immediately. This can cause the fork database to grow very large when syncing and shutdown due to hitting the new `max-reversible-blocks` limit (before #545). When a node is shutdown, in Leap, it was assumed that when restarting a node if that node did not receive any blocks from the network it would be at the same height as when it was shutdown. The existing `tests/nodeos_read_terminate_at_block_test.py` integration test verifies this behavior.

However, if the node shutdown because of `max-reversible-blocks` you would like on restart for the node to process blocks out of the fork database potentially allowing LIB to progress and reversible blocks to be consumed and shrink in size. This is at odds with the expected behavior of a node starting up after a `terminate-at-block`. A user might find it very odd to terminate a node at a block, but find on restart that the node is actually beyond that block.

To fix the issue reported in #565, we would like to process blocks out of the fork database on restart. However, we want to maintain the existing expectations of starting a node after a shutdown via `terminate-at-block` being at the same block height on restart. 

Therefore, this PR modifies nodeos to normally process blocks out of the fork database on startup unless the node has no configured peers. The idea being that if a user has terminated a node with `terminate-at-block` and wants the node to remain at that block, they will restart the node without any `p2p-peer-address` configured. This is a bit of a hack until #570 can be added. Since #570 is a new feature, it will not come until a future release and will not be part of Spring 1.0.0.

Resolves #565

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1969,14 +1969,34 @@ struct controller_impl {
       auto finish_init = [&](auto& forkdb) {
          if( read_mode != db_read_mode::IRREVERSIBLE ) {
             auto pending_head = forkdb.head();
-            if ( pending_head && pending_head->id() != chain_head.id() && chain_head.id() == forkdb.root()->id() ) {
-               ilog( "read_mode has changed from irreversible: applying best branch from fork database" );
-
-               for( ; pending_head->id() != chain_head.id(); pending_head = forkdb.head() ) {
-                  ilog( "applying branch from fork database ending with block: ${id}", ("id", pending_head->id()) );
-                  controller::block_report br;
-                  maybe_switch_forks( br, pending_head, controller::block_status::complete, {}, trx_meta_cache_lookup{} );
+            if ( pending_head && pending_head->id() != chain_head.id() ) {
+               // chain_head equal to root means that read_mode was changed from irreversible mode to head/speculative
+               bool chain_head_is_root = chain_head.id() == forkdb.root()->id();
+               if (chain_head_is_root) {
+                  ilog( "read_mode has changed from irreversible: applying best branch from fork database" );
                }
+
+               // See comment below about pause-at-block for why `|| conf.num_configured_p2p_peers > 0`
+               if (chain_head_is_root || conf.num_configured_p2p_peers > 0) {
+                  for( ; pending_head->id() != chain_head.id(); pending_head = forkdb.head() ) {
+                     ilog( "applying branch from fork database ending with block: ${id}", ("id", pending_head->id()) );
+                     controller::block_report br;
+                     maybe_switch_forks( br, pending_head, controller::block_status::complete, {}, trx_meta_cache_lookup{} );
+                  }
+               }
+            }
+         } else {
+            // It is possible that the node was shutdown with blocks to process in the fork database. For example, if
+            // it was syncing and had processed blocks into the fork database but not yet applied them. In general,
+            // it makes sense to process those blocks on startup. However, if the node was shutdown via
+            // terminate-at-block, the current expectation is that the node can be restarted to examine the state at
+            // which it was shutdown. For now, we will only process these blocks if there are peers configured. This
+            // is a bit of a hack for Spring 1.0.0 until we can add a proper pause-at-block (issue #570) which could
+            // be used to explicitly request a node to not process beyond a specified block.
+            if (conf.num_configured_p2p_peers > 0) {
+               ilog("Process blocks out of forkdb if needed");
+               log_irreversible();
+               transition_to_savanna_if_needed();
             }
          }
       };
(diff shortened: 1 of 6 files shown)


libraries/chain/controller.cpp
```diff
@@ -1969,15 +1969,35 @@ struct controller_impl {
...
                }
             }
+         } else {
+            // It is possible that the node was shutdown with blocks to process in the fork database. For example, if
+            // it was syncing and had processed blocks into the fork database but not yet applied them. In general,
+            // it makes sense to process those blocks on startup. However, if the node was shutdown via
+            // terminate-at-block, the current expectation is that the node can be restarted to examine the state at
+            // which it was shutdown. For now, we will only process these blocks if there are peers configured. This
+            // is a bit of a hack for Spring 1.0.0 until we can add a proper pause-at-block (issue #570) which could
+            // be used to explicitly request a node to not process beyond a specified block.
```
heifner, line 1995: This compromise came after a long conversation with @arhag on potential alternatives.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2224,6 +2224,8 @@ namespace eosio {
          set_state( lib_catchup );
          sync_last_requested_num = 0;
          sync_next_expected_num = chain_info.lib_num + 1;
+      } else if (sync_next_expected_num >= sync_last_requested_num) {
+         // break
```
heifner, line 2228: This change causes the net_plugin to request more blocks if possible instead of remaining in a mode where it thinks it is already syncing.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -345,6 +346,18 @@ def executeSnapshotBlocklogTest(cluster, testNodeId, resultMsgs, nodeArgs, termA
...
+    if not cluster.biosNode.relaunch():
+        Utils.errorExit("Unable to restart bios node")
+
+    if not producingNode.relaunch():
+        Utils.errorExit("Unable to restart producing node")
+
+    if success:
+        for nodeId, nodeArgs in {**regularNodeosArgs, **replayNodeosArgs}.items():
+            assert cluster.getNode(nodeId).relaunch(), f"Unable to relaunch {nodeId}"
+            assert cluster.getNode(nodeId).waitForLibToAdvance(), f"LIB did not advance for {nodeId}"
```
heifner, line 359: This extra section on restarting nodes failed before the fixes in this PR.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -266,10 +267,10 @@ def executeSnapshotBlocklogTest(cluster, testNodeId, resultMsgs, nodeArgs, termA
...
     }
     regularNodeosArgs = {
-        1 : "--read-mode irreversible --terminate-at-block 75",
-        2 : "--read-mode head --terminate-at-block 100",
-        3 : "--read-mode speculative --terminate-at-block 125",
-        4 : "--read-mode irreversible --terminate-at-block 155"
+        1 : "--read-mode irreversible --terminate-at-block 100",
+        2 : "--read-mode head --terminate-at-block 125",
+        3 : "--read-mode speculative --terminate-at-block 150",
+        4 : "--read-mode irreversible --terminate-at-block 180"
```
heifner, line 273: These block values don't really matter, I could revert to old values if desired. They were changed when I was doing some initial testing to try and reproduce the issue in this test. The changes below are what is needed.

greg7mdp (approved):
> this PR modifies nodeos to normally process blocks out of the fork database on startup unless the node has no configured peers

Only in irreversible mode, right?

heifner:
> Only in irreversible mode, right?

Also non-irreversible mode:

https://github.com/AntelopeIO/spring/pull/572/files#diff-42e9f97eed543dd784b1b77a536c50ff8e2403d5bbd3b1fb7f5dd90201391061R1988

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Until pause at block height is available this fix enables use of forkdb to advance LIB on startup when connected to peers, but does not read from forkdb on startup when no peers are connected.
Note:end

Issue #552, opened 2024-08-15
Improve restart production integration test


arhag:
The integration test introduced in https://github.com/AntelopeIO/spring/issues/549 has a minor problem in that the test may not stop the node in the middle of a pending policy. If it misses it and the pending policy becomes active, then the test will still pass but it won't test the behavior the test was intended for.

The test can be improved as follows:

Instead of changing the order of the finalizer keys in order to force a finalizer policy change, change the finalizer policy to replace one of the keys (the one on the node to be shutdown). Then if the node to be shutdown does not have the new replacement key, we can guarantee that the pending policy will not become active. That guarantees that we can shut down the node in a state where it has a pending policy.

Then the shutdown node can be restarted with the new finalizer key also configured. This allows finality to advance, at least in the situation where the original bug was fixed.

PR #577, opened 2024-08-18
[1.0] Improve restart production integration test

linh2931:
The existing restart production  test may not stop the node in the middle of a pending policy. If it misses it and the pending policy becomes active, then the test will still pass but it won't test the behavior the test was intended for.

Instead of changing the order of the finalizer keys in order to force a finalizer policy change, the test is changed to use an unconfigured key in the new finalizer policy such that it is guaranteed that the node is in a state where pending policy exists before the node is shut down.

Resolves https://github.com/AntelopeIO/spring/issues/552


tests/production_restart.py
```diff
@@ -64,28 +64,54 @@
...
+       # in the configured key list (index 0)
+       newBlsPubKey = producerNode.keys[1].blspubkey
+       newBlsPrivKey = producerNode.keys[1].blsprivkey
+       newBlsPop = producerNode.keys[1].blspop
+       producerNode.keys[1].blspubkey = producerNode.keys[0].blspubkey
+       producerNode.keys[1].blsprivkey = producerNode.keys[0].blsprivkey
+       producerNode.keys[1].blspop = producerNode.keys[0].blspop
+       producerNode.keys[0].blspubkey = producerNode.keys[1].blspubkey
+       producerNode.keys[0].blsprivkey = producerNode.keys[1].blsprivkey
+       producerNode.keys[0].blspop = producerNode.keys[1].blspop
```
greg7mdp, line 88: This doesn't really switch keys, right? I think you meant to use `newBls...`.

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve finalizer policy changes in test by changing keys multiple times. Previously changed the order of keys.
Note:end

tests/production_restart.py
linh2931: My bad. Thanks!

tests/production_restart.py
```diff
@@ -64,28 +64,54 @@
...
+       # Check if a pending policy exists
+       finalizerInfo = producerNode.getFinalizerInfo()
+       Print(f"{finalizerInfo}")
+       if finalizerInfo["payload"]["pending_finalizer_policy"] is not None and finalizerInfo["payload"]["pending_finalizer_policy"]["finalizers"][0]["public_key"] == newBlsPubKey:
+          Print(f"Got a pending policy in {i+1} attempts")
+          break
+       else:
+          Print(f"Trying to get a pending policy, {i+1} attempts")
+
+    # It is OK if pending policy does not exist. During manual tests,
```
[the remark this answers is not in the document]
linh2931: Thanks. I missed that. I am going to make the changes.

heifner (approved):

tests/production_restart.py
```diff
@@ -18,9 +18,10 @@
 # Once everything has been confirmed to be working correctly and finality is advancing, cleanly shut down the producer
 # defproducera node but keep the finalizer node of defproducerb running.
 #
-# Then change the finalizer policy (e.g. switch the order of the two finalizers) to get the nodes into a state where
-# they have a pending finalizer policy. At that point restart the producer node defproducera (with stale production
-# enabled so it produces blocks again).
+# Then change the finalizer policy using an unconfigured key in node defproducera to gurantee
```
arhag, line 21: Typo: `gurantee`.

Also, the description above (while from the original issue) is confusing because both nodes have a producer and finalizer each. The description makes it seem like the first node is only a producer node.

tests/production_restart.py
```diff
@@ -64,28 +65,46 @@
     cluster.biosNode.kill(signal.SIGTERM)
     cluster.waitOnClusterSync(blockAdvancing=5)
 
-    node0 = cluster.getNode(0) # producer
-    node1 = cluster.getNode(1) # finalizer
+    producerNode = cluster.getNode(0)
+    finalizerNode = cluster.getNode(1)
```
arhag, line 69: These names (and in fact the original comments) are misleading because each node has its own producer and finalizer configured.

tests/production_restart.py
linh2931: Thanks.

tests/production_restart.py
linh2931: Renamed.

arhag (approved):

PR #587, opened 2024-08-19
[1.0] Update max-reversible-blocks to only enforce savanna blocks

### LINKED ISSUE #579: max-reversible-blocks syncing old blocks
Matthew Darwin
> Trying to sync kylin from genesis fails with Exceeded max reversible blocks allowed, fork db size 3634 >= max-reversible-blocks 3600.  There is a period at the start of the chain where LIB fails to move forward.   
> 
> Should max-reversible-blocks be applied on p2p sync ancient blocks?

> LIB drifts by at least 65K blocks.

heifner:
Update the `max-reversible-blocks` option to only be enforced for blocks after the Savanna transition.

Resolves #579

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1117,6 +1117,13 @@ struct controller_impl {
       return prev->block_num();
    }
 
+   // returns 0 for legacy
+   size_t fork_db_savanna_size() const {
+      return fork_db.apply_s<size_t>([&](const auto& forkdb) {
+         return forkdb.size();
+      });
+   }
+
    bool fork_db_block_exists( const block_id_type& id ) const {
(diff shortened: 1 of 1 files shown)


arhag (approved):

libraries/chain/controller.cpp
```diff
@@ -1117,6 +1117,13 @@ struct controller_impl {
       return prev->block_num();
    }
 
+   // returns 0 for legacy
```
greg7mdp, line 1120: I wonder if this is guaranteed by the standard?

greg7mdp (approved):

libraries/chain/controller.cpp
heifner: Yes, https://en.cppreference.com/w/cpp/language/zero_initialization

PR #597, opened 2024-08-20
[1.0 -> main] Replay from retained directory with no block log

### LINKED ISSUE #586: Loading from snapshot with block strides
Nodes with `blocks-log-stride` option can be in a state where a snapshot does not align with the blocks in the blocks directory. Currently, blocks in the `blocks-retained-dir` are not considered for replay on startup if the block log head does not exist. Even though the node has the blocks needed to start from the snapshot (in the retained directory), it is currently unable to do so in that one edge case.

Fix nodeos to replay from blocks in the `blocks-retained-dir` on startup.

heifner:
PR fixes two issues:
- Replaying from a snapshot with no block log, but blocks in the retained directory, issue #586
  - Before this PR, block log head would be nullptr and no blocks would be replayed from the retained directory.
- Interrupting replay of block log (from blocks or retained directory) via ctrl-c, restarting replay when no fork database.
  - Before this PR, would fail by throwing an error that no fork database exists.

Merges `release/1.0` into `main` including #565 and #595 (forward backport back to `main`)

Resolves #586

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -956,7 +956,7 @@ namespace eosio { namespace chain {
             open(log_dir);
             const auto log_size = std::filesystem::file_size(block_file.get_file_path());
 
-            if (log_size == 0 && !catalog.empty()) {
+            if ((log_size == 0 || !head) && !catalog.empty()) {
                basic_block_log::reset(catalog.verifier.chain_id, catalog.last_block_num() + 1);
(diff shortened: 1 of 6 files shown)


greg7mdp (approved):

arhag (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Replay from Blocks in the retained directory.
Note:end

PR #603, opened 2024-08-20
[5.0 -> 1.0] limit the vector size that can be reserved

greg7mdp:
1.0 merge of AntelopeIO/leap#2405 

> The deserialization of arrays involves reading a size, and then the array elements which are `emplace`'d back into a `std::vector`. For normal vectors, since we know the size in advance, it is beneficial to call `std::vector::reserve()` to prevent quadratic reallocations of the vector's buffer as the values are read. Still, limit the size that can be reserved to `1024` `fc::variant` values, or 16KB since `sizeof(fc::variant) == 16`).


arhag (approved):

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Limit vector size that can be reserved.
Note:end

PR #607, opened 2024-08-20
[1.0] Add unittest verifying that proposer policy does not become active when finality has stalled (and a couple other)

### LINKED ISSUE #553: Test proposer policy does not become active when finality has stalled
Add an additional libtester test (though using the Savanna cluster testing utility) to test new behavior of https://github.com/AntelopeIO/spring/issues/454.

Create a test where we partition the network (two nodes on one side and two nodes on the other side). Propose a new proposer policy in this scenario and notice that it does not become active. Then reconnect the network and finality should start advancing again and the proposer policy should automatically be promoted to become active.

greg7mdp:
Resolves #553.

Also added tests verifying that:
1. when finality advances, proposer policy transitions happens on the expected block.
2. if a proposer policy is already pending, it will become active even if finality stalls.


unittests/savanna_proposer_policy_tests.cpp
```diff
@@ -0,0 +1,159 @@
...
+   auto end_num = A.head().block_num();
+
+   // under Savanna, a new policy becomes active on the first block of a 12 block round after:
+   // 1. finishing the current round
+   // 2. a full round
+   // ----------------------------------------------------------------------------------------
+   uint32_t expected_gap = 12 - (start_num % 12);   // 1. finishing the current round
+   expected_gap += 12;                              // 2. a full round
+   BOOST_REQUIRE_EQUAL(end_num, start_num + expected_gap);
+} FC_LOG_AND_RETHROW()
```
arhag, line 89: Where a block is within respect to a round (first, middle, or last block of the round) is determined by the slot number (timestamp) not the block number.

Additionally, even if you changes `head().block_num()` everywhere above to instead grab the slot of the timestamp of that head block, I'm still not sure if there isn't an off-by-one error in the three test cases above. I say this because you can check you get the desired remainder when dividing the slot number by 12, but the transaction that triggers the policy change goes into the next block. How is the first block of a round defined? Is it the one that has a slot which is a multiple of 12?

And you wouldn't really notice the off-by-one-error because what you may believe is on the last block of round N may be on the first block of round N+1. The `BOOST_REQUIRE_EQUAL` is designed to be robust enough that the test would still work fine for the same reasons the tests above currently pass even though you are incorrectly using the block number. But the comments would not be (and currently are not) accurate.

unittests/savanna_proposer_policy_tests.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+   A.create_accounts(producers);
+   A.tester::set_producers(producers);    // push the action to update the producer schedule
+
+   // produce 24 more blocks. If finality was advancing, the new proposer policy would be active,
+   // but with a split network finality will stall, and the new proposer policy should *not* become active.
+   // -----------------------------------------------------------------------------------------------------
+   A.produce_blocks(24);
+   BOOST_REQUIRE_EQUAL(A.lib_number, orig_lib_num);
+   BOOST_REQUIRE_EQUAL(A.control->active_producers().version, orig_version);
```
arhag, line 118: I think you should continue this test by reestablishing the connection between all the nodes and show that after finality advances again, the new producer schedule becomes active at the expected time.

unittests/savanna_proposer_policy_tests.cpp
```diff
@@ -0,0 +1,159 @@
...
+//     Verify that a proposer policy becomes active when finality has advanced enough to make it pending
+// ---------------------------------------------------------------------------------------------------
+BOOST_FIXTURE_TEST_CASE(pending_proposer_policy_becomes_active_without_finality, savanna_cluster::cluster_t) try {
+   auto& A=_nodes[0];
+
+   const vector<account_name> producers { "p0"_n, "p1"_n };
+   auto orig_version = A.control->active_producers().version;
+
+   while (A.head().block_num() % 12 >= 10)
+      A.produce_block();                  // make sure we are not on the last two blocks of a round
```
arhag, line 131: Again, this needs to be corrected to use the slot instead of the block number. 

And I think there may be an off-by-one error again, but it seems like it would only matter in terms of not matching the comment. It does not appear like it would harm the actual test.

For example, if the head block's time slot, `slot1`, happens to satisfy `slot1 % 12 == 9`, then no additional block would be produced. But that means the next pending block after that head block, the one that includes the transaction that will propose a new proposer policy, would have a slot, `slot2`, which satisfies `slot2 % 12 == 10`. If I'm not mistaken, that would mean it would be the second to last block of the round, which is something the comment says it wants to prevent. But anyway, after producing 13 blocks, the new pending block would have a slot, `slot3`, which satisfies `slot3 % 12 == 11` but is on the next round. So it wouldn't yet have the new proposer policy as active. That would happen on the very next block in this case. But because it is a new round, the new proposer policy would be pending.

It would also unnecessarily prevent the proposer policy from being proposed on the first block of a round, but that doesn't really harm anything either.

arhag (changes requested):

unittests/savanna_proposer_policy_tests.cpp
```diff
@@ -0,0 +1,279 @@
...
+   sb = A.produce_block();                // produce one more block for lib final advance (in-flight QC)
+
+   BOOST_REQUIRE_EQUAL(sb->producer, orig_producer);
+   BOOST_REQUIRE_EQUAL(A.control->active_producers().version, orig_version);
+   BOOST_REQUIRE(!!A.control->pending_producers());
+   BOOST_REQUIRE_GT(A.control->pending_producers()->version, orig_version);
+   auto orig_lib_num = A.lib_number;
+
+   // produce `prod_rep` more blocks. Finality will not be advancing anymore, but still the new policy
+   // will become active because it was already pending.
```
arhag, line 268: This is true, but it is also true that the policy will become active as long as it was simply _proposed_ prior to the last final block when finality stalled.

arhag (approved):

unittests/savanna_proposer_policy_tests.cpp
greg7mdp: @arhag I addressed your latest comments. Thanks for the thorough review!

unittests/savanna_proposer_policy_tests.cpp
```diff
@@ -234,6 +234,7 @@ BOOST_FIXTURE_TEST_CASE(no_proposer_policy_change_without_finality_2, savanna_cl
 // ---------------------------------------------------------------------------------------------------
 BOOST_FIXTURE_TEST_CASE(pending_proposer_policy_becomes_active_without_finality, savanna_cluster::cluster_t) try {
    auto& A=_nodes[0];
+   static_assert(prod_rep >= 4);
```
arhag, line 237: This should be `static_assert(prod_rep > 4);` since below you have:
```cpp
while (A.head().timestamp().slot % prod_rep >= prod_rep - 4)
   A.produce_block();
```

Note that if `prod_rep == 4`, then it would get stuck in an infinite loop.

Better yet, you can keep the `static_assert` as is and instead fix the condition in the `while` loop so that it reflects the comment properly. (See first sentence in https://github.com/AntelopeIO/spring/pull/607#discussion_r1725796947.)

For example:
```cpp
while ((A.head().timestamp().slot + 1) % prod_rep >= prod_rep - 3)
   A.produce_block();
```

greg7mdp: Thanks again, done!

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add test to validate proposed finality policy does not become active when chain finality is not progressing.
Note:end

heifner (approved):
Very clean!

PR #609, opened 2024-08-21
[1.0] Add max_reversible_blocks_allowed(), use in net_plugin to limit sync-fetch-span

### LINKED ISSUE #608: max-reversible-blocks exceeded while syncing legacy blocks
Syncing to kylin.
```
max-reversible-blocks 3600 exceeded, remaining -1010, fork_db_size 4609
```

heifner:
Instead of `net_plugin` calculating how many reversible blocks are allowed to limit `sync-fetch-span`, call a method on `controller`. Now logic about which fork database to check is encapsulated to the controller.

Resolves #608

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4840,6 +4840,12 @@ struct controller_impl {
       return conf.block_validation_mode == validation_mode::LIGHT || conf.trusted_producers.count(producer);
    }
 
+   int32_t max_reversible_blocks_allowed() const {
+      if (conf.max_reversible_blocks == 0)
+         return std::numeric_limits<int32_t>::max();
+      return conf.max_reversible_blocks - fork_db_savanna_size();
+   }
+
    bool should_terminate(block_num_type reversible_block_num) const {
       assert(reversible_block_num > 0);
       if (conf.terminate_at_block > 0 && conf.terminate_at_block <= reversible_block_num) {
@@ -4847,7 +4853,7 @@ struct controller_impl {
               ("n", reversible_block_num)("num", conf.terminate_at_block) );
          return true;
       }
-      if (conf.max_reversible_blocks > 0 && fork_db_savanna_size() >= conf.max_reversible_blocks) {
+      if (max_reversible_blocks_allowed() <= 0) {
          elog("Exceeded max reversible blocks allowed, fork db size ${s} >= max-reversible-blocks ${m}",
               ("s", fork_db_savanna_size())("m", conf.max_reversible_blocks));
          return true;
@@ -5658,6 +5664,10 @@ validation_mode controller::get_validation_mode()const {
 
 bool controller::should_terminate() const {
    return my->should_terminate();
+}
+
+int32_t controller:: max_reversible_blocks_allowed() const {
+   return my->max_reversible_blocks_allowed();
 }
 
 const apply_handler* controller::find_apply_handler( account_name receiver, account_name scope, action_name act ) const
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1117,13 +1117,6 @@ struct controller_impl {
       return prev->block_num();
    }
 
-   // returns 0 for legacy
-   size_t fork_db_savanna_size() const {
-      return fork_db.apply_s<size_t>([&](const auto& forkdb) {
-         return forkdb.size();
-      });
-   }
-
    bool fork_db_block_exists( const block_id_type& id ) const {
       return fork_db.apply<bool>([&](const auto& forkdb) {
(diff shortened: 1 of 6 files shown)


libraries/chain/controller.cpp
```diff
@@ -4840,14 +4840,20 @@ struct controller_impl {
       return conf.block_validation_mode == validation_mode::LIGHT || conf.trusted_producers.count(producer);
    }
 
+   int32_t max_reversible_blocks_allowed() const {
+      if (conf.max_reversible_blocks == 0)
+         return std::numeric_limits<int32_t>::max();
+      return conf.max_reversible_blocks - fork_db_savanna_size();
```
greg7mdp, line 4846: shouldn't it also return std::numeric_limits<int32_t>::max() when we have a legacy fork_db instead of max_reversible_blocks?

Otherwise in net_plugin you could have `(reversible_remaining < sync_fetch_span)`. I guess maybe it is not a big deal,  it just may limit `sync_fetch_span` which could seem strange.

heifner: Maybe. As is, if `max-reversible-blocks` is default 3600 it limits `sync-fetch-span` to 3600. I kind of thought that a good thing. But would be ok with changing it to return `MAX_INT32` if legacy fork db. @arhag what do you think?

greg7mdp: Personally, I think `std::numeric_limits<int32_t>::max()` if more bulletproof, as you don't have to check how the return value is used in net_plugin to know it won't add a limitation.

greg7mdp (approved):

libraries/chain/controller.cpp
arhag: I do think it should return the max value from this function if only dealing with the legacy fork db. It is also consistent with your existing comment describing `max_reversible_blocks_allowed` in `controller.hpp`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1989,30 +1988,30 @@ namespace eosio {
...
    uint32_t sync_manager::active_sync_fetch_span() const {
-      auto fork_db_size = my_impl->chain_plug->chain().fork_db_size();
-      int32_t reversible_remaining = max_reversible_blocks - fork_db_size - 1;
+      int32_t reversible_remaining = my_impl->chain_plug->chain().max_reversible_blocks_allowed();
       if (reversible_remaining <= 0) {
-         fc_wlog(logger, "max-reversible-blocks ${m} exceeded, remaining ${r}, fork_db_size ${fs}",
-                 ("m", max_reversible_blocks)("r", reversible_remaining)("fs", fork_db_size));
+         auto fork_db_size = my_impl->chain_plug->chain().fork_db_size();
+         fc_wlog(logger, "max-reversible-blocks exceeded, remaining ${r}, fork_db_size ${fs}",
+                 ("r", reversible_remaining)("fs", fork_db_size));
```
arhag, line 2008: We could still print a negative remaining here. Is that intended?

Alternatively, we could check if `reversible_remaining < 0` and print the exceeded number of blocks as a positive number.

For example:

```cpp
fc_wlog(logger, "max-reversible-blocks exceeded by ${ex}, fork_db_size ${fs}",
                 ("ex", -reversible_remaining)("fs", fork_db_size));
```

libraries/chain/controller.cpp
```diff
@@ -4840,16 +4833,29 @@ struct controller_impl {
...
+   int32_t max_reversible_blocks_allowed() const {
+      if (conf.max_reversible_blocks == 0)
+         return std::numeric_limits<int32_t>::max();
+
+      return fork_db.apply<int32_t>(
+         [&](const fork_database_legacy_t& forkdb) {
+            return std::numeric_limits<int32_t>::max();
+         },
+         [&](const fork_database_if_t& forkdb) {
+            return conf.max_reversible_blocks - forkdb.size();
```
arhag, line 4845: `conf.max_reversible_blocks` is a `uint32_t`.

What if a user provides a value greater than 2^31? The difference may still be greater than 2^31 and not be representable as an `int32_t`.

plugins/net_plugin/net_plugin.cpp
heifner: I thought negative remaining would just indicate the node exceeded the threshold. For a warning like this I figured that would be appropriate. I can change to print out positive.

arhag (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Improve where calculation of max_reversible_blocks_allowed is performed.
Note:end

PR #616, opened 2024-08-22
[1.0 -> main] Add unittest verifying that proposer policy does not become active when finality has stalled (and a couple other)

### LINKED ISSUE #553: Test proposer policy does not become active when finality has stalled

greg7mdp:
Resolves #553.

Also added tests verifying that:
1. when finality advances, proposer policy transitions happens on the expected block.
2. if a proposer policy is already pending, it will become active even if finality stalls.

### DIFF
--- a/unittests/savanna_cluster.hpp
+++ b/unittests/savanna_cluster.hpp
@@ -113,12 +113,10 @@ namespace savanna_cluster {
          BOOST_REQUIRE_EQUAL(lib_num(), pt_block->block_num());
       }
 
-      // updates producers (producer updates will be propagated to connected nodes), and
-      // wait until one of the new producers is pending.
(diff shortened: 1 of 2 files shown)


ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add test to validate proposed finality policy does not become active when chain finality is not progressing.
Note:end

ericpassmore (approved):

PR #619, opened 2024-08-22
Optimize block processing - Part 1

### LINKED ISSUE #529: Optimize processing of incoming blocks

### LINKED ISSUE #821: Test failure: p2p_sync_throttle_test
https://github.com/AntelopeIO/spring/actions/runs/11021862404/job/30610476067

```
2024-09-24T21:20:30.228163      File "/__w/spring/spring/build/tests/p2p_sync_throttle_test.py", line 135, in <module>
    assert throttledNode.waitForBlock(endLargeBlocksHeadBlock, timeout=240), f'Wait for block {endLargeBlocksHeadBlock} on throttled node timed out'
```

heifner:
Change set already large, going to finish the rest of the work for #529 in separate PRs. See #688 

- Renamed `controller::push_block` and `controller::maybe_switch_forks` to `controller::apply_blocks` as they now do the same thing.
- Renamed `create_block_handle` to `accept_block`. `create_block_handle` does more than just create a `block_handle`. It is often paired with `apply_blocks`, so I think `accept_block(b)` .. `apply_blocks()` is better. Removed `create_block_handle_future` as it is no longer needed.
- Removed `net_plugin` `unlinkable_block_state_cache`. I don't think unlinkable blocks should happen normally anymore (especially after #634 is fixed). This was only an optimization/work-around for the unknown possible unlinkable block posted to the main thread. Since unlinkable blocks are no longer posted to the main thread, this is no longer needed.
- Added work-around in `net_plugin` for not being able to add proper savanna blocks to fork database until savanna transition is complete.
- Modified `fork_database::add` to return a boolean indicating if the added block became the new best head.
- Refactored handling of `block_report`. It no longer is passed around, but instead is tracked and reported from `pending_block`.
- Moved `Produced block` log into `log_applied` along with related `Received block` log message.
- Renamed `replay_push_block` to `replay_irreversible_block` and simplified as it is only used for replaying irreversible blocks now.
- Removed info log `Block not applied to head`. All blocks are processed into the fork database immediately and only the best head at any given time is applied. We report fork switching which is a better indication of what this originally was trying to communicate. 

Partially resolves #529
Resolves #821


ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Optimized Block Processing simplifying controller, now possible due to Spring upgrades.
Note:end

libraries/chain/controller.cpp
```diff
@@ -4267,81 +4260,13 @@ struct controller_impl {
...
-   template <class BSP>
-   void replay_push_block( const signed_block_ptr& b, controller::block_status s ) {
+   void replay_irreversible_block( const signed_block_ptr& b ) {
       validate_db_available_size();
 
       EOS_ASSERT(!pending, block_validate_exception, "it is not valid to push a block when there is a pending block");
 
       try {
          EOS_ASSERT( b, block_validate_exception, "trying to push empty block" );
-         EOS_ASSERT( (s == controller::block_status::irreversible || s == controller::block_status::validated),
```
linh2931, line 4343: Should this be replaced with a small assert?

libraries/chain/controller.cpp
```diff
@@ -4267,81 +4260,13 @@ struct controller_impl {
...
-
-   template <class BSP>
-   void replay_push_block( const signed_block_ptr& b, controller::block_status s ) {
+   void replay_irreversible_block( const signed_block_ptr& b ) {
       validate_db_available_size();
 
       EOS_ASSERT(!pending, block_validate_exception, "it is not valid to push a block when there is a pending block");
 
       try {
          EOS_ASSERT( b, block_validate_exception, "trying to push empty block" );
```
linh2931, line 4269: `push` --> `replay`

libraries/chain/controller.cpp
```diff
@@ -4393,145 +4302,126 @@ struct controller_impl {
...
-            auto pending_head = forkdb.head(include_root_t::yes);
-            if (chain_head.id() != pending_head->id()) {
-               dlog("switching forks on controller->maybe_switch_forks call");
-               controller::block_report br;
-               maybe_switch_forks(br, pending_head,
-                                  pending_head->is_valid() ? controller::block_status::validated
-                                                           : controller::block_status::complete,
-                                  cb, trx_lookup);
-            }
+   void apply_blocks(const forked_callback_t& cb, const trx_meta_cache_lookup& trx_lookup) {
```
linh2931, line 4305: Like this new name. The old `maybe_switch_forks` was confusing.

libraries/chain/controller.cpp
```diff
@@ -2838,7 +2836,6 @@ struct controller_impl {
          pending->_block_report.total_net_usage += trace->net_usage;
          pending->_block_report.total_cpu_usage_us += trace->receipt->cpu_usage_us;
          pending->_block_report.total_elapsed_time += trace->elapsed;
-         pending->_block_report.total_time += fc::time_point::now() - start;
```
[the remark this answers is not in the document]
heifner: Because it is now just calculated at time of reporting instead of building it up over time.

libraries/chain/controller.cpp
```diff
@@ -3484,6 +3464,69 @@ struct controller_impl {
...
+            metrics.total_time_us = (now - br.start_time).count();
+            metrics.net_usage_us = br.total_net_usage;
+            metrics.last_irreversible = chain_head.irreversible_blocknum();
+            metrics.head_block_num = chain_head.block_num();
+            _update_produced_block_metrics(metrics);
+         }
+         return;
+      }
+
+      ilog("Received block ${id}... #${n} @ ${t} signed by ${p} " // "Received" instead of "Applied" so it matches existing log output
```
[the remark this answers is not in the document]
heifner: It might break tools but more keeping it `Received/Produced` as it is the expected output people expect.

The values of `block_status` is only needed to distinguish between `Received` and `Produced`.

libraries/chain/controller.cpp
```diff
@@ -3658,20 +3676,29 @@ struct controller_impl {
...
-            auto start = fc::time_point::now();
+            auto start = fc::time_point::now(); // want to report total time of applying a block
+
+            const signed_block_ptr& b = bsp->block;
+            fc::scoped_set_value prod_light_validation(trusted_producer_light_validation, is_trusted_producer(b->producer));
 
             const bool already_valid = bsp->is_valid();
-            // Only need to consider voting if not already validated, if already validated then we have already voted.
-            if (!already_valid)
+            if (!already_valid || replaying) {
```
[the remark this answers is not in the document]
heifner: If it is `valid` then that means it has already been emitted, unless we are replaying. This maintains the existing behavior of replaying validated blocks emitting `accepted_block_header`.

libraries/chain/controller.cpp
```diff
@@ -4393,145 +4302,126 @@ struct controller_impl {
...
-                  throw;
-               } catch ( const boost::interprocess::bad_alloc& ) {
-                  throw;
-               } catch (const fc::exception& e) {
-                  elog("exception thrown while switching forks ${e}", ("e", e.to_detail_string()));
-                  except = std::current_exception();
-               } catch (const std::exception& e) {
-                  elog("exception thrown while switching forks ${e}", ("e", e.what()));
-                  except = std::current_exception();
+         for( auto ritr = branches.first.rbegin(); ritr != branches.first.rend(); ++ritr ) {
```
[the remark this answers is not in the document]
heifner: Yes. It is checked in `apply_block`.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -182,41 +222,20 @@ namespace eosio::chain {
...
-          */
-         void push_block( block_report& br,
-                          const block_handle& b,
-                          const forked_callback_t& cb,
-                          const trx_meta_cache_lookup& trx_lookup );
-
-         /// Accept block into fork_database
-         void accept_block(const block_handle& b);
+         // thread-safe
+         // returns true if new best head and empty optional if block b is unlinkable
```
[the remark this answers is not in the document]
heifner: What do you think of this:
```cpp
struct accepted_block_handle {
   // true if the new block becomes the new best head of the forkdb
   bool is_best_head = false;
   // block state handle wrapping signed block, empty optional if unlinkable block
   std::optional<block_handle> block; 
};
```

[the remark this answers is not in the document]
linh2931: This is much clearer.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: Maybe add a comment.

libraries/chain/include/eosio/chain/controller.hpp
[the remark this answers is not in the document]
greg7mdp: I don't think the default value for the `bool` is useful.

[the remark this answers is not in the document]
heifner: Added

libraries/chain/controller.cpp
heifner: There is no `s` to assert anymore.

libraries/chain/controller.cpp
heifner: Thanks. Done.

linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -4394,145 +4293,126 @@ struct controller_impl {
...
-               apply_block( br, new_head, s, trx_lookup );
-            } catch ( const std::exception& e ) {
-               dlog("Exception during apply block: ${e}", ("e", e.what()));
-               forkdb.remove( new_head->id() );
-               throw;
+      auto do_apply_blocks = [&](auto& forkdb) {
+         auto new_head = forkdb.head(); // use best head
+         if (!new_head)
+            return; // nothing to do, forkdb at root
+         auto branches = forkdb.fetch_branch_from( new_head->id(), chain_head.id() );
```
greg7mdp, line 4313: I think this might be clearer:

```suggestion
         auto [new_head_branch, old_head_branch] = forkdb.fetch_branch_from( new_head->id(), chain_head.id() );
```

libraries/chain/controller.cpp
```diff
@@ -987,6 +994,10 @@ struct controller_impl {
    vote_signal_t                             voted_block;     // emitted when a local finalizer votes on a block
    vote_signal_t                             aggregated_vote; // emitted when a vote received from the network is aggregated
 
+   std::function<void(produced_block_metrics)> _update_produced_block_metrics;
+   std::function<void(speculative_block_metrics)> _update_speculative_block_metrics;
+   std::function<void(incoming_block_metrics)> _update_incoming_block_metrics;
```
greg7mdp, line 999: Any reason why these are not signals?

libraries/chain/controller.cpp
```diff
@@ -4143,65 +4157,34 @@ struct controller_impl {
...
       
-      auto f = [&](auto& forkdb) -> std::optional<block_handle> {
-         // previous not found could mean that previous block not applied yet
+      auto f = [&](auto& forkdb) -> controller::accepted_block_handle {
+         // previous not found, means it is unlinkable
          auto prev = forkdb.get_block( b->previous, include_root_t::yes );
          if( !prev ) return {};
 
-         return create_block_state_i( forkdb, id, b, *prev );
+         auto [best_head, obh] = create_block_state_i( forkdb, id, b, *prev );
```
greg7mdp, line 4176: Why not have `create_block_state_i` return a `controller::accepted_block_handle`?

libraries/chain/fork_database.cpp
```diff
@@ -277,20 +277,22 @@ namespace eosio::chain {
          EOS_RETHROW_EXCEPTIONS( fork_database_exception, "serialized fork database is incompatible with configured protocol features" )
       }
 
+      auto prev_head = head_impl(include_root_t::no);
```
greg7mdp, line 280: We don't need this call to `head_impl()` (which may be costly) if `ignore_duplicate == ignore_duplicate_t::no`

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -67,8 +67,9 @@ namespace eosio::chain {
       /**
        *  Add block state to fork database.
        *  Must link to existing block in fork database or the root.
+       *  @return true if n becomes the new best head.
```
greg7mdp, line 70:
```suggestion
       *  @return true if n becomes the new best head (and was not the best head before)
```

libraries/chain/controller.cpp
heifner: Much cleaner.

libraries/chain/controller.cpp
heifner: They were moved from `producer_plugin`. I think the main reason they are not signaled is you don't want to populate them if prometheus is not enabled. Signals are aways signaled (as far as I know), so would need to create the struct to signal it. Or would need some other mechanism to know if we should signal or not.

libraries/chain/controller.cpp
heifner: Not sure how I missed that.

libraries/chain/controller.cpp
greg7mdp: You could use [signal::empty()](https://www.boost.org/doc/libs/1_86_0/doc/html/boost/signals2/signal.html)  which is true if no slots are connected to the signal.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Maybe update the member of `accepted_block_handle` to be: `is_new_best_head` for clarity throughout.

libraries/chain/controller.cpp
heifner: I was looking for that, didn't look in the right place apparently. That would require a mutex I assume. Regardless this is all going away rather soon, so would prefer to just leave as is.

libraries/chain/fork_database.cpp
```diff
@@ -277,20 +277,22 @@ namespace eosio::chain {
...
+
       auto inserted = index.insert(n);
       EOS_ASSERT(ignore_duplicate == ignore_duplicate_t::yes || inserted.second, fork_database_exception,
                  "duplicate block added: ${id}", ("id", n->id()));
+
+      return inserted.second && prev_head != head_impl(include_root_t::no);
```
greg7mdp, line 286: Actually we never need `prev_head`.

```suggestion
      return inserted.second && head_impl(include_root_t::no) == n;
```

plugins/prometheus_plugin/metrics.hpp
```diff
@@ -311,16 +311,18 @@ struct catalog_type {
...
-      producer.register_update_incoming_block_metrics(
-          [&strand, this](const producer_plugin::incoming_block_metrics& metrics) {
+
+      auto& chain = app().get_plugin<chain_plugin>().chain();
+      chain.register_update_produced_block_metrics(
+          [&strand, this](const produced_block_metrics& metrics) {
+             strand.post([metrics, this]() { update(metrics); });
+          });
+      chain.register_update_incoming_block_metrics(
+          [&strand, this](const incoming_block_metrics& metrics) {
```
greg7mdp, line 325: Could use a generic lambda instead of the 3 almost identical lambdas.

```
      auto metrics_cb =  [&strand, this](const auto& metrics) {
         strand.post([metrics, this]() { update(metrics); });
      };
      
      auto& producer = app().get_plugin<producer_plugin>();
      producer.register_update_speculative_block_metrics(metrics_cb);

      auto& chain = app().get_plugin<chain_plugin>().chain();
      chain.register_update_produced_block_metrics(metrics_cb);
      chain.register_update_incoming_block_metrics(metrics_cb);
```

libraries/chain/fork_database.cpp
heifner: Or better yet, use iterator and avoid creating a `block_state_ptr`.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -459,6 +483,9 @@ namespace eosio::chain {
       // is the bls key a registered finalizer key of this node, thread safe
       bool is_node_finalizer_key(const bls_public_key& key) const;
 
+      void register_update_produced_block_metrics(std::function<void(produced_block_metrics)>&&);
+      void register_update_speculative_block_metrics(std::function<void(speculative_block_metrics)>&&);
+      void register_update_incoming_block_metrics(std::function<void(incoming_block_metrics)>&&);
```
greg7mdp, line 488: These really should take values I think, not rvalues.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -3190,16 +3146,8 @@ const std::set<account_name>& producer_plugin::producer_accounts() const {
    return my->_producers;
 }
 
-void producer_plugin::register_update_produced_block_metrics(std::function<void(producer_plugin::produced_block_metrics)>&& fun) {
-   my->_update_produced_block_metrics = std::move(fun);
-}
-
 void producer_plugin::register_update_speculative_block_metrics(std::function<void(speculative_block_metrics)> && fun) {
```
greg7mdp, line 3149: I think should take a value, not a rvalue.

plugins/prometheus_plugin/metrics.hpp
heifner: Hopefully, this will all die soon. Worth updating?

libraries/chain/controller.cpp
greg7mdp: That's fine, you can ignore my comments related to these metrics callbacks.

libraries/chain/include/eosio/chain/controller.hpp
heifner: Does it matter?

plugins/producer_plugin/producer_plugin.cpp
heifner: why?

libraries/chain/fork_database.cpp
greg7mdp: I think creating the `shared_ptr` is very minimal overhead, not worth making the code more complicated.

heifner: What do you think of the latest?

libraries/testing/tester.cpp
```diff
@@ -1214,10 +1213,11 @@ namespace eosio::testing {
...
             auto block = a.control->fetch_block_by_number(i);
             if( block ) { //&& !b.control->is_known_block(block->id()) ) {
-               auto btf = b.control->create_block_handle_future( block->calculate_id(), block );
+               auto id = block->calculate_id();
+               auto [best_head, obh] = b.control->accept_block( id, block );
                b.control->abort_block();
-               controller::block_report br;
-               b.control->push_block(br, btf.get(), {}, trx_meta_cache_lookup{}); //, eosio::chain::validation_steps::created_block);
+               EOS_ASSERT(obh, unlinkable_block_exception, "block did not link ${b}", ("b", id));
+               b.control->apply_blocks({}, trx_meta_cache_lookup{}); //, eosio::chain::validation_steps::created_block);
```
greg7mdp, line 1220: Do we need that comment at the end?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3821,39 +3736,50 @@ namespace eosio {
...
-         if( block_num != 0 ) {
-            assert(obt);
-            fc_dlog( logger, "validated block header, broadcasting immediately, connection - ${cid}, blk num = ${num}, id = ${id}",
-                     ("cid", cid)("num", block_num)("id", obt->id()) );
-            my_impl->dispatcher.add_peer_block( obt->id(), cid ); // no need to send back to sender
-            my_impl->dispatcher.bcast_block( obt->block(), obt->id() );
-         }
+         if (best_head) {
+            fc_dlog(logger, "posting block ${n} to app thread", ("n", ptr->block_num()));
+            app().executor().post(priority::medium, exec_queue::read_write,
```
greg7mdp, line 3764: I'm not understanding why we are posting this to the app thread?

libraries/chain/fork_database.cpp
```diff
@@ -277,20 +277,22 @@ namespace eosio::chain {
          EOS_RETHROW_EXCEPTIONS( fork_database_exception, "serialized fork database is incompatible with configured protocol features" )
       }
 
+      const auto prev_indx = index.template get<by_best_branch>().begin();
```
greg7mdp, line 280: You don't need the variable `prev_indx` or the call to `index.template get<by_best_branch>().begin();` here.

libraries/chain/fork_database.cpp
greg7mdp: You didn't remove the previous call to get the old head.

plugins/prometheus_plugin/metrics.hpp
greg7mdp: up to you.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Why accept only rvalues? On this case it works because if you pass a lambda, and intermediate `std::function` is created so it is always a rvalue.

I think we should only require `rvalues` when we have a specific reason to not want any copying.

libraries/chain/fork_database.cpp
heifner: The `.begin()` should be almost no cost. Do you think the `n == head_impl(include_root_t::no)` is actually better? Seems like the `by_best_branch` is pretty clear.

greg7mdp: Why do we need this line at all?

heifner: For the ` prev_indx != index.template get<by_best_branch>().begin()` below.

greg7mdp: But we don't need `prev_indx` below.

If we inserted a new `bsp`, and that `bsp` is the new best head, then we do have a new best head.
The last line can be:

```
      return inserted.second && head_impl(include_root_t::no) == n;
```

heifner: Right. My point is:
```cpp
const auto prev_indx = index.template get<by_best_branch>().begin();
...
return inserted.second && prev_indx != index.template get<by_best_branch>().begin();
```
Is more efficient when inserted which is the normal case. But I can change to `head_impl(include_root_t::no) == n` if you think it is better.

heifner: That is not a lookup it just returns the named index reference. It is implemented as `return *this`.

greg7mdp: Oh I see you're right! Thanks for enlightening me :-).
I think I still prefer my version (or  `return inserted.second && n == *index.template_ get<by_best_branch>().begin();`) because it doesn't rely on the specific behavior of the multi-index implementation. It looks like all the indices are updated whenever an element is added, but I though it could be done lazily maybe (only when the indices are accessed), although now that I think more about it it probably would be tricky to implement.

heifner: Also relies on stable iterators which `multi_index` also provides. 

Since you think it is clearer, I'll go with comparing it to `n`.

If comparing with `n`, I think `head_impl(include_root_t::no) == n` should be almost as efficient as accessing the index directly (one additional if check).

plugins/net_plugin/net_plugin.cpp
heifner: `chain_plugin->accept_block()` calls into the controller to process the block which has to be on the main thread. In Part 2 this only calls controller to apply blocks without a specific block.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Optimize block processing - Part 1.
Note:end

PR #624, opened 2024-08-23
[1.0] Update FSI serialization to support dynamically sized data and verify a checksum

### LINKED ISSUE #617: Update fsi serialization implementation to support dynamically size serialized blobs
Currently the serializer for fsi assumes each finalizer's fsi is fixed size when serialized. We want to update this to work even when each finalizer's serialized fsi can vary in size.

To preserve the optimization that we do not write out the fsi of inactive finalizers on each write, their serialized data should be written first in the file and the fsi for the active finalizers should come after.

The version for the fsi file should be bumped to version 1 so that we can still support version 0 fsi files.

For now, the fsi record for each finalizer that is being serialized does not change. That will happen in a later issue. But this should still create the infrastructure needed to upgrade a version 0 fsi struct to a version 1 fsi struct (which will be useful later) even though currently they would be two different structs that have the same exact members.

heifner:
- Add version 1 of FSI safety file.
  - Supports variable size FSI with inactive FSIs written at the top of the file
  - Adds checksum verification of FSI data

- Added a new `fc::datastream_crc` which uses `boost::crc` for checksum. Defaults to `boost::crc_32_type`, but any `boost::crc` type will work.

Resolves #617


libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -70,21 +71,34 @@ namespace eosio::chain {
...
    public:
-      static constexpr uint64_t current_safety_file_version = 0;
+      ///
+      /// Version 0: Spring 1.0.0 RC2 - File has fixed packed sizes with inactive safety info written to the end
+      ///                               of the file. Consists of [finalizer public_key, FSI]..
+      /// Version 1: Spring 1.0.0 RC3 - Variable length FSI with inactive safety info written at the beginning of
+      ///                               the file. Uses crc32 checksum to verify data on read.
+      ///
+      static constexpr uint64_t safety_file_version_0       = 0;
+      static constexpr uint64_t safety_file_version_1       = 1;
```
greg7mdp, line 81: I don't think these are very useful, but OK.

libraries/chain/finality/finalizer.cpp
```diff
@@ -175,47 +175,80 @@ void my_finalizers_t::maybe_update_fsi(const block_state_ptr& bsp, const qc_t& r
...
+   uint64_t num_finalizers {0};
+   fc::raw::unpack(persist_file, num_finalizers);
+   for (size_t i=0; i<num_finalizers; ++i) {
+      bls_public_key pubkey;
+      my_finalizers_t::fsi_t fsi;
+      fc::raw::unpack(persist_file, pubkey);
+      fc::raw::unpack(persist_file, fsi);
+      res.emplace(pubkey, fsi);
+   }
+}
```
greg7mdp, line 244: I don't see the point in having two functions if they are exactly the same. I feel that the only change may be the type of  `my_finalizers_t::fsi_t`, in which case maybe a template function would be better imo.

libraries/libfc/include/fc/io/datastream_crc.hpp
```diff
@@ -0,0 +1,109 @@
...
+      return ds_.pos();
+   }
+
+   bool valid() const {
+      return ds_.valid();
+   }
+
+   // only use with p==0, otherwise use seekp() below
+   bool seekp(size_t p) {
+      if (p == 0) {
```
greg7mdp, line 74: probably should be `assert(p == 0)`.

libraries/chain/finality/finalizer.cpp
heifner: They were different until I created the dataset_crc. They are different in the PR that targets this one.

libraries/libfc/include/fc/io/datastream_crc.hpp
heifner: Not really sure what to do. I'll add the assert.

libraries/chain/include/eosio/chain/finality/finalizer.hpp
heifner: Maybe they represent all valid options? 🤷  Not sure, but seems better than just 0 and 1 hard coded.

greg7mdp: I feel since `version_0` will always be `0` and `version_1` will always be `1`, hardcoded is just as good and one less indirection, but personal preference so no need to change.

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: New version of Finalizer Safety Information file to support dynamically sized data, with checksum.
Note:end

libraries/chain/finality/finalizer.cpp
linh2931: Add a comment about why seemingly identical functions?

libraries/chain/finality/finalizer.cpp
```diff
@@ -257,17 +290,30 @@ my_finalizers_t::fsi_map my_finalizers_t::load_finalizer_safety_info() {
...
+         verify_checksum = false;
+         break;
+      case safety_file_version_1:
+         load_finalizer_safety_info_v1(res);
+         break;
+      default:
+         assert(0);
+      }
+
+      if (verify_checksum) {
```
linh2931, line 306: Should `verify_checksum` be done before loading?

libraries/chain/finality/finalizer.cpp
heifner: They are not identical in the follow-on PR.

libraries/chain/finality/finalizer.cpp
heifner: That is in this PR. The `datastream_crc` does checksum as it reads.

linh2931 (approved):

arhag (approved):

PR #625, opened 2024-08-23
[1.0] various different signedness warning fixes

spoonincode:
grabbag of different signedness warning fixes I noticed. I'm fine sending these to main instead, but most/all of these were introduced at the 11th hour before cutting 1.0.0 rc

Needs AntelopeIO/chainbase#53

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2008,7 +2008,7 @@ namespace eosio {
                  ("ex", -reversible_remaining)("fs", fork_db_size));
          reversible_remaining = 0;
       }
-      if (reversible_remaining < sync_fetch_span) {
+      if ((uint32_t)reversible_remaining < sync_fetch_span) {
          auto fork_db_size = my_impl->chain_plug->chain().fork_db_size();
          fc_wlog(logger, "sync-fetch-span ${sfs} restricted to ${r} by max-reversible-blocks, fork_db_size ${fs}",
                  ("sfs", sync_fetch_span)("r", reversible_remaining)("fs", fork_db_size));
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2001,18 +2001,22 @@ namespace eosio {
    }
 
    uint32_t sync_manager::active_sync_fetch_span() const {
-      int32_t reversible_remaining = my_impl->chain_plug->chain().max_reversible_blocks_allowed();
-      if (reversible_remaining <= 0) {
-         auto fork_db_size = my_impl->chain_plug->chain().fork_db_size();
-         fc_wlog(logger, "max-reversible-blocks exceeded by ${ex}, fork_db_size ${fs}",
-                 ("ex", -reversible_remaining)("fs", fork_db_size));
-         reversible_remaining = 0;
-      }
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2008,7 +2008,7 @@ namespace eosio {
                  ("ex", -reversible_remaining)("fs", fork_db_size));
          reversible_remaining = 0;
       }
-      if (reversible_remaining < sync_fetch_span) {
+      if ((uint32_t)reversible_remaining < sync_fetch_span) {
```
spoonincode, line 2011: I usually don't like casting away signedness warnings like this but here we've established just above that `reversible_remaining >=0` so it's completely safe. Not sure what would be a better pattern in this case.

tests/test_snapshot_scheduler.cpp
```diff
@@ -233,7 +233,7 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_old_json) {
    unsigned found = 0;
    for(const std::filesystem::directory_entry& dir_entry : std::filesystem::directory_iterator(temp / "snapshots"))
       found += std::regex_search(dir_entry.path().filename().string(), snapshotfile_regex);
-   BOOST_REQUIRE_GE(found, 3);
+   BOOST_REQUIRE_EQUAL(found, 3u);
```
spoonincode, line 236: The test was originally more sloppy and it's why it was `>=3`; I forgot to change this to `==3` after making it more strict. Not a big enough deal to go back to 5.0 to fix this imo

plugins/net_plugin/net_plugin.cpp
heifner: I don't care for the cast either. Maybe something like:

```
      int32_t reversible_remaining = my_impl->chain_plug->chain().max_reversible_blocks_allowed();
      uint32_t constrained_reversible_remaining = 0;
      if (reversible_remaining <= 0) {
         auto fork_db_size = my_impl->chain_plug->chain().fork_db_size();
         fc_wlog(logger, "max-reversible-blocks exceeded by ${ex}, fork_db_size ${fs}",
                 ("ex", -reversible_remaining)("fs", fork_db_size));
      } else {
         constrained_reversible_remaining = reversible_remaining;
      }
      if (constrained_reversible_remaining < sync_fetch_span) {
         auto fork_db_size = my_impl->chain_plug->chain().fork_db_size();
         fc_wlog(logger, "sync-fetch-span ${sfs} restricted to ${r} by max-reversible-blocks, fork_db_size ${fs}",
                 ("sfs", sync_fetch_span)("r", constrained_reversible_remaining )("fs", fork_db_size));
         return constrained_reversible_remaining ;
      }
```

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
heifner: Reminds me of programming in Ada.

heifner (approved):

greg7mdp (approved):

spoonincode:
ah crud sorry I got way to trigger happy and forgot to merge in chainbase first

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Fix signesdess warnings.
Note:end

PR #640, opened 2024-08-26
resurrect `compiler_builtins_tests` for compiler builtins host functions

spoonincode:
Resurrect `compiler_builtins_tests` removed in bedf133. This test is exercising host functions in spring so it does belong here in spring. Removal of these tests lost coverage for host function behavior such as
https://github.com/AntelopeIO/spring/blob/22d9892fbb32f8ee46ea8428427266e49bff6399/libraries/chain/webassembly/compiler_builtins.cpp#L46
(afaict anyways, maybe there is some other test somewhere I missed)


unittests/test-contracts/test_api/test_compiler_builtins.cpp
```diff
@@ -0,0 +1,424 @@
+/**
+ *    @file test_compiler_builtins.cpp
+ *    @copyright defined in eos/LICENSE
+ */
+#include <eosio/eosio.hpp>
+#include <eosio/print.hpp>
+
+#include "test_api.hpp"
+
+//these are no longer exposed in cdt
```
spoonincode, line 10: Similarly, we need to explicitly call the host functions so we know we're actually testing the host functions. It's possible future cdt versions won't ever implicitly call the host functions given something like
```c++
__int128 a = 0, b = 0;
a/b;
```

greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add compiler buildtins tests to Spring. Previously included in Leap CI.
Note:end

PR #647, opened 2024-08-27
explicitly handle inputs in `__divti3` and `__modti3` that result in UB overflow

spoonincode:
Division (or modulo) of INT_MIN by `-1` is ultimately UB due to integer overflow. But not only that, it can cause a fault. For example, the following will fail with `SIGFPE` on x86_64 (build with `-O0` for this trivial case),
```c++
int64_t doit(int64_t in, int64_t x) {
	return in / x;
}
int main() {
	int64_t x = std::numeric_limits<int64_t>::min();
	int64_t y = doit(x, -1);
	return 0;
}
```
So what's interesting  to consider is,
https://github.com/AntelopeIO/spring/blob/abc65cef7117f83e769641f4a47d171689b41a08/libraries/chain/webassembly/compiler_builtins.cpp#L36-L51
Thus, it is possible for a contract to pass `std::numeric_limits<__int128>::min()` and `-1` resulting in a division performed via inputs to a library (libgcc/compiler-rt) that has undefined results.

Fortunately, the `int128` impl in both libgcc and compiler-rt seem to behave the same here from my tests: these inputs cause a "noop" (i.e. as if the division was by `1`) for `__divti3`, and  `__modti3` sets the result to 0.

The first commit in this PR adds a unit test that validates the above statement is currently true: for our supported platforms `__divti3` and `__modti3` behave a certain way given these inputs. Notice that the tests pass with this new unit test, except for the UBSAN build which correctly flags we are performing undefined behavior.

The second commit enforces the current behavior forever with an explicit check in the host functions.

### DIFF
--- a/libraries/chain/webassembly/compiler_builtins.cpp
+++ b/libraries/chain/webassembly/compiler_builtins.cpp
@@ -45,6 +45,12 @@ namespace eosio { namespace chain { namespace webassembly {
 
       EOS_ASSERT(rhs != 0, arithmetic_exception, "divide by zero");
 
+      //force integer overflow to return dividend unchanged
+      if(lhs == std::numeric_limits<__int128>::min() && rhs == -1) {
+         *ret = lhs;
+         return;
+      }
+
       lhs /= rhs;
 
       *ret = lhs;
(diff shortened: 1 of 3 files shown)


linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve div and mod operations in webassemble to reduce overflows.
Note:end

greg7mdp (approved):

PR #648, opened 2024-08-27
[1.0 -> main] Update FSI serialization to support dynamically sized data and verify a checksum

### LINKED ISSUE #617: Update fsi serialization implementation to support dynamically size serialized blobs

heifner:
- Add version 1 of FSI safety file.
  - Supports variable size FSI with inactive FSIs written at the top of the file
  - Adds checksum verification of FSI data

- Added a new `fc::datastream_crc` which uses `boost::crc` for checksum. Defaults to `boost::crc_32_type`, but any `boost::crc` type will work.

Merges `release/1.0` into `main` including #624

Resolves #617


arhag (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: New version of Finalizer Safety Information file to support dynamically sized data, with checksum.
Note:end

PR #652, opened 2024-08-27
[1.0] Pause production when max-reversible-blocks reached

### LINKED ISSUE #641: max-reversible-blocks should only prevent block production
Nodeos should no longer shut down if the number of blocks in the fork database exceeds the configured `max-reversible-blocks`. 

Instead, when producer plugin is about to produce a block, it should check to the `block_header_state` `finality_core` for the number of reversible blocks in that branch. If that exceeds `max-reversible-blocks`, then it should avoid producing the block (can just be speculating) and log a warning message.

heifner:
Change the behavior of `max-reversible-blocks`. Instead of shutting down a node when `max-reversible-blocks` is reached; pause production. 
* A node will no longer shutdown when it reaches `max-reversible-blocks`. 
* A block production node will not produce additional blocks when `max-reversible-blocks` is reached.

A block production node will report an `error`: `Not producing block because max-reversible-blocks ${m} reached, head ${h}, lib ${l}.` and will instead speculatively execute transactions until LIB advances. Note LIB can advance by finalizers being restored and other block producers producing blocks. Once the number of reversible blocks is reduced on the network, the block producer will un-pause and start producing blocks again. 

If all block producers have the same default value for `max-reversible-blocks` (3600) then after approximately 30 minutes of LIB not advancing, all block producers will pause production. Manual restart of at least one block producer will be required at that point, after finality issues have been resolved, with a higher `max-reversible-blocks`. Once a block producer with network finalizers are able to advance finality the number of reversible blocks should be reduced back down to ~2 and all other block producers will automatically un-pause and begin producing again.

Resolves #641


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1930,7 +1935,8 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       return start_block_result::failed;
    }
 
-   block_num_type             head_block_num    = chain.head().block_num();
+   block_handle               head              = chain.head();
+   block_num_type             head_block_num    = head.block_num();
```
greg7mdp, line 1939: detail, I think it would be nicer to define `head` in the if condition where it is used, instead of so much earlier.

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1976,6 +1982,11 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
               ("ov", _last_other_vote_received.load(std::memory_order_relaxed))("bt", _accepted_block_time));
       _pending_block_mode = pending_block_mode::speculating;
       not_producing_when_time = true;
+   } else if (head_block_num - head.irreversible_blocknum() >= _max_reversible_blocks) {
```
linh2931, line 1985: Should `>=` be `>` as the option description says beyond?

linh2931: Need to check `_max_reversible_blocks` not 0. Otherwise always pause.

linh2931 (changes requested):

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I was thinking moving just `block_handle head = chain.head();`, but never mind that is not nicer.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Change behavior to pause production when max-reversible-blocks are reached.
Note:end

PR #658, opened 2024-08-28
[1.0 -> main] Improve restart production integration test

linh2931:
The existing restart production  test may not stop the node in the middle of a pending policy. If it misses it and the pending policy becomes active, then the test will still pass but it won't test the behavior the test was intended for.

Instead of changing the order of the finalizer keys in order to force a finalizer policy change, the test is changed to use an unconfigured key in the new finalizer policy such that it is guaranteed that the node is in a state where pending policy exists before the node is shut down.

Forwards https://github.com/AntelopeIO/spring/pull/577 to `main`

Resolves https://github.com/AntelopeIO/spring/issues/552


greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve finalizer policy changes in test by changing keys multiple times. Previously changed the order of keys.
Note:end

Issue #672, opened 2024-08-30
Test failure: production_restart 


heifner:
https://github.com/AntelopeIO/spring/actions/runs/10621940100/job/29446218336

PR #674, opened 2024-08-30
[1.0] Fix production_restart test flakiness

linh2931:
The test requires pending policy to be in place. It waits on LIB. But a proposed policy is promoted to pending only after the block where it was proposed becomes final; that LIB is not necessarily the LIB where the policy was proposed.

In the test failure reported by https://github.com/AntelopeIO/spring/issues/672, it was LIB on the block before setfinalizer because the setfinalizer was sent to the node that was not producing at the time. On the producer the setfinalizer is in the next block.

The solution is wait until the block where a finalizer policy was proposed becomes final before checking pending policy.

Resolves https://github.com/AntelopeIO/spring/issues/672


heifner (approved):

tests/production_restart.py
```diff
@@ -81,11 +81,12 @@
...
-    assert node0.waitForLibToAdvance(), "node0 did not advance LIB after setfinalizers"
-    # Wait for head to advance twice to make sure pending policy is in place
-    node0.waitForHeadToAdvance()
-    node0.waitForHeadToAdvance()
+    transId = cluster.setFinalizers([node0, node1], node0)
+    assert transId, "setfinalizers failed"
+    # A proposed finalizer policy is promoted to pending only after the block where it
+    # was proposed become irreversible.
+    # Wait for pending policy is in place.
+    assert node0.waitForTransFinalization(transId), f'setfinalizer transaction {transId} failed to be rolled into a LIB block'
```
greg7mdp, line 89: `waitForTransFinalization` seems like an odd name to me? Isn't it just `waitForBlockFinal`?

greg7mdp (approved):

tests/production_restart.py
linh2931: The function takes a transaction ID as a parameter. The name seems ok.

greg7mdp: OK, makes sense, thanks!

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Upgrade test to resolve race condition where test results were checked after policy change was applied.
Note:end

PR #675, opened 2024-08-30
[1.0 -> main] Fix production_restart test flakiness

linh2931:
The test requires pending policy to be in place. It waits on LIB. But a proposed policy is promoted to pending only after the block where it was proposed becomes final; that LIB is not necessarily the LIB where the policy was proposed.

In the test failure reported by https://github.com/AntelopeIO/spring/issues/672, it was LIB on the block before setfinalizer because the setfinalizer was sent to the node that was not producing at the time. On the producer the setfinalizer is in the next block.

The solution is wait until the block where a finalizer policy was proposed becomes final before checking pending policy.

Forwards https://github.com/AntelopeIO/spring/pull/674

Resolves https://github.com/AntelopeIO/spring/issues/672


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Upgrade test to resolve race condition where test results were checked after policy change was applied.
Note:end

PR #681, opened 2024-08-31
[1.0] Test accept_block_header, accept_block, and irreversible_block signals

heifner:
Add verification to libtester that `controller` signals `accept_block_header`, `accept_block`, `irreversible_block` happen in order and do not repeat.

Working on #619 and wanted a test to verify signals continue to work the same as before.

### DIFF
--- a/libraries/testing/include/eosio/testing/tester.hpp
+++ b/libraries/testing/include/eosio/testing/tester.hpp
@@ -8,6 +8,7 @@
 #include <fc/io/json.hpp>
 #include <boost/test/unit_test.hpp>
 #include <boost/tuple/tuple_io.hpp>
+#include <boost/unordered/unordered_flat_map.hpp>
 
 #include <eosio/testing/bls_utils.hpp>
 
@@ -591,7 +592,7 @@ namespace eosio::testing {
          map<transaction_id_type, transaction_receipt> chain_transactions;
          map<account_name, block_id_type>              last_produced_block;
          enum class block_signal { block_start, accepted_block_header, accepted_block, irreversible_block };
-         map<block_id_type, block_signal>              blocks_signaled;
+         boost::unordered_flat_map<block_id_type, block_signal> blocks_signaled;
          unapplied_transaction_queue                   unapplied_transactions;
 
       public:
--- a/libraries/testing/include/eosio/testing/tester.hpp
+++ b/libraries/testing/include/eosio/testing/tester.hpp
@@ -576,6 +576,9 @@ namespace eosio::testing {
          signed_block_ptr       _finish_block();
          void                   _check_for_vote_if_needed(controller& c, const block_handle& bh);
 
+         enum class block_signal { block_start, accepted_block_header, accepted_block, irreversible_block };
+         bool                   _check_signal(const block_id_type& id, block_signal sig);
+
       // Fields:
       protected:
          bool                   _expect_votes {true};                          // if set, ensure the node votes on each block
@@ -591,9 +594,8 @@ namespace eosio::testing {
          controller::config                            cfg;
          map<transaction_id_type, transaction_receipt> chain_transactions;
          map<account_name, block_id_type>              last_produced_block;
-         enum class block_signal { block_start, accepted_block_header, accepted_block, irreversible_block };
-         boost::unordered_flat_map<block_id_type, block_signal> blocks_signaled;
          unapplied_transaction_queue                   unapplied_transactions;
+         boost::unordered_flat_map<block_id_type, block_signal> blocks_signaled;
 
       public:
          vector<digest_type>                           protocol_features_to_be_activated_wo_preactivation;
--- a/libraries/testing/include/eosio/testing/tester.hpp
+++ b/libraries/testing/include/eosio/testing/tester.hpp
@@ -590,6 +590,8 @@ namespace eosio::testing {
          controller::config                            cfg;
          map<transaction_id_type, transaction_receipt> chain_transactions;
          map<account_name, block_id_type>              last_produced_block;
+         enum class block_signal { block_start, accepted_block_header, accepted_block, irreversible_block };
+         map<block_id_type, block_signal>              blocks_signaled;
          unapplied_transaction_queue                   unapplied_transactions;
 
       public:
--- a/libraries/testing/tester.cpp
+++ b/libraries/testing/tester.cpp
@@ -391,7 +391,7 @@ namespace eosio::testing {
          auto i = blocks_signaled.find(id);
          if (i == blocks_signaled.end()) {
             // can be signaled on restart as the first thing since other signals happened before shutdown
-            blocks_signaled[id] = block_signal::accepted_block_header;
+            blocks_signaled[id] = block_signal::irreversible_block;
          } else {
             BOOST_TEST((i->second == block_signal::accepted_block), "should get irreversible_block signal only once");
             i->second = block_signal::irreversible_block;
--- a/libraries/testing/tester.cpp
+++ b/libraries/testing/tester.cpp
@@ -359,11 +359,10 @@ namespace eosio::testing {
             BOOST_TEST(block->block_num() > lib_number);
             auto i = blocks_signaled.find(id);
             BOOST_TEST_REQUIRE((i != blocks_signaled.end()), "should get accepted_block signal after accepted_block_header signal");
-            if (i->second == block_signal::accepted_block) {
-               // fork switch, accepted block signaled when block re-applied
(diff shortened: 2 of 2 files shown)


libraries/testing/tester.cpp
```diff
@@ -341,20 +342,39 @@ namespace eosio::testing {
...
-          for( auto receipt : block->transactions ) {
-              if( std::holds_alternative<packed_transaction>(receipt.trx) ) {
-                  auto &pt = std::get<packed_transaction>(receipt.trx);
+      [[maybe_unused]] auto accepted_block_connection = control->accepted_block().connect([this](const block_signal_params& t) {
+            const auto& [block, id] = t;
+            FC_ASSERT(block);
+            BOOST_TEST(block->block_num() > lib_number);
+            auto i = blocks_signaled.find(id);
+            BOOST_TEST_REQUIRE((i != blocks_signaled.end()), "should get accepted_block signal after accepted_block_header signal");
+            if (i->second == block_signal::accepted_block) {
```
linh2931, line 362: Why empty `if`?

linh2931 (commented):
Should those tests be moved out and be treated like other unit tests, rather than running them every time `open` is executed?

libraries/testing/tester.cpp
heifner: For the comment. I could pull the `BOOST_TEST` into an `if-else` if you think that would be better.

heifner:
> Should those tests be moved out and be treated like other unit tests, rather than running them every time `open` is executed?

I thought it better to make sure these constraints hold for all tests.

libraries/testing/tester.cpp
heifner: I reworked it to get rid of the empty `if`.

libraries/testing/tester.cpp
```diff
@@ -341,20 +342,38 @@ namespace eosio::testing {
...
-        const auto& [ block, id ] = t;
-        FC_ASSERT( block );
-          for( auto receipt : block->transactions ) {
-              if( std::holds_alternative<packed_transaction>(receipt.trx) ) {
-                  auto &pt = std::get<packed_transaction>(receipt.trx);
+      [[maybe_unused]] auto accepted_block_connection = control->accepted_block().connect([this](const block_signal_params& t) {
+            const auto& [block, id] = t;
+            FC_ASSERT(block);
+            BOOST_TEST(block->block_num() > lib_number);
+            auto i = blocks_signaled.find(id);
```
greg7mdp, line 360: Small detail, for some reason I find `i` a stange name for an iterator. Would prefer `it`.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -590,6 +590,8 @@ namespace eosio::testing {
          controller::config                            cfg;
          map<transaction_id_type, transaction_receipt> chain_transactions;
          map<account_name, block_id_type>              last_produced_block;
+         enum class block_signal { block_start, accepted_block_header, accepted_block, irreversible_block };
+         map<block_id_type, block_signal>              blocks_signaled;
```
greg7mdp, line 594: Why not use the more efficient `unordered_map`, or even better `boost::unordered_flat_map`

greg7mdp (approved):
PR title is a bit misleading. This does not add a test, but modifies the test harness so that these checks are performed for every libtester test (which is a good thing imo).

libraries/testing/include/eosio/testing/tester.hpp
heifner: The existing maps here were `map`, I'll change it to `boost::unordered_flat_map`

libraries/testing/tester.cpp
```diff
@@ -341,20 +342,38 @@ namespace eosio::testing {
...
       control->testing_allow_voting(true);
       if (lambda) lambda();
+      [[maybe_unused]] auto block_start_connection = control->block_start().connect([](block_num_type num) {
+         // only block number is signaled, in forking tests will get the same block number more than once.
+      });
+      [[maybe_unused]] auto accepted_block_header_connection = control->accepted_block_header().connect([this](const block_signal_params& t) {
+            const auto& [block, id] = t;
+            FC_ASSERT(block);
+            auto i = blocks_signaled.find(id);
+            BOOST_TEST((i == blocks_signaled.end()), "should get accepted_block_header signal only once, and before accepted_block signal");
```
greg7mdp, line 352: maybe:
```suggestion
            BOOST_TEST(!blocks_signaled.contains(id)), "should get accepted_block_header signal only once, and before accepted_block signal");
```

linh2931:
> > Should those tests be moved out and be treated like other unit tests, rather than running them every time `open` is executed?
> 
> I thought it better to make sure these constraints hold for all tests.

I don't have a strong opinion about this. But prefer to not packing non-essential checks into libtester. It would be sufficient to have a specific test for it. @arhag ?

linh2931 (approved):

greg7mdp (commented):
I almost feel there should be a single `verify_signal` function, which would take a `id` and `block_signal` parameter, and verify that the signal is expected. Maybe that could be a `base_tester` menber instead of a lambda. I think it would be simpler. What do you think?

heifner:
> I almost feel there should be a single `verify_signal` function, which would take a `id` and `block_signal` parameter, and verify that the signal is expected. Maybe that could be a `base_tester` menber instead of a lambda. I think it would be simpler. What do you think?

There are pros and cons. The lambdas keep it close to signal handlers. The other member abstracts it away into one location. I'm fine either way. @linh2931 what do you think?

libraries/testing/tester.cpp
```diff
@@ -348,20 +348,28 @@ namespace eosio::testing {
...
             const auto& [block, id] = t;
             FC_ASSERT(block);
-            BOOST_TEST(block->block_num() > lib_number);
-            auto itr = blocks_signaled.find(id);
-            BOOST_TEST_REQUIRE((itr != blocks_signaled.end()), "should get accepted_block signal after accepted_block_header signal");
-            if (itr->second != block_signal::accepted_block) { // on fork switch, accepted block signaled when block re-applied
-               BOOST_TEST((itr->second == block_signal::accepted_block_header));
-               itr->second = block_signal::accepted_block;
-            }
+            assert(block->block_num() > lib_number);
```
linh2931, line 362: Why FC_ASSERT a line above while assert here?

linh2931: Maybe it is better to make it an assert to be consistent.

greg7mdp:
> There are pros and cons. The lambdas keep it close to signal handlers. The other member abstracts it away into one location. I'm fine either way. @linh2931 what do you think?

I was thinking of something like this:

```
   bool base_tester::check_signal(const block_id_type&id,  block_signal sig) {
      bool valid = true;
      auto upd = fc::make_scoped_exit([&]() { blocks_signaled[id] = sig; });

      auto itr = blocks_signaled.find(id);
      bool present = itr != blocks_signaled.end();

      switch(sig) {
      case block_signal::block_start:
         return true;  // only block number is signaled,

      case block_signal::accepted_block_header:
         // should get accepted_block_header signal only once, and before accepted_block signal;
         return !present;

      case block_signal::accepted_block:
         // should get accepted_block signal after accepted_block_header signal
         // or after accepted_block (on fork switch, accepted block signaled when block re-applied)
         return present && (itr->second == block_signal::accepted_block_header ||
                            itr->second == block_signal::accepted_block);

      case block_signal::irreversible_block:
         return (!present && itr->second == block_signal::irreversible_block) ||
                ( present && itr->second == block_signal::accepted_block);
      }
   }

```

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add a test against controller to ensure `accept_block_header`, `accept_block`, `irreversible_block` are applied in the correct order.
Note:end

libraries/testing/tester.cpp
```diff
@@ -322,6 +323,34 @@ namespace eosio::testing {
...
+         return !present;
+
+      case block_signal::accepted_block:
+         // should get accepted_block signal after accepted_block_header signal
+         // or after accepted_block (on fork switch, accepted block signaled when block re-applied)
+         return present && (itr->second == block_signal::accepted_block_header ||
+                            itr->second == block_signal::accepted_block);
+
+      case block_signal::irreversible_block:
+         return !present ||
```
greg7mdp, line 347: restore comment I forgot in my suggestion
```suggestion
         // can be signaled on restart as the first thing since other signals happened before shutdown
         // otherwise should get irreversible_block signal only once
         return !present ||
```

Issue #651, opened 2024-08-27
Add integration test for max-reversible-blocks


heifner:
Add to #542 to also test that production is paused when `max-reversible-blocks` is hit. Restart one producer node with larger `max-reversible-blocks` and verify production can continue to new max. Restart finalizer so LIB can update and verify production un-pauses on production node.

PR #685, opened 2024-09-03
[1.0] Add integration test for `--max-reversible-blocks`

linh2931:
Setup:

The test network consits of 3 nodes. 2 of them are producer nodes, with producera and producerb. Configure `--max-reversible-blocks` but disable `--production-pause-vote-timeout-ms` to avoid interfeernce.

* node0: Enables block production for producera and has the finalizer key with description of producera. Has vote-threads enabled. Connect to producerbNode.
* producerbNode: Enables block production for producerb. Has vote-threads enabled. Connect to finalizerbNode and node0.
* finalizerbNode: Has the finalizer key with description of producerb. Has vote-threads enabled. Connect to producerbNode.

Test cases:

1. Bring down finalizerbNode. producerbNode and node0 should automatically pause production due to the number of reversible blocks keeping growing because not receiving votes from finalizerbNode and LIB stopping advancings.
2. Bring down node0 and restart it with a larger `--max-reversible-blocks`. node0 should resume production until the new limit is reached.
3. Restart finalizerbNode. producerbNode and node0 should automatically unpause production because the number of reversible blocks decreases due to LIB advances on both nodes.

Resolves https://github.com/AntelopeIO/spring/issues/651


tests/production_pause_max_rev_blks_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+####################################################################################
+# production_pause_max_rev_blks_test
+# Test --max-reversible-blocks works as expected.
+#
+# Setup:
+#
+# The test network consits of 3 nodes. 2 of them are producer nodes, with
+# producera and producerb. Configure --max-reversible-blocks but
+# disable --production-pause-vote-timeout-ms to avoid interfernce.
```
heifner, line 17: interference

tests/production_pause_max_rev_blks_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+    cluster.setWalletMgr(walletMgr)
+
+    Print(f'producing nodes: {pnodes}, totalNodes: {totalNodes}, prodCount: {prodCount}, delay between nodes launch: {delay} second{"s" if delay != 1 else ""}')
+
+    # Set --max-reversible-blocks to a small number so it is exceeded quickly
+    # but it needs to be big enough for transitioning to Savanna..
+    # Disable production-pause-vote-timeout so that production pause is only
+    # caused by max-reversible-blocks reached.
+    extraNodeosArgs="--max-reversible-blocks 48 --production-pause-vote-timeout-ms 0"
```
heifner, line 67: Instead maybe shutdown/relaunch after transition; then you don't have to guess on a number of blocks and the test runs quicker.

tests/TestHarness/Node.py
```diff
@@ -273,6 +273,54 @@ def isProducer():
             f"Waited for {time.perf_counter()-start} sec but never found producer: {producer}. Started with {initialProducer} and ended with {self.getInfo()['head_block_producer']}"
         return found
 
+    # returns True if the node has paused production.
+    def paused(self):
```
heifner, line 277: Maybe a better name is `missedNextProductionRound()`

tests/TestHarness/Node.py
```diff
@@ -273,6 +273,54 @@ def isProducer():
...
+        # Cannot use producer_plugin's paused() endpoint as it does not
+        # include paused due to max-reversible-blocks exceeded.
+        # The idea is to find the scheduled start block of node's producer's
+        # next round. If that block is not produced, it means block production
+        # on the node is paused.
+
```
heifner, line 283: Should assert `isProducer`

tests/TestHarness/Node.py
```diff
@@ -273,6 +273,54 @@ def isProducer():
...
+                currProducerPos=i
+            if scheduled_producers[i] == self.producerName:
+                nodeProducerPos=i
+
+        # find the number of the blocks to node producer's next scheduled round
+        blocksToNextScheduledRound = 0
+        if currProducerPos < nodeProducerPos:
+            # nodeProducerPos - currProducerPos - 1 is the number of producers
+            # from current producer to the node producer in the schedule
+            blocksToNextScheduledRound = (nodeProducerPos - currProducerPos - 1) * 12 + blocksRemainedInCurrRound + 1
```
heifner, line 310: Assumes 12 blocks per round. Many tests do assume 12 blocks per round, but would be nice to stop assuming that is always the case. At the very least make it a const at top of function.

tests/production_pause_max_rev_blks_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # Verify LIB stalled on node0 and producerbNode due to finalizerNode was shutdown
+    Print("Verify LIB stalled  after shutdown of finalizerbNode")
+    if producerbNode.waitForLibToAdvance(timeout=5): # LIB can advance for a few blocks first
+        assert not producerbNode.waitForLibToAdvance(timeout=5), "LIB should not advance on producerbNode after finalizerbNode was shutdown"
+    if node0.waitForLibToAdvance(timeout=5): # LIB can advance for a few blocks first
+        assert not node0.waitForLibToAdvance(timeout=5), "LIB should not advance on node0 after finalizerbNode was shutdown"
+
+    # give time for max reversible blocks to be exceeded
+    time.sleep(24) # 24 is the time to produce 48 blocks which is the configured max-reversible-blocks
```
heifner, line 117: Instead of a sleep, wait for a head block above lib that exceeds max reversible.

tests/production_pause_max_rev_blks_test.py
```diff
@@ -0,0 +1,166 @@
...
+    # nodeos can restart with a higher --max-reversible-blocks
+
+    node0.getInfo()
+    prevHeadBlockNum=node0.lastRetrievedHeadBlockNum
+
+    Print("Shutdown node0")
+    node0.kill(signal.SIGTERM)
+    assert not node0.verifyAlive(), "node0 did not shutdown"
+
+    addSwapFlags={"--max-reversible-blocks": "96"}
```
heifner, line 134: 48 - 96 seems like a large jump. Seems something as small as 20 blocks would be more than enough.

tests/production_pause_max_rev_blks_test.py
```diff
@@ -0,0 +1,166 @@
...
+    Print("Restart node0 with a higher max-reversible-blocks")
+    node0.relaunch(chainArg="--enable-stale-production", addSwapFlags=addSwapFlags)
+
+    Print("Wait for one round")
+    time.sleep(7)
```
heifner, line 139: Instead of a sleep, wait for a head block above lib that exceeds max reversible.

heifner (changes requested):

tests/production_pause_max_rev_blks_test.py
```diff
@@ -0,0 +1,177 @@
...
+    # nodeos can restart with a higher --max-reversible-blocks
+
+    node0.getInfo()
+    prevHeadBlockNum=node0.lastRetrievedHeadBlockNum
+
+    Print("Shutdown node0")
+    node0.kill(signal.SIGTERM)
+    assert not node0.verifyAlive(), "node0 did not shutdown"
+
+    higherMaxReversibleBlocks=maxReversibleBlocks+12
```
heifner, line 145: Add a comment that `12` is just a value chosen for test. Might be better to use 10 so no confusion that it is related to blocks per round.

tests/TestHarness/Node.py
```diff
@@ -273,6 +273,58 @@ def isProducer():
...
+            # len(scheduled_producers) - currProducerPos - 1 is the number
+            # of producers after node producer in the schedule
+            blocksToNextScheduledRound = (nodeProducerPos + (len(scheduled_producers)  - currProducerPos - 1)) * blocksPerProducer + blocksRemainedInCurrRound + 1
+
+        # find the block number of the node producer's next scheduled round
+        nextScheduledRoundBlockNum=currBlockNum + blocksToNextScheduledRound
+        timeout=blocksToNextScheduledRound/2 + 2 # leave 2 seconds for avoid flakiness
+        if Utils.Debug: Utils.Print(f'blocksToNextScheduledRound {blocksToNextScheduledRound}, nextScheduledRoundBlockNum {nextScheduledRoundBlockNum}, timeout {timeout}')
+
+        return not self.waitForBlock(nextScheduledRoundBlockNum, timeout=timeout)
```
greg7mdp, line 326: Even if the node doesn't produce during its round, wouldn't that same block number be produced by the next producer?

greg7mdp (approved):

tests/TestHarness/Node.py
greg7mdp: Does the `self.waitForBlock` only waits for a block produced on the current node?

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Additional test cases for option `max-reversible-blocks`.
Note:end

tests/TestHarness/Node.py
heifner: No, any block produced or received

greg7mdp: So I wonder if the function `def missedNextProductionRound(self):` is correct?
It looks like it assumes that all producers will produce all the blocks of their rounds except for self.
For example if the other producer was paused, we would still time out and (incorrectly) believe this mode didn't produce during its round.

heifner: Yes, seems like this could just use `waitForProducer()` with a timeout instead.

PR #688, opened 2024-09-03
Optimize block processing - Part 2

### LINKED ISSUE #529: Optimize processing of incoming blocks

heifner:
Part 2 of 3 of #529. See #619 for part 1.

This PR only notifies producer plugin that a new incoming block is ready to be processed instead of notifying that a specific block is available.

- Removed unused `rejected_block` channel. If needed in the future this will need to be a signal out of controller instead of a channel.
- Removed unused `voted_block` channel. There is a controller signal for this and the channel was not used.
- Send `on_incoming_block` notifications directly to `producer_plugin` instead of routing through the `chain_plugin`. I think this makes the code easier to understand.
- Modified `sync_recv_block` to be called from `on_accepted_block` for tracking applied blocks.
- Additional logging of exceptions in controller to compensate for no logging in `net_plugin`.

Partially resolves #529


ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Optimized Block Processing notifying only when new blocks are ready for processing. 
Note:end

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1208,10 +1206,6 @@ chain_apis::read_only chain_plugin::get_read_only_api(const fc::microseconds& ht
 }
 
 
-bool chain_plugin::accept_block(const signed_block_ptr& block, const block_id_type& id, const block_handle& bh ) {
```
linh2931, line 1211: Like removing this indirection. No side effects by the removal?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3714,7 +3727,7 @@ namespace eosio {
                      ("cid", cid)("n", ptr->block_num())("id", id.str().substr(8,16)));
          }
          if( exception || !obh) {
-            if (!obh) {
+            if (!exception && !obh) {
```
linh2931, line 3730: This condition `!exception && !obh` is confusing with the `exception || !obh` above.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3714,7 +3727,7 @@ namespace eosio {
...
                if (prev_is_proper_svnn_block || !ptr->is_proper_svnn_block()) {
                   fc_dlog(logger, "unlinkable_block ${bn} : ${id}, previous ${pn} : ${pid}",
                           ("bn", ptr->block_num())("id", id)("pn", block_header::num_from_id(ptr->previous))("pid", ptr->previous));
```
linh2931, line 3733: This comment is for the `assert(obh);` on line 3743 (I could not put a comment there as it is not changed by the PR). That assert is not necessarily true is `exception` is true.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -888,47 +888,34 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-      // de-dupe here... no point in aborting block if we already know the block
-      if (chain.validated_block_exists(id)) {
-         fc_dlog(_log, "already have validated block ${n} ${id}", ("n", blk_num)("id", id));
-         _time_tracker.add_other_time();
-         return true; // return true because the block was already accepted
-      }
-
-      EOS_ASSERT(block->timestamp < (now + fc::seconds(7)), block_from_the_future, "received a block from the future, ignoring it: ${id}", ("id", id));
-
       // start processing of block
```
linh2931, line 897: Remove this comment, as no block is processed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -888,47 +888,34 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       _time_tracker.clear();
    }
 
-   bool on_incoming_block(const signed_block_ptr& block, const block_id_type& id, const block_handle& bh) {
+   bool on_incoming_block() {
```
linh2931, line 891: Should this be renamed `on_any_incoming_blocks` or add a comment saying the function does not process a specific block.

plugins/chain_plugin/chain_plugin.cpp
heifner: No, `net_plugin` already depends on `producer_plugin`, so it can call it directly.

plugins/net_plugin/net_plugin.cpp
heifner: True. I added a `bool unlinkable` which makes it clearer.

plugins/net_plugin/net_plugin.cpp
heifner: If `exception` is `true` it will return in the `if( exception ` statement.

plugins/producer_plugin/producer_plugin.cpp
heifner: Added a comment.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3728,12 +3730,10 @@ namespace eosio {
...
                      ("cid", cid)("n", ptr->block_num())("id", id.str().substr(8,16)));
          }
-         if( exception || !obh) {
-            if (!exception && !obh) {
-               if (prev_is_proper_svnn_block || !ptr->is_proper_svnn_block()) {
-                  fc_dlog(logger, "unlinkable_block ${bn} : ${id}, previous ${pn} : ${pid}",
-                          ("bn", ptr->block_num())("id", id)("pn", block_header::num_from_id(ptr->previous))("pid", ptr->previous));
-               }
+         if( exception || unlinkable) {
+            if (unlinkable && (prev_is_proper_svnn_block || !ptr->is_proper_svnn_block())) {
```
linh2931, line 3757: Why this condition `&& (prev_is_proper_svnn_block || !ptr->is_proper_svnn_block())`?

plugins/net_plugin/net_plugin.cpp
linh2931: I missed the `return`.

plugins/net_plugin/net_plugin.cpp
heifner: See comment lines 3722-3727 above.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,28 +2415,41 @@ namespace eosio {
...
+      if (c) {
+         c->latest_blk_time = sync_active_time = std::chrono::steady_clock::now(); // reset when we receive a block
+         if (blk_applied)
+            c->block_status_monitor_.accepted();
+         if (blk_latency.count() < config::block_interval_us && c->peer_syncing_from_us) {
+            // a peer will not send us a recent block unless it is synced
+            c->peer_syncing_from_us = false;
+         }
+      } else {
+         sync_active_time = std::chrono::steady_clock::now(); // reset when we apply a block as well
```
greg7mdp, line 2446: why?

greg7mdp (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,28 +2415,41 @@ namespace eosio {
...
       }
-      c->latest_blk_time = sync_active_time = std::chrono::steady_clock::now(); // reset when we receive a block
-      if (blk_applied)
-         c->block_status_monitor_.accepted();
-      if (blk_latency.count() < config::block_interval_us && c->peer_syncing_from_us) {
-         // a peer will not send us a recent block unless it is synced
-         c->peer_syncing_from_us = false;
+      if (c) {
+         c->latest_blk_time = sync_active_time = std::chrono::steady_clock::now(); // reset when we receive a block
+         if (blk_applied)
```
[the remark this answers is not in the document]
heifner: Thanks. Need to fix that.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3736,12 +3753,10 @@ namespace eosio {
...
                      ("cid", cid)("n", ptr->block_num())("id", id.str().substr(8,16)));
          }
-         if( exception || !obh) {
-            if (!obh) {
-               if (prev_is_proper_svnn_block || !ptr->is_proper_svnn_block()) {
-                  fc_dlog(logger, "unlinkable_block ${bn} : ${id}, previous ${pn} : ${pid}",
-                          ("bn", ptr->block_num())("id", id)("pn", block_header::num_from_id(ptr->previous))("pid", ptr->previous));
-               }
+         if( exception || unlinkable) {
+            if (unlinkable && (prev_is_proper_svnn_block || !ptr->is_proper_svnn_block())) {
```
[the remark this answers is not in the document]
heifner: It is initialized to `false`.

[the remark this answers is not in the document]
heifner: Done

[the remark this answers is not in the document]
greg7mdp: That whole optimization about making `prev_is_proper_svnn_block` static and initializing it after is unnecessarily confusing.

Also why not make it non `static` and initialize it in the block `if( exception || unlinkable) {`.

The `is_proper_svnn_block` check is trivial anyways, and the static variable access requiring synchronization is likely more expensive.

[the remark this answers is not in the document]
heifner: How do you know the previous was legacy?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3758,94 +3773,24 @@ namespace eosio {
                   ("bt", best_head)("cid", cid)("num", block_num)("id", obh->id()) );
          my_impl->dispatcher.add_peer_block( obh->id(), cid ); // no need to send back to sender
          my_impl->dispatcher.bcast_block( obh->block(), obh->id() );
+         c->block_status_monitor_.accepted();
 
          if (best_head) {
-            fc_dlog(logger, "posting block ${n} to app thread", ("n", ptr->block_num()));
+            ++c->unique_blocks_rcvd_count;
```
greg7mdp, line 3779: what is the meaning of `unique_blocks_rcvd_count`, i.e. what does it count?

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
heifner: How many blocks this connection received first. Prometheus metric that will be removed soon.

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
heifner: renamed

linh2931 (approved):

linh2931 (approved):

PR #690, opened 2024-09-03
[1.0] Do not allow Legacy blocks to have QC block extensions

linh2931:
Currently `verify_qc_claim` is only called on Savanna blocks. On Legacy blocks, we need to verify they do not have a block QC extension.


libraries/chain/controller.cpp
```diff
@@ -4003,12 +4003,13 @@ struct controller_impl {
...
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
-         verify_qc_claim(id, b, prev);
-      }
+
+      // Verify claim made by finality_extension in block header extension and
+      // quorum_certificate_extension in block extension are valid.
+      // This is the only place the evaluation is done.
+      // Note: the purpose of running verify_qc_claim on Legacy blocks too is to
+      // safe guard Legacy blocks.
```
greg7mdp, line 4011:
```suggestion
      // For legacy blocks without a finality_extension, ensure that they don't include
      // a qc_extension either, and that the previous block doesn't have a finality_extension either.
```

greg7mdp (approved):

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3999,35 @@ struct controller_impl {
...
+   void verify_legacy_block_exts( const signed_block_ptr& b ) {
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
+                  "Legacy block #${b} includes a QC block extension",
+                  ("b", block_num) );
```
heifner, line 4014: Seems like this should verify that that `qc_claim` is as expected.
If present then `qc_claim.is_qc_strong` must be `false`.
If present and prev is present then their `qc_claim` block numbers are equal.
If prev not present then `qc_claim` block number is this block number and contains a `new_finalizer_policy_diff`.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3999,35 @@ struct controller_impl {
...
+      // Verify claim made by finality_extension in block header extension and
+      // quorum_certificate_extension in block extension are valid.
+      // This is the only place the evaluation is done.
+      // Note: the purpose of running verify_qc_claim on Legacy blocks too is to
+      // safe guard Legacy blocks.
       if constexpr (savanna_mode) {
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
          verify_qc_claim(id, b, prev);
```
heifner, line 4028: You can remove the legacy checks from `verify_qc_claim` since only calling for savanna.

heifner (changes requested):

libraries/chain/controller.cpp
heifner: Looking good. Add check for the new_finalizer_policy_diff.

arhag (dismissed):

libraries/chain/controller.cpp
arhag: > Add check for the new_finalizer_policy_diff.

Good call.

If I am not mistaken, the only protection we have against a malformed Savanna Genesis block that does not have a `new_finalizer_policy_diff` present is this `assert`:
https://github.com/AntelopeIO/spring/blob/31e94812b92acf31717ca0a3b11db82b77888773/libraries/chain/block_state.cpp#L71

But that only runs in debug mode. If it wasn't present in a Savanna Genesis block, then I believe in a release build of nodeos we would get undefined behavior in `create_if_genesis_block`.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,58 @@ struct controller_impl {
...
+                        block_validate_exception,
+                        "Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
+                        ("b", block_num) );
+         } else {
+            EOS_ASSERT( f_ext.qc_claim.block_num == block_num,
+                        block_validate_exception,
+                        "Transition block #${b} QC claim block_num not equal to current block_num",
+                        ("b", block_num) );
+         }
+      }
```
arhag, line 4024: If it doesn't have `header_ext`, then we should verify that the `prev` also does not have the finality extension.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,58 @@ struct controller_impl {
...
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
+                  "Legacy block #${b} includes a QC block extension",
+                  ("b", block_num) );
+
```
arhag, line 4002: We should also check that the previous block's header satisfies `is_proper_svnn_block() == false`.

arhag: And for that matter, we could also check that the current block satisfied `is_proper_svnn_block() == false`.

arhag (changes requested):

libraries/chain/controller.cpp
heifner: I think it should also attempt to apply the `new_finalizer_policy_diff` and verify `generation` is 1.
Also verify that the `new_proposer_policy_diff` is null for all `qc_claim`s. The finalizer_policy apply_diff actually needs some checks as well. I'll get a PR up for that.

libraries/chain/controller.cpp
linh2931: I missed the `new_finalizer_policy_diff` check last night. Adding it.

libraries/chain/controller.cpp
linh2931: At the time when `verify_legacy_block_exts()` is called, we only have previous `block_header_state_legacy` available. We don't have access to previous proposed finalizer policy which is needed to apply_diff.

linh2931: But I will do a check on Genesis block and apply a diff to make sure its generation is 1.

heifner: The diff would be against an empty container. Just do this:
```
finalizer_policy{}.apply_diff(std::move(*f_ext.new_finalizer_policy_diff));
```

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,93 @@ struct controller_impl {
...
+                     block_validate_exception,
+                     "Savanna Genesis block #${b} QC claim block_num not equal to current block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( f_ext.new_finalizer_policy_diff,
+                     block_validate_exception,
+                     "Savanna Genesis block #${b} finality block header extension misses new_finalizer_policy_diff",
+                     ("b", block_num) );
+
+         finalizer_policy no_policy;
+         finalizer_policy genesis_policy = no_policy.apply_diff(*f_ext.new_finalizer_policy_diff);
```
heifner, line 4054: Please add a comment that this will FC_ASSERT if invalid.

arhag (changes requested):

linh2931:
Close this because of wrong base. https://github.com/AntelopeIO/spring/pull/697 is the replacement.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,86 @@ struct controller_impl {
...
+   // or Transition block.
+   void verify_legacy_block_exts( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
+                  "Legacy block #${b} includes a QC block extension",
```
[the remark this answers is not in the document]
linh2931: Done in https://github.com/AntelopeIO/spring/pull/697

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,86 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy block #${b} may not be Proper Savnanna block",
```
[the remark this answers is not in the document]
linh2931: Done in https://github.com/AntelopeIO/spring/pull/697

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,86 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy block's #${b} previous block may not be Proper Savnanna block",
```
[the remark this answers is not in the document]
linh2931: Done in https://github.com/AntelopeIO/spring/pull/697

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: No, it is not called for Transition blocks.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: No, it is not called for Transition blocks.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: No, it is not called for Transition blocks.

Issue #691, opened 2024-09-03
Verify important block invariants (specifically in finality and QC extensions) for Legacy and Transition blocks


arhag:
## Background and suggest protocol compatible simplifications to `verify_qc_claim`

`verify_qc_claim` was designed to work whether no matter whether the block it was called on was a Legacy block, Transition block, or Proper Savanna block.

But it turns out the current implementation only calls `verify_qc_claim` on Proper Savanna blocks (in fact it, from a type perspective, it cannot even call it on a Legacy block).

https://github.com/AntelopeIO/spring/blob/21cec1fa496587d825e45862ec1d09c895c4362e/libraries/chain/controller.cpp#L4006-L4011

But consider this part of `verify_qc_claim`: https://github.com/AntelopeIO/spring/blob/21cec1fa496587d825e45862ec1d09c895c4362e/libraries/chain/controller.cpp#L3897-L3916

If `verify_qc_claim` is only called for Proper Savanna blocks, then that code can be simplified to just an `EOS_ASSERT` that checks that `header_ext` is present.

Furthermore, consider this part of `verify_qc_claim`: https://github.com/AntelopeIO/spring/blob/21cec1fa496587d825e45862ec1d09c895c4362e/libraries/chain/controller.cpp#L3928-L3934

If `verify_qc_claim` is only called for Proper Savanna blocks, then that code can also be simplfied to just an `EOS_ASSERT` that checks that `prev_finality_ext` is present.

The code is still safe for blocks other Proper Savanna blocks, especially when considering there is also the later check that happens in `block_header_state::next`: https://github.com/AntelopeIO/spring/blob/21cec1fa496587d825e45862ec1d09c895c4362e/libraries/chain/block_header_state.cpp#L442-L443

But the above simplifications are valuable to make the code clearer.

## Undesired blockchain validation rules in existing Spring implementation

None of the above so far is an actual bug. However, the fact that `verify_qc_claim` is not called for Legacy or Transition blocks means that the current implementation of Spring will accept Legacy blocks (i.e. blocks without a finality block header extension) and Transition blocks (i.e. block with a finality block header extension but that has a `schedule_version` that is not equal to the `proper_svnn_schedule_version` constant) even if they have a QC block extension present. In fact, there are other block invariants that can be violated for Legacy and Transition blocks that the current Spring implementation does not check for but really should.

[...]

PR #695, opened 2024-09-04
[1.0] Harden ordered_diff

heifner:
`ordered_diff` is used to process the finalizer and proposer policy diffs from block header extension `finality_extension`. Since the `finality_extension`, is provided to a node to validate, validate the diff during `apply_diff`.

Related to #691


heifner:
Sorry targeted wrong branch

PR #696, opened 2024-09-04
[1.0] Harden ordered_diff 

heifner:
`ordered_diff` is used to process the finalizer and proposer policy diffs from block header extension `finality_extension`. Since the `finality_extension`, is provided to a node to validate, validate the diff during `apply_diff`.

Related to #691

### DIFF
--- a/libraries/libfc/include/fc/container/ordered_diff.hpp
+++ b/libraries/libfc/include/fc/container/ordered_diff.hpp
@@ -41,51 +41,51 @@ public:
 
       diff_result result;
       while (s < source.size() || t < target.size()) {
+         FC_ASSERT(s < std::numeric_limits<SizeType>::max());
+         FC_ASSERT(t < std::numeric_limits<SizeType>::max());
          if (s < source.size() && t < target.size()) {
-            FC_ASSERT(s < std::numeric_limits<SizeType>::max());
-            FC_ASSERT(t < std::numeric_limits<SizeType>::max());
             if (source[s] == target[t]) {
                // nothing to do, skip over
                ++s;
                ++t;
             } else { // not equal
                if (s == source.size() - 1 && t == target.size() - 1) {
+                  FC_ASSERT(result.remove_indexes.size() < MAX_NUM_ARRAY_ELEMENTS);
+                  FC_ASSERT(result.insert_indexes.size() < MAX_NUM_ARRAY_ELEMENTS);
                   // both at end, insert target and remove source
                   result.remove_indexes.push_back(s);
                   result.insert_indexes.emplace_back(t, target[t]);
-                  FC_ASSERT(result.remove_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
-                  FC_ASSERT(result.insert_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
                   ++s;
                   ++t;
                } else if (s + 1 < source.size() && t + 1 < target.size() && source[s + 1] == target[t + 1]) {
+                  FC_ASSERT(result.remove_indexes.size() < MAX_NUM_ARRAY_ELEMENTS);
+                  FC_ASSERT(result.insert_indexes.size() < MAX_NUM_ARRAY_ELEMENTS);
                   // misalignment, but next value equal, insert and remove
                   result.remove_indexes.push_back(s);
                   result.insert_indexes.emplace_back(t, target[t]);
-                  FC_ASSERT(result.remove_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
-                  FC_ASSERT(result.insert_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
                   ++s;
                   ++t;
                } else if (t + 1 < target.size() && source[s] == target[t + 1]) {
                   // source equals next target, insert current target
+                  FC_ASSERT(result.insert_indexes.size() < MAX_NUM_ARRAY_ELEMENTS);
                   result.insert_indexes.emplace_back(t, target[t]);
-                  FC_ASSERT(result.insert_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
                   ++t;
                } else { // source[s + 1] == target[t]
                   // target matches next source, remove current source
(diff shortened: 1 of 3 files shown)


libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -39,48 +44,56 @@ class ordered_diff {
          if (s < source.size() && t < target.size()) {
             if (source[s] == target[t]) {
                // nothing to do, skip over
-               assert(s <= std::numeric_limits<SizeType>::max());
-               assert(t <= std::numeric_limits<SizeType>::max());
+               FC_ASSERT(s <= std::numeric_limits<SizeType>::max());
+               FC_ASSERT(t <= std::numeric_limits<SizeType>::max());
```
greg7mdp, line 48: Can we move these two `FC_ASSERT`  right after the `while` line instead of repeating them in almost every case?

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -94,16 +107,24 @@ class ordered_diff {
    template <typename X>
    requires std::same_as<std::decay_t<X>, diff_result>
    static Container<T> apply_diff(Container<T>&& container, X&& diff) {
+      FC_ASSERT(diff.remove_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
+      FC_ASSERT(diff.insert_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
+
       // Remove from the source based on diff.remove_indexes
       std::ptrdiff_t offset = 0;
       for (SizeType index : diff.remove_indexes) {
+         FC_ASSERT(index + offset <= container.size(), "diff.remove_indexes index ${idx} + offset ${o} not in range ${s}",
```
greg7mdp, line 116: should be `<` I believe.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -39,48 +44,56 @@ class ordered_diff {
...
+               FC_ASSERT(t <= std::numeric_limits<SizeType>::max());
                ++s;
                ++t;
             } else { // not equal
                if (s == source.size() - 1 && t == target.size() - 1) {
                   // both at end, insert target and remove source
-                  assert(s <= std::numeric_limits<SizeType>::max());
-                  assert(t <= std::numeric_limits<SizeType>::max());
+                  FC_ASSERT(s <= std::numeric_limits<SizeType>::max());
+                  FC_ASSERT(t <= std::numeric_limits<SizeType>::max());
```
greg7mdp, line 55: All these should be `<` since we increment `s` or `t` afterwards.

greg7mdp (changes requested):

libraries/libfc/include/fc/container/ordered_diff.hpp
heifner: `.end()` i.e. `.begin() + 0` is allowed and needed.

greg7mdp: doc of `std::vector::erase`:

The iterator pos must be valid and dereferenceable. Thus the end() iterator (which is valid, but is not dereferenceable) cannot be used as a value for pos.

heifner: Oh, right, I was thinking of insert.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -37,50 +42,50 @@ class ordered_diff {
...
                   result.insert_indexes.emplace_back(t, target[t]);
+                  FC_ASSERT(result.remove_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
+                  FC_ASSERT(result.insert_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
                   ++s;
                   ++t;
                } else if (t + 1 < target.size() && source[s] == target[t + 1]) {
                   // source equals next target, insert current target
-                  assert(t <= std::numeric_limits<SizeType>::max());
                   result.insert_indexes.emplace_back(t, target[t]);
+                  FC_ASSERT(result.insert_indexes.size() <= MAX_NUM_ARRAY_ELEMENTS);
```
greg7mdp, line 71: Issue is that we insert *before* the FC_ASSERT. So the FC_ASSERT can be fine because `result.insert_indexes.size() == MAX_NUM_ARRAY_ELEMENTS`, and the next `emplace_back` will go over.

=> change the `FC_ASSERT` to `<` and move before the `emplace_back`.

This applies to the other locations as well.

greg7mdp (changes requested):

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -94,16 +99,24 @@ class ordered_diff {
...
          container.erase(container.begin() + index + offset);
          --offset;
       }
 
       // Insert into the source based on diff.insert_indexes
       for (auto& [index, value] : diff.insert_indexes) {
+         FC_ASSERT(index <= container.size(), "diff.insert_indexes index ${idx} not in range ${s}",
+                   ("idx", index)("s", container.size()));
          container.insert(container.begin() + index, std::move(value));
+         FC_ASSERT(container.size() <= MAX_NUM_ARRAY_ELEMENTS);
```
greg7mdp, line 119: same here, change to `<` and move before line 118.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -37,50 +42,50 @@ class ordered_diff {
       diff_result result;
       while (s < source.size() || t < target.size()) {
          if (s < source.size() && t < target.size()) {
+            FC_ASSERT(s < std::numeric_limits<SizeType>::max());
+            FC_ASSERT(t < std::numeric_limits<SizeType>::max());
```
greg7mdp, line 46: Please move these two *before* the `if`, otherwise we skip the increments at line 84 and 89.

heifner:
Replaced by #698 which targets updated branch.

PR #697, opened 2024-09-04
[1.0] Verify finality and QC extensions in Legacy and Transition blocks

linh2931:
Currently verify_qc_claim is only called on Savanna blocks. On Legacy and Transitions blocks, we need to verify finality and QC extensions too.

Resolves https://github.com/AntelopeIO/spring/issues/691

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4046,7 +4046,7 @@ struct controller_impl {
          const auto& prev_finality_ext = std::get<finality_extension>(it->second);
          EOS_ASSERT( f_ext.qc_claim.block_num == prev_finality_ext.qc_claim.block_num,
                      invalid_qc_claim,
-                     "Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
+                     "Non Genesis Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
                      ("b", block_num) );
          EOS_ASSERT( !f_ext.new_finalizer_policy_diff, invalid_qc_claim,
                      "Non Genesis Transition block #${b} finality block header extension may not have new_finalizer_policy_diff",
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4061,10 +4061,10 @@ struct controller_impl {
    // thread safe, expected to be called from thread other than the main thread
    template<typename ForkDB, typename BS>
    block_handle create_block_state_i( ForkDB& forkdb, const block_id_type& id, const signed_block_ptr& b, const BS& prev ) {
-      constexpr bool prev_is_of_type_block_state = std::is_same_v<typename std::decay_t<BS>, block_state>;
-      assert(prev_is_of_type_block_state == b->is_proper_svnn_block());
+      constexpr bool is_proper_savanna_block = std::is_same_v<typename std::decay_t<BS>, block_state>;
+      assert(is_proper_savanna_block == b->is_proper_svnn_block());
 
-      if constexpr (prev_is_of_type_block_state) {
+      if constexpr (is_proper_savanna_block) {
          EOS_ASSERT( b->is_proper_svnn_block(), block_validate_exception,
                      "create_block_state_i cannot be called on block #${b} which is not a Proper Savanna block unless the prev block state provided is of type block_state",
                      ("b", b->block_num()) );
@@ -4081,7 +4081,7 @@ struct controller_impl {
          }
       }
 
-      auto trx_mroot = calculate_trx_merkle( b->transactions, prev_is_of_type_block_state );
+      auto trx_mroot = calculate_trx_merkle( b->transactions, is_proper_savanna_block );
       EOS_ASSERT( b->transaction_mroot == trx_mroot,
                   block_validate_exception,
                   "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
@@ -4101,14 +4101,14 @@ struct controller_impl {
       EOS_ASSERT( id == bsp->id(), block_validate_exception,
                   "provided id ${id} does not match block id ${bid}", ("id", id)("bid", bsp->id()) );
 
-      if constexpr (prev_is_of_type_block_state) {
+      if constexpr (is_proper_savanna_block) {
          integrate_received_qc_to_block(bsp); // Save the received QC as soon as possible, no matter whether the block itself is valid or not
          consider_voting(bsp, use_thread_pool_t::no);
       }
 
       if (!should_terminate(bsp->block_num())) {
          forkdb.add(bsp, ignore_duplicate_t::yes);
-         if constexpr (prev_is_of_type_block_state)
+         if constexpr (is_proper_savanna_block)
             vote_processor.notify_new_block(async_aggregation);
       }
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3894,26 +3894,13 @@ struct controller_impl {
       bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
       uint32_t block_num = b->block_num();
 
-      if( !header_ext ) {
-         // If there is no header extension, ensure the block does not have a QC and also the previous
-         // block doesn't have a header extension either. Then return early.
-         // ------------------------------------------------------------------------------------------
-         EOS_ASSERT( !qc_extension_present,
-                     invalid_qc_claim,
-                     "Block #${b} includes a QC block extension, but doesn't have a finality header extension",
-                     ("b", block_num) );
+      // This function is called only in Savanna. Finality block header
+      // extension must exist
+      EOS_ASSERT( header_ext,
+                  invalid_qc_claim,
+                  "Block #${b} doesn't have a finality header extension",
+                  ("b", block_num) );
 
-         EOS_ASSERT( !prev_finality_ext,
-                     invalid_qc_claim,
-                     "Block #${b} doesn't have a finality header extension even though its predecessor does.",
-                     ("b", block_num) );
-
-         dlog("received block: #${bn} ${t} ${prod} ${id}, no qc claim, previous: ${p}",
-              ("bn", block_num)("t", b->timestamp)("prod", b->producer)("id", id)("p", b->previous));
-         return;
-      }
-
-      assert(header_ext);
       const auto& f_ext        = std::get<finality_extension>(*header_ext);
       const auto  new_qc_claim = f_ext.qc_claim;
 
@@ -3999,15 +3986,94 @@ struct controller_impl {
       bsp->verify_qc(qc_proof);
    }
 
+   // Verify QC block extension and finality block header extension on a Legacy
+   // or Transition block.
+   void verify_legacy_block_exts( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
+                  "Legacy block #${b} includes a QC block extension",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy block #${b} may not be Proper Savnanna block",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy block's #${b} previous block may not be Proper Savnanna block",
+                  ("b", block_num) );
+
+      auto f_ext_id = finality_extension::extension_id();
+      std::optional<block_header_extension> header_ext = b->extract_header_extension(f_ext_id);
+
+      if (!header_ext) {
+         auto it = prev.header_exts.find(finality_extension::extension_id());
+         EOS_ASSERT( it == prev.header_exts.end(),
+                     block_validate_exception,
+                     "Block #${b} does not have finality block header extension but its previous block does have",
+                     ("b", block_num) );
+
+         return;
+      }
+
+      // Transition Block
+      const auto& f_ext = std::get<finality_extension>(*header_ext);
+      EOS_ASSERT( !f_ext.qc_claim.is_strong_qc,
+                  block_validate_exception,
+                  "Transition block #${b} has a strong QC claim",
+                  ("b", block_num) );
+      EOS_ASSERT( !f_ext.new_proposer_policy_diff,
+                  block_validate_exception,
+                  "Transition block #${b} has new_proposer_policy_diff",
+                  ("b", block_num) );
+
+      if (auto it = prev.header_exts.find(finality_extension::extension_id()); it != prev.header_exts.end()) {
+         const auto& prev_finality_ext = std::get<finality_extension>(it->second);
+         EOS_ASSERT( f_ext.qc_claim.block_num == prev_finality_ext.qc_claim.block_num,
+                     block_validate_exception,
+                     "Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
+                     ("b", block_num) );
+      } else {
+         // Savanna Genesis Block
+         EOS_ASSERT( f_ext.qc_claim.block_num == block_num,
+                     block_validate_exception,
+                     "Savanna Genesis block #${b} QC claim block_num not equal to current block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( f_ext.new_finalizer_policy_diff,
+                     block_validate_exception,
+                     "Savanna Genesis block #${b} finality block header extension misses new_finalizer_policy_diff",
+                     ("b", block_num) );
+
+         finalizer_policy no_policy;
+         // apply_diff will FC_ASSERT if new_finalizer_policy_diff is malformated
+         finalizer_policy genesis_policy = no_policy.apply_diff(*f_ext.new_finalizer_policy_diff);
+         EOS_ASSERT( genesis_policy.generation == 1,
+                     block_validate_exception,
+                     "Savanna Genesis block #${b} finalizer policy generation (${g}) not 1",
+                     ("b", block_num)("g", genesis_policy.generation) );
+      }
+   }
+
    // thread safe, expected to be called from thread other than the main thread
    template<typename ForkDB, typename BS>
    block_handle create_block_state_i( ForkDB& forkdb, const block_id_type& id, const signed_block_ptr& b, const BS& prev ) {
       constexpr bool savanna_mode = std::is_same_v<typename std::decay_t<BS>, block_state>;
+
+      // Verify claim made by finality_extension in block header extension and
+      // quorum_certificate_extension in block extension are valid.
+      // This is the only place the evaluation is done.
+      // Note: the purpose of running verify_qc_claim on Legacy blocks too is to
+      // safe guard Legacy blocks.
       if constexpr (savanna_mode) {
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
          verify_qc_claim(id, b, prev);
+      } else {
+         verify_legacy_block_exts(b, prev);
       }
 
       auto trx_mroot = calculate_trx_merkle( b->transactions, savanna_mode );
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3908,17 +3908,14 @@ struct controller_impl {
            ("bn", block_num)("t", b->timestamp)("prod", b->producer)("id", id)
            ("qc", new_qc_claim)("p", b->previous));
 
-      // If there is a header extension, but the previous block does not have a header extension,
-      // ensure the block does not have a QC and the QC claim of the current block has a block_num
-      // of the current block’s number and that it is a claim of a weak QC. Then return early.
+      // The only time a block should have a finality block header extension but
+      // its parent block does not, is if it is a Savanna Genesis block (which is
+      // necessarily a Transition block). Since verify_qc_claim will not be called
+      // on Transition blocks, prev_finality_ext should always be present
       // -------------------------------------------------------------------------------------------------
-      if (!prev_finality_ext) {
-         EOS_ASSERT( !qc_extension_present && new_qc_claim.block_num == block_num && new_qc_claim.is_strong_qc == false,
-                     invalid_qc_claim,
-                     "Block #${b}, which is the finality transition block, doesn't have the expected extensions",
-                     ("b", block_num) );
-         return;
-      }
+      EOS_ASSERT( prev_finality_ext, invalid_qc_claim,
+                  "Previous block of Block #${b} doesn't have finality block header extension",
+                  ("b", block_num) );
 
       // at this point both current block and its parent have IF extensions, and we are past the
       // IF transition block
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3882,7 +3882,7 @@ struct controller_impl {
    // and quorum_certificate_extension in block extension are valid.
    // Called from net-threads. It is thread safe as signed_block is never modified after creation.
    // -----------------------------------------------------------------------------
-   void verify_qc_claim( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
+   void verify_proper_block_exts( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
       auto qc_ext_id = quorum_certificate_extension::extension_id();
       auto f_ext_id  = finality_extension::extension_id();
 
@@ -3910,7 +3910,7 @@ struct controller_impl {
 
       // The only time a block should have a finality block header extension but
       // its parent block does not, is if it is a Savanna Genesis block (which is
-      // necessarily a Transition block). Since verify_qc_claim will not be called
+      // necessarily a Transition block). Since verify_proper_block_exts will not be called
       // on Transition blocks, prev_finality_ext should always be present
       // -------------------------------------------------------------------------------------------------
       EOS_ASSERT( prev_finality_ext, invalid_qc_claim,
@@ -4101,7 +4101,7 @@ struct controller_impl {
          EOS_ASSERT( b->is_proper_svnn_block(), block_validate_exception,
                      "create_block_state_i cannot be called on block #${b} which is not a Savanna block while its parent is a Savanna block",
                      ("b", b->block_num()) );
-         verify_qc_claim(id, b, prev);
+         verify_proper_block_exts(id, b, prev);
       } else {
          EOS_ASSERT( !b->is_proper_svnn_block(), block_validate_exception,
                      "create_block_state_i cannot be called on block #${b} which is a Savanna block while its parent is not a Savanna block",
@@ -4199,7 +4199,7 @@ struct controller_impl {
       return fork_db.apply<std::optional<block_handle>>(unlinkable, f);
    }
 
-   // thread safe, QC already verified by verify_qc_claim
+   // thread safe, QC already verified by verify_proper_block_exts
    void integrate_received_qc_to_block(const block_state_ptr& bsp_in) {
       // extract QC from block extension
       assert(bsp_in->block);
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4001,13 +4001,13 @@ struct controller_impl {
                   "Legacy block #${b} has invalid schedule_version",
                   ("b", block_num) );
 
-      EOS_ASSERT( !prev.header.is_proper_svnn_block(), invalid_qc_claim,
-                  "Legacy block #${b} may not have previous block that is a Proper Savanna block",
-                  ("b", block_num) );
-
+      // Verify we don't go back from Savanna (Transition or Proper) block to Legacy block
       auto it = prev.header_exts.find(finality_extension::extension_id());
       EOS_ASSERT( it == prev.header_exts.end(), invalid_qc_claim,
                   "Legacy block #${b} has previous block which contains finality block header extension",
+                  ("b", block_num) );
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(), invalid_qc_claim,
+                  "Legacy block #${b} may not have previous block that is a Proper Savanna block",
                   ("b", block_num) );
    }
 
@@ -4043,6 +4043,7 @@ struct controller_impl {
                   ("b", block_num) );
 
       if (auto it = prev.header_exts.find(finality_extension::extension_id()); it != prev.header_exts.end()) {
+         // Transition block other than Genesis Block
          const auto& prev_finality_ext = std::get<finality_extension>(it->second);
          EOS_ASSERT( f_ext.qc_claim.block_num == prev_finality_ext.qc_claim.block_num,
                      invalid_qc_claim,
@@ -4098,9 +4099,6 @@ struct controller_impl {
       // A block that has a Transition block as a parent must either be a Transition
       // block or a Proper Savanna block.
 
-      // Verify claim made by finality_extension in block header extension and
-      // quorum_certificate_extension in block extension are valid.
-      // This is the only place the evaluation is done.
       if constexpr (is_proper_savanna_block) {
          EOS_ASSERT( b->is_proper_svnn_block(), block_validate_exception,
                      "create_block_state_i cannot be called on block #${b} which is not a Savanna block while its parent is a Savanna block",
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3891,19 +3891,19 @@ struct controller_impl {
       // extract current block extension and previous header extension
       auto block_exts = b->validate_and_extract_extensions();
       const finality_extension* prev_finality_ext = prev.header_extension<finality_extension>();
-      std::optional<block_header_extension> header_ext  = b->extract_header_extension(f_ext_id);
+      std::optional<block_header_extension> finality_ext  = b->extract_header_extension(f_ext_id);
 
       bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
       uint32_t block_num = b->block_num();
 
       // This function is called only in Savanna. Finality block header
       // extension must exist
-      EOS_ASSERT( header_ext,
-                  invalid_qc_claim,
-                  "Block #${b} doesn't have a finality header extension",
+      EOS_ASSERT( finality_ext, block_validate_exception,
+                  "Proper Savanna block #${b} does not have a finality header extension",
                   ("b", block_num) );
 
-      const auto& f_ext        = std::get<finality_extension>(*header_ext);
+      assert(finality_ext);
+      const auto& f_ext        = std::get<finality_extension>(*finality_ext);
       const auto  new_qc_claim = f_ext.qc_claim;
 
       dlog("received block: #${bn} ${t} ${prod} ${id}, qc claim: ${qc}, previous: ${p}",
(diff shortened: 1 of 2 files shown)


libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
+   // or Transition block.
+   void verify_legacy_block_exts( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
+                  "Legacy or Transition block #${b} includes a QC block extension",
```
heifner, line 3999: This is not called for `Transition blocks`.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy or Transition block #${b} has invalid schedule_version",
```
heifner, line 4004: This is not called for `Transition block`.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy or Transition block #${b} may not have previous block that is a Proper Savanna block",
```
heifner, line 4009: This is not called for `Transition block`.

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,94 @@ struct controller_impl {
...
+                  block_validate_exception,
+                  "Transition block #${b} has new_proposer_policy_diff",
+                  ("b", block_num) );
+
+      if (auto it = prev.header_exts.find(finality_extension::extension_id()); it != prev.header_exts.end()) {
+         const auto& prev_finality_ext = std::get<finality_extension>(it->second);
+         EOS_ASSERT( f_ext.qc_claim.block_num == prev_finality_ext.qc_claim.block_num,
+                     block_validate_exception,
+                     "Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
+                     ("b", block_num) );
```
greg7mdp, line 4042: I thing we should assert that `!f_ext.new_finalizer_policy_diff` here

greg7mdp (changes requested):

libraries/chain/controller.cpp
linh2931: The function is called before a block becomes proposer Savanna block. It can be a transition block.

heifner: Yes, sorry. My mistake.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
+      // Verify claim made by finality_extension in block header extension and
+      // quorum_certificate_extension in block extension are valid.
+      // This is the only place the evaluation is done.
+      // Note: the purpose of running verify_qc_claim on Legacy blocks too is to
+      // safe guard Legacy blocks.
       if constexpr (savanna_mode) {
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
          verify_qc_claim(id, b, prev);
```
arhag, line 4075: If `savanna_mode` is true, then it means that the parent block of block `b` has a `block_state`.

For a block to have a `block_state` it means the block must be a Savanna block, which means either Transition block or a Proper Savanna block.

A block that has a Savanna block as a parent must also be a Savanna block. A block that has a Proper Savanna block as a parent must also be a Proper Savanna block.

So given that the parent block has a `block_state`, we know that block `b` must either be a Transition block or a Proper Savanna block.

Perhaps there are some other restrictions in the caller of `create_block_state_i` that guarantee that block `b` will never be a Transition block. But that is not obvious purely from the context in this function.

Regardless, `verify_qc_claim` (as it is currently implemented) is appropriate to call on any Savanna block (including Transition blocks) that has a Savanna block as a parent.

@linh2931: Maybe you could put a comment like the above here to make these clearer for the reader. This will especially be useful when we later refactor how these validations work.

By the way, if we are sure that `create_block_state_i` will never be called with a Transition block `b` and with a `prev` that decays to type `block_state`, then we should assert that prior to the call to `verify_qc_claim`. The assert could simply check that `b` is a Proper Savanna block by ensuring `b->is_proper_svnn_block() == true`.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
+   // Verify QC block extension and finality block header extension on a Legacy
+   // or Transition block.
+   void verify_legacy_block_exts( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
```
heifner, line 3998: I think all of these in this method should be `invalid_qc_claim`. Also I think it would be more readable to put the `invalid_qc_claim` on the same line as the first arg (styles vary here obviously, but seems that is more common if it fits).

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
+      // This is the only place the evaluation is done.
+      // Note: the purpose of running verify_qc_claim on Legacy blocks too is to
+      // safe guard Legacy blocks.
       if constexpr (savanna_mode) {
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
          verify_qc_claim(id, b, prev);
+      } else {
+         verify_legacy_block_exts(b, prev);
```
arhag, line 4077: If `savanna_mode` is false, then it means that the parent block of block `b` has a `block_state_legacy`.

For a block to have a `block_state_legacy` it means the block must either be a Legacy block or a Transition block.

A block that has a Legacy block as a parent must either be a Legacy block or a Transition block. 

A block that has a Transition block as a parent must either be a Transition block or a Proper Savanna block.

So `savanna_mode == false` does not, by itself, tell us anything about whether block `b` is a Legacy block, Transition block, or Proper Savanna block.

We do not wish to call `verify_legacy_block_exts` (which should really be renamed) on Proper Savanna blocks. So we should check if `b` is a Proper Savanna block to avoid making the function call.

Ideally, the caller of `create_block_state_i` never makes a call with a Proper Savanna block `b` and a `prev` that does not decay to type `block_state`. It is a programming error to do so. So we could add an `assert` here (or an `EOS_ASSERT` if we want to be extra cautious) that ensures that `b->is_proper_svnn_block() == false`.

arhag (changes requested):

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,94 @@ struct controller_impl {
...
       }
```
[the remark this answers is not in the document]
linh2931: I think we should keep `verify_qc_claim` as is, because it does check QC claims.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,94 @@ struct controller_impl {
...
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present,
+                  block_validate_exception,
+                  "Legacy block #${b} includes a QC block extension",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(),
+                  block_validate_exception,
+                  "Legacy block #${b} may not be Proper Savnanna block",
```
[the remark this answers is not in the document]
linh2931: It was removed due to rewording of the error message.

libraries/chain/controller.cpp
linh2931: In transition blocks other than Genesis block, it can be null if policy has not changed.

greg7mdp: I think you misread what I suggested. I think we should verify that it is **always null** in transition blocks other than the genesis block.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,95 @@ struct controller_impl {
...
       if constexpr (savanna_mode) {
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
          verify_qc_claim(id, b, prev);
+      } else {
+         verify_legacy_block_exts(b, prev);
       }
 
       auto trx_mroot = calculate_trx_merkle( b->transactions, savanna_mode );
```
arhag, line 4080: I don't think we actually care about `savanna_mode` here.

What we really care about is if `b` is a Proper Savanna block or not. We can determine that using `b->is_proper_svnn_block()`.

And if we add the asserts I mentioned above earlier, we would have confidence that this is doing the same thing as passing in `savanna_mode`.

heifner: `is_proper_svnn_block()` is better. This was before we had that method.

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,86 @@ struct controller_impl {
...
+      uint32_t block_num = b->block_num();
+
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+      EOS_ASSERT( !qc_extension_present, invalid_qc_claim,
+                  "Legacy or Transition block #${b} includes a QC block extension",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(), invalid_qc_claim,
```
greg7mdp, line 4001: wrong exception type `invalid_qc_claim` used here and in some following checks.

greg7mdp (changes requested):

libraries/chain/controller.cpp
greg7mdp: I still go back to having two functions with the same name:

```
// called for all non-proper blocks received
verify_block_extensions(const signed_block_ptr& b, const block_state_legacy& prev);

// called for all proper blocks received
verify_block_extensions(const signed_block_ptr& b, const block_state& prev);
```

and just call:

```
   block_handle create_block_state_i( ForkDB& forkdb, const block_id_type& id, const signed_block_ptr& b, const BS& prev ) {
      verify_block_extensions(b, prev);
...
```

libraries/chain/controller.cpp
arhag: Yes, we should verify it is always null for transition blocks other than the genesis block.

libraries/chain/controller.cpp
arhag: We would still need to special case the other `if constexpr` below that integrates a QC and considers voting. These only make sense with a `block_state`.

Not to mention that the future optimizations that delay the actual verification of the attached QC to the chain threads.

arhag: However, we could at least rename `savanna_mode` to `is_proper_savanna_block` and add the assert at the top:

```cpp
assert(is_proper_savanna_block == b->is_proper_svnn_block());
```

libraries/chain/controller.cpp
```diff
@@ -3999,15 +3986,99 @@ struct controller_impl {
...
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
          verify_qc_claim(id, b, prev);
+      } else {
+         EOS_ASSERT( !b->is_proper_svnn_block(), block_validate_exception,
+                     "create_block_state_i cannot be called on block #${b} which is not a Proper Savanna block while its parent is a Legacy block or a Transition block",
+                     ("b", b->block_num()) );
+
+         verify_legacy_and_transition_block_exts_common(b, prev);
```
arhag, line 4075: I suggest removing this and duplicating the checks in `verify_legacy_block_exts` and `verify_transition_block_exts`. That way you can be more specific in the error messages.

libraries/chain/controller.cpp
linh2931: I used `block_validate_exception` for some of them. Kevin suggested using `invalid_qc_claim` for all the checks here. It makes sense as they are all related QC.

libraries/chain/controller.cpp
linh2931: My bad. I missed this. Adding it now.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3983,138 @@ struct controller_impl {
...
+      // block or a Proper Savanna block.
+
+      // Verify claim made by finality_extension in block header extension and
+      // quorum_certificate_extension in block extension are valid.
+      // This is the only place the evaluation is done.
+      if constexpr (is_proper_savanna_block) {
+         EOS_ASSERT( b->is_proper_svnn_block(), block_validate_exception,
+                     "create_block_state_i cannot be called on block #${b} which is not a Savanna block while its parent is a Savanna block",
+                     ("b", b->block_num()) );
          verify_qc_claim(id, b, prev);
```
greg7mdp, line 4104: rename `verify_proper_block_exts`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+      //
+      // A block that has a Legacy block as a parent must either be a Legacy block
+      // or a Transition block.
+      //
+      // A block that has a Transition block as a parent must either be a Transition
+      // block or a Proper Savanna block.
+
+      if constexpr (is_proper_savanna_block) {
+         EOS_ASSERT( b->is_proper_svnn_block(), block_validate_exception,
+                     "create_block_state_i cannot be called on block #${b} which is not a Savanna block while its parent is a Savanna block",
```
arhag, line 4104: Comment should be:

`"create_block_state_i cannot be called on block #${b} which is not a Proper Savanna block unless the prev block state provided is of type block_state"`

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                     ("b", b->block_num()) );
+         verify_proper_block_exts(id, b, prev);
+      } else {
+         EOS_ASSERT( !b->is_proper_svnn_block(), block_validate_exception,
+                     "create_block_state_i cannot be called on block #${b} which is a Savanna block while its parent is not a Savanna block",
```
arhag, line 4109: Comment should be:

`"create_block_state_i cannot be called on block #${b} which is a Proper Savanna block unless the prev block state provided is of type block_state_legacy"`

libraries/chain/controller.cpp
arhag: I agree with Greg. 

I think this particular `EOS_ASSERT` should use `block_validate_exception` since it does not actually have to do with QC claims in the block.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  "Legacy block #${b} includes a QC block extension",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(), invalid_qc_claim,
+                  "Legacy block #${b} has invalid schedule_version",
+                  ("b", block_num) );
+
+      // Verify we don't go back from Savanna (Transition or Proper) block to Legacy block
+      auto it = prev.header_exts.find(finality_extension::extension_id());
+      EOS_ASSERT( it == prev.header_exts.end(), invalid_qc_claim,
```
arhag, line 4006: This should also use `block_validate_exception`.

The assert could have also instead just checked `prev.header.is_legacy_block()`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  "Legacy block #${b} has previous block which contains finality block header extension",
+                  ("b", block_num) );
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(), invalid_qc_claim,
```
arhag, line 4009: This check is redundant. So it should be removed.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+      uint32_t block_num = b->block_num();
+      auto block_exts = b->validate_and_extract_extensions();
+      auto qc_ext_id = quorum_certificate_extension::extension_id();
+      bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
+
+      EOS_ASSERT( !qc_extension_present, invalid_qc_claim,
+                  "Transition block #${b} includes a QC block extension",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(), invalid_qc_claim,
```
arhag, line 4026: This check is redundant given it is part of the definition of the Transition block and therefore must be true before the caller of this function decides it is even appropriate to pass block `b` to `verify_transition_block_exts`. So it should be removed.

But to detect programmer errors in DEBUG mode, it is important to make the modification I noted to the `assert` above:

```cpp
assert(!b->is_legacy_block() && !b->is_proper_svnn_block());
```

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  "Transition block #${b} has invalid schedule_version",
+                  ("b", block_num) );
+
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(), invalid_qc_claim,
```
arhag, line 4030: This should also use `block_validate_exception`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  "Transition block #${b} may not have previous block that is a Proper Savanna block",
+                  ("b", block_num) );
+
+      auto f_ext_id = finality_extension::extension_id();
+      std::optional<block_header_extension> finality_ext = b->extract_header_extension(f_ext_id);
+      const auto& f_ext = std::get<finality_extension>(*finality_ext);
```
arhag, line 4036: Even though it is implied by the earlier `assert` that guarantees (in DEBUG mode) that `b->is_legacy_block() == false`, it would be better to also `assert` here that finality header extension is actually present before you try to dereference.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      EOS_ASSERT( !b->is_proper_svnn_block(), invalid_qc_claim,
+                  "Legacy block #${b} has invalid schedule_version",
+                  ("b", block_num) );
+
+      // Verify we don't go back from Savanna (Transition or Proper) block to Legacy block
+      auto it = prev.header_exts.find(finality_extension::extension_id());
+      EOS_ASSERT( it == prev.header_exts.end(), invalid_qc_claim,
+                  "Legacy block #${b} has previous block which contains finality block header extension",
```
arhag, line 4007: Change comment to:

`"Legacy block #${b} must have previous block that is also a Legacy block"`

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+
+      EOS_ASSERT( !prev.header.is_proper_svnn_block(), invalid_qc_claim,
+                  "Transition block #${b} may not have previous block that is a Proper Savanna block",
+                  ("b", block_num) );
+
+      auto f_ext_id = finality_extension::extension_id();
+      std::optional<block_header_extension> finality_ext = b->extract_header_extension(f_ext_id);
+      const auto& f_ext = std::get<finality_extension>(*finality_ext);
+
+      EOS_ASSERT( !f_ext.qc_claim.is_strong_qc, invalid_qc_claim,
```
arhag, line 4038: I think this check should be removed from here. Instead we can have other more appropriate checks on the strong/weak status of the QC claim in the appropriate branches of the if statement below.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  "Transition block #${b} has a strong QC claim",
+                  ("b", block_num) );
+      EOS_ASSERT( !f_ext.new_proposer_policy_diff, invalid_qc_claim,
+                  "Transition block #${b} has new_proposer_policy_diff",
+                  ("b", block_num) );
+
+      if (auto it = prev.header_exts.find(finality_extension::extension_id()); it != prev.header_exts.end()) {
```
arhag, line 4045: This if condition is equivalent to `!prev.header.is_legacy_block()`. But I'm saying that just for clarity (and to make it clear how it maps to the conditions to check listed in https://github.com/AntelopeIO/spring/issues/691). I'm not suggesting to change the code because this approach also gives you `it` which you need to get the `prev_finality_ext`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+         // Transition block other than Genesis Block
+         const auto& prev_finality_ext = std::get<finality_extension>(it->second);
+         EOS_ASSERT( f_ext.qc_claim.block_num == prev_finality_ext.qc_claim.block_num,
```
arhag, line 4048: This assert should be changed to ensure `f_ext.qc_claim == prev_finality_ext.qc_claim`. And if they are not, the error message should provide the `qc_claim` values of both.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                     invalid_qc_claim,
+                     "Non Genesis Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( !f_ext.new_finalizer_policy_diff, invalid_qc_claim,
+                     "Non Genesis Transition block #${b} finality block header extension may not have new_finalizer_policy_diff",
+                     ("b", block_num) );
+      } else {
+         // Savanna Genesis Block
+         EOS_ASSERT( f_ext.qc_claim.block_num == block_num, invalid_qc_claim,
+                     "Savanna Genesis block #${b} QC claim block_num not equal to current block_num",
```
arhag, line 4058: Check that `f_ext.qc_claim` is equivalent to `qc_claim{block_num, false}`. If it is not, the error message should provide the value of `f_ext.qc_claim`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      auto f_ext_id = finality_extension::extension_id();
+      std::optional<block_header_extension> finality_ext = b->extract_header_extension(f_ext_id);
+      const auto& f_ext = std::get<finality_extension>(*finality_ext);
+
+      EOS_ASSERT( !f_ext.qc_claim.is_strong_qc, invalid_qc_claim,
+                  "Transition block #${b} has a strong QC claim",
+                  ("b", block_num) );
+      EOS_ASSERT( !f_ext.new_proposer_policy_diff, invalid_qc_claim,
```
arhag, line 4041: This should also use `block_validate_exception`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                  ("b", block_num) );
+
+      if (auto it = prev.header_exts.find(finality_extension::extension_id()); it != prev.header_exts.end()) {
+         // Transition block other than Genesis Block
+         const auto& prev_finality_ext = std::get<finality_extension>(it->second);
+         EOS_ASSERT( f_ext.qc_claim.block_num == prev_finality_ext.qc_claim.block_num,
+                     invalid_qc_claim,
+                     "Non Genesis Transition block #${b} QC claim block_num not equal to previous QC claim block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( !f_ext.new_finalizer_policy_diff, invalid_qc_claim,
```
arhag, line 4052: This should also use `block_validate_exception`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                     "Non Genesis Transition block #${b} finality block header extension may not have new_finalizer_policy_diff",
+                     ("b", block_num) );
+      } else {
+         // Savanna Genesis Block
+         EOS_ASSERT( f_ext.qc_claim.block_num == block_num, invalid_qc_claim,
+                     "Savanna Genesis block #${b} QC claim block_num not equal to current block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( f_ext.new_finalizer_policy_diff, invalid_qc_claim,
```
arhag, line 4060: This should also use `block_validate_exception`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+                     "Savanna Genesis block #${b} finality block header extension misses new_finalizer_policy_diff",
+                     ("b", block_num) );
+
+         finalizer_policy no_policy;
+         // apply_diff will FC_ASSERT if new_finalizer_policy_diff is malformated
+         finalizer_policy genesis_policy = no_policy.apply_diff(*f_ext.new_finalizer_policy_diff);
+         EOS_ASSERT( genesis_policy.generation == 1, invalid_qc_claim,
```
arhag, line 4067: This should also use `block_validate_exception`.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+         EOS_ASSERT( f_ext.qc_claim.block_num == block_num, invalid_qc_claim,
+                     "Savanna Genesis block #${b} QC claim block_num not equal to current block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( f_ext.new_finalizer_policy_diff, invalid_qc_claim,
+                     "Savanna Genesis block #${b} finality block header extension misses new_finalizer_policy_diff",
+                     ("b", block_num) );
+
+         finalizer_policy no_policy;
+         // apply_diff will FC_ASSERT if new_finalizer_policy_diff is malformated
+         finalizer_policy genesis_policy = no_policy.apply_diff(*f_ext.new_finalizer_policy_diff);
```
arhag, line 4066: Maybe we should wrap this in one of those try catch macros where it catches the exception and rethrows it as a `block_validate_exception` but includes some extra log information to make it clear that it was a bad genesis diff.

libraries/chain/controller.cpp
```diff
@@ -3894,44 +3896,28 @@ struct controller_impl {
...
-                     ("b", block_num) );
-
-         dlog("received block: #${bn} ${t} ${prod} ${id}, no qc claim, previous: ${p}",
-              ("bn", block_num)("t", b->timestamp)("prod", b->producer)("id", id)("p", b->previous));
-         return;
-      }
+      // This function is called only in Savanna. Finality block header
+      // extension must exist
+      EOS_ASSERT( header_ext,
+                  invalid_qc_claim,
```
arhag, line 3902: This should also use `block_validate_exception`.

Also perhaps we should also rename `header_ext` to `finality_ext` here?

libraries/chain/controller.cpp
```diff
@@ -3894,44 +3896,28 @@ struct controller_impl {
...
+                  "Block #${b} doesn't have a finality header extension",
```
arhag, line 3903: Change message to:

`"Proper Savanna block #${b} does not have a finality header extension"`

libraries/chain/controller.cpp
```diff
@@ -3894,44 +3896,28 @@ struct controller_impl {
...
+      // on Transition blocks, prev_finality_ext should always be present
       // -------------------------------------------------------------------------------------------------
-      if (!prev_finality_ext) {
-         EOS_ASSERT( !qc_extension_present && new_qc_claim.block_num == block_num && new_qc_claim.is_strong_qc == false,
-                     invalid_qc_claim,
-                     "Block #${b}, which is the finality transition block, doesn't have the expected extensions",
-                     ("b", block_num) );
-         return;
-      }
+      EOS_ASSERT( prev_finality_ext, invalid_qc_claim,
```
arhag, line 3918: This should also use `block_validate_exception`.

The assert could have also instead just checked `!prev.header.is_legacy_block()`.

libraries/chain/controller.cpp
```diff
@@ -3894,44 +3896,28 @@ struct controller_impl {
...
+                  "Previous block of Block #${b} doesn't have finality block header extension",
```
arhag, line 3919: Change message to:

`"Proper Savanna block #${b} may not have previous block that is a Legacy block"`

arhag (changes requested):

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3985,138 @@ struct controller_impl {
...
+         // Savanna Genesis Block
+         EOS_ASSERT( f_ext.qc_claim.block_num == block_num, invalid_qc_claim,
+                     "Savanna Genesis block #${b} QC claim block_num not equal to current block_num",
+                     ("b", block_num) );
+         EOS_ASSERT( f_ext.new_finalizer_policy_diff, invalid_qc_claim,
+                     "Savanna Genesis block #${b} finality block header extension misses new_finalizer_policy_diff",
+                     ("b", block_num) );
+
+         finalizer_policy no_policy;
+         // apply_diff will FC_ASSERT if new_finalizer_policy_diff is malformated
```
heifner, line 4065: malformed

libraries/chain/controller.cpp
heifner: Something like:
```cpp
         // apply_diff will FC_ASSERT if new_finalizer_policy_diff is malformated
         finalizer_policy genesis_policy;
         try {
            finalizer_policy no_policy;
            genesis_policy = no_policy.apply_diff(*f_ext.new_finalizer_policy_diff);
         } EOS_RETHROW_EXCEPTIONS(block_validate_exception, "applying diff of Savanna Genesis Block")
```

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3984,113 @@ struct controller_impl {
...
    // thread safe, expected to be called from thread other than the main thread
    template<typename ForkDB, typename BS>
    block_handle create_block_state_i( ForkDB& forkdb, const block_id_type& id, const signed_block_ptr& b, const BS& prev ) {
-      constexpr bool savanna_mode = std::is_same_v<typename std::decay_t<BS>, block_state>;
-      if constexpr (savanna_mode) {
-         // Verify claim made by finality_extension in block header extension and
-         // quorum_certificate_extension in block extension are valid.
-         // This is the only place the evaluation is done.
-         verify_qc_claim(id, b, prev);
+      constexpr bool is_proper_savanna_block = std::is_same_v<typename std::decay_t<BS>, block_state>;
```
greg7mdp, line 4071: I think `prev_is_of_type_block_state` would be a better name for this boolean (what we actually check, not what we think it implies), but no need to delay this PR for this.

libraries/chain/controller.cpp
linh2931: I missed Kevin's suggested code earlier.

libraries/chain/controller.cpp
```diff
@@ -3999,18 +3984,113 @@ struct controller_impl {
...
+                     ("b", block_num)("qc_claim", f_ext.qc_claim) );
+         EOS_ASSERT( f_ext.new_finalizer_policy_diff, block_validate_exception,
+                     "Savanna Genesis block #${b} finality block header extension misses new_finalizer_policy_diff",
+                     ("b", block_num) );
+
+         // apply_diff will FC_ASSERT if new_finalizer_policy_diff is malformed
+         try {
+            finalizer_policy no_policy;
+            finalizer_policy genesis_policy = no_policy.apply_diff(*f_ext.new_finalizer_policy_diff);
+            EOS_ASSERT( genesis_policy.generation == 1, block_validate_exception,
```
greg7mdp, line 4061: I believe this `EOS_ASSERT` should be after the try/catch block.

libraries/chain/controller.cpp
```diff
@@ -3913,17 +3913,16 @@ struct controller_impl {
...
       // -------------------------------------------------------------------------------------------------
-      EOS_ASSERT( prev_finality_ext, invalid_qc_claim,
-                  "Previous block of Block #${b} doesn't have finality block header extension",
+      EOS_ASSERT( !prev.header.is_legacy_block(), block_validate_exception,
+                  "Proper Savanna block #${b} may not have previous block that is a Legacy block",
                   ("b", block_num) );
 
       // at this point both current block and its parent have IF extensions, and we are past the
       // IF transition block
       // ----------------------------------------------------------------------------------------
```
arhag, line 3924: The above comment can be removed.

greg7mdp (approved):

libraries/chain/controller.cpp
```diff
@@ -4061,9 +4061,7 @@ struct controller_impl {
             EOS_ASSERT( genesis_policy.generation == 1, block_validate_exception,
                         "Savanna Genesis block #${b} finalizer policy generation (${g}) not 1",
                         ("b", block_num)("g", genesis_policy.generation) );
-         } catch( ... ) {
-            EOS_THROW(block_validate_exception, "Failed to apply Genesis block finalizer_policy_diff");
-         }
+         } EOS_RETHROW_EXCEPTIONS(block_validate_exception, "applying diff of Savanna Genesis Bloc")
```
greg7mdp, line 4064: block.

libraries/chain/controller.cpp
```diff
@@ -3993,80 +3992,78 @@ struct controller_impl {
       auto qc_ext_id = quorum_certificate_extension::extension_id();
       bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
 
-      EOS_ASSERT( !qc_extension_present, invalid_qc_claim,
+      EOS_ASSERT( !qc_extension_present, block_validate_exception,
```
arhag, line 3995: For consistency, this should either remain `invalid_qc_claim` or we should change similar checks involving the integrity of the QC block extension in `verify_proper_block_exts`.

libraries/chain/controller.cpp
heifner: You have to declare `genesis_policy` outside the try then. I think it is fine as is.

libraries/chain/controller.cpp
```diff
@@ -3993,80 +3992,78 @@ struct controller_impl {
...
-      assert(!b->is_legacy_block());
+      assert(!b->is_legacy_block() && !b->is_proper_svnn_block());
 
       uint32_t block_num = b->block_num();
       auto block_exts = b->validate_and_extract_extensions();
       auto qc_ext_id = quorum_certificate_extension::extension_id();
       bool qc_extension_present = block_exts.count(qc_ext_id) != 0;
 
-      EOS_ASSERT( !qc_extension_present, invalid_qc_claim,
+      EOS_ASSERT( !qc_extension_present, block_validate_exception,
```
arhag, line 4017: For consistency, this should either remain `invalid_qc_claim` or we should change similar checks involving the integrity of the QC block extension in `verify_proper_block_exts`.

heifner (approved):

libraries/chain/controller.cpp
linh2931: If apply_diff does not throw, we will execute this EOS_ASSERT.

arhag (changes requested):

libraries/chain/controller.cpp
arhag: The assert after it means it represents what it says. It is a programming error to call this function in a way that does not respect that semantic, which is why we have the assert.

libraries/chain/controller.cpp
linh2931: For now I reverted back. I will clean up exception types in `verify_proper_block_exts` in future PR.

arhag (approved):

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Refactor code to verify finality and QC extensions on Legacy and Transitions blocks, in addition to Savanna blocks.
Note:end

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: .
Note:end

ericpassmore:
Note:start
category: Other
component: Internal
summary: Refactor code to verify finality and QC extensions on Legacy and Transitions blocks, in addition to Savanna blocks.
Note:end

PR #699, opened 2024-09-04
[1.0 -> main] Test accept_block_header, accept_block, and irreversible_block signals

heifner:
Add verification to libtester that `controller` signals `accept_block_header`, `accept_block`, `irreversible_block` happen in order and do not repeat.

Working on #619 and wanted a test to verify signals continue to work the same as before.

Merges `release/1.0` into `main` including #681


greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Add a test against controller to ensure accept_block_header, accept_block, irreversible_block are applied in the correct order.
Note:end

PR #700, opened 2024-09-04
[1.0.1] Verify block signature first when creating block_state

heifner:
When creating a `block_state`, verify the block signature before ingesting anything from the block.


greg7mdp:
Shouldn't we verify the block signature even before we start validating the block header extensions?

heifner:
> Shouldn't we verify the block signature even before we start validating the block header extensions? First thing in `create_block_state_i`?

I think in general we should consider verification to be allowed in parallel. The point of this PR is that we complete block signature verification before integrating the block. The order of any verification before integration shouldn't matter.

greg7mdp:
> > Shouldn't we verify the block signature even before we start validating the block header extensions? First thing in `create_block_state_i`?
> 
> I think in general we should consider verification to be allowed in parallel. The point of this PR is that we complete block signature verification before integrating the block. The order of any verification before integration shouldn't matter.

In general I agree, except that our extension/qc verification code itself could have bugs causing nodeos to abort. It seems safer to be to validate signatures first (old reliable code), to ensure that the block header extensions that we verify next were not tampered with.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: When creating a `block_state`, verify the block signature before ingesting anything from the block..
Note:end

libraries/chain/block_state.cpp
```diff
@@ -9,20 +9,66 @@
...
+
+} // namespace detail
+
+using namespace eosio::chain::detail;
+
+// ASSUMPTION FROM controller_impl::apply_block = all untrusted blocks will have their signatures pre-validated here
 block_state::block_state(const block_header_state& prev, signed_block_ptr b, const protocol_feature_set& pfs,
                          const validator_t& validator, bool skip_validate_signee)
-   : block_header_state(prev.next(*b, validator))
+   : block_header_state(verify_block_sig(prev, b, skip_validate_signee) ? prev.next(*b, validator) : block_header_state{})
```
greg7mdp, line 66: `verify_block_sig` always returns true, so the ternary operator is weird. I think the following is clearer maybe:
```suggestion
   : block_header_state(verify_block_sig(prev, b, skip_validate_signee), prev.next(*b, validator))
```

greg7mdp (approved):

libraries/chain/block_state.cpp
heifner: I tried that first, but compiler doesn't like it:
```
/home/heifner/ext/leap/libraries/chain/block_state.cpp:66:25: error: no viable conversion from 'bool' to 'block_id_type' (aka 'fc::sha256')
   : block_header_state(verify_block_sig(prev, b, skip_validate_signee), prev.next(*b, validator))
```
However, an extra set of `()` seems to do the trick.

linh2931 (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: When creating a block_state, verify the block signature before ingesting anything from the block.
Note:end

PR #704, opened 2024-09-05
[1.0.1] P2P: Sync timeout

### LINKED ISSUE #682: P2P: Attemping to sync from peer that sent go_away
`received go_away_message, reason = some other non-fatal condition, possibly unknown block`
Spring syncing from Leap. Leap does not have new feature of peer reporting first block available. Regardless, it should mark a connection so that on re-connect it is not chosen as a peer to sync from. Looks like my syncing peer keeps trying to sync from this peer because it has the lowest ping time, but it does not have the blocks needed to sync.

Somewhat related, I noticed that `peer_ping_time_ns` is not reset on close. Should have the following in `_close()`.
```
      peer_ping_time_ns = std::numeric_limits<uint64_t>::max();
```

heifner:
- Reset `peer_ping_time_ns` on closing a connection. This peer ping time is used to determine which peers to sync from.
- Update `sync_active_time` on `accepted_block` signal
  - It is possible for a sync to timeout a connection because it had not recently received a block. However, with a large `sync-fetch-span` and/or a slow node and/or large blocks to process, the connection can timeout because it had not received a block recently due to the main thread being busy applying blocks. Reset the `sync_active_time` on each applied block so that the `net_plugin` knows the blocks are still being processed and there is no reason to give up on the connection.

Note: `should_sync_from` already took into account the `no_retry` `go_away_reason`.

Resolves #682

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2224,7 +2224,7 @@ namespace eosio {
    // thread safe
    bool sync_manager::sync_recently_active() const {
       auto time_since_active = std::chrono::steady_clock::now() - sync_active_time.load();
-      bool active = std::chrono::steady_clock::now() - sync_active_time.load() < my_impl->resp_expected_period;
+      bool active = time_since_active < my_impl->resp_expected_period;
       if (!active) {
          fc_dlog(logger, "sync has not received a block recently, time since sync block ${t}ms",
                  ("t", std::chrono::duration_cast<std::chrono::milliseconds>(time_since_active).count()));
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2226,7 +2226,7 @@ namespace eosio {
       auto time_since_active = std::chrono::steady_clock::now() - sync_active_time.load();
       bool active = time_since_active < my_impl->resp_expected_period;
       if (!active) {
-         fc_dlog(logger, "sync has not received a block recently, time since sync block ${t}ms",
+         fc_dlog(logger, "sync not recently active, time since last sync block ${t}ms",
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2222,7 +2223,13 @@ namespace eosio {
 
    // thread safe
    bool sync_manager::sync_recently_active() const {
-      return std::chrono::steady_clock::now() - sync_active_time.load() < my_impl->resp_expected_period;
+      auto time_since_active = std::chrono::steady_clock::now() - sync_active_time.load();
+      bool active = std::chrono::steady_clock::now() - sync_active_time.load() < my_impl->resp_expected_period;
```
greg7mdp, line 2227:
```suggestion
      bool active = time_since_active < my_impl->resp_expected_period;
```

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2222,7 +2223,13 @@ namespace eosio {
 
    // thread safe
    bool sync_manager::sync_recently_active() const {
-      return std::chrono::steady_clock::now() - sync_active_time.load() < my_impl->resp_expected_period;
+      auto time_since_active = std::chrono::steady_clock::now() - sync_active_time.load();
+      bool active = time_since_active < my_impl->resp_expected_period;
+      if (!active) {
+         fc_dlog(logger, "sync has not received a block recently, time since sync block ${t}ms",
```
linh2931, line 2229: `time since sync block`: Would it be more clear `time since last synced block`?

linh2931 (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Other
summary: Improve syncing so that capability of peer in addition to latency is taken into account.
Note:end

PR #706, opened 2024-09-05
[1.0.1 -> main] P2P: Sync timeout

### LINKED ISSUE #682: P2P: Attemping to sync from peer that sent go_away

heifner:
- Reset `peer_ping_time_ns` on closing a connection. This peer ping time is used to determine which peers to sync from.
- Update `sync_active_time` on `accepted_block` signal
  - It is possible for a sync to timeout a connection because it had not recently received a block. However, with a large `sync-fetch-span` and/or a slow node and/or large blocks to process, the connection can timeout because it had not received a block recently due to the main thread being busy applying blocks. Reset the `sync_active_time` on each applied block so that the `net_plugin` knows the blocks are still being processed and there is no reason to give up on the connection.

Note: `should_sync_from` already took into account the `no_retry` `go_away_reason`.

Merges `release/1.0` into `main` including #704 

Resolves #682

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ nodeos --full-version
 ```
 You should see a [semantic version](https://semver.org) string followed by a `git` commit hash with no errors. For example:
 ```
-v3.1.2-0b64f879e3ebe2e4df09d2e62f1fc164cc1125d1
+v1.0.1-9026a03c09c9b4f93edca696b5eef259f0ab96b3
 ```
 
 ## Build and Install from Source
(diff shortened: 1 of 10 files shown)


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
summary: Improve syncing so that capability of peer in addition to latency is taken into account.
Note:end

PR #713, opened 2024-09-06
[1.0.1] Fix pause production on missed round

### LINKED ISSUE #712: Production pause for one block after prior BP misses round
If the producer before your producer missed their entire production round it causes you to pause production and miss producing your second block. I think the issue there is we produce the first block because we are not paused, but the vote is not received on that block from the vote thread pool by the time we start the second block. Remember we start the second block immediately. So we think we should pause. No recent votes from producer. Also no recent votes from others as there was no blocks produced for 6 seconds.

image

heifner:
Change the implicit pause production due to votes algorithm:

Track four timestamps: 
- producer vote `latest_vote`, `first_block_after_vote`
- other votes `latest_vote`, `first_block_after_vote`

All default to value of the minimum representable time.

See new `production_pause_vote_tracker` for how these timestamps are tracked and processed.

Includes a test to verify that producer is not paused if the producer before it misses their round.
Example failure of test before fix: https://github.com/AntelopeIO/spring/actions/runs/10739300625/job/29785359524?pr=713

Also make `production_pause_vote_timeout.py` less flaky by having the center node connect to the peers instead of the peers connecting to the center node. When the center node is restarted you want it to as quickly as possible reconnect to its peers instead of waiting for the peers connection timeout for them to reconnect to the center node.

Resolves #712


greg7mdp (approved):

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -80,7 +80,7 @@ const static uint16_t   default_max_inline_action_depth              = 4;
 const static uint16_t   default_max_auth_depth                       = 6;
 const static uint32_t   default_sig_cpu_bill_pct                     = 50 * percent_1; // billable percentage of signature recovery
 const static uint32_t   default_produce_block_offset_ms              = 450;
-const static uint32_t   default_production_pause_vote_timeout_ms     = 6u*1000u; // 6 seconds
+const static uint32_t   default_production_pause_vote_timeout_ms     = 13u*1000u; // 13 seconds
```
linh2931, line 83: Maybe add a comment here why 13 is chosen

linh2931 (approved):

ericpassmore:
Note:start
category: Other
summary: Change the default implicit pause of production due to lack of votes from 6 seconds to 13 seconds.
Note:end

libraries/chain/include/eosio/chain/config.hpp
heifner: Updated to new algorithm. Reverted back to 6 seconds.

arhag (changes requested):
Still reviewing. Please don't merge yet.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1098,9 +1195,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    void resume() {
       _pause_production = false;
       // reset vote received so production can be explicitly resumed, will pause again when received vote time limit hit again
-      auto now = fc::time_point::now();
-      _last_other_vote_received.store(now, std::memory_order_relaxed);
-      _last_producer_vote_received.store(now, std::memory_order_relaxed);
+      if (_is_savanna_active)
+         _implicit_pause_vote_tracker.activate();
```
arhag, line 1199: Here you would instead make the call:

```cpp
_implicit_pause_vote_tracker.force_unpause();
```

You may still wish to guard it with `if(_is_savanna_active)` as an optimization, but it would be safe to call it regardless of whether Savanna was active or not.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -642,52 +842,28 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+      chain.get_mutable_subjective_billing().on_block(_log, block, now);
    }
 
-   void on_block_header(const signed_block_ptr& block) {
+   void on_accepted_block_header(const signed_block_ptr& block) {
       if (!block->is_proper_svnn_block()) {
          if (_producers.contains(block->producer))
             _producer_watermarks.consider_new_watermark(block->producer, block->block_num(), block->timestamp);
+      } else {
+         _implicit_pause_vote_tracker.on_accepted_block_header();
```
arhag, line 866: I think this should still first call the `producer_plugin_impl::update_active_finalizers();` function which should be returned back.

And then after that you would instead make the call:

```cpp
_implicit_pause_vote_tracker.record_received_block(fc::time_point::now());
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -698,89 +874,10 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-      // need a vote within timeout of last accepted block
-      if (_is_producer_active_finalizer &&
-          _accepted_block_time - last_producer_vote_received > _production_pause_vote_timeout) {
-         return implicit_pause::prod_paused;
-      }
-      if (_other_active_finalizers &&
-          _accepted_block_time - last_other_vote_received    > _production_pause_vote_timeout) {
-         return implicit_pause::other_paused;
+         if (_is_savanna_active) {
+            _implicit_pause_vote_tracker.activate();
```
arhag, line 878: There would be no need for this in the simplified `production_pause_vote_tracker` class I am proposing. In fact, `activate` and `init` member functions (among others) would go away entirely.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1986,11 +2083,9 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
       _pending_block_mode = pending_block_mode::speculating;
       not_producing_when_time = true;
-   } else if (implicit_pause p = implicitly_paused(); p != implicit_pause::not_paused) {
-      std::string reason = p == implicit_pause::prod_paused ? "producer votes" : "votes received";
-      fc_elog(_log, "Not producing block because no recent ${r}, block ${t}, last producer vote ${pv}, other votes ${ov}, last accepted block time ${bt}",
-              ("r", std::move(reason))("t", block_time)("pv", _last_producer_vote_received.load(std::memory_order_relaxed))
-              ("ov", _last_other_vote_received.load(std::memory_order_relaxed))("bt", _accepted_block_time));
+   } else if (auto p = _implicit_pause_vote_tracker.implicitly_paused(); p != production_pause_vote_tracker::implicit_pause::not_paused) {
+      // log 'Not producing block because ...'
+      _implicit_pause_vote_tracker.log_not_producing(p, block_time);
```
arhag, line 2088: This would be replaced with:

```cpp
   } else if (const auto status = _implicit_pause_vote_tracker.check_pause_status(now); status.should_pause()) {
      assert(status.earliest_conflicting_block_received_time.has_value());
      const auto conflict_block_time = status.earliest_conflicting_block_received_time.value();

      // "Not producing block because no recent" log message looked for in production_pause_vote_timeout.py
      std::string reason = (status.latest_other_vote < conflict_block_time) 
                           ? ((status.latest_producer_vote < conflict_block_time) 
                              ? "votes"
                              : "votes received from others") 
                           : "votes received from configured producer";
      fc_elog(_log, "Not producing block because no recent ${r}, block ${t}, last producer vote ${pv}, other votes ${ov}, "
                    "earliest conflicting block time ${cb}",
              ("r", std::move(reason))("t", block_time)
              ("pv", status.latest_producer_vote_received_time)
              ("ov", status.latest_other_vote_received_time)
              ("cb", conflict_block_time));
      // ...
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -642,52 +842,28 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-            }
-            if (!_other_active_finalizers) {
-               _other_active_finalizers = std::ranges::any_of(fin_policy->finalizers, [&](const auto& f) {
-                  return !_producers.contains(to_account_name_safe(f.description));
-               });
-            }
-         }
-      }
+   bool implicitly_paused() const {
+      return _implicit_pause_vote_tracker.implicitly_paused() != production_pause_vote_tracker::implicit_pause::not_paused;
```
arhag, line 846: Here you would instead return:

```cpp
return _implicit_pause_vote_tracker.check_pause_status(fc::time_point::now()).should_pause();
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +360,213 @@ struct block_time_tracker {
...
+   // are there any active finalizers that are not part of producers of this node (not expected to be false outside tests)
+   bool                                              _other_active_finalizers = false;
+   alignas(hardware_destructive_interference_sz)
+   std::atomic<bool>                                 _active = false;
+   std::atomic<fc::time_point>                       _last_producer_vote_time;
+   std::atomic<fc::time_point>                       _last_other_vote_time;
+   std::atomic<fc::time_point>                       _first_accepted_block_time_since_last_producer_vote;
+   std::atomic<fc::time_point>                       _first_accepted_block_time_since_last_other_vote;
+   fc::microseconds                                  _production_pause_vote_timeout;
+};
```
arhag, line 568: My review comments offer an alternative structure that I believe is simpler to reason through, but assume an alternative `production_pause_vote_tracker` class which permits an interface to record when blocks and votes and received and query the pause status according to the tracker.

Particularly, the class does not concern itself with matters like what is the definition of a producer vote vs an other vote. And it has no access to the `controller`.

The interface would look like the following:

```cpp
class production_pause_vote_tracker
{
public:
    // By default start with maximum vote_timeout so that production pause mechanism is effectively disabled.
    explicit production_pause_vote_tracker(fc::microseconds block_acceptance_tolerance = fc::milliseconds(250),
                                           fc::microseconds vote_timeout = fc::microseconds::maximum());

    void set_vote_timeout(fc::microseconds vote_timeout);
    void set_block_acceptance_tolerance(fc::microseconds block_acceptance_tolerance);

    // The member functions below can be called concurrently with respect to
    // themselves and one another, but not generally with respect to the functions above.

    bool record_received_other_vote(fc::time_point vote_received_time);
    bool record_received_producer_vote(fc::time_point vote_received_time);
    bool record_received_block(fc::time_point block_received_time);
    void force_unpause();

    struct pause_status
    {
        fc::time_point                latest_other_vote_received_time;
        fc::time_point                latest_producer_vote_received_time;
        std::optional<fc::time_point> earliest_conflicting_block_received_time;

        bool should_pause() const { return earliest_conflicting_block_received_time.has_value(); }
    };

    pause_status check_pause_status(fc::time_point current_time) const;

private:
    // member variables omitted
};
```

I omitted the actual implementation above for clarity. But I have actually put one together already. You can find it here: https://gist.github.com/arhag/cbf7bbe16b1cea43c419cdcb01d23646

arhag (changes requested):

tests/production_pause_vote_timeout.py
```diff
@@ -196,6 +198,29 @@
...
+    node0.kill(signal.SIGTERM)
+    assert not node0.verifyAlive(), "node0 did not shutdown"
+    
+    Print("Verify defproducerb does not pause and produces all blocks of its round")
+    assert node1.waitForHeadToAdvance(), "node1 paused after finalizercNode was shutdown"
+    assert node1.processUrllibRequest("producer", "paused", returnType=ReturnType.raw) == b'false', "node1 paused after node0 was shutdown"
+
+    # wait for c to make sure b is not currently producing
+    node1.waitForProducer("defproducerc", exitOnError=True)
+    # wait for b now to make sure a should have produced before it
```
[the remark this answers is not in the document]
heifner: Yes, `a` especially.

plugins/producer_plugin/producer_plugin.cpp
heifner: The tracking of active finalizers is rather expensive, seems like we should avoid tracking them on validation nodes. I've wrapped your `production_pause_vote_trakcer` in `implicit_production_pause_vote_tracker`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -517,11 +721,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    bool                                              _disable_subjective_p2p_billing              = true;
    bool                                              _disable_subjective_api_billing              = true;
    fc::time_point                                    _irreversible_block_time;
-   fc::time_point                                    _accepted_block_time;
    bool                                              _is_savanna_active                           = false;
-   bool                                              _is_producer_active_finalizer                = false;
-   // are there any active finalizers that are not part of producers of this node (not expected to be false outside tests)
-   bool                                              _other_active_finalizers                     = false;
```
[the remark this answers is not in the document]
heifner: The tracking of active finalizers is rather expensive, seems like we should avoid tracking them on validation nodes. I've wrapped your `production_pause_vote_trakcer` in `implicit_production_pause_vote_tracker`. Also add a `pause_check` enum to `check_pause_status` since which `vote_timing` to check depends on active finalizers.

plugins/producer_plugin/producer_plugin.cpp
heifner: Thanks. I like that this can be tested independly.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,181 @@ struct block_time_tracker {
...
+         return;
+      // disable implicit pause for eosio
+      if (_producers.contains(eosio::chain::config::system_account_name))
+         return;
+      // no producers configured so there is no possible production to pause
+      if (_producers.empty())
+         return;
+
+      _active = true;
+   }
```
arhag, line 392: I don't think the `activate` function is necessary and it would be better to remove it.

But if we did want to keep it for some reason, the implementation should be:

```cpp
void activate() {
   const auto vtm = get_vote_track_mode();

   if (vtm == vote_track_mode::inactive) {
      // Change to `vote_track_mode::all_votes` for now so that a future call to
      // `update_active_finalizers` has an opportunity to set it correctly going forward.
      _vote_track_mode.store(vote_track_mode::all_votes, std::memory_order_relaxed);
   }
   // Otherwise:
   // It is disabled, in which case we do not want to activate it.
   // Or it is already activated, in which case we do not want to change the vote_track_mode.
}
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1098,9 +1171,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    void resume() {
       _pause_production = false;
       // reset vote received so production can be explicitly resumed, will pause again when received vote time limit hit again
-      auto now = fc::time_point::now();
-      _last_other_vote_received.store(now, std::memory_order_relaxed);
-      _last_producer_vote_received.store(now, std::memory_order_relaxed);
+      if (_is_savanna_active)
+         _implicit_pause_vote_tracker.force_unpause();
```
arhag, line 1175: Guarding this with an if statement is unnecessary (though also it does no harm to keep it as is).

arhag (changes requested):

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+   // Can be called concurrently with all member functions except:
+   //    + set_vote_timeout
+   //    + check_pause_status
+   //
+   // Sets the vote timeout duration which drives the production pause status.
+   //
+   // Pre-condition: vote_timeout should not be negative.
+   void set_vote_timeout(fc::microseconds vote_timeout) {
+      if (vote_timeout.count() > 0) {
```
arhag, line 28: I missed this earlier but this should be `>=`.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,229 @@
...
+   // main thread while votes can be signaled off the main thread. This allows votes to be signaled for a block before
+   // the block is signaled. We could track last accepted block id and correlate that with the vote signal, but simpler
+   // to add a signal_tolerance. Normally the tolerance only needs to be a few milliseconds, but no real harm in making
+   // it larger. Half a block interval is a nice value.
+   //
+   // Pre-condition: block_acceptance_tolerance should not be negative.
+   void set_block_acceptance_tolerance(fc::microseconds block_acceptance_tolerance) {
+      assert(block_acceptance_tolerance.count() >= 0);
+
+      if (block_acceptance_tolerance.count() > 0) {
```
arhag, line 49: I missed this earlier but this should be `>=`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,181 @@ struct block_time_tracker {
...
+                  return _producers.contains(to_account_name_safe(f.description));
+               });
+            }
+            if (!_other_active_finalizers) {
+               _other_active_finalizers = std::ranges::any_of(fin_policy->finalizers, [&](const auto& f) {
+                  return !_producers.contains(to_account_name_safe(f.description));
+               });
+            }
+         }
+      }
```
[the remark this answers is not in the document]
heifner: It can become inactive as well. Reflected in my changes.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,181 @@ struct block_time_tracker {
...
+ */
+struct implicit_production_pause_vote_tracker {
+
+   void init(const controller& chain,
+             const std::set<chain::account_name>& producers,
+             fc::microseconds production_pause_vote_timeout)
+   {
+      _chain = &chain;
+      _producers = producers; // copy, normally contains at most one
+      _vt.set_vote_timeout(production_pause_vote_timeout);
```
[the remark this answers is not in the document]
heifner: `inactive` as the comment says seems better.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -698,89 +844,16 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-      if (_is_producer_active_finalizer &&
-          _accepted_block_time - last_producer_vote_received > _production_pause_vote_timeout) {
-         return implicit_pause::prod_paused;
-      }
-      if (_other_active_finalizers &&
-          _accepted_block_time - last_other_vote_received    > _production_pause_vote_timeout) {
-         return implicit_pause::other_paused;
-      }
-      return implicit_pause::not_paused;
+      _implicit_pause_vote_tracker.on_vote(connection_id, status, msg, active_auth, pending_auth);
```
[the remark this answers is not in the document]
heifner: Routing it through `producer_plugin_impl` like the other signals seems cleaner to me.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Why? `>=` seems correct to me. Why set it to `-0` ?

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Why? `>=` seems correct to me. Why set it to `-0` ?

arhag: It wouldn't change the value in state if `block_acceptance_tolerance` is 0. 0 is a valid value to pass in to the function.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
arhag: It wouldn't change the value in state if `vote_timeout` is 0.

We could argue that 0 is not a sensible value to pass in to the function. If we want to go that route instead, then we should update the comment for the pre-condition to say vote_timeout should be positive.

plugins/producer_plugin/producer_plugin.cpp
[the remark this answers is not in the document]
arhag: When could both `finalizer_policy_contains_configured_producer` and `finalizer_policy_contains_other_producer` be false?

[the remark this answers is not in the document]
heifner: Oh right, good point.

arhag (approved):

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+         old_votes         = old_other_vote | old_producer_vote,
+      };
+
+      received_times other_vote{};
+      received_times producer_vote{};
+
+      pause_reason reason() const {
+         unsigned int r = static_cast<unsigned int>(pause_reason::not_paused);
+
+         if (other_vote.earliest_conflict_block.has_value()) {
```
linh2931, line 123: Can those `static_cast` be avoided? You can check if both other_vote and producer_vote have values first and set `r` to `old_votes`, and do the rest of checks. But no need to change.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+         if (!s.first_block_after_vote.has_value())
+            return;
+
+         if (*s.first_block_after_vote < threshold_time) {
+            rt.earliest_conflict_block = *s.first_block_after_vote;
+         }
+      };
+
+      if (check != pause_check::producer) {
```
linh2931, line 174: Rather than using negative (`!=`) comparison, you can do it explicitly
 ```
if (both) {
   process vote for other;
   process vote timing for producer;
} else if (producer) {
   process vote for producer;
} else if (other) {
   process vote for other;
} else {
   assert;
}
```

This way you don't ned complicated asserts and code seems more clearer.

File: plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp [hunk 19]
```diff
@@ -0,0 +1,260 @@
...
+         auto vote_time = latest_vote.load(std::memory_order_relaxed);
+
+         const auto adjusted_block_received_time = fc::time_point(block_received_time).safe_add(block_acceptance_delta);
+
+         if (adjusted_block_received_time <= vote_time)
+            return false;
+
+         auto block_orig = first_block_after_vote.load(std::memory_order_relaxed);
+
+         if ((vote_time < block_orig) && (block_orig <= block_received_time))
```
linh2931, line 214: Why this condition? Can you add a comment?

linh2931 (commented):
Have not gone through producer_plugin changes yet.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,196 @@ struct block_time_tracker {
...
+         assert(false); // should never happen
+      }
+
+      if (!active_auth && !pending_auth) {
+         fc_elog(_log, "vote signal contains no valid authority ${m}", ("m", msg));
+         return;
+      }
+
+      const auto now = fc::time_point::now();
+      account_name auth_desc = active_auth ? to_account_name_safe(active_auth->description)
```
linh2931, line 462: Probably `auth_desc` should be called `finalizer` or `finalizer_name`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,196 @@ struct block_time_tracker {
...
+         check = production_pause_vote_tracker::pause_check::both;
+         break;
+      }
+
+      return _vt.check_pause_status(now, check);
+   }
+
+   // called from multiple threads
+   void on_vote(uint32_t connection_id, vote_result_t status, const vote_message_ptr& msg,
+                const finalizer_authority_ptr& active_auth, const finalizer_authority_ptr& pending_auth)
```
linh2931, line 438: Probably `active_auth` should be named as `active_finalizer` and`pending_auth` as `pending_finalizer`.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+namespace eosio {
+
+class production_pause_vote_tracker {
+public:
+   // By default, start with maximum vote_timeout so that production pause mechanism is effectively disabled.
+   // Pre-conditions: vote_timeout and block_acceptance_tolerance should both not be negative.
+   explicit production_pause_vote_tracker(fc::microseconds block_acceptance_tolerance = fc::milliseconds(250),
+                                          fc::microseconds vote_timeout               = fc::microseconds::maximum())
+      : negative_vote_timeout(), block_acceptance_delta(), latest_other_vote(), latest_producer_vote() {
+      set_vote_timeout(vote_timeout);
```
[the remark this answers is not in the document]
heifner: Added assert to `set_vote_timeout`, so both are covered by `assert` in the setters.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+   //    + set_vote_timeout
+   //    + check_pause_status
+   //
+   // Sets the vote timeout duration which drives the production pause status.
+   //
+   // Pre-condition: vote_timeout should be positive.
+   void set_vote_timeout(fc::microseconds vote_timeout) {
+      if (vote_timeout.count() > 0) {
+         negative_vote_timeout = fc::microseconds(-vote_timeout.count());
+      }
```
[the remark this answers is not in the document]
heifner: Added an assert

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+   // main thread while votes can be signaled off the main thread. This allows votes to be signaled for a block before
+   // the block is signaled. We could track last accepted block id and correlate that with the vote signal, but simpler
+   // to add a signal_tolerance. Normally the tolerance only needs to be a few milliseconds, but no real harm in making
+   // it larger. Half a block interval is a nice value.
+   //
+   // Pre-condition: block_acceptance_tolerance should not be negative.
+   void set_block_acceptance_tolerance(fc::microseconds block_acceptance_tolerance) {
+      assert(block_acceptance_tolerance.count() >= 0);
+
+      if (block_acceptance_tolerance.count() >= 0) {
```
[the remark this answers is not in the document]
heifner: No, but neither does 1 microsecond. Seems ok to allow it.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+   }
+
+   // Can be called concurrently with all member functions except:
+   //    + set_block_acceptance_tolerance
+   //    + record_received_block
+   //
+   // Sets the block acceptance tolerance duration which determines the relevancy of blocks to record.
+   //
+   // votes can be signaled before accepted_block_header, accepted_block_header is currently only signaled from the
+   // main thread while votes can be signaled off the main thread. This allows votes to be signaled for a block before
```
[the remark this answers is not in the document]
heifner: Changed to `from other threads`

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+         old_other_vote    = 0b01,
+         old_producer_vote = 0b10,
+         old_votes         = old_other_vote | old_producer_vote,
+      };
+
+      received_times other_vote{};
+      received_times producer_vote{};
+
+      pause_reason reason() const {
+         unsigned int r = static_cast<unsigned int>(pause_reason::not_paused);
```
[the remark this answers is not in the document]
heifner: Seems simple enough to keep it all in the header.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -0,0 +1,260 @@
...
+   // This function loses track of any previously recorded blocks.
+   void force_unpause() {
+      latest_other_vote.reset_received_blocks();
+      latest_producer_vote.reset_received_blocks();
+   }
+
+   struct pause_status {
+      struct received_times {
+         fc::time_point                latest_vote{};
+         std::optional<fc::time_point> earliest_conflict_block{}; // If present, must be greater than latest_vote.
```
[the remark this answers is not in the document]
heifner: Added comment

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Updated to use if/else instead of `|=`

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Changed to switch.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Added some comments

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,196 @@ struct block_time_tracker {
...
+
+      // Can be disabled either by configuration (if production_pause_vote_timeout.count() == 0)
+      // or because there can never be any possible production to pause (if producers.empty()).
+      // If it is not disabled for either of those two reasons, then change the
+      // initial vote_track_mode from its default value of `vote_track_mode::disabled`
+      // to `vote_track_mode::inactive` so that a future call to `update_active_finalizers`
+      // has an opportunity to set it correctly going forward.
+      if (!producers.empty() && production_pause_vote_timeout.count() != 0) {
+         _vt.set_vote_timeout(production_pause_vote_timeout);
+         _vote_track_mode.store(vote_track_mode::inactive, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
heifner: It is only set here and the constructor.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -360,6 +361,196 @@ struct block_time_tracker {
...
+
+   auto check_pause_status(fc::time_point now) const {
+      const auto vtm = get_vote_track_mode();
+
+      production_pause_vote_tracker::pause_check check = production_pause_vote_tracker::pause_check::both;
+
+      switch (vtm) {
+      case vote_track_mode::disabled:
+      case vote_track_mode::inactive:
+         return production_pause_vote_tracker::pause_status{}; // should_pause() will return false
```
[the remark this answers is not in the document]
heifner: No, not outdated. I updated to make it clearer.

plugins/producer_plugin/producer_plugin.cpp
heifner: I changed it to `finalizer_auth_desc`. It is `finalizer_authority.description`.

plugins/producer_plugin/producer_plugin.cpp
heifner: These are `finalizer_authority`, maybe we could have named that type something else. We have a `finalizer` type, these are not those. I renamed to `active_finalizer_auth` and `pending_finalizer_auth`.

linh2931 (approved):

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -171,13 +180,18 @@ class production_pause_vote_tracker {
...
 
-      if (check != pause_check::producer) {
-         assert(check == pause_check::both || check == pause_check::other);
+      switch (check) {
+      case pause_check::both:
          process_vote_timing(latest_other_vote, status.other_vote);
-      }
-      if (check != pause_check::other) {
-         assert(check == pause_check::both || check == pause_check::producer);
+         // no break
```
arhag, line 186: Could use `[[fallthrough]];`

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -201,20 +215,27 @@ class production_pause_vote_tracker {
       }
 
       // Can be called concurrently with all member functions.
-      bool record_received_block(fc::time_point block_received_time, fc::microseconds block_acceptance_delta) {
+      bool record_received_block(fc::time_point block_received_time, fc::time_point block_time_stamp, fc::microseconds block_acceptance_delta) {
```
arhag, line 218: Again, `block_timestamp` looks a little nicer.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
```diff
@@ -201,20 +215,27 @@ class production_pause_vote_tracker {
...
          if (adjusted_block_received_time <= vote_time)
             return false;
 
          auto block_orig = first_block_after_vote.load(std::memory_order_relaxed);
 
-         if ((vote_time < block_orig) && (block_orig <= block_received_time))
+         // if we have already received a block since our last vote then nothing to do.
+         // Also make sure the block received time is greater than the tracked block, should always
+         // be true unless a clock skew.
+         if ((vote_time < block_orig) && (block_orig <= received_time))
```
arhag, line 235: I added this check before because I wanted to allow `record_received_block` to be thread-safe.

That means two threads might call `record_received_block` block concurrently and so the `store` into the `first_block_after_vote` could come in any order regardless of how their received times were ordered.

But now with the most recent changes of using `received_time = std::min(block_received_time, block_time_stamp)`, there is another reason why they can come out of order.

It is possible that we first accept a block on one branch that has timestamp `t2` and then accept a block on another branch that has timestamp `t1` where `t1 < t2`. Even though the `block_received_time` of the block with timestamp `t1` would presumably be greater than the `block_received_time` of the block with timestamp `t2` (since it was accepted after in nodeos), with the RT being the minimum and both blocks being late, the `received_time` for the first block would be `t2` and the `received_time` for the second block would be the lower value `t1`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -453,36 +455,36 @@ struct implicit_production_pause_vote_tracker {
...
-      if (_producers.contains(auth_desc)) { // _producers not modified, thread safe
+      if (_producers.contains(finalizer_auth_desc)) { // _producers not modified, thread safe
          _vt.record_received_producer_vote(now);
       } else {
          _vt.record_received_other_vote(now);
       }
    }
 
-   void record_received_block(fc::time_point now) {
+   void record_received_block(fc::time_point now, fc::time_point block_time_stamp) {
```
arhag, line 483: Again, `block_timestamp` looks a little nicer.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Currently `record_received_block` is only called from the main thread. Both `accept_block_header` and `accept_block` are currently only signaled from the main thread. I would like to potentially change that for #529. So maybe in the future it could be called from multiple threads.

I added your comment to the code.

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: `error: 'fallthrough' attribute is only allowed on empty statements`

arhag: Did you forget the semicolon at the end of it?

Anyway, it doesn't matter. It's not necessary.

arhag (approved):

plugins/producer_plugin/include/eosio/producer_plugin/production_pause_vote_tracker.hpp
heifner: Yep. Fixed.

arhag (approved):

linh2931 (approved):

Issue #693, opened 2024-09-03
Handle block invariant checks during replay appropriately and in particular ensure --force-all-checks is respected


arhag:
The bug in https://github.com/AntelopeIO/spring/issues/691 revealed that there is a more sensible way of handling the checks for block invariants involving QCs.

Specifically, `verify_qc_claim` was trying to do too much, and it revealed that important block invariants were not being checked for Legacy and Transition blocks. 

Those invariants were captured in the issue description of https://github.com/AntelopeIO/spring/issues/691.

In particular there were four specific categories of block invariant validation:
1. Validations for Legacy blocks
2. Validations for Transition blocks
3. Basic validations for Proper Savanna blocks
4. Complete QC validation for Proper Savanna blocks that have a QC block extension present

The changes of https://github.com/AntelopeIO/spring/issues/691 are now validating all of these categories of block invariants (as appropriate for the specific block type).

Additionally, there was some clean up in the functions that handle these different categories of validation. However, the Basic validation for Proper Savanna blocks (which is cheap to validation) and the complete QC validation for Proper Savanna blocks that have a QC block extension present (which is more computationally expensive to validate) were still both carried out (in order) in the `verify_qc_claim` function.

This monolithic approach to checking the invariants of a Proper Savanna block almost led to a bug where posting the `verify_qc_claim` check to a separate thread (as an optimization) would have prevented a check for a critical block invariant that was assumed as a pre-condition in later parts of the code that was indirectly called when constructing the `block_state`; see https://github.com/AntelopeIO/spring/pull/538/files#r1717405576.

So in addition to any other (protocol compatible) refactoring of these different categories of checks within the implementation that may be desired to do for code clarity reasons, this issue may present an opportunity (for reasons that will be more clear when the issue gets into the actual bug described by this issue) to separate out the checks for basic validation for Proper Savanna blocks from the complete QC validation for Proper Savanna blocks that have a QC block extension present. And maybe even to go further and optimize the validation by carrying out the complete QC validation for Proper Savanna blocks that have a QC block extension present in parallel with the construction of the `block_state`. However, it is still critical to finish the basic validation for Proper Savanna blocks prior to starting the construction the `block_state` and to finish the complete QC validation for Proper Savanna blocks that have a QC block extension present prior to calling `integrate_received_qc_to_block`.

But the main goal of this issue is to address a bug with the way these block invariant checks are handled during replay.

[...]

PR #715, opened 2024-09-06
[1.0.1] Handle block invariant checks during replay appropriately and in particular ensure --force-all-checks is respected

linh2931:
Currently QC block extension and finality header extension are not verified during replay.

This PR performs all the basic checks of Legacy, Transition, and Proper Savanna blocks during replay. If `--force-all-checks` is specified, a complete QC validation is performed.

Resolves https://github.com/AntelopeIO/spring/issues/693

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3878,15 +3878,10 @@ struct controller_impl {
           });
    }
 
-   // basic verifies all aspects of QC except signatures, complete verifies
-   // singatures in addition to basic
-   enum class qc_validation_t { basic, complete };
-
-   // Verify QC claim made by finality_extension in header extension
-   // and quorum_certificate_extension in block extension are valid.
+   // Verify basic proper_block invariants.
    // Called from net-threads. It is thread safe as signed_block is never modified after creation.
    // -----------------------------------------------------------------------------
-   void verify_proper_block_exts( const std::optional<block_id_type>& id, const signed_block_ptr& b, const block_header_state& prev, qc_validation_t qc_validation ) {
+   void verify_basic_proper_block_invariants( const std::optional<block_id_type>& id, const signed_block_ptr& b, const block_header_state& prev ) {
       assert(b->is_proper_svnn_block());
 
       auto qc_ext_id = quorum_certificate_extension::extension_id();
@@ -3975,21 +3970,9 @@ struct controller_impl {
                   invalid_qc_claim,
                   "QC is_strong (${s1}) in block extension does not match is_strong_qc (${s2}) in header extension. Block number: ${b}",
                   ("s1", qc_proof.is_strong())("s2", new_qc_claim.is_strong_qc)("b", block_num) );
-
-      // It is expensive to do signature verification. Do it only when requested.
-      if (qc_validation == qc_validation_t::complete) {
-         // find the claimed block's block state on branch of id
-         auto bsp = fork_db_fetch_bsp_on_branch_by_num( prev.id(), new_qc_claim.block_num );
-         EOS_ASSERT( bsp,
-                     invalid_qc_claim,
-                     "Block state was not found in forkdb for block_num ${q}. Block number: ${b}",
-                     ("q", new_qc_claim.block_num)("b", block_num) );
-
-         bsp->verify_qc(qc_proof);
-      }
    }
 
-   void verify_legacy_block_exts( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+   void verify_legacy_block_invariants( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
       assert(b->is_legacy_block());
 
       uint32_t block_num = b->block_num();
@@ -4011,7 +3994,7 @@ struct controller_impl {
                   ("b", block_num) );
    }
 
-   void verify_transition_block_exts( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+   void verify_transition_block_invariants( const signed_block_ptr& b, const block_header_state_legacy& prev ) {
       assert(!b->is_legacy_block() && !b->is_proper_svnn_block());
 
       uint32_t block_num = b->block_num();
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -3882,7 +3882,7 @@ struct controller_impl {
    // and quorum_certificate_extension in block extension are valid.
    // Called from net-threads. It is thread safe as signed_block is never modified after creation.
    // -----------------------------------------------------------------------------
-   void verify_proper_block_exts( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
+   void verify_proper_block_exts( const std::optional<block_id_type>& id, const signed_block_ptr& b, const block_header_state& prev, bool complete_qc_validation ) {
```
greg7mdp, line 3885: Maybe use `enum class full_qc_validation_t` instead of `bool`

greg7mdp: Sorry, now that I look at your change, I think it would be even cleaner to have:

`enum class qc_validation_t { no_bls_sig_validation, full };`

What do you think?

linh2931: We do two kinds of QC validation: basic (checking claimed block number, ...) and full (basic + BLS signature validation). Probably we should have `enum class qc_validation_t { basic, complete };` and add comments.

greg7mdp: Yes, but I thought it was nice if the `basic` name explicitely stated that we skipped the signature validation (that way no extra comments are needed).
What about: ` enum class qc_validation_t { basic_no_sig, complete };`?
If you prefer your simpler version that's fine too.

libraries/chain/controller.cpp
```diff
@@ -3878,11 +3878,13 @@ struct controller_impl {
...
    }
 
+   enum class full_qc_validation_t { no, yes };
+
    // Verify QC claim made by finality_extension in header extension
    // and quorum_certificate_extension in block extension are valid.
    // Called from net-threads. It is thread safe as signed_block is never modified after creation.
    // -----------------------------------------------------------------------------
-   void verify_proper_block_exts( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
+   void verify_proper_block_exts( const std::optional<block_id_type>& id, const signed_block_ptr& b, const block_header_state& prev, full_qc_validation_t qc_validation ) {
```
heifner, line 3887: I think you should not pass in `id` and have this method (and the others) return `qc_claim_t`. Then log from `create_block_handle_i` if not a `legacy_block`.

heifner (changes requested):

libraries/chain/controller.cpp
linh2931: Thanks. This will avoid the misleading logging when replaying.

libraries/chain/controller.cpp
```diff
@@ -4081,6 +4086,17 @@ struct controller_impl {
          }
       }
 
+      return;
```
arhag, line 4089: Unnecessary return.

libraries/chain/controller.cpp
```diff
@@ -3944,52 +3938,39 @@ struct controller_impl {
...
       }
 
-      // At this point, we are making a new claim in this block, so it better include a QC to justify this claim.
-      EOS_ASSERT( qc_extension_present,
-                  invalid_qc_claim,
+      // At this point, we are making a new claim in this block, so it must includes a QC to justify this claim.
+      EOS_ASSERT( qc_extension_present, block_validate_exception,
                   "Block #${b} is making a new finality claim, but doesn't include a qc to justify this claim", ("b", block_num) );
 
       const auto& qc_ext   = std::get<quorum_certificate_extension>(block_exts.find(qc_ext_id)->second);
```
heifner, line 3950: If you use `block_exts.find` above instead of `count` you can reuse the iterator here.

libraries/chain/controller.cpp
```diff
@@ -3878,11 +3878,10 @@ struct controller_impl {
           });
    }
 
-   // Verify QC claim made by finality_extension in header extension
-   // and quorum_certificate_extension in block extension are valid.
+   // Verify basic proper_block invariants.
    // Called from net-threads. It is thread safe as signed_block is never modified after creation.
    // -----------------------------------------------------------------------------
-   void verify_proper_block_exts( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
+   void verify_basic_proper_block_invariants( const signed_block_ptr& b, const block_header_state& prev ) {
```
heifner, line 3884: I think this should return a `std::optional<qc_t>`. Then if the optional contains a `qc_t`, pass it into `verify_qc()`. Note `qc_t` has a method `to_qc_claim()` which is very cheap to call. That will be much faster and eliminate duplicate code.

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -4058,28 +4040,79 @@ struct controller_impl {
       }
    }
 
-   // thread safe, expected to be called from thread other than the main thread
-   template<typename ForkDB, typename BS>
-   block_handle create_block_state_i( ForkDB& forkdb, const block_id_type& id, const signed_block_ptr& b, const BS& prev ) {
+   // verify basic_block invariants
+   template<typename BS>
+   void verify_basic_block_invariants(const signed_block_ptr& b, const BS& prev) {
```
greg7mdp, line 4045: I don't think we need `basic` in the names `verify_basic_block_invariants` and `verify_basic_proper_block_invariants`.

greg7mdp (approved):

libraries/chain/controller.cpp
linh2931: Thanks. I was trying to make the function independent. But your method is shorter and faster.

libraries/chain/controller.cpp
linh2931: I think the `basic` here indicates we have more to check (like full qc) if required.

heifner (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Enhance checks for QC on blocks across block types Legacy, Transition, and Proper Savanna.
Note:end

libraries/chain/controller.cpp
```diff
@@ -3925,71 +3921,59 @@ struct controller_impl {
...
-      auto bsp = fork_db_fetch_bsp_on_branch_by_num( prev.id(), new_qc_claim.block_num );
-      EOS_ASSERT( bsp,
-                  invalid_qc_claim,
-                  "Block state was not found in forkdb for block_num ${q}. Block number: ${b}",
-                  ("q", new_qc_claim.block_num)("b", block_num) );
-
-      // verify the QC proof against the claimed block
-      bsp->verify_qc(qc_proof);
+      return qc_proof;
    }
```
heifner, line 3965: Be explicit with optional type otherwise you don't get return value optimization

libraries/chain/controller.cpp
```diff
@@ -3925,71 +3921,59 @@ struct controller_impl {
...
 
-      const auto& qc_ext   = std::get<quorum_certificate_extension>(block_exts.find(qc_ext_id)->second);
+      assert(qc_ext_itr != block_exts.end() );
+      const auto& qc_ext   = std::get<quorum_certificate_extension>(qc_ext_itr->second);
       const auto& qc_proof = qc_ext.qc;
 
       // Check QC information in header extension and block extension match
-      EOS_ASSERT( qc_proof.block_num == new_qc_claim.block_num,
-                  invalid_qc_claim,
+      EOS_ASSERT( qc_proof.block_num == new_qc_claim.block_num, invalid_qc_claim,
```
arhag, line 3955: It would probably make more sense if this was a `block_validate_exception`.

libraries/chain/controller.cpp
```diff
@@ -3925,71 +3921,59 @@ struct controller_impl {
...
                   "Block #${b}: Mismatch between qc.block_num (${n1}) in block extension and block_num (${n2}) in header extension",
                   ("n1", qc_proof.block_num)("n2", new_qc_claim.block_num)("b", block_num) );
 
       // Verify claimed strictness is the same as in proof
-      EOS_ASSERT( qc_proof.is_strong() == new_qc_claim.is_strong_qc,
-                  invalid_qc_claim,
+      EOS_ASSERT( qc_proof.is_strong() == new_qc_claim.is_strong_qc, invalid_qc_claim,
```
arhag, line 3960: It would probably make more sense if this was a `block_validate_exception`.

arhag (commented):
LGTM

libraries/chain/controller.cpp
linh2931: Thanks. Did not notice this before I pressed the green button. Will do a followup PR tomorrow.

libraries/chain/controller.cpp
linh2931: Will be in a follwoup PR.

libraries/chain/controller.cpp
linh2931: Will be in a follwoup PR.

libraries/chain/controller.cpp
```diff
@@ -4058,26 +4043,59 @@ struct controller_impl {
...
+      constexpr bool is_proper_savanna_block = std::is_same_v<typename std::decay_t<BS>, block_state>;
+      assert(is_proper_savanna_block == b->is_proper_svnn_block());
+
+      std::optional<qc_t> qc = verify_basic_block_invariants(b, prev);
+
+      if constexpr (is_proper_savanna_block) {
+         if (qc) {
+            verify_qc(b, prev, *qc);
+
+            const auto  qc_claim = qc->to_qc_claim();
```
linh2931, line 4095: In the followup PR, will remove this statement and move `qc->to_qc_claim()` into the `dlog` below so it is not executed if `dlog` is not enabled

libraries/chain/controller.cpp
linh2931: On a second thought, because the new QC claim's block number does not match the block number in the attached QC, can we consider it as an invalid QC claim?

libraries/chain/controller.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/735

libraries/chain/controller.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/735

PR #719, opened 2024-09-07
[1.0.1] Add missing data to block_header_state and change verify_qc_claim to use block_header_state data rather than only looking up a block_state in the fork database

### LINKED ISSUE #694: Add missing data to block_header_state and change verify_qc_claim to use block_header_state data rather than only looking up a block_state in the fork database

greg7mdp:
Resolves #694.

Implements updates described in issue #694:
  - store policy generation numbers in every `block_ref` of the `finality_core`
  - maintains `latest_qc_claim_block_active_finalizer_policy` in `block_header_state`
  - new snapshot format (`v8`) which adds these new members. `v7` format of spring 1.0  is unsupported.
  - new fork_database format (version 3). version 2 not loaded anymore, but version 1 (legacy fork_db only) OK.
  - update the serialization of the finalizer safety file, so it doesn't change even though the `block_ref` struct has two new members.
  - add version to `chain_head.dat` file and do not load existing version.
  - updated qc validation which does not require the fork database anymore, all needed info is in the parent `block_header_state`
  - add new test (`verify_spring_1_0_block_compatibitity`) validating that this PR's changes do not affect the block ids generated.
  - update tests impacted by the change, including adding to repo new snapshot v8 reference files.


libraries/chain/block_handle.cpp
```diff
@@ -13,25 +23,34 @@ void block_handle::write(const std::filesystem::path& state_file) {
...
+   if (magic == chain_head_magic && version == chain_head_version) {
+      fc::raw::unpack(f, *this);
+      ilog("Loading chain_head block ${bn} ${id}", ("bn", block_num())("id", id()));
+      res = true;
+   }
 
    std::filesystem::remove(state_file);
 
-   return true;
+   return res;
```
heifner, line 53: If the `unpack` fails it throws an exception and the `chain_head.dat` is not removed. If the `chain_head.dat` is less than `2*sizeof(chain_head_magic)` it is not read but the file is not removed. If the magic or version is wrong which it will be for existing 1.0.0 `chain_head.dat`, the file is removed and then the error from `controller::startup` will be `No existing chain_head.dat file`.

Seems we should wrap the whole thing in a try-catch and consistently NOT remove the file. They could downgrade to 1.0.0, and be able to startup. 

I think we should report an error telling user to consider starting from a snapshot if this fails for any reason.

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,61 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
...
+      assert(pending_finalizer_policy && pending_finalizer_policy->second->generation == pending_gen);
+      EOS_ASSERT(pending_finalizer_policy && pending_finalizer_policy->second->generation == pending_gen, chain_exception,
+                 "Logic error in finalizer policy retrieval");  // just in case
+      res.pending_finalizer_policy = pending_finalizer_policy->second;
+   }
+
+   return res;
+}
+
+// Only defined for core.latest_qc_claim().block_num <= num <= core.current_block_num()
```
heifner, line 187: How about adding an `assert` of this.

libraries/chain/block_state.cpp
```diff
@@ -233,7 +238,14 @@ vote_status_t block_state::has_voted(const bls_public_key& key) const {
...
 void block_state::verify_qc(const qc_t& qc) const {
-   aggregating_qc.verify_qc(qc, strong_digest, weak_digest);
+   // Do not use `block_state::aggregating_qc` which applies only for `this` block.
+   // `verify_qc()` can be called on a descendant `block_state` of `qc.block_num`, so we need
+   // to create a new `aggregating_qc_t` with the finalizer policies of the claimed block.
+   // ---------------------------------------------------------------------------------------
+   finalizer_policies_t policies = get_finalizer_policies(qc.block_num);
+   aggregating_qc_t aggregating_qc(policies.active_finalizer_policy, policies.pending_finalizer_policy);
+
+   aggregating_qc.verify_qc(qc, policies.finality_digest, create_weak_digest(policies.finality_digest));
```
heifner, line 248: With this usage, probably makes sense to factor out qc verification out of `aggregating_qc_t` into another type that `aggregating_qc_t` and this can use. The `active_policy_sig` and `pending_policy_sig` of `aggregating_qc_t` are not used. Seems wasteful to create a `aggregating_qc_t` for this.

Maybe add a `finalizer_policy_pair_t` that both `aggregating_qc_t` and this method can use. The `finalizer_policy_pair_t` can have the `verify_qc` and `verify_dual_finalizers_votes` methods.

libraries/chain/controller.cpp
```diff
@@ -2285,8 +2285,13 @@ struct controller_impl {
          } else {
             EOS_THROW(snapshot_exception, "Unsupported block_state version");
          }
+      } else if (header.version == 7) {
+         // snapshot created with Spring 1.0, which was very soon superseded by Spring 1.0.1
+         // and a new snapshot format.
+         // --------------------------------------------------------------------------------
+         EOS_THROW(snapshot_exception, "v7 snapshots are not supported anymore in Spring 1.01 and above");
```
heifner, line 2292: `1.0.1`

libraries/chain/block_state.cpp
heifner: Instead of `block_state` having a `verify_qc`, maybe it makes more sense for it to return a `finalizer_policy_pair_t` given a `block_num`.

libraries/chain/finality/finalizer.cpp
```diff
@@ -267,7 +300,7 @@ void my_finalizers_t::load_finalizer_safety_info_v1(fsi_map& res) {
       bls_public_key pubkey;
       my_finalizers_t::fsi_t fsi;
       fc::raw::unpack(persist_file, pubkey);
-      fc::raw::unpack(persist_file, fsi);
+      unpack_v1(persist_file, fsi);
       res.emplace(pubkey, fsi);
```
heifner, line 304: Why these safey.dat file changes? Just clean up or is there an actual change here?

libraries/chain/fork_database.cpp
```diff
@@ -646,10 +646,16 @@ namespace eosio::chain {
...
-                                                   // version == 1 -> legacy
-                                                   // version == 2 -> savanna (two possible fork_db, one containing `block_state_legacy`,
-                                                   //                          one containing `block_state`)
+
+      // write out current version which is always max_supported_version
+      // version == 1 -> legacy
+      // version == 2 -> Spring 1.0
+      //                 (two possible fork_db, one containing `block_state_legacy`, one containing `block_state`)
+      //                  unsupported by Spring 1.01 and above
+      // version == 3 -> Spring 1.01 updated block_header_state (core with policy gen #)
```
heifner, line 655: `1.0.1`

libraries/chain/fork_database.cpp
```diff
@@ -691,6 +697,8 @@ namespace eosio::chain {
 
             uint32_t version = 0;
             fc::raw::unpack( ds, version );
+            EOS_ASSERT( version != 2, fork_database_exception,
+                        "Version 2 of fork_database (created by Spring 1.0) is not supported" );
```
heifner, line 701: I think better to say `Spring 1.0.0`

unittests/snapshot_tests.cpp
```diff
@@ -382,10 +382,10 @@ void compatible_versions_test()
    std::filesystem::copy(source_log_dir / "blocks.index", config.blocks_dir / "blocks.index");
    TESTER base_chain(config, *genesis);
 
-   std::string current_version = "v7";
+   std::string current_version = "v8";
 
    int ordinal = 0;
-   for(std::string version : {"v2", "v3", "v4" , "v5", "v6", "v7"})
+   for(std::string version : {"v2", "v3", "v4", "v5", "v6", "v8"}) // v7 version not supported in spring 1.01 and above
```
heifner, line 388: `1.0.1`

heifner (changes requested):

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,61 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
    return *active_finalizer_policy;
 }
 
+// Only defined for core.latest_qc_claim().block_num <= ref.block_num() <= core.current_block_num()
+// Retrieves the finalizer policies applicatble for the block referenced by `ref`.
+// See full explanation in issue #694.
+// ------------------------------------------------------------------------------------------------
+finalizer_policies_t block_header_state::get_finalizer_policies(const block_ref& ref) const {
```
linh2931, line 148: The preconditions of this function can be asserted here like the asserts in `core`.

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,61 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
...
+
+   if (active_finalizer_policy->generation == active_gen)
+      res.active_finalizer_policy = active_finalizer_policy;      // the one active at block_num is still active
+   else {
+      // cannot be the pending one as it never was active
+      assert(!pending_finalizer_policy || pending_finalizer_policy->second->generation > active_gen);
+
+      // has to be the one in latest_qc_claim_block_active_finalizer_policy
+      assert(latest_qc_claim_block_active_finalizer_policy != nullptr);
+      assert(latest_qc_claim_block_active_finalizer_policy->generation == active_gen);
```
linh2931, line 164: Since EOS_ASSERT on line 164 is always on, is this small assert redundant?

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,61 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
...
+
+   auto pending_gen = ref.pending_policy_generation;
+   if (pending_gen == 0)
+      res.pending_finalizer_policy = nullptr;                    // no pending policy at block_num.
+   else if (pending_gen == active_finalizer_policy->generation)
+      res.pending_finalizer_policy = active_finalizer_policy;    // policy pending at block_num became active
+   else {
+      // cannot be the one in latest_qc_claim_block_active_finalizer_policy since it was active at
+      // core.latest_qc_claim().block_num. So it must be the one still pending.
+      assert(pending_finalizer_policy && pending_finalizer_policy->second->generation == pending_gen);
```
linh2931, line 178: Is this small assert redundant?

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,61 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
...
+}
+
+// Only defined for core.latest_qc_claim().block_num <= num <= core.current_block_num()
+// Retrieves the active finalizer policy generation applicatble for the block `num`, which
+// can be the current block or one of its ancestors up to core.latest_qc_claim().block_num (incl).
+// -----------------------------------------------------------------------------------------------
+uint32_t block_header_state::get_active_finalizer_policy_generation(block_num_type num) const {
+   if (num == block_num())
+      return active_finalizer_policy->generation;
+   const block_ref& ref = core.get_block_reference(num);
```
linh2931, line 194: Can we assert `get_block_reference(num)` always find the block_ref for the `num`?

libraries/chain/block_header_state.cpp
```diff
@@ -325,6 +380,22 @@ void finish_next(const block_header_state& prev,
       next_header_state.finalizer_policy_generation = prev.finalizer_policy_generation;
    }
 
+   // now populate next_header_state.latest_qc_claim_block_active_finalizer_policy
+   // this keeps track of the finalizer policy which was active @ latest_qc_claim().block_num, but which
+   // can be overwritten by a previously pending police (member `active_finalizer_policy`)
```
linh2931, line 385: police -> policy

libraries/chain/block_header_state.cpp
```diff
@@ -325,6 +380,22 @@ void finish_next(const block_header_state& prev,
...
+   // --------------------------------------------------------------------------------------------------
+   const auto& next_core                 = next_header_state.core;
+   auto        latest_qc_claim_block_num = next_core.latest_qc_claim().block_num;
+   const auto  active_generation_num     = next_header_state.active_finalizer_policy->generation;
+   if (prev.get_active_finalizer_policy_generation(latest_qc_claim_block_num) != active_generation_num) {
+      const auto& latest_qc_claim_block_ref = next_header_state.core.get_block_reference(latest_qc_claim_block_num);
+      next_header_state.latest_qc_claim_block_active_finalizer_policy =
+         prev.get_finalizer_policies(latest_qc_claim_block_ref).active_finalizer_policy;
+   } else {
+      next_header_state.latest_qc_claim_block_active_finalizer_policy = nullptr;
```
linh2931, line 396: why set to `nullptr`?

linh2931 (commented):
Submit partial reviews. Will continue the rest shortly.

libraries/chain/block_handle.cpp
greg7mdp: done.

libraries/chain/block_header_state.cpp
greg7mdp: done.

libraries/chain/block_state.cpp
greg7mdp: I don't disagree, but I think we should refrain from refactoring the code in a 1.01 release. 
I tried to do what was necessary for cleany fixing the issue in this PR.
I'll be happy to clean up this further in a separate PR targeted to the main branch.

libraries/chain/finality/finalizer.cpp
greg7mdp: Not a cleanup. We cannot use the `fc::raw::unpack(persist_file, fsi);` because the `fsi` includes `last_vote` and `lock`, which are of type `block_ref`, and `block_ref` now includes two new members (the active and pending generations)

libraries/chain/fork_database.cpp
greg7mdp: updated.

libraries/chain/fork_database.cpp
greg7mdp: updated

unittests/snapshot_tests.cpp
greg7mdp: updated.

libraries/chain/block_handle.cpp
```diff
@@ -4,6 +4,16 @@
 
 namespace eosio::chain {
 
+// -------------------------------------------------------------------------------------------
+// prior to writing magic and version numbers, we simply serialized the class (*this) to
+// file. Let's call this the implicit version 0, which is not supported anymore in
+// Spring 1.01 and above.
```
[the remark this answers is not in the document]
greg7mdp: updated.

libraries/chain/block_handle.cpp
```diff
@@ -4,6 +4,16 @@
...
+// However, we need to make sure that `chain_head_magic` can't match the tag of a std::variant
```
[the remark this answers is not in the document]
greg7mdp: [this one](https://github.com/AntelopeIO/spring/blob/43e7ab54332336a0188aa5a0871cf9263569365e/libraries/chain/include/eosio/chain/block_handle.hpp#L13)

libraries/chain/block_handle.cpp
```diff
@@ -13,25 +23,34 @@ void block_handle::write(const std::filesystem::path& state_file) {
...
+   fc::raw::pack(f, chain_head_version);
    fc::raw::pack(f, *this);
 }
 
 bool block_handle::read(const std::filesystem::path& state_file) {
-   if (!std::filesystem::exists(state_file))
-      return false;
+   bool res = false;
+
+   if (!std::filesystem::exists(state_file) || std::filesystem::file_size(state_file) <= 2 * sizeof(chain_head_magic))
```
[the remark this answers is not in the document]
greg7mdp: I have an EOS_ASSERT now.

libraries/chain/block_header_state.cpp
greg7mdp: I added the assert.

libraries/chain/block_header_state.cpp
greg7mdp: No, I have both on purpose.
The `assert` to clearly state this property always hold. 
The `EOS_ASSERT` in case I am wrong  and that is not true. Which is why I added the comment `// just in case`.

libraries/chain/block_header_state.cpp
greg7mdp: No, I have both on purpose.
The assert to clearly state this property always hold. And the fact that this small `assert`  is there indicates that the property holds in all our tests.

The EOS_ASSERT in case I am wrong and that is not true. Which is why I added the comment // just in case.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Enhance Block Header State and only use provided block data to verify QC claims.
Note:end

libraries/chain/block_header_state.cpp
greg7mdp: `get_block_reference` returns a referencem, and already has the following asserts:

```
   assert(last_final_block_num() <= block_num); // Satisfied by precondition.
   assert(block_num < current_block_num()); // Satisfied by precondition.
```

Because the assert I added at the start of the function is stricter, I think we're fine.

libraries/chain/block_header_state.cpp
greg7mdp: updated.

libraries/chain/block_header_state.cpp
greg7mdp: We reset it when it doesn't need to be tracked anymore. This will allow the memory holding that policy to be release when the last block holding it as an active policy is also removed from fork_db.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -121,12 +129,22 @@ struct block_header_state : fc::reflect_init {
...
    std::vector<std::pair<block_num_type, finalizer_policy_ptr>> proposed_finalizer_policies;
+
    // Track in-flight pending finalizer policy. At most one pending
    // finalizer policy at any moment.
    // When the block associated with the pending finalizer policy becomes final,
    // it becomes active.
    std::optional<std::pair<block_num_type, finalizer_policy_ptr>> pending_finalizer_policy;
 
+   // It may be that the `finality_core` references a finalizer policy generation which is neither the active
+   // or pending one. This can happen when a pending policy became active, replacing the previously active
```
arhag, line 140: To be clear, it is possible for the latest final block (which also is tracked in  `finality_core` to have an active finalizer policy that is still not being tracked.

But we don't care about that.

We just care about any finalizer policies tracked within `finality_core` that are restricted to the block range from `core.latest_qc_claim().block_num` to `core.current_block_num()` inclusive.

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,65 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
    return *active_finalizer_policy;
 }
 
+// Only defined for core.latest_qc_claim().block_num <= ref.block_num() <= core.current_block_num()
+// Retrieves the finalizer policies applicatble for the block referenced by `ref`.
```
arhag, line 145: Typo: applicable

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,65 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
...
+   if (active_finalizer_policy->generation == active_gen)
+      res.active_finalizer_policy = active_finalizer_policy;      // the one active at block_num is still active
+   else {
+      // cannot be the pending one as it never was active
+      assert(!pending_finalizer_policy || pending_finalizer_policy->second->generation > active_gen);
+
+      // has to be the one in latest_qc_claim_block_active_finalizer_policy
+      assert(latest_qc_claim_block_active_finalizer_policy != nullptr);
+      assert(latest_qc_claim_block_active_finalizer_policy->generation == active_gen);
+      EOS_ASSERT(latest_qc_claim_block_active_finalizer_policy->generation == active_gen, chain_exception,
```
arhag, line 167: Should be consistent with the asserts above.

This should never happen, but if we somehow are wrong and it does, we would like to avoid the undefined behavior. 

So the condition for the `EOS_ASSERT` should be: `latest_qc_claim_block_active_finalizer_policy != nullptr && latest_qc_claim_block_active_finalizer_policy->generation == active_gen`.

unittests/snapshot_tests.cpp
```diff
@@ -382,10 +382,10 @@ void compatible_versions_test()
    std::filesystem::copy(source_log_dir / "blocks.index", config.blocks_dir / "blocks.index");
    TESTER base_chain(config, *genesis);
 
-   std::string current_version = "v7";
+   std::string current_version = "v8";
 
    int ordinal = 0;
-   for(std::string version : {"v2", "v3", "v4" , "v5", "v6", "v7"})
+   for(std::string version : {"v2", "v3", "v4", "v5", "v6", "v8"}) // v7 version not supported in spring 1.0.1 and above
```
arhag, line 388: Capitalize Spring.

arhag: Also, maybe change the example in the comment below of how to support a new snapshot version in this test to use `snap_v8` instead of `snap_v7`.

unittests/snapshots/CMakeLists.txt
```diff
@@ -23,3 +23,7 @@ configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/snap_v6.bin.json.gz ${CMAKE_CURRENT_
 configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/snap_v7.bin.gz ${CMAKE_CURRENT_BINARY_DIR}/snap_v7.bin.gz COPYONLY )
 configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/snap_v7.json.gz  ${CMAKE_CURRENT_BINARY_DIR}/snap_v7.json.gz  COPYONLY )
 configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/snap_v7.bin.json.gz ${CMAKE_CURRENT_BINARY_DIR}/snap_v7.bin.json.gz COPYONLY )
```
arhag, line 25: Should we get rid of the `snap_v7` files now?

unittests/savanna_misc_tests.cpp
```diff
@@ -537,9 +535,10 @@ BOOST_FIXTURE_TEST_CASE(validate_qc_requiring_finalizer_policies, savanna_cluste
    BOOST_REQUIRE_EQUAL(active->generation, p2);         // b6 has active finalizer policy p2
    BOOST_REQUIRE(!A.head_pending_finalizer_policy());   // and no pending finalizer policy.
 
-   auto b6_snapshot = A.snapshot();
+   // At this point, in the Spring 1.0 implementation, policy  P2 is lost from the block_header_state
```
arhag, line 538: Should say `in the Spring 1.0.0 implementation (which has the bug described in issue #694)`.

unittests/savanna_misc_tests.cpp
```diff
@@ -585,12 +584,107 @@ BOOST_FIXTURE_TEST_CASE(validate_qc_requiring_finalizer_policies, savanna_cluste
...
-#endif
+
+} FC_LOG_AND_RETHROW()
+
+
+
+// ---------------------------------------------------------------------------------------------------
+// For issue #694, we need to change the finality core of the `block_header_state`, but we want to
+// insure that this doesn't create a consensus incompatibility with spring 1.0, so the blocks created
+// with newer versions remain compatible (and linkable) by spring 1.0.
```
arhag, line 599: `insure` -> `ensure`
`spring 1.0` -> `Spring 1.0.0` (in two places)

unittests/savanna_misc_tests.cpp
```diff
@@ -585,12 +584,107 @@ BOOST_FIXTURE_TEST_CASE(validate_qc_requiring_finalizer_policies, savanna_cluste
...
+   push_block(0, b8);                                   // nodes received b6
+
+   auto b9 = A.produce_block();
+   print("b9", b9);
+   BOOST_REQUIRE_EQUAL(qc_s(qc(b9)), strong_qc(b8));    // b9 claims a strong QC on b8
+   BOOST_REQUIRE_EQUAL(A.lib_number, b6->block_num());  // b9 makes B6 final
+
+   // check that the block id of b9 match what we got with spring 1.0
+   auto b9_id = b9->calculate_id();
+   BOOST_REQUIRE_EQUAL(b9_id, block_id_type{"00000013725f3d79bd4dd4091d0853d010a320f95240981711a942673ad87918"});
```
arhag, line 687: This is okay for now. But not quite what was requested in https://github.com/AntelopeIO/spring/issues/694.

The goal for this test was to commit the blocks generated by Spring 1.0.0 to the repo and then have the test successfully synchronize the blocks. That way if there was a problem, the error would help us reveal where the problem was. And we could always step-through debug to get the bottom of the issue.

With this approach we just detect something is wrong but the test does not permit us to figure out what could be the cause.

Anyway, you do not need to do what was originally requested as part of this issue. This achieves the goal of detecting unintentional consensus changes. I created a new issue https://github.com/AntelopeIO/spring/issues/742 to track the further work needed to augment this test, and that work can go to `main` instead.

arhag (changes requested):
By the way, your PR description says "all needed info is in the root `block_header_state`" but it should say "all needed info is in the parent `block_header_state`".

libraries/chain/block_handle.cpp
```diff
@@ -13,24 +23,39 @@ void block_handle::write(const std::filesystem::path& state_file) {
...
 
 bool block_handle::read(const std::filesystem::path& state_file) {
    if (!std::filesystem::exists(state_file))
       return false;
 
-   fc::datastream<fc::cfile> f;
-   f.set_file_path(state_file);
-   f.open("rb");
+   EOS_ASSERT(std::filesystem::file_size(state_file) >= 2 * sizeof(chain_head_magic), chain_exception,
+              "File `chain_head.dat` seems to be corrupted. The best course of action might be to restart from a snapshot (v8 or above)" );
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/008e1e5b6

libraries/chain/block_handle.cpp
```diff
@@ -13,24 +23,39 @@ void block_handle::write(const std::filesystem::path& state_file) {
...
 
-   ilog("Loading chain_head block ${bn} ${id}", ("bn", block_num())("id", id()));
+      uint64_t magic, version;
+      fc::raw::unpack(f, magic);
+      fc::raw::unpack(f, version);
 
-   std::filesystem::remove(state_file);
+      EOS_ASSERT(magic == chain_head_magic && version == chain_head_version, chain_exception,
+                 "Error reading `chain_head.dat` file. It is likely a Spring 1.0 version which is not supported by Spring 1.01 and above"
+                 "The best course of action might be to restart from a snapshot (v8 or above)" );
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/008e1e5b6

libraries/chain/block_state.cpp
```diff
@@ -184,25 +184,30 @@ block_state_ptr block_state::create_transition_block(
    return result_ptr;
 }
 
-block_state::block_state(snapshot_detail::snapshot_block_state_v7&& sbs)
+// Spring 1.01 to ? snapshot v8 format. Updated `finality_core` to include finalizer policies
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/e3cdb7c8127968243dd824f66a392d0e3456aa2b

libraries/chain/fork_database.cpp
```diff
@@ -646,10 +646,16 @@ namespace eosio::chain {
...
-      fc::raw::pack( out, max_supported_version ); // write out current version which is always max_supported_version
-                                                   // version == 1 -> legacy
-                                                   // version == 2 -> savanna (two possible fork_db, one containing `block_state_legacy`,
-                                                   //                          one containing `block_state`)
+
+      // write out current version which is always max_supported_version
+      // version == 1 -> legacy
+      // version == 2 -> Spring 1.0.0
+      //                 (two possible fork_db, one containing `block_state_legacy`, one containing `block_state`)
+      //                  unsupported by Spring 1.01 and above
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/e3cdb7c8127968243dd824f66a392d0e3456aa2b

libraries/chain/controller.cpp
```diff
@@ -2291,8 +2291,13 @@ struct controller_impl {
          } else {
             EOS_THROW(snapshot_exception, "Unsupported block_state version");
          }
+      } else if (header.version == 7) {
+         // snapshot created with Spring 1.0, which was very soon superseded by Spring 1.0.1
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/fa686db91f87c71ea08255f6a194b3c8587ada42

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -192,6 +213,10 @@ struct block_header_state : fc::reflect_init {
    const finalizer_policy& get_last_pending_finalizer_policy() const;
    const proposer_policy& get_last_proposed_proposer_policy() const;
 
+   // Only defined for core.latest_qc_claim().block_num <= num <= core.current_block_num()
+   finalizer_policies_t get_finalizer_policies(const block_ref& ref) const;
+   uint32_t get_active_finalizer_policy_generation(block_num_type block_num) const;
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/43622f20d47511d7e70b27c76e429ca2cf794326

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -178,9 +182,15 @@ struct block_state : public block_header_state {     // block_header_state provi
          bool                              skip_validate_signee,
          const std::optional<digest_type>& action_mroot_savanna);
 
-   explicit block_state(snapshot_detail::snapshot_block_state_v7&& sbs);
+   explicit block_state(snapshot_detail::snapshot_block_state_v8&& sbs);
 
    void sign(const signer_callback_type& signer, const block_signing_authority& valid_block_signing_authority);
+
+   finalizer_policies_t get_finalizer_policies(block_num_type num) const {
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/43622f20d47511d7e70b27c76e429ca2cf794326

libraries/chain/block_header_state.cpp
```diff
@@ -141,6 +141,65 @@ const finalizer_policy& block_header_state::get_last_pending_finalizer_policy()
...
+   return res;
+}
+
+// Only defined for core.latest_qc_claim().block_num <= num <= core.current_block_num()
+// Retrieves the active finalizer policy generation applicatble for the block `num`, which
+// can be the current block or one of its ancestors up to core.latest_qc_claim().block_num (incl).
+// -----------------------------------------------------------------------------------------------
+uint32_t block_header_state::get_active_finalizer_policy_generation(block_num_type num) const {
+   assert(core.links.empty() ||   // called from a bogus block_state constructed in a test
+          (core.latest_qc_claim().block_num <= num && num <= core.current_block_num()));
```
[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/4164a34b1f15871df1885844128842619d023dec

libraries/chain/include/eosio/chain/block_header_state.hpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/0dd7a3639f78f350fae5bcf7ecbaa20972337226

libraries/chain/block_header_state.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/ced3e553b6a4b0500bc80efb05ae6c781e24c641

unittests/snapshot_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/edf294dee52b5f314bab2a52928de9d2a06a1eb6

unittests/snapshots/CMakeLists.txt
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/1bdd27645b5ab83eaa630970d9ccc55e7241d2f8

libraries/chain/block_state.cpp
heifner: Please create an issue for it if not doing it in this PR. It is wasteful as is.

unittests/savanna_misc_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/d272c2517ef206052e74fb9bec037c073e6d5be7

unittests/savanna_misc_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/d272c2517ef206052e74fb9bec037c073e6d5be7

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/719/commits/43e7ab54332336a0188aa5a0871cf9263569365e

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -127,7 +127,11 @@ struct block_state : public block_header_state {     // block_header_state provi
       return timestamp() > fc::time_point::now() - fc::seconds(30);
    }
 
-   block_ref make_block_ref() const { return block_ref{block_id, timestamp(), strong_digest }; } // use the cached `finality_digest`
+   // use the cached `finality_digest`
+   block_ref make_block_ref() const {
```
linh2931, line 131: Why do we define a new `make_block_ref` here? We already have one in block_header_state.

heifner (approved):

linh2931 (approved):

PR #722, opened 2024-09-09
[1.0.1] Verify index >= 0 in apply_diff and fix warnings

linh2931:
`index` in `apply_diff` can be a signed int and can be negative. Currently index is only checked not greater than the size. 

This also results in warnings in numerous places as `apply_diff` is in a header file which is included by a number of files:

```
libraries/libfc/include/fc/container/ordered_diff.hpp:111:35: warning: comparison of integer expressions of different signedness: ‘std::ptrdiff_t’ {aka ‘long int’} and ‘std::vector<char>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
...
ibraries/libfc/include/fc/container/ordered_diff.hpp:119:26: warning: comparison of integer expressions of different signedness: ‘std::tuple_element<0, std::pair<int, char> >::type’ {aka ‘int’} and ‘std::vector<char>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
```

Resolves https://github.com/AntelopeIO/spring/issues/723

### DIFF
--- a/libraries/libfc/include/fc/container/ordered_diff.hpp
+++ b/libraries/libfc/include/fc/container/ordered_diff.hpp
@@ -108,7 +108,7 @@ public:
       // Remove from the source based on diff.remove_indexes
       std::ptrdiff_t offset = 0;
       for (SizeType index : diff.remove_indexes) {
-         SizeType updated_index = index + offset;
+         auto updated_index = index + offset;
          // Safe to do static_cast as `updated_index >= 0` is verified first
          FC_ASSERT(updated_index >= 0 && (static_cast<Container<T>::size_type>(updated_index) < container.size()), "diff.remove_indexes index ${idx} + offset ${o} not in range ${s}",
                    ("idx", index)("o", offset)("s", container.size()));
--- a/libraries/libfc/include/fc/container/ordered_diff.hpp
+++ b/libraries/libfc/include/fc/container/ordered_diff.hpp
@@ -28,11 +28,12 @@ namespace fc {
 template <typename T, typename SizeType = size_t, template<typename Y, typename...> typename Container = std::vector>
 requires std::equality_comparable<T>
          && std::random_access_iterator<typename Container<T>::iterator>
-         && std::is_unsigned<SizeType>::value
+         && std::is_unsigned_v<SizeType>
          && (std::numeric_limits<typename Container<T>::size_type>::max() >= std::numeric_limits<SizeType>::max())
 class ordered_diff {
 public:
    using size_type = SizeType;
+   using container_size_type = typename Container<T>::size_type;
 
    struct diff_result {
       Container<SizeType>                remove_indexes;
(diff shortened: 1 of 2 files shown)


libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -108,15 +108,17 @@ class ordered_diff {
       // Remove from the source based on diff.remove_indexes
       std::ptrdiff_t offset = 0;
       for (SizeType index : diff.remove_indexes) {
-         FC_ASSERT(index + offset < container.size(), "diff.remove_indexes index ${idx} + offset ${o} not in range ${s}",
+         SizeType updated_index = index + offset;
```
heifner, line 111: This should not be `SizeType` as that is often going to be unsigned. Use `auto` instead.

heifner (changes requested):

heifner (approved):

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -26,7 +26,10 @@ namespace fc {
...
  */
 template <typename T, typename SizeType = size_t, template<typename Y, typename...> typename Container = std::vector>
-requires std::equality_comparable<T> && std::random_access_iterator<typename Container<T>::iterator>
+requires std::equality_comparable<T>
+         && std::random_access_iterator<typename Container<T>::iterator>
+         && std::is_unsigned<SizeType>::value
+         && (std::numeric_limits<typename Container<T>::size_type>::max() >= std::numeric_limits<SizeType>::max())
 class ordered_diff {
 public:
    using size_type = SizeType;
```
greg7mdp, line 35: maybe add:
    `using cont_size_t = typename Container<T>::size_type;` 

so that you can use `cont_size_t` from now on.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -106,19 +109,26 @@ class ordered_diff {
...
-         // Safe to do static_cast as `updated_index >= 0` is verified first
-         FC_ASSERT(updated_index >= 0 && (static_cast<Container<T>::size_type>(updated_index) < container.size()), "diff.remove_indexes index ${idx} + offset ${o} not in range ${s}",
-                   ("idx", index)("o", offset)("s", container.size()));
+      for (typename Container<T>::size_type i = 0; i < diff.remove_indexes.size(); ++i) {
+         FC_ASSERT(i == 0 || diff.remove_indexes[i] > diff.remove_indexes[i-1],
+                   "diff.remove_indexes not strictly monotonically increasing: current index ${c}, previous index ${p}",
+                   ("c", diff.remove_indexes[i])("p", diff.remove_indexes[i-1]));
+
+         assert(diff.remove_indexes[i] >= i);
+         assert(std::numeric_limits<typename Container<T>::size_type>::min() + i <= diff.remove_indexes[i]); // check for underflow of diff.remove_indexes[i] - i
```
greg7mdp, line 118: Because of the assert at line 117, there can be no underflow.
Furthermore `std::numeric_limits<typename Container<T>::size_type>::min()` is zero if the type is unsigned, which you should require for this class.

heifner (approved):

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -26,10 +26,14 @@ namespace fc {
  * @param Container container type for ordered diff and for diff_result
  */
 template <typename T, typename SizeType = size_t, template<typename Y, typename...> typename Container = std::vector>
-requires std::equality_comparable<T> && std::random_access_iterator<typename Container<T>::iterator>
+requires std::equality_comparable<T>
+         && std::random_access_iterator<typename Container<T>::iterator>
+         && std::is_unsigned_v<SizeType>
```
greg7mdp, line 31: Since we use `Container<T>::size_type` as type of the index of our loops, let's require that it is an unsigned type as well:
`&& std::is_unsigned_v<typename Container<T>::size_type>`

This will ensure we don't compare signed and unsigned values.

libraries/libfc/include/fc/container/ordered_diff.hpp
```diff
@@ -106,16 +110,24 @@ class ordered_diff {
...
       // Insert into the source based on diff.insert_indexes
-      for (auto& [index, value] : diff.insert_indexes) {
+      for (container_size_type i = 0; i < diff.insert_indexes.size(); ++i) {
+         FC_ASSERT(i == 0 || diff.insert_indexes[i].first > diff.insert_indexes[i-1].first,
+                   "diff.insert_indexes not strictly monotonically increasing: current index ${c}, previous index ${p}",
+                   ("c", diff.insert_indexes[i].first)("p", diff.insert_indexes[i-1].first));
+         auto& [index, value] = diff.insert_indexes[i];
          FC_ASSERT(index <= container.size(), "diff.insert_indexes index ${idx} not in range ${s}",
                    ("idx", index)("s", container.size()));
          container.insert(container.begin() + index, std::move(value));
```
greg7mdp, line 133: that `std::move` seems incorrect to me, since we accept `diff` as a forwarding reference.
I think we should `std::forward` `diff` to a local value, and then use that.

greg7mdp (changes requested):

arhag (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Assert non-negative index in `apply_diff`.
Note:end

greg7mdp (approved):

PR #726, opened 2024-09-09
[1.0.1] Add tests for --force-all-checks option in replay

linh2931:
Add tests for --force-all-checks option in replay.

As expected, those tests currently fail pending to https://github.com/AntelopeIO/spring/pull/719. Will be merged in after it.


tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -275,7 +276,10 @@ def executeSnapshotBlocklogTest(cluster, testNodeId, resultMsgs, nodeArgs, termA
     replayNodeosArgs = {
         5 : "--read-mode irreversible",
         6 : "--read-mode head",
-        7 : "--read-mode speculative"
+        7 : "--read-mode speculative",
+        8 : "--read-mode irreversible --force-all-checks",
+        9 : "--read-mode head --force-all-checks",
+        10 : "--read-mode speculative --force-all-checks"
```
heifner, line 282: This verifies it doesn't SEGFAULT or error out. But it does not verify that all checks were actually performed.

heifner (changes requested):

tests/nodeos_read_terminate_at_block_test.py
heifner: This type of test will come in via #727

heifner (approved):

greg7mdp (approved):

PR #732, opened 2024-09-10
[1.0.1] Add ability to deep-mind log to stderr

### LINKED ISSUE #724: When starting firehose from a snapshot on kylin, nodeos terminates with `DMLOG FPRINTF_FAILURE_TERMINATED`
When starting firehose from a snapshot on kylin, nodeos terminates with 

```
Sep  9 14:30:37 kylin-sfdm42 nodeos[2291398]: DMLOG FPRINTF_FAILED failed written=0 remaining=28 1 Bad file descriptor
Sep  9 14:30:37 kylin-sfdm42 nodeos[2291398]: DMLOG FPRINTF_FAILURE_TERMINATED
```

Nodeos will exit and if restarted will continue where it left off, so eventually it stops exiting and runs live against latest blocks.
(note there is a bug currently in firehose-core where firehose doesn't exit on EOF, this is already fixed and will be merged into future version of fireantelope... ref https://github.com/streamingfast/firehose-core/pull/66)

This nodeos stopping seems reproducible on spring 1.0-final and 1.0-rc3, but not on 1.0-rc2.

My crude understanding of how this is meant to work is that if the nodeos STDOUT buffer gets full it it is supposed to apply backpressure to nodeos to stop syncing blocks and wait until STDOUT buffer empties enough to continue syncing.  Maybe some recent refactoring with forked blocks handling changed behaviour?

nodeos config.ini:

```
... (usual stuff)
deep-mind = true
contracts-console = true
api-accept-transactions = false
```

fireantelope config.yml:

```
start:
  args:
    - reader-node-stdin
  flags:
    log-verbosity: 0
    log-to-file: false
```

heifner:
The theory is that #724 is caused by too much buffered output to the pipe. Add ability to write to stderr which is unbuffered.

Add an example `logging.json` that can be used to configure deep-mind to log to stderr.

Seems to resolve #724

### DIFF
--- a/libraries/libfc/src/log/dmlog_appender.cpp
+++ b/libraries/libfc/src/log/dmlog_appender.cpp
@@ -24,9 +24,13 @@ namespace fc {
 
    dmlog_appender::dmlog_appender( const std::optional<dmlog_appender::config>& args )
    :dmlog_appender(){
-      if (!args || args->file == "-")
+      if (!args || (args->file == "-" || args->file == "-stdout"))
       {
          my->out = stdout;
+      }
+      else if (args && args->file == "-stderr")
+      {
+         my->out = stderr;
       }
       else
       {
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3427,15 +3427,13 @@ struct controller_impl {
          }
 
          if (auto* dm_logger = get_deep_mind_logger(false)) {
-            // forkdb not available during replay
             block_handle_accessor::apply<void>(chain_head,
                         [&](const block_state_legacy_ptr& head) {
                            if (head->block->contains_header_extension(finality_extension::extension_id())) {
-                              block_num_type lib = replaying ? chain_head.irreversible_blocknum() : fork_db_root_block_num();
                               auto bsp = get_transition_savanna_block(head);
                               assert(bsp);
                               assert(bsp->active_finalizer_policy);
-                              dm_logger->on_accepted_block_v2(head->id(), lib, head->block,
+                              dm_logger->on_accepted_block_v2(head->id(), fork_db_root_block_num(), head->block,
                                                               bsp->get_finality_data(),
                                                               bsp->active_proposer_policy,
                                                               finalizer_policy_with_string_key{*bsp->active_finalizer_policy});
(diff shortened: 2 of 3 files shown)


libraries/libfc/src/log/dmlog_appender.cpp
```diff
@@ -24,10 +24,14 @@ namespace fc {
 
    dmlog_appender::dmlog_appender( const std::optional<dmlog_appender::config>& args )
    :dmlog_appender(){
-      if (!args || args->file == "-")
+      if (!args || (args->file == "-" || args->file == "-stdout"))
```
greg7mdp, line 27: detail... I think the extra parenthesis are unneeded.

libraries/libfc/src/log/dmlog_appender.cpp
```diff
@@ -24,10 +24,14 @@ namespace fc {
...
       {
          my->out = stdout;
       }
+      else if (args && args->file == "-stderr")
```
greg7mdp, line 31: to be pedantic checking `args` is not needed here

greg7mdp (approved):

spoonincode (changes requested):
Adding the ability for deepmind to go to stderr seems useful but also seems like a new feature (i.e. not release branch material).

So far I've not seen any evidence that a backed up `FILE` buffer can cause this failure. From the testing I've done, when the buffer is full nodeos blocks as expected. Is there a way to reproduce this error just from buffering? The `EBADF` errno from the original report also would not immediately suggest a user-space buffering problem to me.

Since deepmind output is, I take it, always expected to be piped to some other process, maybe dmlog_appender should `fflush` after each output (or maybe more optimally at some other delineation point). That does seems like a reasonable thing we could add. But lack of that only means that worst case the consumer is a little "behind" the most recent deepmind output.

imo we need more understanding of the issue -- we can't just say deepmind only works properly on stderr instead of stdout.

libraries/libfc/src/log/dmlog_appender.cpp
heifner: Not needed, but I think it reads better.

heifner:
@spoonincode re-targeted to `main`.

spoonincode (approved):

ericpassmore:
Note:start
category: Logging
component: DeepMind
summary: Add the ability for deepmind to go to stderr, which is unbuffered.
Note:end

PR #733, opened 2024-09-10
[1.0.1] Flush deep-mind logger after each message

### LINKED ISSUE #724: When starting firehose from a snapshot on kylin, nodeos terminates with `DMLOG FPRINTF_FAILURE_TERMINATED`

heifner:
`fflush` each deep-mind log message.

Attempt to resolve #724, reported from issue's creator that this fixes the issue they were seeing.

### DIFF
--- a/libraries/libfc/src/log/dmlog_appender.cpp
+++ b/libraries/libfc/src/log/dmlog_appender.cpp
@@ -11,8 +11,7 @@
 #include <boost/thread/mutex.hpp>
 #include <fc/exception/exception.hpp>
 #include <iomanip>
-#include <mutex>
-#include <sstream>
+#include <cstdio>
 
 namespace fc {
    class dmlog_appender::impl {
@@ -93,5 +92,8 @@ namespace fc {
          message_ptr = &message_ptr[written];
          remaining_size -= written;
       }
+      // attempt a flush, ignore any error
+      if (!my->is_stopped)
+         fflush(my->out);
    }
 }


spoonincode (approved):
I am extremely skeptical this fixes any `EBADF` failure. If it's empirically behaving better it's probably just some timing difference or such.

But it does seem like we should be `fflush()`ing to ensure the reader always gets the final log messages of a block promptly.

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: DeepMind
summary: Flush DeepMind Logger after each message.
Note:end

PR #734, opened 2024-09-10
[1.0.1 -> main] Log calculated LIB for deep-mind

### LINKED ISSUE #728: Undefined behavior replaying with deep-mind savanna blocks
Replaying with deep-mind logging enabled accesses the fork database which is not allowed during replay. Need to log the calculated LIB instead of the `fork_db_root_block_num` for the deep-mind logger. 

Also should add debug `assert`s to all forkdb access so that `asserton` run of tests will catch similar issues.

heifner:
For the deep-mind log, log the caculated LIB instead of the node's realized LIB. This more closely reflects legacy behavior and is more appropriate for deep-mind logging as it cares more about the state of the chain than any particular node.

Also adds debug `assert`s to all forkdb calls that expect fork database `root` to exist.

Merges `release/1.0` into `main` including #731 

Resolves #728

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1066,6 +1066,7 @@ struct controller_impl {
    }
 
    uint32_t fork_db_head_block_num() const {
+      assert(fork_db_has_root());
       return fork_db.apply<uint32_t>(
          [&](const auto& forkdb) {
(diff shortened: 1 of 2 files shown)


ericpassmore:
Note:start
category: Other
component: SHiP
summary: Log calculated LIB for deep-mind support.
Note:end

PR #735, opened 2024-09-10
[1.0.1] Small improvements left in PR 715 

linh2931:
There are several improvements identified in https://github.com/AntelopeIO/spring/pull/715 before and after it was merged in:
* Be explicit about return type to get return value optimization
* Avoid `qc->to_qc_claim()` in normal execution path
* Proper uses of `invalid_qc_claim` exception type

Completely resolves https://github.com/AntelopeIO/spring/issues/693


libraries/chain/controller.cpp
```diff
@@ -3967,7 +3967,7 @@ struct controller_impl {
                   "QC is_strong (${s1}) in block extension does not match is_strong_qc (${s2}) in header extension. Block number: ${b}",
```
greg7mdp, line 3967: On line 3965, I think `strength` is the right term instead of `strictness`

greg7mdp (approved):

arhag (approved):

PR #736, opened 2024-09-10
Move verify_qc onto the chain thread pool

heifner:
During syncing one `net` thread was at 100% CPU. Moving the `verify_qc` onto the `chain-threads` thread pool gives about 10% CPU on a net thread and 50% on two `chain-threads`. Did not do detailed timing, but sync time improvement on my machine for 1000 block sync from ~ 3.5 seconds to ~3 seconds.

Creates the `block_state` and verifies producer signature on the `net` thread while the `qc` is being verified in the chain thread pool.


ericpassmore:
Note:start
category: Other
component: Internal
summary: Increase capacity during heavy sync by moving `verify_qc` onto the chain thread pool.
Note:end

libraries/chain/controller.cpp
```diff
@@ -3954,7 +3954,7 @@ struct controller_impl {
 
             // if previous block's header extension has the same claim, just return
             // (previous block already validated the claim)
-            return {};
+            return std::optional<qc_t>{};
```
greg7mdp, line 3957: I don't think this change is warranted. If we were returning a value in the optional, yes, but here we return an empty optional so there is no extra object construction.

libraries/chain/controller.cpp
```diff
@@ -4086,7 +4086,7 @@ struct controller_impl {
          }
       }
 
-      return {};
+      return std::optional<qc_t>{};
```
greg7mdp, line 4089: Same as above.

libraries/chain/controller.cpp
```diff
@@ -4128,11 +4130,19 @@ struct controller_impl {
       EOS_ASSERT( id == bsp->id(), block_validate_exception,
                   "provided id ${id} does not match block id ${bid}", ("id", id)("bid", bsp->id()) );
 
+
       if constexpr (is_proper_savanna_block) {
+         assert((qc && verify_qc_future.valid()) || (!qc && !verify_qc_future.valid()));
```
greg7mdp, line 4135: or 
```suggestion
         assert(!!qc == verify_qc_future.valid());
```

libraries/chain/controller.cpp
```diff
@@ -4101,10 +4101,12 @@ struct controller_impl {
       assert(is_proper_savanna_block == b->is_proper_svnn_block());
 
       std::optional<qc_t> qc = verify_basic_block_invariants(id, b, prev);
-
+      log_and_drop_future<void> verify_qc_future;
       if constexpr (is_proper_savanna_block) {
          if (qc) {
-            verify_qc(prev, *qc);
+            verify_qc_future = post_async_task(thread_pool.get_executor(), [this, &qc, &prev] {
+               verify_qc(prev, *qc);
```
greg7mdp, line 4108: What happens if there is an exception in `verify_qc()`. Will we just log it and go ahead and create a `block_state`?

libraries/chain/controller.cpp
heifner: Hmm, ok. Not sure that is more readable, but sure.

libraries/chain/controller.cpp
heifner: Yes, `block_state` will be created while the `verify_qc()` is running. If `verify_qc()` does throw then the created `block_state` is thrown away and not added to the fork database.

greg7mdp: > the created `block_state` is thrown away and not added to the fork database.

What makes this happen?

linh2931 (approved):

libraries/chain/controller.cpp
heifner: If `verify_qc` throws it is captured by the future. The future rethrows the exception, either here:
https://github.com/AntelopeIO/spring/blob/dbff5a6892bd98256242b26e5a5950690d07c057/libraries/chain/controller.cpp#L4137-L4137
Which will throw out of the method before `forkdb.add()`


Or if another exception is thrown, then the destructor of `log_and_drop_future` logs it. But the other exception that is thrown will also happen before line 4137 and will prevent the `block_state` from being added to the forkdb.

greg7mdp: Cool, thanks, learned something.

greg7mdp (approved):

PR #744, opened 2024-09-11
[1.0.1] Make signed_block_ptr const

heifner:
Change `signed_block_ptr` to be `std::shared_ptr<const signed_block>` instead of `std::signed_block>`.
Requires allowing `fc::unpack` of `std::shared_ptr<T>` to cast away const so it can be overwritten.
Didn't turn out too bad.


unittests/vote_processor_tests.cpp
```diff
@@ -67,15 +67,15 @@ auto create_test_block_state(const block_state_ptr& prev) {
    static block_timestamp_type timestamp;
    timestamp = timestamp.next(); // each test block state will be unique
    signed_block_ptr block = std::make_shared<signed_block>(prev->block->clone());
```
greg7mdp, line 69: Can you make this `auto` to avoid the following `const_pointer_cast`s?

greg7mdp (approved):

greg7mdp:
Maybe create a type `using mutable_signed_block_ptr = std::shared_ptr<signed_block>;`?

libraries/chain/block_state.cpp
```diff
@@ -80,7 +105,7 @@ block_state::block_state(const block_header_state&                bhs,
                          const block_signing_authority&           valid_block_signing_authority,
                          const digest_type&                       action_mroot)
    : block_header_state(bhs)
-   , block(std::make_shared<signed_block>(signed_block_header{bhs.header}))
+   , block()
```
greg7mdp, line 108: No big deal, but why even have this line?

libraries/chain/controller.cpp
```diff
@@ -5514,7 +5514,7 @@ bool controller::validated_block_exists(const block_id_type& id) const {
 
 std::optional<signed_block_header> controller::fetch_block_header_by_id( const block_id_type& id )const {
    auto sb_ptr = my->fork_db_fetch_block_by_id(id);
-   if( sb_ptr ) return std::optional<signed_block_header>{*static_cast<signed_block_header*>(sb_ptr.get())};
+   if( sb_ptr ) return std::optional<signed_block_header>{*static_cast<const signed_block_header*>(sb_ptr.get())};
```
greg7mdp, line 5517: I don't know why we need this case. This seems to work fine with clang:
```suggestion
   if( sb_ptr ) return std::optional<signed_block_header>{*sb_ptr};
```

unittests/vote_processor_tests.cpp
```diff
@@ -56,7 +56,7 @@ auto create_genesis_block_state() { // block 2
    genesis->block = block;
    genesis->activated_protocol_features = std::make_shared<protocol_feature_activation_set>();
    genesis->active_finalizer_policy = std::make_shared<finalizer_policy>(new_finalizer_policy);
-   genesis->block->previous = make_block_id(1);
+   std::const_pointer_cast<signed_block>(genesis->block)->previous = make_block_id(1);
```
greg7mdp, line 59:
```suggestion
   block->previous = make_block_id(1);
```

unittests/producer_schedule_tests.cpp
```diff
@@ -711,8 +711,9 @@ BOOST_AUTO_TEST_CASE( extra_signatures_test ) try {
       additional_sigs.emplace_back( remote.get_private_key("alice"_n, "bs4").sign(sig_digest) );
 
       // Serialize the augmented additional signatures back into the block extensions.
-      b->block_extensions.clear();
-      emplace_extension(b->block_extensions, additional_sigs_eid, fc::raw::pack( additional_sigs ));
+      std::const_pointer_cast<signed_block>(b)->block_extensions.clear();
```
greg7mdp, line 714: I think we should declare `b` as `mutable_signed_block_ptr` and make a deep copy on line 697:

`b = valid_block;` => `b = std::make_shared<block_ptr>(*valid_block);`

as it is probably not a good idea to modify the block returned by `produce_block`.

An then the cast below is not needed.

libraries/chain/block_state.cpp
heifner: Documentation. Makes it clear what the `std::move` is replacing below.

libraries/chain/controller.cpp
heifner: I think it is more documentation that a slice is happening.

unittests/vote_processor_tests.cpp
heifner: This one is block_state->block so it is needed.

unittests/producer_schedule_tests.cpp
heifner: We could, but the test is about adding an extra signature to an existing block. Seems more in the spirit of the test to manipulate the block itself.

libraries/chain/controller.cpp
greg7mdp: To me it feels like unnecessary obfuscation. I don't think making the code unnecessarily more complex is usually a good idea.

heifner: I think making the slice obvious is being less obfuscated🤷. Personally I prefer if the language made you explicitly say you wanted a slice.

unittests/vote_processor_tests.cpp
greg7mdp: I don't think so, `block` is `auto block = std::make_shared<signed_block>();`

It becomes a `signed_block_ptr` when the shared_ptr is copied to `genesis`, but `block` is still mutable as line 40 demonstrates.

heifner: This is `genesis->block`. so `block_state::block` which is a `signed_block_ptr`

```
/home/heifner/ext/leap/unittests/vote_processor_tests.cpp:59:29: error: no viable overloaded '='
   genesis->block->previous = make_block_id(1);
   ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~
/home/heifner/ext/leap/libraries/libfc/include/fc/crypto/sha256.hpp:13:7: note: candidate function (the implicit copy assignment operator) not viable: 'this' argument has type 'const block_id_type' (aka 'const fc::sha256'), but method is not marked const
class sha256 
      ^
/home/heifner/ext/leap/libraries/libfc/include/fc/crypto/sha256.hpp:13:7: note: candidate function (the implicit move assignment operator) not viable: 'this' argument has type 'const block_id_type' (aka 'const fc::sha256'), but method is not marked const

```

heifner: .. but can do it before the assignment.

greg7mdp (approved):

unittests/vote_processor_tests.cpp
heifner: Which is what you meant all along probably. Sorry.

greg7mdp: > genesis->block->previous = make_block_id(1);

That was not my suggestion :-)

unittests/producer_schedule_tests.cpp
arhag: I like Greg's suggestion.

The test is taking a copy of a valid block from the `remote` node, corrupting it, and then pushing the corrupted block to the `main` node.

So, I don't see what Greg suggested as going against the spirit of this test.

greg7mdp (approved):
Nice!

arhag (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Cleaner code by making `signed_block_ptr` const. 
Note:end

PR #761, opened 2024-09-12
[1.0.1 -> main] Make signed_block_ptr const

heifner:
Change `signed_block_ptr` to be `std::shared_ptr<const signed_block>` instead of `std::signed_block>`.
Requires allowing `fc::unpack` of `std::shared_ptr<T>` to cast away const so it can be overwritten.
Didn't turn out too bad.

Merges `release/1.0` into `main` including #744


arhag (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Cleaner code by making `signed_block_ptr` const. 
Note:end

PR #762, opened 2024-09-12
remove EOS VM OC workaround for LLVM 7.0.x ABI mixup

spoonincode:
Remove an old workaround for LLVM 7.0.x that resolved run time crashes due to ABI breakage inside of LLVM when spring was compiled with clang and LLVM was compiled with gcc.

This workaround was useful in the EOSIO 2.0 era because of RHEL7 support: the only official-ish package for an LLVM version compatible with EOSIO 2.0 was from SCL with just LLVM 7.0.1 (as a 'system' package of course this was compiled with gcc). But during this era of EOSIO we would often recommend building EOSIO was clang. Thus.. runtime breakage would occur in this configuration without the workaround.

We can remove this old workaround now.


greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
category: Other
component: WebAssembly
summary: Remove unneeded LLVM v7 workaround.
Note:end

PR #763, opened 2024-09-12
[1.0.1 -> main] Update v1 fsi reference file with the one generated by Spring built at tag `v1.0.0`

greg7mdp:
Note that the file is 9 bytes bigger, because it save 3 fsi, each having a 4 byte timestamp instead of a 1 byte bool.


linh2931 (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Update test data to match current version of Finality Safety Information.
Note:end

PR #765, opened 2024-09-12
[1.0.1 -> main] Fix pause production on missed round

### LINKED ISSUE #712: Production pause for one block after prior BP misses round

heifner:
Change the implicit pause production due to votes algorithm:

Track four timestamps: 
- producer vote `latest_vote`, `first_block_after_vote`
- other votes `latest_vote`, `first_block_after_vote`

All default to value of the minimum representable time.

See new `production_pause_vote_tracker` for how these timestamps are tracked and processed.

Includes a test to verify that producer is not paused if the producer before it misses their round.
Example failure of test before fix: https://github.com/AntelopeIO/spring/actions/runs/10739300625/job/29785359524?pr=713

Also make `production_pause_vote_timeout.py` less flaky by having the center node connect to the peers instead of the peers connecting to the center node. When the center node is restarted you want it to as quickly as possible reconnect to its peers instead of waiting for the peers connection timeout for them to reconnect to the center node.

Merges `release/1.0` into `main` including #713

Resolves #712


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
summary: Change the implicit pause production algorithm incorporating new timestamps. This ensures producer continues without a pause following a missed round by the previous producer.
Note:end

Issue #727, opened 2024-09-09
Add unit test to verify replay of blocks with bad QC data


arhag:
Adds tests to check that the changes of https://github.com/AntelopeIO/spring/issues/693 are implemented correctly.

Bad QC data includes:
+ Invalid QC claim (one that goes backward for example).
+ Irrelevant QC (claims a block number other than the one claimed in the block header, or mismatching weak/strong claim)
+ Bad QC (signature validation fails)

Note that the last one should fail to replay only if `--force-all-checks` is not provided. The other two should always fail to replay.

PR #766, opened 2024-09-12
[1.0.1 -> main] Add unit tests to verify replay of blocks with bad QC data

linh2931:
Forwards https://github.com/AntelopeIO/spring/pull/759

Adds tests for QC validation during replay of blocks (implemented by https://github.com/AntelopeIO/spring/issues/693).

Cases are tested:
+ Invalid QC claim (one that goes backward for example). Replay always fails.
+ Irrelevant QC (claims a block number other than the one claimed in the block header, or mismatching weak/strong claim). Replay always fails.
+ Bad QC (signature validation fails). Replay fails only if `--force-all-checks` is provided.

Resolves https://github.com/AntelopeIO/spring/issues/727

### DIFF
--- a/unittests/replay_block_invariants_tests.cpp
+++ b/unittests/replay_block_invariants_tests.cpp
@@ -1,0 +1,247 @@
+#include <eosio/testing/tester.hpp>
+#include <boost/test/unit_test.hpp>
+
+// Test scenarios
+//    * replay through a block containing an invalid QC claim block num (backward).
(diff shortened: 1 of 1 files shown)


arhag (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Add unit tests to verify replay of blocks with bad QC data.
Note:end

heifner (approved):

Issue #635, opened 2024-08-26
Node requesting blocks way ahead of forkdb head


heifner:
`requesting range 157319140 to 157322736, fhead 157238220, lib 157238217`

Note that is some 80K blocks ahead.

Looks like this: https://github.com/AntelopeIO/spring/pull/572#discussion_r1720822685 is unconstrained and allows net_plugin to keep requesting as the node catches up.

```
debug 2024-08-26T14:03:36.311 net-3     net_plugin.cpp:1185           operator()           ] ["89.38.98.37:9876 - cbb90ba" - 1 89.38.98.37:9876] handle notice_message
debug 2024-08-26T14:03:36.311 net-3     net_plugin.cpp:3625           handle_message       ] ["89.38.98.37:9876 - cbb90ba" - 1 89.38.98.37:9876] this is a last irreversible notice with 157731556 pending blocks: 157731556 d247112a413f0ac3...
debug 2024-08-26T14:03:36.311 net-3     net_plugin.cpp:3645           handle_message       ] ["89.38.98.37:9876 - cbb90ba" - 1 89.38.98.37:9876] this is a last irreversible notice with 157731556 blocks
debug 2024-08-26T14:03:36.311 net-3     net_plugin.cpp:2437           sync_recv_notice     ] ["89.38.98.37:9876 - cbb90ba" - 1 89.38.98.37:9876] sync_manager got last irreversible block notice
debug 2024-08-26T14:03:36.311 net-3     net_plugin.cpp:2200           is_sync_required     ] last req = 157319139, last recv = 157319140 known = 157731554 our fhead = 157238220
debug 2024-08-26T14:03:36.311 net-3     net_plugin.cpp:2132           request_next_chunk   ] sync_last_requested_num: 157319139, sync_next_expected_num: 157319140, sync_known_lib_num: 157731554, sync-fetch-span: 5000, fhead: 157238220, lib: 157238217
info  2024-08-26T14:03:36.311 net-3     net_plugin.cpp:2137           request_next_chunk   ] ["89.38.98.37:9876 - cbb90ba" - 1 89.38.98.37:9876] Catching up with chain, our last req is 157319139, theirs is 157731554, next expected 157319140, fhead 157238220
warn  2024-08-26T14:03:36.311 net-3     net_plugin.cpp:2018           active_sync_fetch_sp ] sync-fetch-span 5000 restricted to 3597 by max-reversible-blocks, fork_db_size 3
[...]

PR #768, opened 2024-09-12
[1.0.2] Remove explicit net_plugin pause during snapshot

### LINKED ISSUE #592: Do not timeout P2P connections when creating a snapshot
Creating a snapshot ties up the main thread for the time it takes to create the snapshot. During that time, new blocks can't be processed by the main thread. This can cause the net_plugin's sync timer and/or block timer to fire and disconnect from its peers. The connection timer then has to fire for the connections to be re-established. This can cause a node that is syncing and periodically creating snapshots to pause after each snapshot waiting for connections to be re-established and syncing to continue.

heifner:
As noted on the addition of pausing during snapshot:
https://github.com/AntelopeIO/leap/pull/2029/files#r1440851488

Pausing does cause peers to disconnect. You have to wait on the connection timer to reconnect and continue syncing. 

Now that we no longer sync ahead: #635 
The issue https://github.com/AntelopeIO/leap/issues/2021 is not an issue anymore as we do not keep syncing ahead. The sync is limited by the `sync-fetch-span`. 

Tested with EOS Mainnet node syncing and with an EOS Mainnet node in-sync.

Resolves #592

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -553,14 +553,6 @@ namespace eosio {
       void start_expire_timer();
       void start_monitors();
 
-      // we currently pause on snapshot generation
-      void wait_if_paused() const {
-         controller& cc = chain_plug->chain();
-         while (cc.is_writing_snapshot()) {
-            std::this_thread::sleep_for(std::chrono::milliseconds(10));
-         }
-      }
-
       void expire();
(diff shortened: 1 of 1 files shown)


ericpassmore:
Note:start
category: Tests
summary: Continue P2P connections while performing a snapshot.
Note:end

greg7mdp (approved):

PR #777, opened 2024-09-13
[1.0.2] P2P: Allow irreversible mode syncing to continue when LIB is paused

### LINKED ISSUE #773: Kylin sync stuck
Trying to sync kylin from genesis with spring-1.0.1 release.  It is  stuck at the point LIB does not advance for a bunch of blocks. 

image

heifner:
- Allow a node to continue syncing ahead when LIB is paused. This is important when running in `irreversible` mode as the `chain_head` does not advance until LIB advances. 
- Added some additional debug logging to net_plugin. Also added an info level log when it allows a node to sync ahead past `sync-fetch-span` of head because LIB is paused.
- Add new test cases for startup-catchup test.
  - Sync with a sync-fetch-span of 1 while in irreversible mode.
    - Note this only tripped the issue in Savanna mode as the test doesn't run long enough to trigger a LIB catchup in legacy mode. It does a HEAD catchup in legacy mode (at least in my local testing). I didn't modify the test to produce more blocks to force a LIB catchup in legacy mode as the test does already trigger LIB catchup in Savanna mode.

Tested syncing Kylin past the large LIB pause.

Resolves #773


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2182,14 +2182,27 @@ namespace eosio {
...
+   // called from c's connection strand
    bool sync_manager::is_sync_request_ahead_allowed(block_num_type blk_num) const REQUIRES(sync_mtx) {
       if (blk_num >= sync_last_requested_num) {
          // do not allow to get too far ahead (sync_fetch_span) of chain head
          // use chain head instead of fork head so we do not get too far ahead of applied blocks
          uint32_t head = my_impl->get_chain_head_num();
          if (blk_num < head + sync_fetch_span)
             return true;
+
+         // might be in irreversible mode
```
linh2931, line 2194: Do you want to check if it is really in LIB mode?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2182,14 +2182,27 @@ namespace eosio {
...
+         controller& cc = my_impl->chain_plug->chain();
+         auto calculated_lib = cc.fork_db_head().irreversible_blocknum();
+         if (calculated_lib <= my_impl->get_chain_lib_num()) {
+            fc_ilog(logger, "sync ahead allowed past sync-fetch-span ${sp} for paused LIB ${l}, forkdb size ${s}",
+                    ("sp", sync_fetch_span)("l", calculated_lib)("s", cc.fork_db_size()));
+            return true;
+         }
```
linh2931, line 2201: You can also dlog here why sync ahead not allowed.

plugins/net_plugin/net_plugin.cpp
heifner: I was originally going to let `net_plugin` determine if in `irreversible` mode, but it is not needed. This condition can't be true unless we are in `irreversible` mode. If it somehow was true we would want to also allow syncing ahead.

plugins/net_plugin/net_plugin.cpp
heifner: Done

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2182,14 +2184,37 @@ namespace eosio {
...
+         uint32_t head_num = my_impl->get_chain_head_num();
+         if (blk_num < head_num + sync_fetch_span)
             return true;
+
+         // might be in irreversible mode
+         controller& cc = my_impl->chain_plug->chain();
+         if (cc.get_read_mode() == db_read_mode::IRREVERSIBLE) {
+            // chain head == lib == fork_db_root in irreversible
+            auto forkdb_head = cc.fork_db_head();
+            if (forkdb_head.block_num() - head_num >= sync_fetch_span) {
```
linh2931, line 2201: Add a small assert ensuring `forkdb_head.block_num() >= head_num`

heifner: It is possible we catch forkdb in the middle of a fork switch. We could guard against that or let this if pass, worse case it allows an extra request of blocks. Although to avoid confusion likely better to make it `if (forkdb_head.block_num() > head_num && forkdb_head.block_num() - head_num >= sync_fetch_span) {`

linh2931 (approved):

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2182,14 +2184,38 @@ namespace eosio {
...
+         uint32_t head_num = my_impl->get_chain_head_num();
+         if (blk_num < head_num + sync_fetch_span)
             return true;
+
+         // might be in irreversible mode
+         controller& cc = my_impl->chain_plug->chain();
+         if (cc.get_read_mode() == db_read_mode::IRREVERSIBLE) {
+            // chain head == lib == fork_db_root in irreversible
+            auto forkdb_head = cc.fork_db_head();
+            if (forkdb_head.block_num() > head_num && forkdb_head.block_num() - head_num >= sync_fetch_span) {
```
greg7mdp, line 2201: why do we need `&& forkdb_head.block_num() - head_num >= sync_fetch_span`?

If lib doesn't advance, i.e. `calculated_lib == head_num` below, wouldn't this be already covered by the test above at line 2193?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2208,7 +2234,10 @@ namespace eosio {
          return;
       }
 
-      if( sync_state != lib_catchup || !sync_recently_active()) {
+      stages current_sync_state = sync_state;
```
greg7mdp, line 2237: why this new variable?

heifner: Because it is logged and it is an atomic.

plugins/net_plugin/net_plugin.cpp
heifner: Line 2193 is comparing `head_num` which is the same as `lib` and `fork_db_root` in irreversible mode. This is comparing the last block in the forkdb on the best head to what has been applied. If that is not greater than `sync_fetch_span` then we don't want more blocks.

ericpassmore:
Note:start
category: Other
summary: P2P: Allow irreversible mode syncing to continue when LIB is paused.
Note:end

plugins/net_plugin/net_plugin.cpp
greg7mdp: Isn't `blk_num` the last block we received, so similar to `forkdb_head`?

If so, `blk_num < head_num + sync_fetch_span` is similar to `forkdb_head.block_num() < head_num + sync_fetch_span`

plugins/net_plugin/net_plugin.cpp
greg7mdp: This is not done consistently, not [here]( https://github.com/AntelopeIO/spring/blob/dab3534bdb5202a8ce288eaaded4c137456dc91a/plugins/net_plugin/net_plugin.cpp#L2012-L2019) for example

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2182,14 +2184,38 @@ namespace eosio {
               sync_next_expected_num < sync_last_requested_num );
    }
 
+   // called from c's connection strand
    bool sync_manager::is_sync_request_ahead_allowed(block_num_type blk_num) const REQUIRES(sync_mtx) {
```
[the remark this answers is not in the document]
heifner: It doesn't wait till we have 0 blocks that can be applied before requesting more. But I reworked it to use your `num_blocks_not_applied` and `num_blocks_that_can_be_applied` variables to be clearer.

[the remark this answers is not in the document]
greg7mdp: > It doesn't wait till we have 0 blocks that can be applied before requesting more.

I'm pretty sure it did. Before if would not return true unless the following condition is true:

`if (calculated_lib <= head_num) {`

which occurs only when we have applied all the blocks possible.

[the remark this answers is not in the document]
heifner: In the case of LIB stuck, yes. But in the nominal case it would return `true` via the `num_block_not_applied` `if`.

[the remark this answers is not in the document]
greg7mdp: correct, that's the case I was referring to. 

When lib is not stuck this second part is unnecessary anyways, since `num_blocks_not_applied` and `num_blocks_that_can_be_applied` are essentially the same (modulo the number of blocks that lib trials back for).

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2174,14 +2176,42 @@ namespace eosio {
...
       if (blk_num >= sync_last_requested_num) {
          // do not allow to get too far ahead (sync_fetch_span) of chain head
          // use chain head instead of fork head so we do not get too far ahead of applied blocks
-         uint32_t head = my_impl->get_chain_head_num();
-         if (blk_num < head + sync_fetch_span)
+         uint32_t head_num = my_impl->get_chain_head_num();
+         if (blk_num < head_num) { // avoid underflow on blk_num - head_num
+            return true;
+         }
+         block_num_type num_blocks_not_applied = blk_num - head_num;
```
greg7mdp, line 2188:
```suggestion
         block_num_type num_blocks_not_applied = blk_num > head_num ? blk_num - head_num : 0;
```

greg7mdp (approved):

linh2931 (approved):

Issue #776, opened 2024-09-13
trace-history/chain-state-history option of state_history_plugin breaking with v1.0.1's snapshot-import/node running


jun0tpyrc:
did like 30 times using different snapshot sources to confirm this 

Both of these flags are not safe to add in import/migrate process
```
trace-history = true 
chain-state-history = true   
```
not ok to add back afterwards as well
we using them as suggested for hyperion indexer setup

What's worse is as long as `chain-state-history = true` is involved , it will not print error detais but only the `appbase quit called` result ... 

Otherwise( just using `trace-history = true` ) , could see
```
wrote != -1: write failure on file /usr/opt/eosio/data/state-history/trace_history.log: Operation not supported
    {"fn":"/usr/opt/eosio/data/state-history/trace_history.log","e":"Operation not supported"}
    nodeos  random_access_file.hpp:136 write_to

info  2024-09-13T03:56:48.831 nodeos    main.cpp:172                  operator()           ] appbase quit called
warn  2024-09-13T03:56:48.831 nodeos    controller.hpp:493            emit                 ] /__w/spring/spring/libraries/chain/controller.cpp:3419 controller_emit_signal_exception: 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:153 on_accepted_block

error 2024-09-13T03:56:48.831 nodeos    controller.cpp:3809           apply_block          ] e.to_detail_string(): 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:153 on_accepted_block

error 2024-09-13T03:56:48.831 nodeos    controller.cpp:4473           operator()           ] exception thrown while switching forks 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:153 on_accepted_block

    {}
    nodeos  controller.cpp:3817 apply_block
```

jun0tpyrc:
tested with 1.0.0/1.0.1

jun0tpyrc:
mainnet only , jungle4 testnet is fine

spoonincode:
> `Operation not supported`

Are you running on Ubuntu 18? Leap 4.0 was the last version to support Ubuntu 18.

[...]

PR #780, opened 2024-09-13
[1.0.2] non-atomic append fallback to `random_access_file` for old kernels

spoonincode:
`random_access_file`, used for state_history in Spring, makes use of a kernel feature added in 4.16. I didn't really think much of that requirement because 4.16 was released well over 6 years (before EOS genesis!), and is well outside the scope of our officially supported platforms.

Yet we've now seen two reports of users stumbling on this:
* A user running Ubuntu 18.04[^1], which for non-desktop installs will use 4.15 by default. It's possible to opt in to the HWE kernel (default on desktop installs) that moves up to 5.4. This was seen on Telegram somewhere I can't find the link at the moment.
* A user running a Ubuntu 22.04 container on Amazon Linux 2. AL2[^2] actually has three supported kernels: 4.14, 5.10, and 5.15, with 5.10 being the default for new installs now. But the user had an old configuration still set to 4.14. Seen in #776

state_history doesn't actually need this atomic append, the implementation was just done this way for cleanliness, conciseness, and performance. So adding a fallback for old kernels here doesn't impact correctness for state_history. Still not really sure I want to add this additional chatter to the code though.

[^1]: I regret a little not bumping the reproducible build's runtime requirement before 1.0.0 release so that it simply would not work on Ubuntu 18 -- other projects have made this bump recently, for example `https://github.com/bitcoin/bitcoin/pull/29987`
[^2]: AL2 is in its 11th hour of support. It was originally scheduled to be EOL July 2023 but they extended it to July 2025.

### DIFF
--- a/libraries/libfc/include/fc/io/random_access_file.hpp
+++ b/libraries/libfc/include/fc/io/random_access_file.hpp
@@ -124,14 +124,17 @@ struct random_access_file_context {
 
          ssize_t wrote = -1;
          do {
-            if(offs == append_t)
+            if(offs == append_t) {
 #ifdef __linux__
                wrote = pwritev2(fd, iov, i, 0, RWF_APPEND);   //linux *not* opened with O_APPEND, appending requires special flag
+               if(wrote == -1 && errno == EOPNOTSUPP)         //fallback for kernels before 4.16
+                  wrote = pwritev(fd, iov, i, size());
 #else
                wrote = writev(fd, iov, i);                    //opened with O_APPEND, just write
 #endif
-            else
+            } else {
                wrote = pwritev(fd, iov, i, offs);
+            }
          } while(wrote == -1 && errno == EINTR);
          FC_ASSERT(wrote != -1, "write failure on file ${fn}: ${e}", ("fn", display_path)("e", strerror(errno)));
 


greg7mdp (approved):

libraries/libfc/include/fc/io/random_access_file.hpp
```diff
@@ -124,14 +124,17 @@ struct random_access_file_context {
 
          ssize_t wrote = -1;
          do {
-            if(offs == append_t)
+            if(offs == append_t) {
 #ifdef __linux__
                wrote = pwritev2(fd, iov, i, 0, RWF_APPEND);   //linux *not* opened with O_APPEND, appending requires special flag
+               if(wrote == -1 && errno == EOPNOTSUPP)         //fallback for kernels before 4.16
+                  wrote = pwritev(fd, iov, i, size());
```
heifner, line 131: `pwritev2` ignores offset (except -1) when given `RWF_APPEND`.
`pwritev` you need size because you want to append. I agree it is cleaner and more concise. Is it actually more performant?

I wonder if `pwritev` should just be used.

heifner: Oh, never mind. It is more performant because you don't have to call `size()`.

heifner (approved):
When I saw the report about amznlinux2, I wondered if we shouldn't try to support it. This resolves #776 then. My vote would be to put it in 1.0.2.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Support Linux Kernel 4.15, by falling back to `random_access_file` for the older kernel.
Note:end

Issue #778, opened 2024-09-13
fork_db check prevents voting on blocks including a qc claim on a block not in fork_db.


greg7mdp:
While trying to reproduce the use case of issue #751:

```
/* -----------------------------------------------------------------------------------------------------
            Finality advancing past block claimed on alternate branch
            =========================================================

Time:        t1      t2      t3      t4      t5      t6      t7      t8
Blocks:
    B0 <---  B1 <--- B2 <--- B3 <-|- B4 <--- B5
                                  |
                                  \----------------- B6 <--- B7 <--- B8 <--- B9
QC claim:
           Strong          Strong  Strong  Strong  Strong   Weak   Strong  Strong
             B0              B1      B3      B4      B2      B6      B7      B8

Vote:                      Strong  Strong   Strong  Weak   Strong  Strong  Strong

                                                     ^
                                                     |
                                                 validating those weak votes on b2
                                                 would fail if nodes have received b4 and b5
                                                 which advanced lib to b3

    - Node D is isolated and has not seen B3, B4, and B5
    - it received B3 via push_block, (so it can make it its head and produce a child of B3), but has not
      received votes on b3 (only on b2), so b6 includes a strong QC on b2.
    - when b6 is pushed to A, B and C, qc validation should fail.
--------------------------------------------------------------------------------------------------------*/
```

I was unable to obtain a QC on `B6` (which itself has a QC claim on `B2`).

This is because finalizers who have voted on `B5` (top fork), have advanced `lib` to `B3`, and therefore `B2` is not in their fork_db anymore.

Unfortunately, we have the [following check]( https://github.com/AntelopeIO/spring/blob/0e4cb8ed0d490e6b2a9d64d33ad56bc005d36b40/libraries/chain/controller.cpp#L4244) (also [here](https://github.com/AntelopeIO/spring/blob/0e4cb8ed0d490e6b2a9d64d33ad56bc005d36b40/libraries/chain/controller.cpp#L4251)), which prevents these finalizers from voting.

This check was intended to find an ancestor block of the current block, with `block_num >= latest_qc_claim__block_ref.block_num()`, that is validated. However the check is incorrect and only does a lookup for `latest_qc_claim__block_ref.block_id`, which fails in this case. 

[...]

PR #782, opened 2024-09-13
[1.0.2] Add new testcase where finality advances on alternate branch with lesser lib

greg7mdp:
Add a testcase where finalizers cannot vote on a block from a different branch because they are locked on a block from their own branch.

```

Time:        t1      t2      t3      t4      t5      t6      t7   
Blocks:
                                  |
                                  \----------------- B6 <--- B7
QC claim:
           Strong          Strong  Strong  Strong  Strong   No QC
             B0              B1      B3      B4      B2     achieved

Vote:                      Strong  Strong   Strong  Weak     -

                                                     ^
                                                     |
                                                 which advanced lib to b3

    - when b6 is pushed to A, B and C, finalizers of A, B, and C are unable to vote on it, because they
      are locked on B4,
          -> liveness check fails because: `B6' s core.latest_qc_block_timestamp() <  fsi.lock.timestamp`
             because `B2 timestamp < B4 timestamp`.
          -> safety check fails because `B6` does not extend `B4`.
```

After voting on `B5` (which makes `B3` final), the finalizers who voted on `B5` are locked on `B4`, and therefore cannot vote on `B6`:

- liveness check fails because: `B6' s core.latest_qc_block_timestamp() <  fsi.lock.timestamp` because
 `B2 timestamp < B4 timestamp`.
- safety check fails because `B6` does not extend `B4`.

As a result, this testcase reproduces the scenario described above, and verifies that we lack votes to form a QC on `B6`. 

Although until issue #778 is addressed, the problem described in issue #778 is the cause of the lack of votes, and `decide_vote` is not even called.


ericpassmore:
Note:start
category: Tests
component: Internal
summary: New test case where finality advances on lesser LIB.
Note:end

linh2931 (approved):

unittests/savanna_cluster.hpp
```diff
@@ -127,6 +129,8 @@ namespace savanna_cluster {
 
       size_t& vote_delay() { return _vote_delay; }
 
+      void propagate_delayed_votes_to(node_t &to);
```
heifner, line 132: `node_t& to`

unittests/savanna_misc_tests.cpp
```diff
@@ -688,5 +687,106 @@ BOOST_FIXTURE_TEST_CASE(verify_spring_1_0_block_compatibitity, savanna_cluster::
...
+                                                 validating those weak votes on b2
+                                                 would fail if nodes have received b4 and b5
+                                                 which advanced lib to b3
+
+    - Node D is isolated and has not seen B3, B4, and B5
+    - it received B3 via push_block, (so it can make it its head and produce a child of B3), but has not
+      received votes on b3 (only on b2), so b6 includes a strong QC on b2.
+    - when b6 is pushed to A, B and C, qc validation would fail in Spring 1.0.0 because the fork_db
+      lookup, but succeeds in Spring 1.0.1 as only the finality core of the claiming block is needed
+      to validate the QC.
```
heifner, line 716: This comment needs to be updated for this test.

unittests/savanna_misc_tests.cpp
```diff
@@ -688,5 +687,106 @@ BOOST_FIXTURE_TEST_CASE(verify_spring_1_0_block_compatibitity, savanna_cluster::
...
+   // - liveness check fails because: `b6' s core.latest_qc_block_timestamp() <  fsi.lock.timestamp`
+   //   because `b2 timestamp < b4 timestamp`.
+   // - safety check fails because `b6` does not extend `b4`.
+   //
+   // As a result, we don't expect the next block (b7) to include a QC
+   // ---------------------------------------------------------------------------------------------------
+
+   auto b7 = D.produce_block();                         // D produces a block. It still has not seen b4 and b5.
+   print("b7", b7);
+   BOOST_REQUIRE(!qc(b7));                              // b7 should include a QC
```
heifner, line 787: Comment should be `b7 should not include a QC`, right?

heifner (changes requested):

unittests/savanna_misc_tests.cpp
greg7mdp: You're right, thanks. I also updated the PR description.

unittests/savanna_misc_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/782/commits/78c1053e75ccde144fc424429c38e4800d760aa9

unittests/savanna_misc_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/782/commits/78c1053e75ccde144fc424429c38e4800d760aa9

unittests/savanna_cluster.hpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/782/commits/41ceebcdd7364e21f84f613c872aa112d1845abe

heifner (approved):

Issue #612, opened 2024-08-21
P2P send signed_block directly from disk


heifner:
Currently, when a peer asks for a block:
- signed_block is deserialized off disk
- signed_block is serialized to net

This can be expensive and add up to considerable CPU, see #611 

For blocks retrieved from the block log this is unnecessary. Also note with Savanna, blocks are moved out of the fork database into the block log much quicker. The serialization format on disk is the same as on P2P.

Add a new controller interfaces: `fetch_serialied_block_by_number` and `fetch_serialized_block_by_id`. These would pull directly from the block log without having to deserialize/serialize or from the forkdb where serialization would be required. Use these new interfaces in `net_plugin` to provide blocks to peers.

arhag:
> or from the forkdb where serialization would be required

We could also cache the serialization in the fork database so it only has to happen once.

Issue #611, opened 2024-08-21
Reject P2P connections from leap after Savanna activated


heifner:
Any Leap node connected to network when it activates Savanna will be hard-forked out. That Leap node will repeatedly request blocks from its peers because it will not be able to process received blocks.

After Savanna is activated (one proper Savanna block received), `net_plugin` could reject connections from net versions pre `proto_savanna`. Then the node would not waste time sending blocks to a node that can't process them.

matthewdarwin:
jungle-dm12 - leap 5.0 
jungle-trxrly173 - spring 1.0-rc2

from leap 5.0 side:

![image](https://github.com/user-attachments/assets/aae36714-c842-4a28-bb57-f610fce35439)


from spring 1.0 side:

![image](https://github.com/user-attachments/assets/07f8dbf4-09ca-4696-9f5a-b8bc1a56f0e5)

bhazzard:
This should be generalized for future hard fork releases to the case where new protocol features are activated, rejecting p2p connections from nodes without those protocol features.

PR #785, opened 2024-09-16
P2P: Improve performance by not serializing and deserializing blocks from block log

linh2931:
Currently, when a peer asks for a block, if the block is retrieved from the block log, `net plugin`
- reads the block form the block log and deserializes it
- then serializes it before sending it to the peer

This can be expensive and add up to considerable CPU, see #611.

Those conversions are not necessary, as the serialization format in block log is the same as on P2P.

This PR adds `read_serialized_block_by_num()` to `block_log` and `fetch_serialized_block_by_number()` to `controller`, and updates `net plugin` to use the new interface.

Resolves https://github.com/AntelopeIO/spring/issues/612


libraries/chain/block_log.cpp
```diff
@@ -615,6 +628,68 @@ namespace eosio { namespace chain {
...
+                  // Just fall back to the regular `retry_read_block_by_num` and then serialize.
+                  auto ret = retry_read_block_by_num(block_num);
+                  if (ret) {
+                     return fc::raw::pack(*ret);
+                  } else {
+                     return {};
+                  }
+               }
+
+               if (verify_block_num == verify_block_num_t::yes) {
```
linh2931, line 646: Or should we always do a full deserialization here to make sure the block is valid but still return the serialized format? This matches the validation behavior of the original `read_block_by_num()` but still saves Net Plugin from doing serialization.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Improve P2P performance by maintaining serialized format from block log to peer syncing.
Note:end

libraries/chain/block_log.cpp
```diff
@@ -197,6 +197,23 @@ namespace eosio { namespace chain {
...
+      uint32_t get_block_num_at(fc::datastream<fc::cfile>& file, uint64_t position) {
+         // to derive blknum_offset==14 see block_header.hpp and note on disk struct is packed
+         //   block_timestamp_type timestamp;                  //bytes 0:3
+         //   account_name         producer;                   //bytes 4:11
+         //   uint16_t             confirmed;                  //bytes 12:13
+         //   block_id_type        previous;                   //bytes 14:45, low 4 bytes is big endian block number
+         //   of previous block
+
+         file.seek_end(0);
+         EOS_ASSERT(position <= file.tellp(), block_log_exception, "Invalid block position ${position}",
```
greg7mdp, line 209: This really should be:
```suggestion
         EOS_ASSERT(position +  blknum_offset + sizeof(uint32_t) <= file.tellp(), block_log_exception, "Invalid block position ${position}",
```

with `blknum_offset` declared before.

libraries/chain/block_log.cpp
```diff
@@ -474,6 +479,12 @@ namespace eosio { namespace chain {
          };
          std::optional<signed_block_with_id> head;
 
+         enum class verify_block_num_t {
+            yes, // Verify block number when read a serialized block.
+            no   // Does not verify block number when read a serialized block.
+                 // This is for the case where block id is vefified seperately.
```
greg7mdp, line 485:
```suggestion
            yes, // Verify the block number when reading a serialized block.
            no   // Do not verify the block number when reading a serialized block.
                 // This is for the case where the block id is verified separately.
```

libraries/chain/block_log.cpp
greg7mdp: Wouldn't a block in the block log be guaranteed to be valid?

libraries/chain/block_log.cpp
```diff
@@ -1230,6 +1305,32 @@ namespace eosio { namespace chain {
...
+   }
+
+   std::vector<char> block_log::read_serialized_block_by_id(const block_id_type& id) const {
+      std::lock_guard g(my->mtx);
+
+      auto block_num = block_header::num_from_id(id);
+
+      std::optional<signed_block_header> h = my->read_block_header_by_num(block_num);
+      if (h) {
+         auto returd_id = h->calculate_id();
```
greg7mdp, line 1320:
```suggestion
         auto read_id = h->calculate_id();
```

libraries/chain/block_log.cpp
greg7mdp: I don't understand whay we need this `verify_block_num`. I would think that the blocks in the block log are correct, and that therefore the block at position ` get_block_pos(block_num)` is guaranteed to have the block number `block_num`. Is that not the case?

libraries/chain/block_log.cpp
```diff
@@ -615,6 +628,68 @@ namespace eosio { namespace chain {
...
+                  // current block is not the last block in the log file.
+                  uint64_t next_block_pos = get_block_pos(block_num + 1);
+                  if (next_block_pos == block_log::npos) {
+                     wlog("no position found for block ${n}", ("n", block_num + 1));
+                     return {};
+                  }
+                  if (next_block_pos <= pos) {
+                     wlog("next block position ${nn} should be greater than current block position ${cn}", ("nn", next_block_pos)("cn", pos));
+                     return {};
+                  }
```
greg7mdp, line 669: I think this can be `assert(next_block_pos > pos);`

libraries/chain/block_log.cpp
```diff
@@ -615,6 +628,68 @@ namespace eosio { namespace chain {
...
+               uint32_t head_block_num = block_header::num_from_id(head->id);
+               constexpr uint32_t block_pos_size = sizeof(uint64_t); // size of block position field in the block log file
+
+               if (block_num < head_block_num) {
+                  // current block is not the last block in the log file.
+                  uint64_t next_block_pos = get_block_pos(block_num + 1);
+                  if (next_block_pos == block_log::npos) {
+                     wlog("no position found for block ${n}", ("n", block_num + 1));
+                     return {};
+                  }
```
greg7mdp, line 665: I think this can be: `assert(next_block_pos != block_log::npos);`

libraries/chain/block_log.cpp
linh2931: The original `read_block_by_num` checks `block_num`. I am little paranoid. Will discuss in design meeting.

libraries/chain/block_log.cpp
linh2931: This EOS_ASSERT was in the original function. Will correct it. Thanks.

libraries/chain/block_log.cpp
greg7mdp: still some typos remain.

linh2931: This section of code was removed.

libraries/chain/block_log.cpp
linh2931: It was decided to use EOS_ASSERT for preventing undefined behavior on our side and for consistency.

libraries/chain/block_log.cpp
linh2931: It was decided to use EOS_ASSERT.

libraries/chain/block_log.cpp
```diff
@@ -615,6 +628,68 @@ namespace eosio { namespace chain {
...
+               }
+
+               if (verify_block_num == verify_block_num_t::yes) {
+                  auto bnum = get_block_num_at(block_file, pos);
+                  if (block_num != bnum) {
+                     wlog("Wrong block num was read from block log. expected block_num: ${n}, returned block num: ${bnum}",
+                           ("n", block_num)("bnum", bnum));
+                     return {};
+                  }
+               }
```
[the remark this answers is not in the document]
linh2931: This code was removed.

libraries/chain/controller.cpp
```diff
@@ -5514,6 +5514,13 @@ signed_block_ptr controller::fetch_block_by_id( const block_id_type& id )const {
    return signed_block_ptr();
 }
 
+std::vector<char> controller::fetch_serialized_block_by_id( const block_id_type& id )const { try {
```
heifner, line 5517: We don't need the `fetch_serialized_block_by_id`. If we do in the future, I think it should verify the retrieved serialized block is the correct block id instead of just assuming the block_num of the id will retrieve the correct block.

heifner: See new PR https://github.com/AntelopeIO/spring/pull/803 that removes its usage.

heifner (changes requested):
Still reviewing, but thought I would get these comments out so you know why PR #803 was created.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1838,6 +1838,25 @@ namespace eosio {
...
+         const uint32_t payload_size = which_size + v.size();
+
+         const char* const header = reinterpret_cast<const char* const>(&payload_size); // avoid variable size encoding of uint32_t
+         const size_t buffer_size = message_header_size + payload_size;
+
+         auto send_buffer = std::make_shared<vector<char>>( buffer_size );
+         fc::datastream<char*> ds( send_buffer->data(), buffer_size );
+         ds.write( header, message_header_size );
+         fc::raw::pack( ds, unsigned_int( signed_block_which ) );
+         ds.write( v.data(), v.size() );
```
heifner, line 1855: Instead of `vector<char> read_serialized_block_by_num(block_num)` how about `void read_serialized_block_by_num(block_num, vector<char>& data)`. 

`read_serialized_block_by_num` would append to the `data` buffer. We can then avoid the copy here.

heifner (changes requested):
Still reviewing, but thought I would get these comments out so you know why PR #803 was created.

libraries/chain/block_log.cpp
```diff
@@ -245,10 +246,13 @@ namespace eosio { namespace chain {
             //   block_id_type        previous;                   //bytes 14:45, low 4 bytes is big endian block number
             //   of previous block
 
-            EOS_ASSERT(position <= size(), block_log_exception, "Invalid block position ${position}",
-                       ("position", position));
+            file.seek_end(0);
```
greg7mdp, line 249: Why did you change this to not use `size()`, and do a `seek_end()` instead?

libraries/chain/block_log.cpp
```diff
@@ -615,6 +621,63 @@ namespace eosio { namespace chain {
...
+                  auto ret = retry_read_block_by_num(block_num);
+                  if (ret) {
+                     return fc::raw::pack(*ret);
+                  } else {
+                     return {};
+                  }
+               }
+
+               uint32_t head_block_num = block_header::num_from_id(head->id);
+               EOS_ASSERT(block_num <= head_block_num, block_log_exception,
```
greg7mdp, line 640: Doesn't this assert change the previous behavior of  `connection::blk_send`? 

Before, it would: 
`peer_ilog( this, "fetch block by id returned null, id ${id}", ("id", blkid) );`

Now it will:
`peer_elog( this, "caught assert on fetch_serialized_block_by_id, ${ex}, id ${id}", ("ex", ex.to_string())("id", blkid) );`

I don't think Areg said we had to always use `EOS_ASSERT`, but we had to avoid undefined behavior. In this case having an info message stating that the requested block was not found seems fine.

libraries/chain/block_log.cpp
```diff
@@ -615,6 +621,63 @@ namespace eosio { namespace chain {
...
+
+               uint64_t block_size     = 0;
+               constexpr uint32_t block_pos_size = sizeof(uint64_t); // size of block position field in the block log file
+
+               if (block_num < head_block_num) {
+                  // current block is not the last block in the log file.
+                  uint64_t next_block_pos = get_block_pos(block_num + 1);
+
+                  EOS_ASSERT(next_block_pos != block_log::npos, block_log_exception,
+                             "no position found for block ${n}", ("n", block_num + 1));
```
greg7mdp, line 652: I think the next assert at line 653 is sufficient, and this one is not needed.

libraries/chain/block_log.cpp
```diff
@@ -615,6 +621,63 @@ namespace eosio { namespace chain {
...
+                  EOS_ASSERT(next_block_pos > pos + block_pos_size, block_log_exception,
+                             "next block position ${np} should be greater than current block position ${p} plus block position field size ${bps}",
+                             ("np", next_block_pos)("p", pos)("bps", block_pos_size));
+
+                  block_size = next_block_pos - pos - block_pos_size;
+               } else {
+                  // current block is the last block in the file.
+                  assert (block_num == head_block_num);
+
+                  auto file_size = std::filesystem::file_size(block_file.get_file_path());
```
greg7mdp, line 662: why not use `size()`

libraries/chain/block_log.cpp
linh2931: It was a leftover from last move around. Reverted back.

libraries/chain/block_log.cpp
linh2931: `basic_block_log` does not have `size()` to return the size of its block log file.

libraries/chain/block_log.cpp
```diff
@@ -185,6 +185,16 @@ namespace eosio { namespace chain {
          return block;
       }
 
+      template <typename Stream>
+      void read_serialized_block(Stream&& ds, uint64_t block_size, std::vector<char>& buff) {
+         assert(buff.empty());
```
heifner, line 190: The point of taking the vector by reference was to append to the already filled out message header from net_plugin. However, I realize now that the message_header includes the payload size which you don't know until here. 

We could let this method create the exact net_message needed. Call it `vector<char> create_net_message_from_block(block_num_type block_num)` then we could avoid the copy. 

If we don't want to add that knowledge to `block_log.cpp` then I guess we will have to pay for an extra copy of the `vector<char>` or do something more fancy and use two `boost::asio::const_buffer`s, one for the header and one for the payload.

libraries/chain/block_log.cpp
```diff
@@ -185,6 +185,16 @@ namespace eosio { namespace chain {
...
+
+         try {
+            buff.resize(block_size);
+            ds.read(buff.data(), block_size); // append to buff
+         } catch (...) { buff.clear(); } // restore buff to its original state
```
heifner, line 195: With this `catch` and the assert might as well just return the vector as it is the same thing with return value optimization.

libraries/chain/block_log.cpp
```diff
@@ -603,6 +619,43 @@ namespace eosio { namespace chain {
             return pos;
          }
 
+         block_pos_size_t get_block_position_and_size(uint32_t block_num) {
```
heifner, line 622: For this and the other added methods can you add unittests for them.

heifner (changes requested):

libraries/chain/block_log.cpp
linh2931: I was trying to insert the serialized bytes into the Net message but gave up due to complexities. I like the original returning vector approach: it is cleaner, no precondition assumption, and no knowledge of net plugin.

libraries/chain/block_log.cpp
linh2931: Will revert back to return vector.

libraries/chain/block_log.cpp
linh2931: Can the tests be in a separate PR. I plan to write some comprehensive tests.

heifner: There is no rush to get this in, but doesn't matter to me.

linh2931: https://github.com/AntelopeIO/spring/issues/808

libraries/chain/block_log.cpp
```diff
@@ -615,6 +668,18 @@ namespace eosio { namespace chain {
             FC_LOG_AND_RETHROW()
          }
 
+         std::vector<char> read_serialized_block_by_num(uint32_t block_num) final {
+            try {
+               block_pos_size_t pos_size = get_block_position_and_size(block_num);
```
greg7mdp, line 673: I feel it would be cleaner with the following (and also `get_block_position_and_size` could return a `std::pair`)
```suggestion
               auto [position, size] = get_block_position_and_size(block_num);
```

libraries/chain/block_log.cpp
```diff
@@ -603,6 +619,43 @@ namespace eosio { namespace chain {
...
+         block_pos_size_t get_block_position_and_size(uint32_t block_num) {
+            uint64_t pos = get_block_pos(block_num);
+
+            if (pos == block_log::npos) {
+               return block_pos_size_t {.position = block_log::npos, .size = 0};
+            }
+
+            assert(head);
+            uint32_t last_block_num = block_header::num_from_id(head->id);
+            assert(block_num <= last_block_num);
```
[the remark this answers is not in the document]
linh2931: I was debating about this. We have already found the block position for `block_num` so it should be in "good" shape. But as `block_num` comes from Net Plugin and to be paranoid, use EOS_ASSERT just in case.

libraries/chain/block_log.cpp
linh2931: I think struct provides stronger type check, especially when those two members are of the same type, and it is more clear in return statement `return block_pos_size_t {.position = pos, .size = block_size}` than just return `{pos, block_size}`.

linh2931: I can still use structure binding to make it cleaner. Thanks.

libraries/chain/include/eosio/chain/log_catalog.hpp
```diff
@@ -159,6 +164,50 @@ struct log_catalog {
...
+         if (block_num < first_block_num())
+            return {};
+
+         auto it = --collection.upper_bound(block_num);
+
+         if (block_num <= it->second.last_block_num) {
+            auto name = it->second.filename_base;
+            log_data.open(name.replace_extension("log"));
+            log_index.open(name.replace_extension("index"));
+            active_index = std::distance(collection.begin(), it);
```
heifner, line 201: This is all the same code as `get_block_position`. Seems you could call `get_block_position` from `block_position_size` and not have all this duplicate code.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1898,8 +1910,7 @@ namespace eosio {
    }
 
    // called from connection strand
-   size_t connection::enqueue_block( const signed_block_ptr& b, bool to_sync_queue) {
-      peer_dlog( this, "enqueue block ${num}", ("num", b->block_num()) );
+   size_t connection::enqueue_block( const std::vector<char>& b, bool to_sync_queue) {
```
heifner, line 1913: Pass `block_num` into `enqueue_block` so you can add back in the `peer_dlog`.

heifner (changes requested):

libraries/chain/include/eosio/chain/log_catalog.hpp
linh2931: Thanks! Much simpler.

heifner (approved):

libraries/chain/block_log.cpp
```diff
@@ -603,6 +619,46 @@ namespace eosio { namespace chain {
             return pos;
          }
 
+         block_pos_size_t get_block_position_and_size(uint32_t block_num) {
```
greg7mdp, line 622: minor detail, I think it might be cleaner to declare the return value in the beginning and use it instead of intermadiate variables, as below, but no need to change:

```c++
         block_pos_size_t get_block_position_and_size(uint32_t block_num) {
            block_pos_size_t bps { .position = get_block_pos(block_num) };

            if (bps.position == block_log::npos)
               return bps;

            assert(head);
            uint32_t last_block_num = block_header::num_from_id(head->id);
            EOS_ASSERT(block_num <= last_block_num, block_log_exception,
                       "block_num ${bn} should not be greater than last_block_num ${lbn}",
                       ("bn", block_num)("lbn", last_block_num));

            constexpr uint32_t block_pos_size = sizeof(uint64_t); // size of block position field in the block log file

            if (block_num < last_block_num) {
               // current block is not the last block in the log file.
               uint64_t next_block_pos = get_block_pos(block_num + 1);

               EOS_ASSERT(next_block_pos > bps.position + block_pos_size, block_log_exception,
                          "next block position ${np} should be greater than current block position ${p} plus block position field size ${bps}",
                          ("np", next_block_pos)("p", bps.position)("bps", block_pos_size));

               bps.size = next_block_pos - bps.position - block_pos_size;
            } else {
               // current block is the last block in the file.

               block_file.seek_end(0);
               auto file_size = block_file.tellp();
               EOS_ASSERT(file_size > bps.position + block_pos_size, block_log_exception,
                          "block log file size ${fs} should be greater than current block position ${p} plus block position field size ${bps}",
                          ("fs", file_size)("p", bps.position)("bps", block_pos_size));

               bps.size = file_size - bps.position - block_pos_size;
            }

            return bps;
         }
```

libraries/chain/include/eosio/chain/log_catalog.hpp
```diff
@@ -159,6 +164,26 @@ struct log_catalog {
       }
    }
 
+   std::optional<block_pos_size_t> get_block_position_and_size(uint32_t block_num) {
+      std::optional<uint64_t> pos = get_block_position(block_num);
+
+      if (!pos) {
+         return {};
```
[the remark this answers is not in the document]
linh2931: Was trying to follow existing method patterns in the two files: `get_block_position` in `log_catalog.hpp` while `get_block_pos` in `block_log.cpp` returns `block_log::npos`.

libraries/chain/block_log.cpp
linh2931: I think constructing return value right before returning makes it easier to see what to be returned.

greg7mdp (approved):

Issue #751, opened 2024-09-11
Add new testcase where finality advances on alternate branch with lesser lib


greg7mdp:
Here is a whiteboard drawing of the [testcase](https://us06web.zoom.us/wb/doc/dcRtOGBzTtqnhsBoiR1ozw/p/81087274413) .

Same in a block comment:
```
/* -----------------------------------------------------------------------------------------------------
            Finality advancing past block claimed on alternate branch
            =========================================================

Time:        t1      t2      t3      t4      t5      t6      t7      t8
Blocks:
    B0 <---  B1 <--- B2 <--- B3 <-|- B4 <--- B5
                                  |
                                  \----------------- B6 <--- B7 <--- B8 <--- B9
QC claim:
           Strong          Strong  Strong  Strong  Strong   Weak   Strong  Strong
             B0              B1      B3      B4      B2      B6      B7      B8

Vote:                      Strong  Strong   Strong  Weak   Strong  Strong  Strong

                                                     ^
                                                     |
                                                 validating those weak votes on b2
                                                 would fail if nodes have received b4 and b5
                                                 which advanced lib to b3

    - Node D is isolated and has not seen B3, B4, and B5
    - it received B3 via push_block, (so it can make it its head and produce a child of B3), but has not
      received votes on b3 (only on b2), so b6 includes a strong QC on b2.
    - when b6 is pushed to A, B and C, qc validation should fail.
--------------------------------------------------------------------------------------------------------*/
```

Need a testcase similar as above, but where we avoid the problems that nodes which have seen block `B5` (to advance lib to `B3`) have voted on it (so locked on `B4`).

greg7mdp:
@arhag This testcase does not work. After voting on `B5` (which makes `B3` final), the finalizers who voted on `B5` are locked on `B4`, and therefore cannot vote on `B6`:

- liveness check fails because: `B6' s core.latest_qc_block_timestamp() <  fsi.lock.timestamp` because `B2` timestamp < `B4` timestamp.
- safety check fails because `B6` does not extend `B4`.

arhag:
@greg7mdp:

Try this out:

```
/* -----------------------------------------------------------------------------------------------------
            Finality advancing past block claimed on alternate branch
            =========================================================
Producer:    C       C       C       C       C       D       D       D       D
Timestamp:   t1      t2      t3      t4      t5      t6      t7      t8      t9
Blocks:
[...]

PR #788, opened 2024-09-16
[1.0.2] Don't assume that claimed block is in fork_db.

### LINKED ISSUE #778: fork_db check prevents voting on blocks including a qc claim on a block not in fork_db.

greg7mdp:
Resolves #778.

Update `fork_db_validated_block_exists`  to lookup parents instead of just claimed block.

Also updated the `validated` member in both `block_state_legacy` and `block_state` to be `copyable_atomic<bool>` instead of `bool`, which allowed to make the `set_valid()` and `is_valid()` members public, and thus eliminate a couple of `friend` declarations.


libraries/chain/fork_database.cpp
```diff
@@ -603,15 +603,24 @@ namespace eosio::chain {
...
+   // precondition: `id` is already in fork_db, so is a descendent of `root`
+   // ------------------------------------------------------------------------------------------------------
    template<class BSP>
-   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id) const {
-      auto itr = index.find( id );
-      return itr != index.end() && (*itr)->is_valid();
+   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id, uint32_t block_num) const {
+      assert(root && (root->id() == id || index.find(id) != index.end()));
+      if (root->block_num() >= block_num)
+         return true;
```
heifner, line 618: I don't see how you can assume this is true.

If the block you are considering voting on was not on the best branch then this can return true when that branch has been abandoned and pruned from the forkdb.

I was expecting this to walk back from `id` until it hit root or a validated block or block_num < `block_header::num_from_id(latest_qc_claim__block_ref.block_id)` and verifying it reached `latest_qc_claim__block_ref.block_id` or root.

libraries/chain/controller.cpp
```diff
@@ -4240,15 +4250,13 @@ struct controller_impl {
          if (bsp->is_recent() || testing_allow_voting) {
             if (use_thread_pool == use_thread_pool_t::yes && async_voting == async_t::yes) {
                boost::asio::post(thread_pool.get_executor(), [this, bsp=bsp]() {
-                  const auto& latest_qc_claim__block_ref = bsp->core.get_block_reference(bsp->core.latest_qc_claim().block_num);
-                  if (fork_db_validated_block_exists(latest_qc_claim__block_ref.block_id)) {
+                  if (fork_db_validated_block_exists(bsp->previous(), bsp->core.latest_qc_claim().block_num)) {
```
heifner, line 4253: I think you have to pass `latest_qc_claim__block_ref.block_id` instead of the block_num.

libraries/chain/controller.cpp
```diff
@@ -1130,9 +1130,19 @@ struct controller_impl {
       });
    }
 
-   bool fork_db_validated_block_exists( const block_id_type& id ) const {
+   bool fork_db_validated_block_exists( const block_id_type& id ) {
```
heifner, line 1133: This is not commented as thread safe and is currently only called from the main thread. But it used to be thread-safe, but is no longer for `block_state_legacy`.

libraries/chain/controller.cpp
```diff
@@ -1130,9 +1130,19 @@ struct controller_impl {
...
+       return fork_db.apply<bool>([&](const auto& forkdb) {
+         auto bsp = forkdb.get_block(id);
+         return bsp ? bsp->is_valid() : false;
```
heifner, line 1136: Might be good to add a comment that `is_valid()` is not thread safe for `block_state_legacy`.

heifner (changes requested):

libraries/chain/controller.cpp
greg7mdp: Why is it not thread safe?

heifner: `is_valid()` is not thread safe for `block_state_legacy`

greg7mdp: Shall we make it a `copyable_atomic<bool>` as well?

libraries/chain/fork_database.cpp
greg7mdp: > If the block you are considering voting on was not on the best branch then this can return true when that branch has been abandoned and pruned from the forkdb.

Yes, good point, I updated the PR.

libraries/chain/controller.cpp
greg7mdp: I don't think so, as long as we check it is an ancestor of the passed `id`.

libraries/chain/controller.cpp
heifner: Not of 1.0.x I don't think.

greg7mdp: Other uses of `is_valid` that are think are not restricted to `block_state`:
[1](https://github.com/AntelopeIO/spring/blob/c4cd685edc3fd9a604ec2d431313a91aeaaa502b/libraries/chain/controller.cpp#L1510)  [2](https://github.com/AntelopeIO/spring/blob/c4cd685edc3fd9a604ec2d431313a91aeaaa502b/libraries/chain/controller.cpp#L1825) [3](https://github.com/AntelopeIO/spring/blob/c4cd685edc3fd9a604ec2d431313a91aeaaa502b/libraries/chain/controller.cpp#L3673) [4](https://github.com/AntelopeIO/spring/blob/c4cd685edc3fd9a604ec2d431313a91aeaaa502b/libraries/chain/controller.cpp#L4402)  [5](https://github.com/AntelopeIO/spring/blob/c4cd685edc3fd9a604ec2d431313a91aeaaa502b/libraries/chain/controller.cpp#L1358)

libraries/chain/fork_database.cpp
heifner: I think it is "correct" now, but seems overly convoluted and error prone. Seems to me much safer to pass in `latest_qc_claim__block_ref.block_id` which provides a range of blocks similar to `fetch_branch` that takes two ids. Also would be more efficient as you don't have to do all the `shared_ptr` copies.

Also. please add some unit tests for this new forkdb method.

libraries/chain/fork_database.cpp
```diff
@@ -603,15 +603,27 @@ namespace eosio::chain {
...
 
+   // returns true if block `id`, or one of its ancestors >= block_num, is found in fork_db and `is_valid()`
+   // precondition: `id` is already in fork_db, so is a descendent of `root`
+   // ------------------------------------------------------------------------------------------------------
    template<class BSP>
-   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id) const {
-      auto itr = index.find( id );
-      return itr != index.end() && (*itr)->is_valid();
+   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id, uint32_t block_num) const {
+      assert(root && (root->id() == id || index.find(id) != index.end()));
```
heifner, line 616: This seems overly restrictive. It may be true, but I don't see any reason why you need to enforce it. If `id` doesn't exist in the forkdb then the answer is false.

libraries/chain/include/eosio/chain/block_state_legacy.hpp
```diff
@@ -78,7 +78,7 @@ namespace eosio::chain {
          _cached_trxs = std::move( trxs_metas );
       }
 
-      bool                                                validated = false;
+      copyable_atomic<bool>                               validated{false};
```
heifner, line 81: I don't think we should make this change on `release/1.0`

greg7mdp: Why not?

libraries/chain/controller.cpp
heifner: Those are all on the main thread.

libraries/chain/fork_database.cpp
greg7mdp: Personally, I find the `fetch_branch` that takes two ids more convoluted than the `fetch_branch` that takes an `id` and a `block_num`.
But I'll change `validated_block_exists_impl` to take two `id`s. 
I left the `assert` because it makes the code simpler and faster (allows to `return true;` at the end without checks).

heifner: I think once you add some unit tests you will see the current implementation also allows for an abandoned branch to return true.

libraries/chain/include/eosio/chain/block_state_legacy.hpp
heifner: It is private and will serialize the same way. So, I guess I'm ok with changing this.

libraries/chain/fork_database.cpp
greg7mdp: Since `id` *is* in fork_db (per the assert), and walking back from `is` we don't find 'claimed_id`, it means that `claimed_id` is either root or and ancester of root (since we know that `claimed_id` is an ancestor of `id).

heifner:
```
b1 <- b2  <- b3  <-b4
      b2' <- b3'
```
* net-thread
 -- call `create_block_state_i`
  --- find bottom b3', create block state b4'
 * main-thread
-- advance root to top b2, purges b2', b3'
 net-thread
 -- continues `create_block_state_i` execution 
  --- call `consider_voting`
  --- calls `fork_db_validated_block_exists(b3', b1')
  --- returns `true`
  --- Also returns `true` if b2' or b3' happen to be valid.

libraries/chain/controller.cpp
greg7mdp: It is thread safe after https://github.com/AntelopeIO/spring/pull/788/commits/3c417b1f768d390299500a9033f200a3e8ecffaf

libraries/chain/fork_database.cpp
greg7mdp: Thanks for explaining, I finally understand that the `assert` fay be false and `id` may not be present. https://github.com/AntelopeIO/spring/pull/788/commits/3a133d244e24c14f55b3b5ebdf4d936cabaf3b91

libraries/chain/controller.cpp
heifner: We could make `is_valid()` public now and remove many of the `friends`.

libraries/chain/fork_database.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/788/commits/3a133d244e24c14f55b3b5ebdf4d936cabaf3b91

libraries/chain/controller.cpp
greg7mdp: Unfortunately I could only remove:
- `template <typename BS> friend struct fork_database_impl;`
- `friend struct block_state_legacy_accessor;`

https://github.com/AntelopeIO/spring/pull/788/commits/311c94dfb0f29d32168111dd88f97918ff760b02

ericpassmore:
Note:start
category: Other
component: Internal
summary: Validate claimed block in in fork_db.
Note:end

unittests/fork_db_tests.cpp
```diff
@@ -137,4 +140,46 @@ BOOST_AUTO_TEST_CASE(add_remove_test) try {
...
+// -------------------------------------------------------------
+BOOST_FIXTURE_TEST_CASE(validated_block_exists, generate_forkdb_state) try {
+   bsp12b->set_valid(true);
+   bsp13b->set_valid(true);
+   bsp14b->set_valid(true);
+
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp14b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp13b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp12b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp11b->id()));
```
heifner, line 154: Giving `bsp11a->set_valid(true)`
`forkdb.validated_block_exists(bsp14b->id(), bsp13a->id()));`
will return `true`, I think.

greg7mdp: As it should. The call is in error because `claimed_id` is expected to be an ancestor of `id`.

unittests/fork_db_tests.cpp
```diff
@@ -137,4 +140,51 @@ BOOST_AUTO_TEST_CASE(add_remove_test) try {
...
+   bsp14b->set_valid(false);
+   BOOST_REQUIRE_EQUAL(false, forkdb.validated_block_exists(bsp14b->id(), bsp14b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp13b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp12b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp11b->id()));
+
+   bsp12b->set_valid(false);
+   BOOST_REQUIRE_EQUAL(false, forkdb.validated_block_exists(bsp14b->id(), bsp14b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp13b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp12b->id()));
```
heifner, line 170: This should be `false` because 12b is not valid.

unittests/fork_db_tests.cpp
```diff
@@ -137,4 +140,51 @@ BOOST_AUTO_TEST_CASE(add_remove_test) try {
...
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp11b->id()));
```
heifner, line 171: This should be `false` because 11b is not valid.

libraries/chain/fork_database.cpp
```diff
@@ -603,15 +603,27 @@ namespace eosio::chain {
...
+
+      for (auto i = index.find(id); i != index.end(); i = index.find((*i)->previous())) {
+         id_present = true;
+         if ((*i)->is_valid())
+            return true;
+         if ((*i)->id() == claimed_id)
+            return false;
+      }
+      return id_present;
    }
```
heifner, line 627: I think this should be something more like:
```cpp
   template<class BSP>
   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id, const block_id_type& claimed_id) const {
      assert(root);
      bool id_present = false;
      block_num_type claimed_block_num = block_header::num_from_id(claimed_id);

      for (auto i = index.find(id); i != index.end() && (*i)->block_num() >= claimed_block_num; i = index.find((*i)->previous())) {
         id_present = true;
         if ((*i)->id() == claimed_id)
            return (*i)->is_valid();
      }
      return id_present &&
             (root->id() == claimed_id || block_header::num_from_id(root->id()) > claimed_block_num);
   }

```

heifner (changes requested):

unittests/fork_db_tests.cpp
```diff
@@ -137,4 +140,51 @@ BOOST_AUTO_TEST_CASE(add_remove_test) try {
...
+BOOST_FIXTURE_TEST_CASE(validated_block_exists, generate_forkdb_state) try {
+   bsp12b->set_valid(true);
+   bsp13b->set_valid(true);
+   bsp14b->set_valid(true);
+   bsp11a->set_valid(true);
+
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp14b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp13b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp12b->id()));
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp11b->id()));
```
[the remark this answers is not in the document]
greg7mdp: Well not according to the description of what `validated_block_exists` does.
`returns true if block `id`, or one of its ancestors not older than claimed_id, is found in fork_db and `is_valid()`.`

If a block is valid, doesn't it mean that all of its ancestors are valid? Maybe I should update the testcase to reflect that.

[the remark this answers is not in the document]
heifner: Yes, if a block is valid then we have validated its ancestors. The test could be cleaned up to reflect that invariant.

unittests/fork_db_tests.cpp
```diff
@@ -137,4 +140,51 @@ BOOST_AUTO_TEST_CASE(add_remove_test) try {
...
+
+   // incorrect call as bsp13a is not an ancestor of bsp14b. As a result `claimed_id` is
+   // not found, so we assume it is either the root or an older block which is `valid`.
+   BOOST_REQUIRE_EQUAL(true,  forkdb.validated_block_exists(bsp14b->id(), bsp13a->id()));
```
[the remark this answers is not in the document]
greg7mdp: I don't understand what you mean. If you can make it return false when the blocks are in a valid config in fork_db *and* it should return `true`, please let me know.

libraries/chain/fork_database.cpp
greg7mdp: What problem with my implementation does it solve?

The last return in your implementation already relies on the fact that `claimed_id` is an ancestor of `id`, since it compares block numbers.

So if we didn't find `claimed_id` while walking back up to toot, it means that `claimed_id` **has** to be either `root` or one of its ancestors, which are valid (they even are final).

unittests/fork_db_tests.cpp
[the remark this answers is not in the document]
greg7mdp: test updated https://github.com/AntelopeIO/spring/pull/788/commits/45e67aa1097e653ad329e466aa9e651b52cedf38

libraries/chain/fork_database.cpp
heifner: It solves the test cases I mentioned above that return the wrong result.

You can walk all the way back to root and not find `claimed_id` because the `claimed_id` is on a different branch than `id`. That is what the comparison of block numbers tells you.

greg7mdp: There is no testcase returning a wrong result, because  `claimed_id` is by precondition **not** on a different branch than `id`

heifner: But the algorithm above also passes all the tests and returns `false` when not on the same branch so the precondition does not have to be true.

greg7mdp: In the use case we have for this function, the precondition has to be true.
I don't see why we have to make the function more complex to support a use case we don't need, and probably will never need.

heifner: I don't think the algorithm above is more complex. I find it easier to reason that it does the right thing. However, I'll not press any more as your assert now verifies the precondition. You do have to remove the one test case now I would think.

libraries/chain/include/eosio/chain/block_state_legacy.hpp
```diff
@@ -78,7 +75,7 @@ namespace eosio::chain {
...
```
heifner, line 78: Would be good to mention this in the PR description.

heifner (approved):

libraries/chain/include/eosio/chain/block_state_legacy.hpp
greg7mdp: Indeed, done.

libraries/chain/fork_database.cpp
```diff
@@ -603,15 +603,30 @@ namespace eosio::chain {
...
+   // precondition: claimed_id is either id, or an ancestor of id
+   // returns true if block `id`, or one of its ancestors not older than claimed_id, is found in fork_db
+   // and `is_valid()`.
+   // ------------------------------------------------------------------------------------------------------
    template<class BSP>
-   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id) const {
-      auto itr = index.find( id );
-      return itr != index.end() && (*itr)->is_valid();
+   bool fork_database_impl<BSP>::validated_block_exists_impl(const block_id_type& id, const block_id_type& claimed_id) const {
+      bool id_present = false;
```
linh2931, line 617: Maybe we should assert at the beginning claim_id is an ancestor of id.

libraries/chain/fork_database.cpp
```diff
@@ -603,15 +603,30 @@ namespace eosio::chain {
...
+      for (auto i = index.find(id); i != index.end(); i = index.find((*i)->previous())) {
+         id_present = true;
+         if ((*i)->is_valid())
+            return true;
+         if ((*i)->id() == claimed_id)
+            return false;
+      }
+
+      // if we return `true`, let's validate the precondition and make sure claimed_id is not in another branch
+      assert(!id_present || block_header::num_from_id(claimed_id) <= block_header::num_from_id(root->id()));
```
linh2931, line 628: Why does `block_header::num_from_id(claimed_id) <= block_header::num_from_id(root->id()))` imply `claim_id` not on another branch?

Probably it should return false if claim_id is not on this branch.

libraries/chain/fork_database.cpp
greg7mdp: We can't check for certain at the beginning, because we don't know if the `claimed_id` is in fork_db.
And if we don't find the `claimed_id`, we do assert at the end that its block number is <= than the root block number, which means it is not on another branch and still present in fork_db.
I think we check all we can check.

libraries/chain/fork_database.cpp
greg7mdp: > Why does `block_header::num_from_id(claimed_id) <= block_header::num_from_id(root->id()))` imply `claim_id` not on another branch?

because all blocks in fork_db have block numbers greater than the root block number (because they are all root's descendants).

linh2931: I was thinking about the claimed_id's corresponding block number was a corrupted large number but it was not in the forkdb. Probably we don't need to worry about it.

linh2931 (approved):

greg7mdp:
@arhag got two approvals on this. Please let me know if you'd like me to merge this without waiting for the updated testcase (if I don't hear from you I'll wait).

arhag (dismissed):
@greg7mdp:

See https://github.com/AntelopeIO/spring/issues/751#issuecomment-2372540610 for the test case that works for both https://github.com/AntelopeIO/spring/issues/751 and https://github.com/AntelopeIO/spring/issues/778.

I think that test should be added here to demonstrate that this PR does in fact resolve the bug reported in https://github.com/AntelopeIO/spring/issues/778 properly.

heifner (approved):

linh2931 (approved):

greg7mdp:
@arhag got two new approvals from Lin and Kevin. Time to merge?

PR #801, opened 2024-09-18
[1.0.2 -> main] Add new testcase where finality advances on alternate branch with lesser lib

greg7mdp:
Forwarding PR #782 .

Add a testcase where finalizers cannot vote on a block from a different branch because they are locked on a block from their own branch.

```

Time:        t1      t2      t3      t4      t5      t6      t7   
Blocks:
                                  |
                                  \----------------- B6 <--- B7
QC claim:
           Strong          Strong  Strong  Strong  Strong   No QC
             B0              B1      B3      B4      B2     achieved

Vote:                      Strong  Strong   Strong  Weak     -

                                                     ^
                                                     |
                                                 which advanced lib to b3

    - when b6 is pushed to A, B and C, finalizers of A, B, and C are unable to vote on it, because they
      are locked on B4,
          -> liveness check fails because: `B6' s core.latest_qc_block_timestamp() <  fsi.lock.timestamp`
             because `B2 timestamp < B4 timestamp`.
          -> safety check fails because `B6` does not extend `B4`.
```

After voting on `B5` (which makes `B3` final), the finalizers who voted on `B5` are locked on `B4`, and therefore cannot vote on `B6`:

- liveness check fails because: `B6' s core.latest_qc_block_timestamp() <  fsi.lock.timestamp` because
 `B2 timestamp < B4 timestamp`.
- safety check fails because `B6` does not extend `B4`.

As a result, this testcase reproduces the scenario described above, and verifies that we lack votes to form a QC on `B6`. 

Although until issue #778 is addressed, the problem described in issue #778 is the cause of the lack of votes, and `decide_vote` is not even called.


heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: New test case where finality advances on lesser LIB.
Note:end

PR #802, opened 2024-09-18
Optimize block processing - Part 3

### LINKED ISSUE #529: Optimize processing of incoming blocks

heifner:
Part 3 of 3 of #529. See #688 for part 2 and #619 for part 1.

- Add optional handler id to `app().executor().post()`.
  -  If handler id is provided and a existing function with the same handler id has already been posted with the same priority the function is not posted.
  - This allows for deduplication of redundant posts.
  -  Specifically if an existing `on_incoming_block()` has been posted to the main thread, then another does not need to be added as the `on_incoming_block` will execute any blocks ready from the fork database.

Resolves #529


libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -61,22 +76,47 @@ class exec_pri_queue : public boost::asio::execution_context
...
+      }
+      prio_queue& que = priority_que(q);
+      std::unique_lock g( mtx_, std::defer_lock );
+      if (lock_enabled_ || q == exec_queue::read_exclusive) {
+         // called directly from any thread for read_exclusive
+         g.lock();
+      }
+      if (!que.empty()) {
+         // find the associated priority
+         auto i = std::lower_bound(que.ordered_begin(), que.ordered_end(), priority, [](const auto& h, int priority) {
```
linh2931, line 109: I thought you would use `id` to find a duplicated task.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -61,22 +76,47 @@ class exec_pri_queue : public boost::asio::execution_context
...
+            return h->priority() > priority;
+         });
+         if (i != que.ordered_end() && (*i)->priority() == priority && (*i)->id() == id)
```
linh2931, line 112: Does the find above imply `(*i)->priority() == priority`?

heifner: It is a `lower_bound`, so no.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
linh2931: What happens if a non-unique-id task and a unique-id task with the same priority exist, and the non-unique-id task is found first? You would insert the unique-id task into the queue?

heifner: This goes with the more conservative approach of only de-dup on matching id for a given `priority` and only if it is the next item to be processed (but seems like that should be if it is the last in the priority range not the first). Seems it should only not add if the last thing added with that priority is right before it in the queue.

q9 q8 q7 q6

If q9 is non-unique and q8 is something that changes state so that the non-unique is needed then a q5 should be posted. So only de-dup if q6 is non-unique.

I'm having a hard time getting the ordered iterators to work at the moment.

heifner: Fixed https://github.com/AntelopeIO/spring/pull/802/commits/cf8d7c12d680ef00f8d15a4e4a68d0129980e740

ericpassmore:
Note:start
category: Other
component: Internal
summary: Optimize block processing - Part 3: deduplication of redundant posts.
Note:end

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -61,22 +76,56 @@ class exec_pri_queue : public boost::asio::execution_context
...
+         auto i = std::lower_bound(que.ordered_begin(), end, priority, [](const auto& h, int priority) {
+            return h->priority() > priority;
+         });
+         if (i != end) {
+            // ordered iterator appears to only be a forward iterator
+            auto p = i;
+            for (; i != end; p = i, ++i) {
+               if ((*i)->priority() != priority)
+                  break;
+            }
```
greg7mdp, line 119: This seems wrong to me. Shouldn't we want to skip handlers which have the same priority, but a different id?

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -61,22 +76,56 @@ class exec_pri_queue : public boost::asio::execution_context
...
+            if ((*p)->priority() == priority && (*p)->id() == id)
+               return;
+         }
```
greg7mdp, line 122: Could it be as simple as this?

```suggestion
         if (i != end) {
            // ordered iterator appears to only be a forward iterator
            for (; i != end && (*i)->priority() == priority; ++i)
               if ((*i)->id() == id)
                  return;
         }
```

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
linh2931: I have been thinking about this. Is it worthy the complexity to de-dep? Can `on_incoming_block` itself handle duplications? I am worried we might miss processing some blocks too.

heifner: `on_incoming_block` does already handle duplicates. You can call it as often as you like. The idea here is to avoid posting to the queue. Although, we do have to take a mutex anyway, so it only avoids the post to the queue. I'll try and see if I can measure the performance difference.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
heifner: The idea is the find the last posted function. If it same as you are attempting to post, don't bother.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -61,22 +76,56 @@ class exec_pri_queue : public boost::asio::execution_context
...
    }
 
-   // called from appbase::application_base::exec poll_one() or run_one()
    template <typename Function>
-   void add(int priority, exec_queue q, size_t order, Function function) {
+   void add(int priority, exec_queue q, size_t order, Function&& function) {
       assert( num_read_threads_ > 0 || q != exec_queue::read_exclusive);
       prio_queue& que = priority_que(q);
-      std::unique_ptr<queued_handler_base> handler(new queued_handler<Function>(priority, order, std::move(function)));
+      std::unique_ptr<queued_handler_base> handler(new queued_handler<Function>(handler_id::unique, priority, order, std::forward<Function>(function)));
```
greg7mdp, line 83: I think it would be nicer to use the same pattern as the `add()` below, so the two functions are more similar and we don't have this `unique_ptr` stuff. 
```
      std::unique_lock g( mtx_, std::defer_lock );
      if (lock_enabled_ || q == exec_queue::read_exclusive) {
         // called directly from any thread for read_exclusive
         g.lock();
      }
      ...
      que.push( new queued_handler<Function>(...);
      }
```

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
heifner: No. You have to find the last posted handler with the same priority and see if that one has the same id.

greg7mdp: Why does it has to be the last posted one?

greg7mdp:
What was the rationale for not using `unique_ptr` anymore in the `prio_queue`?

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
greg7mdp: OK. I doubt it makes any significant difference though.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
heifner: This goes with the more conservative approach of only de-dup on matching id for a given priority when it was the last id posted.

Say you have items in the queue, u=unique, pb=process_incoming_block
`q9(u) q8(pb) q7(u)`

Say you want to post a `q6(pb)`, it is possible this is not safe to de-dup because if `q7(u)` does something that needs `q6(pb)` to run then you end up with `q6(pb)` not running. So this goes with the more conservative route of only de-dup if `q7(u)` is a `pb`. Granted we could argue that `q7(u)` should post a `pb` if needs that to happen.

heifner:
> What was the rationale for not using `unique_ptr` anymore in the `prio_queue`?

`boost::heap` doesn't support them. It uses copy in some places.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
greg7mdp: The code I sent dedups only if it is the same id (pb)  at the same priority, so in your example posting `q6(pb)` would not dedup as there is no other `q6(pb)` in the queue. Or am I missing something?

heifner: The 9,8,7 are just order numbers. They are all the same priority.
`q9(u) q8(pb) q7(u)`
So your code would find `q8(pb)` and return (de-dup). Want to only compare to `q7(u)` and then non-dedup.

greg7mdp: Is there a reason for this requirement, i.e would the code not work if we didn't check that it was the last posted? It seems to me that either we want to dedup, and then we should just do it, or not.

greg7mdp:
> `boost::heap` doesn't support them. It uses copy in some places.

annoying. Maybe it would be better to use `shared_ptr`, which would make it easier to be confident that there are no memory leaks (although there aren't now).

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
heifner: Been thinking about this.  Maybe I am being too conservative. If the last item running needs to have a follow non-unique item it should post it.

greg7mdp (approved):

linh2931 (approved):

PR #811, opened 2024-09-23
[1.0.2] Prevent sync ahead in irreversible mode

### LINKED ISSUE #789: Add test that net_plugin does not sync ahead beyond limit
`net_plugin` is constrained not to sync ahead of `sync-fetch-span*2-1`, add a test that verifies this behavior.

heifner:
Issue #789 was to create a test that verifies #777. The new test exposed an issue with the current implementation that would allow it to sync ahead too far when blocks have not made it to the fork database yet. Includes a fix that verifies fork database has the block being evaluated before considering syncing ahead.

Failing test before fix: https://github.com/AntelopeIO/spring/actions/runs/11002592099

Resolves #789

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2192,15 +2192,15 @@ namespace eosio {
          controller& cc = my_impl->chain_plug->chain();
          if (cc.get_read_mode() == db_read_mode::IRREVERSIBLE) {
             auto forkdb_head = cc.fork_db_head();
-            if (forkdb_head.block_num() >= blk_num) {
-               auto calculated_lib = forkdb_head.irreversible_blocknum();
-               auto num_blocks_that_can_be_applied = calculated_lib > head_num ? calculated_lib - head_num : 0;
-               if (num_blocks_that_can_be_applied < sync_fetch_span) {
-                  if (head_num )
-                     fc_ilog(logger, "sync ahead allowed past sync-fetch-span ${sp}, block ${bn} for paused LIB ${l}, chain_lib ${cl}, forkdb size ${s}",
-                             ("bn", blk_num)("sp", sync_fetch_span)("l", calculated_lib)("cl", head_num)("s", cc.fork_db_size()));
-                  return true;
-               }
+            auto calculated_lib = forkdb_head.irreversible_blocknum();
+            auto num_blocks_that_can_be_applied = calculated_lib > head_num ? calculated_lib - head_num : 0;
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2192,13 +2192,15 @@ namespace eosio {
...
-            auto num_blocks_that_can_be_applied = calculated_lib > head_num ? calculated_lib - head_num : 0;
-            if (num_blocks_that_can_be_applied < sync_fetch_span) {
-               if (head_num )
-               fc_ilog(logger, "sync ahead allowed past sync-fetch-span ${sp}, block ${bn} for paused LIB ${l}, chain_lib ${cl}, forkdb size ${s}",
-                       ("bn", blk_num)("sp", sync_fetch_span)("l", calculated_lib)("cl", head_num)("s", cc.fork_db_size()));
-               return true;
+            if (forkdb_head.block_num() >= blk_num) {
+               auto calculated_lib = forkdb_head.irreversible_blocknum();
+               auto num_blocks_that_can_be_applied = calculated_lib > head_num ? calculated_lib - head_num : 0;
+               if (num_blocks_that_can_be_applied < sync_fetch_span) {
```
greg7mdp, line 2198: I wonder is the test at line 2195 is not too strict. If this function is always evaluated before the last received block is added to fork_db, might we never return true?

Instead, what about the following change:
```suggestion
               auto num_blocks_that_can_be_applied = calculated_lib > head_num ? calculated_lib - head_num : 0;
               
               // add the blocks that can potentially be applied because they are not in fork_db yet
               num_blocks_that_can_be_applied += blk_num > forkdb_head.block_num() ? blk_num - forkdb_head.block_num() : 0;
               
               if (num_blocks_that_can_be_applied < sync_fetch_span) {
```

heifner: I like that, thanks.

heifner: Done

greg7mdp (approved):

tests/TestHarness/Node.py
```diff
@@ -687,6 +687,27 @@ def findInLog(self, searchStr):
                         return True
         return False
 
+    def linesInLog(self, searchStr):
```
linh2931, line 690: Like those two common functions are defined here.

linh2931 (approved):

ericpassmore:
Note:start
category: Other
component: P2P
summary: Prevent sync ahead in irreversible mode.
Note:end

PR #813, opened 2024-09-24
[Test only, don't merge] test ship streamer

taokayan:
example stuck behavior:
```
info  2024-09-24T04:02:12.400 trx_gener trx_provider.cpp:71           connect              ] Attempting P2P connection to localhost:9879.

info  2024-09-24T04:02:12.400 trx_gener trx_provider.cpp:75           connect              ] Connected to localhost:9879.

2024-09-24T04:02:13.399179    

2024-09-24T04:02:13.399297  Found first trx record: /home/kayan-u20/workspaces/spring/build/TestLogs/ship_streamer_test22340/first_trx_22757.txt

2024-09-24T04:02:13.399383  Found first trx record: /home/kayan-u20/workspaces/spring/build/TestLogs/ship_streamer_test22340/first_trx_22758.txt

2024-09-24T04:02:13.399450  first transactions: ['978ea4db0cce34065ca959d3c25636fd7bc7e9432b97f71651a572746fc1efb3', '662959e19ce2e9d19322eacde0b020db95aa16abd63c2004c2e940f11953adeb']

2024-09-24T04:02:13.749440  {'server_version': '711d7a78', 'chain_id': '3a9172c91a02325927ed5082eab363678a762be5cdd6e281dee3549cde107357', 'head_block_num': 129, 'last_irreversible_block_num': 73, 'last_irreversible_block_id': '000000497cd81faa49b07dd38c35776ec1b53c7bd3db144bb356853a354bbc42', 'head_block_id': '00000081e5e8e45c5dafef2f427dea26643b788f1317b17777114c9c849ffbb8', 'head_block_time': '2024-09-24T04:02:13.500', 'head_block_producer': 'defproducerc', 'virtual_block_cpu_limit': 567683, 'virtual_block_net_limit': 1191770, 'block_cpu_limit': 500000, 'block_net_limit': 1048576, 'server_version_string': 'v1.0.0-rc3', 'fork_db_head_block_num': 129, 'fork_db_head_block_id': '00000081e5e8e45c5dafef2f427dea26643b788f1317b17777114c9c849ffbb8', 'server_full_version_string': 'v1.0.0-rc3-711d7a786b42ff89eaeb9dc91a7c7ed42a5cf7eb', 'total_cpu_weight': 0, 'total_net_weight': 0, 'earliest_available_block_num': 1, 'last_irreversible_block_time': '2024-09-24T04:01:45.500'}

[...]

PR #816, opened 2024-09-24
[1.0.2] SHiP: Fix hang on shutdown

### LINKED ISSUE #815: SHiP hang on exit
See #813 

- Create a repeatable test based on `ship_streamer_test.py`
- Investigate and find fix.

heifner:
`state_history_plugin` (SHiP) would sometimes hang on shutdown. The `nodeos` process would get stuck trying to lock a mutex of the main application `io_context` strand while destroying a `boost::beast::websocket::stream`. 

Created a test that reproduces the failure (thanks @taokayan for #813). The failure was not reliably repeatable, but it did fail on every CI/CD run for at least one of the platforms. See for example: https://github.com/AntelopeIO/spring/actions/runs/11017331583/job/30595880305

Modified SHiP `boost::beast::websocket` to use the SHiP thread instead of the main application thread executor strand.

Resolves #815

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -100,12 +100,11 @@ public:
    template <typename Protocol>
    void create_listener(const std::string& address) {
       const boost::posix_time::milliseconds accept_timeout(200);
-      // connections set must only be modified by main thread; run listener on main thread to avoid needing another post()
-      auto strand = boost::asio::make_strand(thread_pool.get_executor());
-      fc::create_listener<Protocol>(strand, _log, accept_timeout, address, "", [this, strand](Protocol::socket&& socket) {
-         boost::asio::post(app().get_io_service(), [this, strand, socket{std::move(socket)}]() mutable {
-            catch_and_log([this, &socket, &strand]() {
-               connections.emplace(new session(std::move(socket), std::move(strand), chain_plug->chain(),
+      // connections set must only be modified by main thread; run listener on ship thread so sockets use default executor of the ship thread
+      fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
+         boost::asio::post(app().get_io_service(), [this, socket{std::move(socket)}]() mutable {
+            catch_and_log([this, &socket]() {
+               connections.emplace(new session(std::move(socket), boost::asio::make_strand(thread_pool.get_executor()), chain_plug->chain(),
                                                trace_log, chain_state_log, finality_data_log,
                                                [this](const chain::block_num_type block_num) {
                                                   return get_block_id(block_num);
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -101,21 +101,24 @@ public:
    void create_listener(const std::string& address) {
       const boost::posix_time::milliseconds accept_timeout(200);
       // connections set must only be modified by main thread; run listener on main thread to avoid needing another post()
-      fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
-         catch_and_log([this, &socket]() {
-            connections.emplace(new session(std::move(socket), boost::asio::make_strand(thread_pool.get_executor()), chain_plug->chain(),
-                                            trace_log, chain_state_log, finality_data_log,
-                                            [this](const chain::block_num_type block_num) {
-                                               return get_block_id(block_num);
-                                            },
-                                            [this](const chain::block_id_type& block_id) {
-                                               return chain_plug->chain().fetch_block_by_id(block_id);
-                                            },
-                                            [this](session_base* conn) {
-                                               boost::asio::post(app().get_io_service(), [conn, this]() {
-                                                  connections.erase(connections.find(conn));
-                                               });
-                                            }, _log));
+      auto strand = boost::asio::make_strand(thread_pool.get_executor());
+      fc::create_listener<Protocol>(strand, _log, accept_timeout, address, "", [this, strand](Protocol::socket&& socket) {
+         boost::asio::post(app().get_io_service(), [this, strand, socket{std::move(socket)}]() mutable {
+            catch_and_log([this, &socket, &strand]() {
+               connections.emplace(new session(std::move(socket), std::move(strand), chain_plug->chain(),
+                                               trace_log, chain_state_log, finality_data_log,
+                                               [this](const chain::block_num_type block_num) {
+                                                  return get_block_id(block_num);
+                                               },
+                                               [this](const chain::block_id_type& block_id) {
+                                                  return chain_plug->chain().fetch_block_by_id(block_id);
+                                               },
+                                               [this](session_base* conn) {
+                                                  boost::asio::post(app().get_io_service(), [conn, this]() {
+                                                     connections.erase(connections.find(conn));
+                                                  });
+                                               }, _log));
+            });
          });
       });
    }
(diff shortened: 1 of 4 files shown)


ericpassmore:
Note:start
category: Other
component: SHiP
summary: Fix hang on shutdown.
Note:end

spoonincode (changes requested):
> Modified SHiP `boost::beast::websocket` to use the SHiP thread instead of the main application thread executor strand

I think really what this changes is what I'd call the "default executor" of the websocket (I am not sure what the proper terminology is). My understanding was that the default executor was never used based on how the code is structured because the bound executor for every async call on the websocket was a `use_awaitable`, and that will use the coroutine's executor (which is the strand created from ship thread, in this case). The websocket certainly uses the ship thread and strand prior to this change. But if this does fix the problem there must be some other case where the default executor matters. And I do think it's good to make the default executor match up.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -101,7 +101,7 @@ struct state_history_plugin_impl {
    void create_listener(const std::string& address) {
       const boost::posix_time::milliseconds accept_timeout(200);
       // connections set must only be modified by main thread; run listener on main thread to avoid needing another post()
-      fc::create_listener<Protocol>(app().get_io_service(), _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
+      fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
```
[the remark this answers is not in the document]
heifner: In this case there is only one SHiP thread and its implicit strand. I agree that it would be better if `fc::create_listener` took a strand.

tests/ship_kill_client_test.py
```diff
@@ -0,0 +1,132 @@
...
+# Connect a number of clients and then kill the clients and shutdown nodoes.
+# nodeos should exit cleanly and not hang or SEGfAULT.
+#
+###############################################################
+
+Print=Utils.Print
+
+appArgs = AppArgs()
+extraArgs = appArgs.add(flag="--num-clients", type=int, help="How many ship_streamers should be started", default=1)
+args = TestHelper.parse_args({"--activate-if","--dump-error-details","--keep-logs","-v","--leave-running","--unshared"}, applicationSpecificArgs=appArgs)
```
linh2931, line 26: I think we can start to make new tests run only under Savanna so we don't need to change again when testing under Legacy is removed. So `--activate-if` is not needed; just start the cluster using `activateIF=True`.

plugins/state_history_plugin/state_history_plugin.cpp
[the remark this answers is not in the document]
spoonincode: Yeah for 1.0.x that's fine but we should really get it right on main -- the code is currently structured with the assumption that increasing ship threads 'just works'

tests/ship_kill_client_test.py
```diff
@@ -0,0 +1,132 @@
...
+    cmd = f"{shipClient} --start-block-num {start_block_num} --end-block-num {end_block_num} --fetch-block --fetch-traces --fetch-deltas --fetch-finality-data"
+    if Utils.Debug: Utils.Print(f"cmd: {cmd}")
+    clients = []
+    files = []
+    shipTempDir = os.path.join(Utils.DataDir, "ship")
+    os.makedirs(shipTempDir, exist_ok = True)
+    shipClientFilePrefix = os.path.join(shipTempDir, "client")
+
+    for i in range(0, args.num_clients):
+        start = time.perf_counter()
```
linh2931, line 103: This `start` is not used.

tests/ship_kill_client_test.py
```diff
@@ -0,0 +1,132 @@
...
+extraArgs = appArgs.add(flag="--num-clients", type=int, help="How many ship_streamers should be started", default=1)
+args = TestHelper.parse_args({"--activate-if","--dump-error-details","--keep-logs","-v","--leave-running","--unshared"}, applicationSpecificArgs=appArgs)
+
+Utils.Debug=args.v
+cluster=Cluster(unshared=args.unshared, keepRunning=args.leave_running, keepLogs=args.keep_logs)
+activateIF=args.activate_if
+dumpErrorDetails=args.dump_error_details
+walletPort=TestHelper.DEFAULT_WALLET_PORT
+
+totalProducerNodes=2
```
linh2931, line 34: Why do we need 2 producer nodes?

tests/ship_kill_client_test.py
```diff
@@ -0,0 +1,132 @@
...
+                      specificExtraNodeosArgs=specificExtraNodeosArgs) is False:
+        Utils.cmdError("launcher")
+        Utils.errorExit("Failed to stand up cluster.")
+
+    # verify nodes are in sync and advancing
+    cluster.waitOnClusterSync(blockAdvancing=5)
+    Print("Cluster in Sync")
+
+    prodNode0 = cluster.getNode(0)
+    nonprodNode = cluster.getNode(1)
```
linh2931, line 65: should this be prodNode1?

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -101,21 +101,24 @@ struct state_history_plugin_impl {
...
-                                            [this](session_base* conn) {
-                                               boost::asio::post(app().get_io_service(), [conn, this]() {
-                                                  connections.erase(connections.find(conn));
-                                               });
-                                            }, _log));
+      auto strand = boost::asio::make_strand(thread_pool.get_executor());
+      fc::create_listener<Protocol>(strand, _log, accept_timeout, address, "", [this, strand](Protocol::socket&& socket) {
+         boost::asio::post(app().get_io_service(), [this, strand, socket{std::move(socket)}]() mutable {
+            catch_and_log([this, &socket, &strand]() {
+               connections.emplace(new session(std::move(socket), std::move(strand), chain_plug->chain(),
```
spoonincode, line 108: The changes here seem to make only a single strand for all connections. And then this single strand is moved here? Doesn't look right. I think what you had before was better even though it still didn't make the websocket's default executor be a strand.

plugins/state_history_plugin/state_history_plugin.cpp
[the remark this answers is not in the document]
heifner: Seems passing a strand works.

tests/ship_kill_client_test.py
heifner: We don't, updated.

tests/ship_kill_client_test.py
heifner: Changed to have two non-producer nodes.

spoonincode (approved):

linh2931 (approved):

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -100,22 +100,24 @@ struct state_history_plugin_impl {
...
-                                               return chain_plug->chain().fetch_block_by_id(block_id);
-                                            },
-                                            [this](session_base* conn) {
-                                               boost::asio::post(app().get_io_service(), [conn, this]() {
-                                                  connections.erase(connections.find(conn));
-                                               });
-                                            }, _log));
+      // connections set must only be modified by main thread; run listener on ship thread so sockets use default executor of the ship thread
+      fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "", [this](Protocol::socket&& socket) {
+         boost::asio::post(app().get_io_service(), [this, socket{std::move(socket)}]() mutable {
```
spoonincode, line 105: Thinking about this more, I think this creates a (fantastically remote) possibility of the socket being destroyed after its executor is destroyed. This occurs when this callback is `post()`ed after the main thread is stopped, and then the plugin is destroyed, and then appbase is destroyed which will destroy pending callbacks. Obviously something we can tolerate for 1.0.x but another good reason to improve `create_listener`

PR #823, opened 2024-09-25
fix Python 3.12 `SyntaxWarning`s due to `invalid escape sequence`

spoonincode:
for example,
```
Node.py:674: SyntaxWarning: invalid escape sequence '\.'
  match=re.match("stderr\..+\.txt", entry.name)
```

I don't claim this PR does an exhaustive fix, these are just some that I saw flagged recently


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: fix Python 3.12 `SyntaxWarning` due to invalided escape sequence.
Note:end

Issue #742, opened 2024-09-11
Test regression of syncing and replay with actual committed reference blockchain data


arhag:
To be confident that changes do not unintentionally introduce changes the consensus protocol, we would like to add a new test (and testing utility) that checks for regression in compatibility of syncing and replaying reference blockchain data.

The test could store a reference blockchain (with QCs) generated and serialized (and committed to the source control) at some earlier point in time. Then it could read that blockchain into a nodeos (replay it) and use it as a source of blocks from which another nodeos instance can sync those blocks. The test should involve both a replay and sync to test both of those code paths in controller.

We would not normally commit that reference blockchain to source control. The test/utility could allow us to manually regenerate it if necessary, but we would really only need to do it once (or whenever we want to change the nature of the reference blockchain for the test case). Then the test, when it normally runs, just ensures it can validate that blockchain. However, it would be good for the test to generate the reference blockchain anyway, even though it is going to throw it away under typical use, just as an ongoing check in our CI to ensure the reference blockchain generation portion of the test does not break or regress.

This test (should be done with the Boost testing framework so that the generation of the reference blockchain is deterministic which provides multiple advantages. First, anytime we need to change the reference blockchain (which remember is binary data committed to source control), a reviewer could easily validate the changed data in a pull request by regenerating it themself locally. Second, it allows for the test to include a final check at the end where it compares the block ID of the reference chain it generates to the one that was in the source control to ensure they are the same.

Note that an early iteration of such a test was introduced as a `verify_spring_1_0_block_compatibitity` test as part of the work for https://github.com/AntelopeIO/spring/issues/694. But that version did not write out block data to source control and test that syncing and replaying those blocks still worked. And it also did not provide a utility to easily persist the generated block data to source control if desired. It just generated the test blockchain and verified the last block IDs matched a hard-coded value. That was a good start. But this issue is to go further and implement the test/utility as is described in this issue.

PR #827, opened 2024-09-26
[1.0.2] Test regression of syncing and replay with actual committed reference blockchain data

linh2931:
To be confident that changes do not unintentionally introduce changes of the consensus protocol, we would like to add a testing utility that generates reference blockchain data and a new test that checks for regression in compatibility of syncing and replaying the reference blockchain data.

The PR
* provides a command line option `--save-blockchain` to generate and save the reference blockchain data to `unittests/test-data/consensus_blockchain` directory. The full command is `unittests/unit_test -t savanna_misc_tests/verify_block_compatibitity -- --save-blockchain`.
* loads the reference blockchain into a nodeos by replaying it, and and verifies the block ID of the replayed LIB matches with the ID of LIB in the reference blocks log.
* starts another nodeos instance from reference snapshot, sync blocks from the replayed nodeos, and verifies resulting block ID.

Resolves https://github.com/AntelopeIO/spring/issues/742


unittests/savanna_misc_tests.cpp
```diff
@@ -591,18 +592,141 @@ BOOST_FIXTURE_TEST_CASE(validate_qc_requiring_finalizer_policies, savanna_cluste
...
+   std::vector<char> buf;
+   buf.resize(size);
+
+   ref_file.seek(0);
+   ref_file.read(buf.data(), size);
+
+   ref_file.close();
+
+   return buf;
+}
```
greg7mdp, line 648: You could use `fc::read_file_contents()` here, as I learned last week from Kevin!

linh2931: Thanks!

unittests/savanna_misc_tests.cpp
```diff
@@ -681,10 +799,30 @@ BOOST_FIXTURE_TEST_CASE(verify_spring_1_0_block_compatibitity, savanna_cluster::
    BOOST_REQUIRE_EQUAL(qc_s(qc(b9)), strong_qc(b8));    // b9 claims a strong QC on b8
    BOOST_REQUIRE_EQUAL(A.lib_number, b6->block_num());  // b9 makes B6 final
 
-   // check that the block id of b9 match what we got with Spring 1.0.0
+   std::filesystem::path test_data_path { UNITTEST_TEST_DATA_DIR };
+   auto ref_blockchain_path = test_data_path / "consensus_blockchain";
+
+   // check that the block id of b9 match what we got before.
    auto b9_id = b9->calculate_id();
-   BOOST_REQUIRE_EQUAL(b9_id, block_id_type{"00000013725f3d79bd4dd4091d0853d010a320f95240981711a942673ad87918"});
```
greg7mdp, line 686: I think we can leave the comparison with a hardcoded id.

greg7mdp: Never mind, comparing with the stored id is fine

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -521,6 +521,13 @@ namespace eosio::testing {
             return {cfg, gen};
          }
 
+         static bool arguments_contains(std::string arg) {
```
heifner, line 524: `const&`

unittests/savanna_misc_tests.cpp
```diff
@@ -591,18 +594,133 @@ BOOST_FIXTURE_TEST_CASE(validate_qc_requiring_finalizer_policies, savanna_cluste
...
+   auto source_index_file = blocks_path / "blocks.index";
+
+   auto ref_log_file             = ref_blockchain_path / "blocks.log";
+   auto ref_index_file           = ref_blockchain_path / "blocks.index";
+   auto ref_id_file_name         = ref_blockchain_path / "id";
+   auto ref_snapshot_file_name   = ref_blockchain_path / "snapshot";
+
+   // save reference blocks log
+   std::filesystem::copy_file(source_log_file, ref_log_file, std::filesystem::        copy_options::overwrite_existing);
+   std::filesystem::copy_file(source_index_file, ref_index_file, std::filesystem::        copy_options::overwrite_existing);
```
heifner, line 611: Remove extra spaces after `std::filesystem::`

heifner (changes requested):

heifner (approved):

unittests/savanna_misc_tests.cpp
```diff
@@ -681,10 +799,30 @@ BOOST_FIXTURE_TEST_CASE(verify_spring_1_0_block_compatibitity, savanna_cluster::
...
+
+   // check that the block id of b9 match what we got before.
    auto b9_id = b9->calculate_id();
-   BOOST_REQUIRE_EQUAL(b9_id, block_id_type{"00000013725f3d79bd4dd4091d0853d010a320f95240981711a942673ad87918"});
+   BOOST_REQUIRE_EQUAL(b9_id, read_reference_id(ref_blockchain_path));
+
+   if (save_blockchain) {
+      save_blockchain_data(ref_blockchain_path, A.get_config().blocks_dir, b9_id, snapshot);
+      return;
+   }
```
greg7mdp, line 812: I believe this should be before line 807, otherwise is there is a need to save the blockchain (meaning that it is different from what is already saved), line 807 would fail first.

unittests/savanna_misc_tests.cpp
```diff
@@ -681,10 +799,30 @@ BOOST_FIXTURE_TEST_CASE(verify_spring_1_0_block_compatibitity, savanna_cluster::
...
+   // replay the reference blockchain and make sure LIB id in the replayed
+   // chain matches reference LIB id
+   // --------------------------------------------------------------------
+   fc::temp_directory temp_dir; // need to pass in temp_dir. otherwise it would be destroyed after replay_reference_blockchain returns
+   std::unique_ptr<tester> replay_chain = replay_reference_blockchain(ref_blockchain_path, temp_dir, blog);
 
+   // start another blockchain using reference snapshot, and sync with the blocks
+   // from the replayed blockchain
+   // ---------------------------------------------------------------------------
+   sync_replayed_blockchain(ref_blockchain_path, std::move(replay_chain), blog);
```
greg7mdp, line 825: This should check that the result of replaying the blockchain matches `b1` through `b9`. Unless I'm mistaken, here it just checks the consistency of what is stored in the reference chain.

linh2931: At the end of both `replay_reference_blockchain()` and `sync_replayed_blockchain()`, replayed/synched block ID is verified against with reference block ID

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Enhance test framework and test cases to sync and replay using actual committed reference blockchain data.
Note:end

PR #845, opened 2024-09-30
[1.0.2] SHiP: Fix hang on exit

### LINKED ISSUE #822: Test failure: ship_kill_client_test
https://github.com/AntelopeIO/spring/actions/runs/11037427972/job/30658923368

Looks like the nodeos with SHiP hung on exit. Appears it may be same issue as #815.

heifner:
This is a temporary fix until issue #842 can be worked. Some alternatives were attempted, but no simple fix could be found that didn't require more work toward #842 then desired for 1.0.x. See https://github.com/AntelopeIO/appbase/pull/34

Test runs of a hacked up state_history_plugin with sleeps that caused it to hang on almost every CI/CD run.
Example failures: https://github.com/AntelopeIO/spring/actions/runs/11076070901
With this fix: https://github.com/AntelopeIO/spring/actions/runs/11124229769

Resolves #822

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -380,6 +380,14 @@ void state_history_plugin_impl::plugin_shutdown() {
    accepted_block_connection.reset();
    block_start_connection.reset();
    thread_pool.stop();
+
+   // This is a temporary fix until https://github.com/AntelopeIO/spring/issues/842 can be worked.
+   // Drain the io_service of anything that could be referencing state_history_plugin
+   if (app().executor().get_io_service().stopped()) {
+      app().executor().get_io_service().restart();
+      while (app().executor().get_io_service().poll())
+         ;
+   }
 }
 
 void state_history_plugin::plugin_shutdown() {
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -387,6 +387,8 @@ void state_history_plugin_impl::plugin_shutdown() {
       app().executor().get_io_service().restart();
       while (app().executor().get_io_service().poll())
          ;
+      // clear priority queue of anything pushed by poll(), see application_base exec()
+      app().executor().clear();
    }
 }
 


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -380,6 +380,14 @@ void state_history_plugin_impl::plugin_shutdown() {
...
    block_start_connection.reset();
    thread_pool.stop();
+
+   // This is a temporary fix until https://github.com/AntelopeIO/spring/issues/842 can be worked.
+   // Drain the io_service of anything that could be referencing state_history_plugin
+   if (app().executor().get_io_service().stopped()) {
+      app().executor().get_io_service().restart();
+      while (app().executor().get_io_service().poll())
+         ;
+   }
```
spoonincode, line 390: It looks like a SIGINT etc here will interrupt it. Probably fine.

I wonder if there is a way to add some sort of sentinel to know that "anything that could be referencing state_history_plugin" _has_ actually been drained, so we don't somehow get in a state where we loop forever here because something else unrelated is `post()`ing forever. I'm not immediately coming up with a good solution though.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -380,6 +380,14 @@ void state_history_plugin_impl::plugin_shutdown() {
    accepted_block_connection.reset();
    block_start_connection.reset();
    thread_pool.stop();
+
```
spoonincode, line 383: With the connections torn down, is there any risk pumping the main thread below might apply a block that ship would then not see?

plugins/state_history_plugin/state_history_plugin.cpp
heifner: Maybe just not do the `poll()` call in a loop. We are not expecting anything to be posting at this point.

plugins/state_history_plugin/state_history_plugin.cpp
heifner: Good call! It does seem like this could push a block into the controller that would be missed by the controller.

heifner: Actually, no I don't think that is not possible. Everything that uses `app().executor().post` when they execute they are only pushed into the priority_queue to execute. So probably should clear that out again here.

plugins/state_history_plugin/state_history_plugin.cpp
heifner: Interesting enough, the `poll()` has to be in a loop. I guess the `co_spawn` must be posting additional as it drains.

greg7mdp (approved):

greg7mdp (approved):

plugins/state_history_plugin/state_history_plugin.cpp
spoonincode: I wonder if we even need those connections to be manually torn down

spoonincode (approved):

plugins/state_history_plugin/state_history_plugin.cpp
heifner: Probably not, but that can be cleaned up when we work the real solution.

ericpassmore:
Note:start
category: Other
component: SHiP
summary: Fix hang on exit, draining references to SHiP found in io_service.
Note:end

Issue #849, opened 2024-10-01
`time` field for produced blocks is wrong and the name `time` is confusing with received block's `time`


linh2931:
The `time` field reported for `Produced block` is wrong:

`node_bios/stderr.2024_10_01_11_52_13.txt:info  2024-10-01T15:53:04.594 nodeos   controller.cpp:3475  log_applied  ] Produced block 0de718e5b389a264... #104 @ 2024-10-01T15:53:05.000 signed by eosio [trxs: 0, lib: 102, confirmed: , net: 0, cpu: 100, elapsed: 234 us, time: 1727797984594382 us]`

This happens in both `main` and `rel 1.0` branches.

Both `produced blocks` and `received blocks` use the same short name `time`, which is confusing. It would be cleaner to use `producing time` and `applying time`.

`cpu` field could add postfix `us` in the same way as other time.

PR #850, opened 2024-10-01
[1.0.2] Fix reported time for produced blocks and fix confusing reported time name

linh2931:
This PR
* fixes incorrect reported time for "Produced blocks"
* changes `time` for "Produced blocks" to `producing time`, and for "Received blocks" to `applying time`
* adds time unit `us` to reported CPU time

Before the changes
`info  2024-10-01T18:31:59.066 nodeos    controller.cpp:3462           commit_block         ] Produced block dd7794184632fdd2... #100 @ 2024-10-01T18:31:59.500 signed by eosio [trxs: 0, lib: 98, net: 0, cpu: 100, elapsed: 342 us, time: 1727797983669080 us]`

After
`info  2024-10-01T19:07:27.667 nodeos    controller.cpp:3461           commit_block         ] Produced block b29e040564b322f2... #100 @ 2024-10-01T19:07:28.000 signed by eosio [trxs: 0, lib: 98, net: 0, cpu: 100 us, elapsed: 291 us, producing time: 451393 us]`

Fix https://github.com/AntelopeIO/spring/issues/849

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3459,7 +3459,7 @@ struct controller_impl {
             fc::time_point now = fc::time_point::now();
 
             ilog("Produced block ${id}... #${n} @ ${t} signed by ${p} "
-                 "[trxs: ${count}, lib: ${lib}${confs}, net: ${net}, cpu: ${cpu}, elapsed: ${et} us, time: ${tt} us]",
+                 "[trxs: ${count}, lib: ${lib}${confs}, net: ${net}, cpu: ${cpu}, elapsed: ${et} us, producing time: ${tt} us]",
                  ("p", new_b->producer)("id", id.str().substr(8, 16))("n", new_b->block_num())("t", new_b->timestamp)
                  ("count", new_b->transactions.size())("lib", chain_head.irreversible_blocknum())("net", br.total_net_usage)
                  ("cpu", br.total_cpu_usage_us)("et", br.total_elapsed_time)("tt", now - br.start_time)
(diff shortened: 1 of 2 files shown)


libraries/chain/controller.cpp
```diff
@@ -5257,6 +5259,7 @@ void controller::assemble_and_complete_block( block_report& br, const signer_cal
       valid_block_signing_authority);
 
    br = my->pending->_block_report;
+   br.start_time = start_time;
```
heifner, line 5262: This does not look correct. You want the total time for producing the block. This should be from the time of start_block() which should be what is in pending.

heifner (changes requested):

libraries/chain/controller.cpp
linh2931: Do you mean we don't need to do anything here, just use the `start_time` in pending as it is set to `now` at construction?

heifner: I would assume so, yes.

heifner (approved):

libraries/chain/controller.cpp
linh2931: After the change, producing time looks close to 500 ms like
`info  2024-10-01T19:07:27.667 nodeos    controller.cpp:3461           commit_block         ] Produced block b29e040564b322f2... #100 @ 2024-10-01T19:07:28.000 signed by eosio [trxs: 0, lib: 98, net: 0, cpu: 100 us, elapsed: 291 us, producing time: 451393 us]`

libraries/chain/controller.cpp
```diff
@@ -3459,7 +3459,7 @@ struct controller_impl {
             fc::time_point now = fc::time_point::now();
 
             ilog("Produced block ${id}... #${n} @ ${t} signed by ${p} "
-                 "[trxs: ${count}, lib: ${lib}${confs}, net: ${net}, cpu: ${cpu}, elapsed: ${et} us, time: ${tt} us]",
+                 "[trxs: ${count}, lib: ${lib}${confs}, net: ${net}, cpu: ${cpu} us, elapsed: ${et} us, producing time: ${tt} us]",
```
greg7mdp, line 3462: I'm not sure how a user would understand the difference between `elapsed` and `producing time`?

greg7mdp (approved):

ericpassmore:
Note:start
category: Logging
component: Internal
summary: Fix reported time for produced blocks and disambiguate time in logs.
Note:end

PR #853, opened 2024-10-01
[1.0.2] P2P: Resolve on reconnect

### LINKED ISSUE #525: P2P: Resolve endpoints on re-connect attempts
Currently `net_plugin` resolves endpoints at startup and then reuses the `tcp::resolver::results_type` for any re-connection attempt. If resolve fails at startup then the resolve is not attempted again. Also `net_plugin` reports an unhelpful "Element not found` error.

See https://github.com/AntelopeIO/spring/pull/523#discussion_r1714117611

heifner:
Resolve address on reconnect. (Take 2, See #825)

Instead of:
````
info  2024-09-25T21:50:50.686 net-1     net_plugin.cpp:2797           operator()           ] connection failed to p2p.telos-bad.zenblocks.io:9876, Element not found
info  2024-09-25T21:50:50.686 net-1     net_plugin.cpp:1461           _close               ] ["p2p.telos-bad.zenblocks.io:9876" - 1 <unknown>:<unknown>] closing
````
Now you get:
```
warn  2024-09-25T21:42:48.642 net-0     net_plugin.cpp:4524           operator()           ] Unable to resolve p2p.telos-bad.zenblocks.io:9876 Host not found (authoritative)
```

This takes a different approach than #825. This approach is closer to Leap 4.0 approach to connection lifetime although it does keep the `connections_manager` added in Leap 5.0.

Plan on backporting to Leap 5.0 once this is merged.

Resolves #525

### CODE REVIEW

File: plugins/net_plugin/net_plugin.cpp [hunk 1]
```diff
@@ -4670,56 +4640,75 @@ namespace eosio {
          return "invalid peer address";
       }
 
-      std::lock_guard g( connections_mtx );
-      if( find_connection_i( peer_address ) )
-         return "already connected";
+      {
+         std::lock_guard g( connections_mtx );
```
greg7mdp, line 4644: If `connections_mtx` is to remain a `shared_mutex` (which I'm not sure is a great idea), then this probably should be `std::shared_lock`.  Same for `connections_manager::number_connections()` and the unique_lock in `connections_manager::connect_supplied_peers`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4670,56 +4640,75 @@ namespace eosio {
...
-            } else {
-               fc_wlog( logger, "Unable to resolve ${host}:${port} ${error}",
-                        ("host", host)("port", port)( "error", err.message() ) );
-               it->c->set_state(connection::connection_state::closed);
-               ++(it->c->consecutive_immediate_connection_close);
-            }
-      } );
+      connection_ptr c = std::make_shared<connection>( peer_address, listen_address );
+      if (c->resolve_and_connect()) {
+         add(c);
```
greg7mdp, line 4653: `c` could be moved I guess, but probably not measurable difference.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4748,21 +4737,31 @@ namespace eosio {
    }
 
    std::optional<connection_status> connections_manager::status( const string& host )const {
-      std::shared_lock g( connections_mtx );
-      auto con = find_connection_i( host );
+      connection_ptr con;
+      {
+         std::shared_lock g( connections_mtx );
+         con = find_connection_i( host );
+      }
```
greg7mdp, line 4744: Should we keep the lock until we are done using `con`?

heifner: No. `get_status` grabs a different mutex. Safer to not hold both.

greg7mdp: So should we do the same [here](https://github.com/AntelopeIO/spring/blob/6d36988895b0b2d148e15e32a801b17cff55fd6f/plugins/net_plugin/net_plugin.cpp#L4741)

heifner: I did do it there.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4670,56 +4640,75 @@ namespace eosio {
...
-      std::lock_guard g( connections_mtx );
-      const auto& index = connections.get<by_connection>();
-      const auto& it = index.find(c);
-      if( it != index.end() ) {
-         it->c->connect( it->ips );
+      string::size_type colon = peer_address().find(':');
+      if (colon == std::string::npos || colon == 0) {
+         fc_elog( logger, "Invalid peer address. must be \"host:port[:<blk>|<trx>]\": ${p}", ("p", peer_address()) );
+         return false;
+      }
```
spoonincode, line 4678: There seems to be a little duplication between this block and the `split_host_port_type()` below. I'm not sure it's worth trying to fix up though.

heifner: That is a bit of a mess. I'll fix it up.

ericpassmore:
Note:start
category: Other
component: P2P
summary: When reconnecting using the net plugin resolve endpoint.
Note:end

plugins/net_plugin/net_plugin.cpp

spoonincode (approved):

greg7mdp (approved):

Issue #578, opened 2023-02-05
Refactor: use ORCv2 for compatibility with modern LLVM versions with EOS VM OC


spoonincode:
EOS VM OC was implemented using LLVM's ORCv1 interface. This was actually already in deprecation at the time of EOSIO 2.0's release (circa LLVM 9). LLVM removed ORCv1 in version 12 meaning Leap cannot be compiled with LLVM 12+

So far this hasn't been too much of a problem: Ubuntu 22.04 and upcoming 23.04 ship with an optional llvm-11-dev package that works. Both Arch & even upcoming Fedora 38 still provide an llvm11 package. But the days of a legacy LLVM 11 package being provided by distros are numbered. Upcoming Debian 12 does _not_ have an LLVM 11 package.

EOS VM OC needs to be refactored to support modern LLVM versions (such refactoring may not be limited to just ORC interface). ORCv2 was available in LLVM 7 I believe, so it's possible the implementation could be refactored to support all versions from 7 to 16. But we could opt to reduce the scope of the support if need be (Ubuntu 20.04 ships with LLVM 10 as the "default" LLVM).

ericpassmore:
Running into this issue now on AWS ubuntu 22.04 for arm

PR #855, opened 2024-10-01
add manually run workflow to exhaustively test all LLVM versions expected to work

spoonincode:
Spring is expected to build with LLVM versions 7 through 11. As #578 is worked this range of versions will change to include newer versions. I'd like to have a way to demonstrate that newer versions do indeed compile and test successfully beyond "trust me bro".

So add a simple manually dispatched workflow that compiles with all supported LLVM major versions. This seems like a good enough tradeoff over doing something vastly more complex in CI, and I certainly do not want this to run on every PR. The main risk I see is that since this workflow is disparate to the primary per-PR build workflow it can potentially get out of date; such as the build dependencies or `cmake`/`ctest` arguments going stale. I feel like that's an acceptable risk at the moment.

A run of this workflow can be found here,
https://github.com/AntelopeIO/spring/actions/runs/11132990881

### DIFF
--- a/.github/workflows/llvm.yaml
+++ b/.github/workflows/llvm.yaml
@@ -1,0 +1,41 @@
+name: Exhaustive LLVM Build
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  doit:
+    name: 'Build & Test (${{matrix.version}})'
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ['10.0.1', '11.1.0']
+    runs-on: ["self-hosted", "enf-x86-beefy"]
+    container: ubuntu:jammy
+    steps:
+      - name: Install packages
+        run: |
+          apt-get update && apt-get -y upgrade
+          apt-get install -y build-essential cmake git libcurl4-openssl-dev libgmp-dev ninja-build python3 zlib1g-dev
+      - name: Clone LLVM
+        run: git clone -b llvmorg-${{matrix.version}} --single-branch --recursive https://github.com/llvm/llvm-project
+      - name: Build LLVM
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/llvm.yaml
```diff
@@ -0,0 +1,41 @@
...
+permissions:
+  contents: read
+
+jobs:
+  doit:
+    name: 'Build & Test (${{matrix.version}})'
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ['10.0.1', '11.1.0']
```
spoonincode, line 15: But wait, if we expect 7 through 11 to work why is this new workflow only testing 10 and 11? ah yes, tune in soon for the next PR...

.github/workflows/llvm.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          path: src
+      - name: Build spring
+        run: |
+            cmake -S src -B build -DCMAKE_BUILD_TYPE=Release -GNinja
+            cmake --build build
+      - name: Test spring
+        run: ctest --test-dir build -j $(nproc) --output-on-failure -LE "(nonparallelizable_tests|long_running_tests)" -E eosvmoc_limits_unit_test_eos-vm-oc --timeout 480
```
spoonincode, line 41: The `eosvmoc_limits_unit_test` is very sensitive to exact code generated since its pass criteria was empirically derived from a single LLVM version with a threshold of 1 byte. Need to think on how to improve this test; disable in here for now.

linh2931 (approved):

heifner (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Test more versions of LLVM.
Note:end

PR #858, opened 2024-10-01
[5.0 -> 1.0.2] remove EOS VM OC's erroneous `GS register is not as expected message`

spoonincode:
spring 1.0 merge of AntelopeIO/leap#2406. This message,
> `x86_64 GS register is not set as expected. EOS VM OC may not run correctly on this platform`

Seems to be even more common on Spring 1.0, I believe because it's more common that forkdb will apply blocks during startup.


linh2931 (approved):

heifner (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Clean up unhelpful GS register error.
Note:end

PR #867, opened 2024-10-02
[1.0.2 -> main] Don't assume that claimed block is in fork_db. 

### LINKED ISSUE #778: fork_db check prevents voting on blocks including a qc claim on a block not in fork_db.

greg7mdp:
Resolves #778 and #751.

Update fork_db_validated_block_exists to lookup parents instead of just claimed block.

Also updated the validated member in both block_state_legacy and block_state to be copyable_atomic<bool> instead of bool, which allowed to make the set_valid() and is_valid() members public, and thus eliminate a couple of friend declarations.

Also adds testcase described in issue #751.


heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Validate claimed block in in fork_db.
Note:end

Issue #846, opened 2024-10-01
Test failure: nodeos_snapshot_diff_if_test 


heifner:
https://github.com/AntelopeIO/spring/actions/runs/11115558731/job/30884557872

Requested to create a snapshot at block 213, but the node was already at block 217.

heifner:
https://github.com/AntelopeIO/spring/actions/runs/11131513435/job/30933803826

linh2931:
https://github.com/AntelopeIO/spring/actions/runs/11131822845/job/30934996338?pr=850#step:4:1735

heifner:
https://github.com/AntelopeIO/spring/actions/runs/11133200464/job/30939501536

greg7mdp:
https://github.com/AntelopeIO/spring/actions/runs/11133625885/job/30940631902
https://github.com/AntelopeIO/spring/actions/runs/11134937237/job/30944403983

spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/11147253627/job/30982140532

spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/11149293818/job/30988479537

spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/11150390339/job/30991966826

spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/11152694055/job/30999205664

PR #875, opened 2024-10-02
[1.0.3] Fix flaky nodeos_snapshot_diff_test due to a race condition

linh2931:
Recently, `nodeos_snapshot_diff_test` fails in almost every CICD run.

This is due to a race condition that at the startup of head mode node, net thread and http thread can start to work in different order. If http thread processes `schedule_snapshot` request after net thread starts to sync with the irreversible mode node, schedule_snapshot request will miss the scheduled block number. If it is before net thread starts to sync with the irreversible mode node, `schedule_snapshot` request will catch the scheduled block number and the snapshot is taken.

This PR changes to start up the head mode node while the irreversible mode node is down so that net thread will not sync, make `schedule_snapshot` request, and finally restart irreversible node.

Test run in CICD for 3 times. No failure happened.

Resolves https://github.com/AntelopeIO/spring/issues/846


greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Snapshot
summary: Fix flaky `nodeos_snapshot_diff_test` due to a race condition.
Note:end

ericpassmore (approved):

PR #881, opened 2024-10-02
[1.0.3] Fix issue with block report log time including the time for `maybe_switch_forks`

### LINKED ISSUE #831: Block report logs incorrect times
```
debug 2024-09-27T03:35:46.775 nodeos    producer_plugin.cpp:2165      start_block          ] Starting block #396367492 2024-09-27T03:35:47.500 producer atticlabeosb, deadline 2024-09-27T03:35:47.275
...
debug 2024-09-27T03:35:47.112 nodeos    producer_plugin.cpp:310       report               ] Block #396367492 atticlabeosb trx idle: 192987us out of 427393us, success: 74, 143230us, fail: 0, 0us, transient: 0, 0us, other: 91176us
```
Looks like the total time of `427393` and the `other` time of `91176` includes the processing of the incoming block:
```
debug 2024-09-27T03:35:46.685 nodeos    producer_plugin.cpp:310       report               ] Block #396367491 atticlabeosb trx idle: 292841us out of 504843us, success: 43, 152771us, fail: 9, 276us, transient: 0, 0us, other: 58955us
```
Not sure if this is always incorrect or only when restarting a speculative block:
```
debug 2024-09-27T03:35:46.684 nodeos    producer_plugin.cpp:987       restart_speculative_ ] Restarting exhausted speculative block #396367491
```

greg7mdp:
Resolves #831.

Also did some cleanup, replacing one instance of `report` with `populate_speculative_block_metrics` (which does just that instead of calling the other `report`.)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2020,11 +2022,10 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    if (!chain_plug->accept_transactions())
       return start_block_result::waiting_for_block;
 
-   abort_block();
-
    chain.maybe_switch_forks([this](const transaction_metadata_ptr& trx) { _unapplied_transactions.add_forked(trx); },
                             [this](const transaction_id_type& id) { return _unapplied_transactions.get_trx(id); });
 
+   abort_block(); // does reset `_time_tracker`, so should be after `maybe_switch_forks`
```
heifner, line 2028: I think the `abort_block()` should be put back where it was. `maybe_switch_forks` EOS_ASSERTs that there is not a pending block. There are a few places where `abort_block()` is not called before `schedule_production_loop`. If we cleaned that up the `abort_block()` would not be needed at all. Regardless you would want to reset the timer instead then as `abort_block()` would not be needed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2954,25 +2955,26 @@ void producer_plugin_impl::produce_block() {
    chain.commit_block(br);
 
    const signed_block_ptr new_b = chain.head().block();
+   fc::time_point now = fc::time_point::now();
+
+   _time_tracker.add_other_time();
+   _time_tracker.report(new_b->block_num(), new_b->producer, now);
```
heifner, line 2961: We want to include the time for updating prometheus in `other`.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/881/commits/b776b9d4462b24e681ef5a93938bd06c1ac1a3cb

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/881/commits/a0c811772ad6b1896840c83921f51b9c5dfc8ba8

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Fix issue with block report log time including the time for `maybe_switch_forks`.
Note:end

PR #895, opened 2024-10-04
fix build with clang19

spoonincode:


libraries/libfc/include/fc/io/datastream.hpp
```diff
@@ -110,8 +110,8 @@ class datastream<Streambuf, typename std::enable_if_t<std::is_base_of_v<std::str
 
    size_t read(char* data, size_t n) { return buf.sgetn(data, n); }
    size_t write(const char* data, size_t n) { return buf.sputn(data, n); }
-   size_t tellp() { return this->pubseekoff(0, std::ios::cur); }
-   bool   skip(size_t p) { this->pubseekoff(p, std::ios::cur);  return true;  }
+   size_t tellp() { return buf.pubseekoff(0, std::ios::cur); }
+   bool   skip(size_t p) { buf.pubseekoff(p, std::ios::cur);  return true;  }
```
spoonincode, line 114: This is wild how it was never detected before.

greg7mdp: Probably because we never create a `datastream` with a class inheriting from `std::streambuf`.

Also, I believe that the way we use `enable_if_t` here is wrong, see the notes [here](https://en.cppreference.com/w/cpp/types/enable_if).

greg7mdp (approved):

libraries/libfc/include/fc/io/datastream.hpp
spoonincode: yeah I was a little worried the reason it was being flagged now was because something with the `enable_if` is goofed up and this was now being improperly selected.. but I couldn't find evidence of that actually occurring.

linh2931 (approved):

libraries/libfc/include/fc/io/datastream.hpp
spoonincode: It might be a fun tech debt item to replace all these across the code with c++20 `requires`

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Fix build to work with clang19.
Note:end

PR #590, opened 2024-08-19
[1.0] P2P: Track sync time to know when to allow a new start sync

heifner:
Track when syncing starts as a backup determination on when to allow a new sync to begin. This continues to prevent multiple syncs to be kicked off from different connections at the same time. It works around a race condition on a connection syncing but with no active sync timer running.

Resolves #583

PR #898, opened 2024-10-07
[1.0.3 -> main] Test failure: p2p_sync_throttle_test

### LINKED ISSUE #882: Test failure: p2p_sync_throttle_test
https://github.com/AntelopeIO/spring/actions/runs/11152149161/job/30997817134

greg7mdp:
Resolves #882.

While moving some code around (see [here](https://github.com/AntelopeIO/spring/pull/590/files)), a call to:

`void connection::close(bool reconnect, bool shutdown)`

became a call to:

`int close(int __fd)`

This restores the intended behavior (close the actual connection, instead of closing fd number 1)

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2188,7 +2188,7 @@ namespace eosio {
       if( !ec ) {
          peer_dlog(c, "sync timed out");
          sync_reassign_fetch( c );
-         close(true);
+         c->close(true);
       } else if( ec != boost::asio::error::operation_aborted ) { // don't log on operation_aborted, called on destroy
          peer_elog( c, "setting timer for sync request got error ${ec}", ("ec", ec.message()) );
       }


heifner (approved):

spoonincode (approved):
obviously we need to merge this in now, but seems the tests are all failing

heifner:
> obviously we need to merge this in now, but seems the tests are all failing

Seems it has exposed a bug in sync wait timeout causing syncing to not continue correctly. I'm looking into it.

greg7mdp:
> Seems it has exposed a bug in sync wait timeout causing syncing to not continue correctly. I'm looking into it.

I wonder if `sync_reassign_fetch`:

https://github.com/AntelopeIO/spring/blob/c431b4961775bc3a56cb6ce6f2cbdf0dd20dd99b/plugins/net_plugin/net_plugin.cpp#L2190-L2191)

calls `request_next_chunk`:

https://github.com/AntelopeIO/spring/blob/c431b4961775bc3a56cb6ce6f2cbdf0dd20dd99b/plugins/net_plugin/net_plugin.cpp#L2200-L2212

which reassigns the same connection, which we close immediately afterwards?

heifner:
> which reassigns the same connection to `sync_source`, which we close immediately afterwards?

Yes, that is possible in general and since only one connection for this test then it would be the same one. However, I think the issue here is that the throttling node is not sending any blocks after it hits the throttle state.

greg7mdp:
> since only one connection for this test then it would be the same one

How can it not crash is we [post a request](https://github.com/AntelopeIO/spring/blob/c431b4961775bc3a56cb6ce6f2cbdf0dd20dd99b/plugins/net_plugin/net_plugin.cpp#L2070-L2073) and then immediately after close the connection?

heifner:
> How can it not be a problem if we post a request and then immediately after close the connection?

The connection has not sent a block in an expected timeframe. The response to that is to disconnect. That is our only connection for this test, but regardless, it is the right thing to do. The real bug here is that the node didn't send a block in the expected time frame, see #899.

If we close the connection then any pending requests are not sent.

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: Fix p2p sync to close connection.
Note:end

PR #897, opened 2024-10-07
[1.0] Test failure: p2p_sync_throttle_test

greg7mdp:
Resolves #882.

While moving some code around (see [here](https://github.com/AntelopeIO/spring/pull/590/files)), a call to:

`void connection::close(bool reconnect, bool shutdown)`

became a call to:

`int close(int __fd)`

This restores the intended behavior (close the actual connection, instead of closing fd number 1)

PR #900, opened 2024-10-07
[1.0.3 -> main] P2P: Schedule a retry of enqueue_sync_block when throttled

### LINKED ISSUE #882: Test failure: p2p_sync_throttle_test

heifner:
When a connection is throttled it will not attempt to enqueue another block unless `enqueue_sync_block` is called. This will happen on any message sent to the connection. However, that can be a long time if the node has to wait for a heart-beat timeout before some message is sent to the peer. 

Instead, schedule a try-again call to `enqueue_sync_block` in 100ms. 

This fixes the current issue with `p2p_sync_throttle_test` which could not make progress because it repeately received a sync wait timeout.

Merges `release/1.0` into `main` including #899 

Partially resolves #882, See #897

### DIFF
--- a/libraries/libfc/include/fc/io/datastream.hpp
+++ b/libraries/libfc/include/fc/io/datastream.hpp
@@ -110,8 +110,8 @@ class datastream<Streambuf, typename std::enable_if_t<std::is_base_of_v<std::str
 
    size_t read(char* data, size_t n) { return buf.sgetn(data, n); }
    size_t write(const char* data, size_t n) { return buf.sputn(data, n); }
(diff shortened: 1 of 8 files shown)


spoonincode:
needs to be based on `gh_882_main` unless you want two PRs merged in one upmerge

greg7mdp (approved):

spoonincode (approved):

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: Schedule a retry of `enqueue_sync_block` when throttled.
Note:end

PR #902, opened 2024-10-07
relax eosvmoc_limits_test's generated_code_size_limit pass/fail criteria; reenable test in exhaustive LLVM workflow

spoonincode:
Relax the pass/fail criteria for this test. Generated code size can vary based on LLVM version being used and this test is not intended to detect regressions or minute differences between those LLVM versions. (put another way, the generated code size is a subjective limit)

Reenable `eosvmoc_limits_unit_test_eos-vm-oc` in the exhaustive LLVM workflow now that the different versions the workflow uses will not get caught up on this test. Example passing run is at,
https://github.com/AntelopeIO/spring/actions/runs/11223260327

### DIFF
--- a/.github/workflows/llvm.yaml
+++ b/.github/workflows/llvm.yaml
@@ -38,4 +38,4 @@ jobs:
             cmake -S src -B build -DCMAKE_BUILD_TYPE=Release -GNinja
             cmake --build build
       - name: Test spring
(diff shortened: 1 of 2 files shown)


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Relax eos vm test limits on generated code size.
Note:end

PR #908, opened 2024-10-08
[1.0.3] Test failure: distributed-transactions-if-test

### LINKED ISSUE #848: Test failure: distributed-transactions-if-test
https://github.com/AntelopeIO/spring/actions/runs/11125907838/job/30915459664

`Duplicate Staring block found: 100 in /__w/spring/spring/build/TestLogs/distributed-transactions-test8/node_bios/stderr.2024_10_01_13_21_16.txt`

Received block 100 late so it restarted speculative block 100. This is expected behavior, so not really sure what we want to do with this. The check was added to guard agaisnt a bug where a block was restarted when it was not suppose to be. Maybe the test should look at the timestamps and only fail if the start block is within some threshold.

greg7mdp:
Resolves #848.

This updates the `verifyStartingBlockMessages` check in `Node.py` to check that we have only one `Starting block` in the log for any block number *unless*:

1. the block was restarted because it was exhausted, 
2. or the second `Starting block` is for a different block time than the first.

I have checked the new regular expression with a small python script against the log and it appears to work fine.

script:
```
import re

def check_file_for_pattern(file_path, pattern):
    regex = re.compile(pattern)

    with open(file_path, 'r') as file:
        for line_num, line in enumerate(file, start=1):
            match = regex.match(line)
            if match:
                blockNumber, time = match.group(1), match.group(2)
                print(f"{blockNumber} -> {time}")

if __name__ == "__main__":
    file_path = "stderr.2024_10_01_13_21_16.txt"
    pattern = r".*Starting block #(\d+) .*(\d\d:\d\d\.\d\d\d) producer"
    check_file_for_pattern(file_path, pattern)
```

execution:

```
❯ python ~/tmp/check_regexp.py
2 -> 21:17.000
3 -> 21:17.500
4 -> 21:18.000
5 -> 21:18.500
6 -> 21:19.000
7 -> 21:19.500
8 -> 21:20.000
9 -> 21:20.500
10 -> 21:21.000
...
```

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -729,7 +729,7 @@ class Node(Transactions):
                     match = starting_block_regexp.match(line)
                     if match:
                         blockNumber, time = match.group(1), match.group(2)
-                        if blockNumber in notRestartedBlockNumbersAndTimes and notRestartedBlockNumbersAndTimes[blockNumber] != time:
+                        if blockNumber in notRestartedBlockNumbersAndTimes and notRestartedBlockNumbersAndTimes[blockNumber] == time:
                             print(f"Duplicate Starting block found: {blockNumber} in {f}")
                             duplicateStartFound = True
                             break
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -708,37 +708,36 @@ class Node(Transactions):
                 count += contents.count(searchStr)
         return count
 
-    # verify only one or two 'Starting block' per block number unless block is restarted
+    # verify only one 'Starting block' per block number unless block is restarted
     def verifyStartingBlockMessages(self):
         dataDir=Utils.getNodeDataDir(self.nodeId)
         files=Node.findStderrFiles(dataDir)
+        restarting_exhausted_regexp = re.compile(r"Restarting exhausted speculative block #(\d+)")
+        starting_block_regexp       = re.compile(r"Starting block #(\d+) .*(\d\d:\d\d\.\d\d\d) producer")
+
         for f in files:
-            blockNumbers = set()
-            duplicateBlockNumbers = set()
-            threeStartsFound = False
-            lastRestartBlockNum = 0
-            blockNumber = 0
+            notRestartedBlockNumbersAndTimes = {}
+            duplicateStartFound = False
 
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

greg7mdp:
> This should be: "2. or the second `Starting block` is for a different block time than the first."

I don't think so. The full sentence is:

we should have only `Starting block` in the log for any block number **unless** the second `Starting block` is for the exact same block time as the first.

heifner:
> we should have only `Starting block` in the log for any block number **unless** the second `Starting block` is for the exact same block time as the first.

It is actually the opposite of that.

This should be allowed:
```
debug 2024-10-01T13:22:05.393 nodeos    producer_plugin.cpp:2165      start_block          ] Starting block #100 2024-10-01T13:22:06.000 producer defproducerc, deadline 2024-10-01T13:22:05.962
debug 2024-10-01T13:22:06.025 nodeos    producer_plugin.cpp:2165      start_block          ] Starting block #100 2024-10-01T13:22:06.500 producer defproducerc, deadline 2024-10-01T13:22:06.525
```

This should not be allowed:
```
debug 2024-10-01T13:22:05.393 nodeos    producer_plugin.cpp:2165      start_block          ] Starting block #100 2024-10-01T13:22:06.000 producer defproducerc, deadline 2024-10-01T13:22:05.962
debug 2024-10-01T13:22:06.025 nodeos    producer_plugin.cpp:2165      start_block          ] Starting block #100 2024-10-01T13:22:06.000 producer defproducerc, deadline 2024-10-01T13:22:06.525
```

heifner (approved):

tests/TestHarness/Node.py
```diff
@@ -708,37 +708,36 @@ def countInLog(self, searchStr) -> int:
                 count += contents.count(searchStr)
         return count
 
-    # verify only one or two 'Starting block' per block number unless block is restarted
+    # verify only one 'Starting block' per block number unless block is restarted
```
linh2931, line 711: Maybe copy the description in the PR here.

linh2931 (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Improve python test harness to better support distributed transactions tests.
Note:end

PR #909, opened 2024-10-08
Create socket with provided executor

### LINKED ISSUE #819: P2P, HTTP, SHiP: Change fc::create_listener to use strand
Currently `fc::create_listener` creates sockets with the provided executor. For our use cases, it would be nice if `fc::create_listener` would create a strand for each socket. The users of `fc::create_listener` could then retrieve the strand from the `socket.get_executor()` for use.

Something like:
https://github.com/AntelopeIO/spring/commit/438ae6223d7012f75f41b8332a0becda79ef723d

Then:
- Update SHiP to use that strand via `socket.get_executor()` instead of making a new strand.
- Update `net_plugin` to use the socket strand instead of creating a different one.

heifner:
Modify `fc::create_listener` to take a lambda that provides the executor (often a strand) for the socket that is created. This makes sure the socket is associated with the desired executor. 

I don't see how this helps: https://github.com/AntelopeIO/spring/pull/816#discussion_r1774606535 Maybe I'm missing something. If this doesn't help that issue, then seems like this should target `main` and not 1.0.3. <= Changed target to `main`.

Resolves #819


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -101,25 +101,27 @@ struct state_history_plugin_impl {
...
-                                               },
-                                               [this](session_base* conn) {
-                                                  boost::asio::post(app().get_io_service(), [conn, this]() {
-                                                     connections.erase(connections.find(conn));
-                                                  });
-                                               }, _log));
+      fc::create_listener<Protocol>(thread_pool.get_executor(), _log, accept_timeout, address, "",
+         [this](const auto&) { return boost::asio::make_strand(thread_pool.get_executor()); },
+         [this](Protocol::socket&& socket) {
+            boost::asio::post(app().get_io_service(), [this, socket{std::move(socket)}]() mutable {
```
spoonincode, line 107: Just for code conciseness, can we go back to placing the listener on main thread to avoid this post() entirely?

heifner:
> I think it lgtm, but what are compelling reasons to make the change in 1.0? It certainly makes asio usage more idiomatic, which is good, but I don't know if we can point to something as being fixed or defective?

I completely agree. I don't think this helps https://github.com/AntelopeIO/spring/pull/816#discussion_r1774606535 . I asked the same question in the PR description.

spoonincode:
I think https://github.com/AntelopeIO/spring/pull/816#discussion_r1774606535 was resolved with #845

heifner:
> I think [#816 (comment)](https://github.com/AntelopeIO/spring/pull/816#discussion_r1774606535) was resolved with #845

Right. At least until we can get a better fix via: https://github.com/AntelopeIO/spring/issues/842

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1176,8 +1176,8 @@ namespace eosio {
 
    connection::connection( const string& endpoint, const string& listen_address )
       : peer_addr( endpoint ),
-        strand( my_impl->thread_pool.get_executor() ),
-        socket( new tcp::socket( my_impl->thread_pool.get_executor() ) ),
+        strand( boost::asio::make_strand(my_impl->thread_pool.get_executor()) ),
+        socket( new tcp::socket( strand ) ),
```
spoonincode, line 1180: Now that the socket is always being created with a strand (both here and via the listener), that should mean that all the `bind_executor`s to the `strand` can be removed. Not sure you want to tackle that now or not (going in to 1.0 probably not, but going in to main maybe)

heifner: I left it mainly because it was targeting 1.0. At this point, maybe it would be cleaner. In many places you would have `boost::asio::post(socket.get_executor(), [](){})` instead which is not really any difference I don't think.

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -58,30 +58,47 @@ struct listener_base<boost::asio::local::stream_protocol> {
...
    const auto& acceptor() const { return acceptor_; }
 
    void do_accept() {
-      acceptor_.async_accept([self = this->shared_from_this()](boost::system::error_code ec, auto&& peer_socket) {
+      boost::system::error_code ec;
+      auto ep = acceptor_.local_endpoint(ec);
+      if (ec) {
+         fc_wlog(logger_, "Unable to retrieve local_endpoint of acceptor, error: ${e}", ("e", ec.message()));
+      }
+      acceptor_.async_accept(socket_executor_fn_(ep),
```
spoonincode, line 100: What is the purpose for the changes here and for passing the local endpoint as a parameter to the `socket_executor_fn_`? (this local endpoint parameter is not used by any of ship/http/p2p)

heifner: I thought it might be useful. If someone was wanting to choose which executor according to which endpoint. I'm happy to remove it since we don't have a current use-case.

plugins/state_history_plugin/state_history_plugin.cpp
heifner: Actually that doesn't work: https://github.com/AntelopeIO/spring/actions/runs/11238499086
We capture the `this` and we need the listener to shutdown when `thread_pool.stop()` is called in the `plugin_shutdown`. This can be cleaned up when we work: https://github.com/AntelopeIO/spring/issues/842

plugins/net_plugin/net_plugin.cpp
spoonincode: You would have to move the timers to using the strand too (afaict not a problem) but after that it does seem like all 5 `bind_executor`s could be removed. I am not seeing the need for `boost::asio::post(socket.get_executor(), [](){})` in those same spots. Anyways, it's not something that needs to be done now just a clean up I noticed since we're constructing these more idiomaticly now.

plugins/state_history_plugin/state_history_plugin.cpp
spoonincode: > We capture the `this` and we need the listener to shutdown when `thread_pool.stop()` is called in the `plugin_shutdown`.

I'm not immediately seeing why that's a problem. `this` is still alive after `plugin_shutdown()` completes. I think something else we're missing. Maybe the already-created `strand` (that is using ship's thread pool's executor) expects to touch that executor when it's destroyed when `async_accept()` is torn down when the main thread's io_context is destroyed which is after the plugin is destroyed and thus the thread pool's executor.

But yeah can wait to touch this further later.

plugins/net_plugin/net_plugin.cpp
heifner: Oh, you were talking specifically about the `bind_executor`. I was thinking of the `boost::asio::post(strand,`.

spoonincode (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Refactor net library to create socket with provided executor. Improves P2P, HTTP, SHiP.
Note:end

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -58,30 +58,47 @@ struct listener_base<boost::asio::local::stream_protocol> {
...
+///                          ExecutorType(const endpoint_type&)
+/// @tparam CreateSession Callback Function to call after each accepted connection, provides newly created socket
+///                       and the Executor of the newly accepted socket: void(Socket&& socket, ExecutorType& ex)
+///                       Note that the ExecutorType provided by SocketExecutorFn can be retrieved from the socket
+///                       via socket.get_executor().
+///
 /////////////////////////////////////////////////////////////////////////////////////////////
-template <typename Protocol, typename Executor, typename CreateSession>
-struct listener : listener_base<Protocol>, std::enable_shared_from_this<listener<Protocol, Executor, CreateSession>> {
+template <typename Protocol, typename Executor, typename SocketExecutorFn, typename CreateSession>
```
greg7mdp, line 71:
```suggestion
template <typename Protocol, typename Executor, typename SocketExecutorFn, typename CreateSession>
requires (std::is_same_v<Protocol, boost::asio::ip::tcp> || std::is_same_v<Protocol, boost::asio::local::stream_protocol>) &&
         std::invocable<SocketExecutorFn&, typename Protocol::endpoint>
```

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -152,16 +169,26 @@ struct listener : listener_base<Protocol>, std::enable_shared_from_this<listener
...
+/// @tparam Executor The executor for the acceptor and acceptor timer
+/// @tparam SocketExecutorFn Function that returns executor to be used for the newly accepted socket:
+///                          ExecutorType(const endpoint_type&)
+/// @tparam CreateSession Callback Function to call after each accepted connection, provides newly created socket
+///                       and the Executor of the newly accepted socket: void(Socket&& socket, ExecutorType& ex)
+///                       Note that the ExecutorType provided by SocketExecutorFn can be retrieved from the socket
+///                       via socket.get_executor().
 /// @throws std::system_error or boost::system::system_error
-template <typename Protocol, typename Executor, typename CreateSession>
+template <typename Protocol, typename Executor, typename SocketExecutorFn, typename CreateSession>
```
greg7mdp, line 188:
```suggestion
template <typename Protocol, typename Executor, typename SocketExecutorFn, typename CreateSession>
requires (std::is_same_v<Protocol, boost::asio::ip::tcp> || std::is_same_v<Protocol, boost::asio::local::stream_protocol>) &&
         std::invocable<SocketExecutorFn&, typename Protocol::endpoint>
```

greg7mdp (approved):

greg7mdp (approved):

PR #910, opened 2024-10-08
[1.0.3 -> main] Test failure: distributed-transactions-if-test

### LINKED ISSUE #848: Test failure: distributed-transactions-if-test

greg7mdp:
Resolves #848.

This updates the `verifyStartingBlockMessages` check in `Node.py` to check that we have only one `Starting block` in the log for any block number *unless*:

1. the block was restarted because it was exhausted, 
2. or the second `Starting block` is for a different block time than the first.

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -708,37 +708,39 @@ class Node(Transactions):
                 count += contents.count(searchStr)
         return count
 
-    # verify only one or two 'Starting block' per block number unless block is restarted
+    # Verify that we have only one "Starting block" in the log for any block number unless:
+    # - the block was restarted because it was exhausted,
+    # - or the second "Starting block" is for a different block time than the first.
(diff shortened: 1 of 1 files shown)


heifner (approved):

spoonincode (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Improve python test harness to better support distributed transactions tests.
Note:end

PR #912, opened 2024-10-08
P2P: Immediately close on bad block

### LINKED ISSUE #643: Race condition in net_plugin that can lead to unnecessary unlinkable block exception
The net_plugin will send a request to peer asking it to return a particular span of blocks referenced by block number. It will then send another request with a later span of block to next peer (round robin). The net_plugin waits until it receives (which includes successful unpacking of the serialized data) all of the blocks from the latest requested span before sending out the request for the next span of blocks. But it does not validate the integrity of the block data received prior to making the next request.

So if the block data was corrupted some how, e.g. missing a transaction, the net_plugin would realize this after it had already sent out the next request. The corrupted block would of course be dropped, but the block that comes after would likely be received by another peer before the net_plugin has a chance to re-request the block that was corrupted. So that block would run into a unlinkable block exception when attempting to add the block to the fork database since its necessary parent block (the corrupted block) would not yet be in the fork database. The unlinkable block exception causes net_plugin to post that block into the queue for the main thread to process along with the other remaining blocks it receives from that peer for the span of blocks requested.

Those sequence of blocks will all ultimately fail to link because the corrupted block is missing. But after a certain number of failure, net_plugin will close the connection with its "bad" peer, and then after some time, after it reconnects, it will re-request the necessary blocks and resolve the issue automatically. But it gets there after a very inefficient process.

It would be more efficient if the net_plugin was aware of the block corruption and the need to request blocks starting from the block it needs.

One simple way to achieve this is to wait until all the received blocks from one requested span of blocks have been minimally verified to have block integrity before requesting the next span. For example, ensuring the last block of the span has been successfully been added into the fork database prior to requesting the next span is one way of achieving this goal. If it encounters a corrupted block, it can disconnect from that peer immediately (ignoring the remainder of the blocks in the span that the peer is sending), and then send a request for a new span of blocks to the next peer starting with the block number of the corrupted block. This solution can be part of the work of https://github.com/AntelopeIO/spring/issues/529 since that also changes the behavior of net_plugin to never post an unlinkable block to the queue for the main thread.

heifner:
Note almost all the work of #643 was done as part of #529. This PR represents the remaining part of closing immediately if a block is invalid.

Resolves #643

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -217,7 +217,6 @@ namespace eosio {
       void bcast_transaction(const packed_transaction_ptr& trx);
       void rejected_transaction(const packed_transaction_ptr& trx);
       void bcast_block( const signed_block_ptr& b, const block_id_type& id );
-      void rejected_block(const block_id_type& id);
 
       void expire_blocks( uint32_t lib_num );
       void recv_notice(const connection_ptr& conn, const notice_message& msg, bool generated);
@@ -2698,10 +2697,6 @@ namespace eosio {
       } );
    }
 
-   void dispatch_manager::rejected_block(const block_id_type& id) {
-      fc_dlog( logger, "rejected block ${bn} ${id}", ("bn", block_header::num_from_id(id))("id", id) );
-   }
-
    // called from any thread
    void dispatch_manager::bcast_transaction(const packed_transaction_ptr& trx) {
       trx_buffer_factory buff_factory;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3723,7 +3718,7 @@ namespace eosio {
          bool exception = false;
          bool best_head = false;
          bool unlinkable = false;
-         sync_manager::closing_mode close_mode = sync_manager::closing_mode::handshake;
+         sync_manager::closing_mode close_mode = sync_manager::closing_mode::immediately;
          try {
             EOS_ASSERT(ptr->timestamp < (fc::time_point::now() + fc::seconds(7)), block_from_the_future,
                        "received a block from the future, ignoring it: ${id}", ("id", id));
```
linh2931, line 3724: We will reject the block if this EOS_ASSERT happens. Should `ignoring` be changed to `rejecting`?

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: With the exception of unlinkable blocks, disconnect immediately instead of allowing for 13 invalid blocks before disconnect.
Note:end

PR #913, opened 2024-10-09
[1.0.3] Optimize handling of invalid vote message

heifner:
Small optimization of handling of errors for vote messages. Shortcut out of loop when connection is found.
Improve error message for invalid vote message.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -4024,7 +4024,7 @@ namespace eosio {
       case vote_result_t::unknown_public_key:
       case vote_result_t::invalid_signature:
       case vote_result_t::max_exceeded:  // close peer immediately
-         fc_elog(vote_logger, "Exceeded max votes per connection for ${c}", ("c", connection_id));
+         fc_elog(vote_logger, "Invalid vote(s), closing connection - ${c}", ("c", connection_id));
          my_impl->connections.any_of_connections([connection_id](const connection_ptr& c) {
             if (c->connection_id == connection_id) {
                c->close( false );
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -4025,19 +4025,25 @@ namespace eosio {
       case vote_result_t::invalid_signature:
       case vote_result_t::max_exceeded:  // close peer immediately
          fc_elog(vote_logger, "Exceeded max votes per connection for ${c}", ("c", connection_id));
```
linh2931, line 4027: Not from your PR. But this logging only lists `Exceeded max votes` as the failure reason. It can be unknown_public_key or invalid_signature

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
heifner: Changed it.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4024,20 +4024,26 @@ namespace eosio {
       case vote_result_t::unknown_public_key:
       case vote_result_t::invalid_signature:
       case vote_result_t::max_exceeded:  // close peer immediately
-         fc_elog(vote_logger, "Exceeded max votes per connection for ${c}", ("c", connection_id));
-         my_impl->connections.for_each_connection([connection_id](const connection_ptr& c) {
+         fc_elog(vote_logger, "Invalid vote(s), closing connection - ${c}", ("c", connection_id));
+         my_impl->connections.any_of_connections([connection_id](const connection_ptr& c) {
```
greg7mdp, line 4028: Maybe define a new API in connections_manager:

```
      template <typename F>
      void for_connection(uint32_t connection_id, F&& f) {
         any_of_connections([connection_id, f = std::forward<F>(f)](const connection_ptr& c) {
            if (c->connection_id == connection_id) {
               f();
               return true;
            }
            return false;
         });
```

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Improve error handling of vote messages, improve log message and shortcut out of loop when connection is found.
Note:end

PR #918, opened 2024-10-09
Normalize usage of LIB

### LINKED ISSUE #806: Normalize LIB nomenclature
There are a number of different things we call `lib` or `last irreversible block`.
- Savanna block's `core.last_final_block_num` also known as `irreversible_blocknum`
- Legacy block's `dpos_irreversible_blocknum` also known as `irreversible_blocknum`
- fork database `root`
- fork database `pending_savanna_lib_id`
- block log `head` (might not exist if running with no block log)

Going forward we would like to use `lib` or `last irreversible block` in logs and documentation to refer to `irreversible_blocknum` which is the lib in respect to a given block. In locations where fork database root or fork database pending savanna lib is required they should be referred to `fork_db_root`/`froot` and `fork_db_pending_lib`. The block log head should be referred to simply as the block log head. 

This issue is to examine and cleanup any log messages, method names, and comments in the code base to correspond to this new nomenclature.

heifner:
Most of the changes involve using `froot` instead of `lib`.

Did not change the interface of `tester` to avoid uses having to update their tests.
`get_info` was unchanged; it still reports `last_irreversible_block_num` and `last_irreversible_block_id` for `forkdb_root_block_num` and `forkdb_root_block_id`.

Resolves #806

### DIFF
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -230,7 +230,7 @@ namespace eosio::chain {
          void set_async_voting(async_t val);
          void set_async_aggregation(async_t val);
 
-         /// Apply any blocks that are ready from the forkdb
+         /// Apply any blocks that are ready from the fork_db
          void apply_blocks(const forked_callback_t& cb, const trx_meta_cache_lookup& trx_lookup);
 
          struct accepted_block_result {
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -329,15 +329,11 @@ namespace eosio::chain {
          void set_savanna_lib_id(const block_id_type& id);
 
          // thread-safe
-         bool   fork_db_has_root() const;
+         bool         fork_db_has_root() const;
          // thread-safe
-         size_t fork_db_size() const;
-         // thread-safe, only valid if fork_db_has_root()
-         uint32_t fork_db_root_block_num() const;
-         // thread-safe, only valid if fork_db_has_root()
-         block_id_type fork_db_root_block_id() const;
-         // thread-safe, only valid if fork_db_has_root()
-         time_point fork_db_root_block_time() const;
+         block_handle fork_db_root()const;
+         // thread-safe
+         size_t       fork_db_size() const;
 
          // thread-safe, retrieves block according to fork db best branch which can change at any moment
          signed_block_ptr fetch_block_by_number( uint32_t block_num )const;
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -328,15 +328,16 @@ namespace eosio::chain {
 
          void set_savanna_lib_id(const block_id_type& id);
 
+         // thread-safe
          bool   fork_db_has_root() const;
+         // thread-safe
          size_t fork_db_size() const;
-
-         // thread-safe, applied LIB, fork db root
-         uint32_t last_irreversible_block_num() const;
-         // thread-safe, applied LIB, fork db root
-         block_id_type last_irreversible_block_id() const;
-         // thread-safe, applied LIB, fork db root
-         time_point last_irreversible_block_time() const;
+         // thread-safe, only valid if fork_db_has_root()
+         uint32_t fork_db_root_block_num() const;
+         // thread-safe, only valid if fork_db_has_root()
+         block_id_type fork_db_root_block_id() const;
+         // thread-safe, only valid if fork_db_has_root()
+         time_point fork_db_root_block_time() const;
 
          // thread-safe, retrieves block according to fork db best branch which can change at any moment
          signed_block_ptr fetch_block_by_number( uint32_t block_num )const;
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1286,17 +1286,17 @@ read_only::get_info_results read_only::get_info(const read_only::get_info_params
 
    auto head = db.head();
    auto head_id = head.id();
-   auto fork_root = db.fork_db_root();
-   auto fork_head = db.fork_db_head();
-   auto fork_root_id = fork_root.id();
-   auto fork_head_id = fork_head.id();
+   auto fork_db_root = db.fork_db_root();
+   auto fork_db_head = db.fork_db_head();
+   auto fork_db_root_id = fork_db_root.id();
+   auto fork_db_head_id = fork_db_head.id();
 
    return {
       itoh(static_cast<uint32_t>(app().version())),
       db.get_chain_id(),
       block_header::num_from_id(head_id),
-      block_header::num_from_id(fork_root_id),
-      fork_root_id,
+      block_header::num_from_id(fork_db_root_id),
+      fork_db_root_id,
       head_id,
       db.head().block_time(),
       db.head().producer(),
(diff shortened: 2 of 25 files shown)


libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -328,15 +328,16 @@ namespace eosio::chain {
...
-         // thread-safe, applied LIB, fork db root
-         block_id_type last_irreversible_block_id() const;
-         // thread-safe, applied LIB, fork db root
-         time_point last_irreversible_block_time() const;
+         // thread-safe, only valid if fork_db_has_root()
+         uint32_t fork_db_root_block_num() const;
+         // thread-safe, only valid if fork_db_has_root()
+         block_id_type fork_db_root_block_id() const;
+         // thread-safe, only valid if fork_db_has_root()
+         time_point fork_db_root_block_time() const;
```
greg7mdp, line 340: I think a better API would be to return a `block_handle` here, even if it involves a `block_ptr` copy.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2233,10 +2233,10 @@ namespace eosio {
 
       auto chain_info = my_impl->get_chain_info();
 
-      sync_reset_lib_num(c, false);
+      sync_reset_froot_num(c, false);
 
       //--------------------------------
-      // sync need checks; (lib == last irreversible block)
+      // sync need checks; (lib == fork database root)
```
linh2931, line 2239: Does this `lib` be changed?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2233,10 +2233,10 @@ namespace eosio {
...
       //
       // 0. my head block id == peer head id means we are all caught up block wise
       // 1. my head block num < peer lib - start sync locally
```
linh2931, line 2242: This comment does not match what the code is doing below.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -420,8 +420,8 @@ namespace eosio {
 
 
       struct chain_info_t {
-         block_id_type lib_id;
-         uint32_t      lib_num = 0;
+         block_id_type fork_root_id;
+         uint32_t      fork_root_num = 0;
```
greg7mdp, line 424: why not `froot`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -439,9 +439,9 @@ namespace eosio {
 
    public:
       void update_chain_info();
-      void update_chain_info(const block_id_type& lib);
+      void update_chain_info(const block_id_type& froot_id);
       chain_info_t get_chain_info() const;
-      uint32_t get_chain_lib_num() const;
+      uint32_t get_fork_root_num() const;
```
greg7mdp, line 444: `froot` here to I think for consistency.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1942,49 +1942,49 @@ namespace eosio {
...
+            sync_known_froot_num = c->peer_froot_num;
          }
       } else {
-         // Closing connection, therefore its view of LIB can no longer be considered as we will no longer be connected.
-         // Determine current LIB of remaining peers as our sync_known_lib_num.
-         uint32_t highest_lib_num = 0;
-         my_impl->connections.for_each_block_connection( [&highest_lib_num]( const auto& cc ) {
+         // Closing connection, therefore its view of froot can no longer be considered as we will no longer be connected.
+         // Determine current froot of remaining peers as our sync_known_froot_num.
+         uint32_t highest_fork_root_num = 0;
```
greg7mdp, line 1958: `froot`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2049,9 +2049,9 @@ namespace eosio {
 
       auto reset_on_failure = [&]() REQUIRES(sync_mtx) {
          sync_source.reset();
-         sync_known_lib_num = chain_info.lib_num;
+         sync_known_froot_num = chain_info.fork_root_num;
```
greg7mdp, line 2052: `froot` here and all other `fork_root` locations.

libraries/chain/include/eosio/chain/controller.hpp
heifner: A `signed_block` does not provide an efficient `.id()` call.

plugins/net_plugin/net_plugin.cpp
heifner: I was matching the existing `fork_head_` members.

plugins/net_plugin/net_plugin.cpp
heifner: Hmm, we have a `get_fork_head_num`. Seems better to spell it out.

plugins/net_plugin/net_plugin.cpp
heifner: Matches the handshake `fork_root_num`.

plugins/net_plugin/net_plugin.cpp
heifner: I think I would rather go the other direction. Change `froot` to `fork_root` except in logging.

linh2931 (approved):

libraries/chain/include/eosio/chain/controller.hpp
greg7mdp: my bad, I misspoke, the `block_handle` contains the `block_state_ptr` not `signed_block_ptr`.

heifner: Oh right. Sure might as well match `fork_db_head()`

plugins/net_plugin/net_plugin.cpp
greg7mdp: whatever we choose let's be consistent throughout in the code.

linh2931: I prefer `fork_root` as it spells out what it really is; it is not long. The `f` in `froot` can be anything, it can even be be thought of `finality`.

libraries/chain/include/eosio/chain/controller.hpp
heifner: Done.

plugins/net_plugin/net_plugin.cpp
heifner: Renamed `froot` to `fork_root`.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1290,16 +1284,19 @@ const string read_only::KEYi64 = "i64";
 read_only::get_info_results read_only::get_info(const read_only::get_info_params&, const fc::time_point&) const {
    const auto& rm = db.get_resource_limits_manager();
 
-   auto head_id = db.head().id();
-   auto lib_id = db.last_irreversible_block_id();
-   auto fhead_id = db.fork_db_head().id();
+   auto head = db.head();
+   auto head_id = head.id();
+   auto fork_root = db.fork_db_root();
+   auto fork_head = db.fork_db_head();
```
greg7mdp, line 1290: Now this seems inconsistent. Why do we have both `fork_root` and `fork_db_root` at different spots in the code?

greg7mdp: I don't really like `fork_root` because it sounds like is the the `root` of a `fork`, not of `fork_db`. If we are going to spell it out, I think it should be `fork_db_root` everywhere, although I did prefer the shorter `froot` :-).

linh2931:
Another thing I noticed we have `fork_db` and `forkdb`. Probably we should use only one of them (at least easier for search)

linh2931:
> Another thing I noticed we have `fork_db` and `forkdb`. Probably we should use only one of them (at least easier for search)

Maybe another PR to clean up  all of those forkdb, fork_db usages, froot, ...

linh2931 (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Normalize usage of LIB.
Note:end

Issue #527, opened 2024-08-13
get_info off the main thread


heifner:
Currently `/v1/chain/get_info` processing is posted to the main thread. `get_info` could be moved off the main thread by copying the current get_info data into the `get_info_results` struct on every `accept_block` signal. Then `get_info` could be handled completely off the main thread.

See https://github.com/AntelopeIO/leap/issues/1212 for original issue which includes EOS Nation API usage graphs.

linh2931:
`get_info` is pushed to the read-only queue. In an API node with `--read-only-thread` configured, `get_info` will be serviced by read-only threads.

Do we want to have a dedicated thread for `get_info`?

If we cache `get_info_results` on every accept_block signal, will it be a waste if no `get_info` is called?

heifner:
I don't think we want a dedicated thread for `get_info`, just the http thread pool.
If `get_info` is not called then yes it is a waste to update the info on every `accept_block`. However, this seems like a good trade-off. We can disable this logic if `chain_api_plugin` is not enabled.

linh2931:
Thanks.

Why do the existing read-only threads not help here?

greg7mdp:
> it is a waste to update the info on every accept_block

I doubt that the cost of doing this is even measurable. I wouldn't worry about it.

Issue #830, opened 2024-09-27
P2P: connection monitor running more often than configured


heifner:
With a `connection-cleanup-period = 30` which is the default:

```
info  2024-09-27T13:53:11.169 net-3     net_plugin.cpp:4899           connection_monitor   ] p2p client connections: 0/4, peer connections: 10/10, block producer peers: 0
warn  2024-09-27T13:53:11.224 net-1     net_plugin.cpp:4711           operator()           ] Unable to resolve p2p.bitmars.one:8080 Host not found (authoritative)
info  2024-09-27T13:53:11.318 net-2     net_plugin.cpp:2862           operator()           ] connection failed to peer.eosn.io:9876, Connection refused
info  2024-09-27T13:53:11.445 net-3     net_plugin.cpp:1596           operator()           ] ["p2p.mainnet.eosrio.io:9876 - 259659b" - 6 187.16.79.235:9876] Sending handshake generation 1, lib 396158153, fhead 396158354, id f7b41de76d6e8ef5
warn  2024-09-27T13:53:11.587 net-3     net_plugin.cpp:3608           handle_message       ] ["p2p.mainnet.eosrio.io:9876 - 259659b" - 6 187.16.79.235:9876] Clock offset is 3495269us, calculation: (rec 1727445195011927000 - org 1727445191445814595 + xmt 1727445195011928000 - dst 1727445191587501533)/2
info  2024-09-27T13:53:11.587 net-3     net_plugin.cpp:1596           operator()           ] ["p2p.mainnet.eosrio.io:9876 - 259659b" - 6 187.16.79.235:9876] Sending handshake generation 2, lib 396158153, fhead 396158354, id f7b41de76d6e8ef5
info  2024-09-27T13:53:11.588 net-1     net_plugin.cpp:3465           handle_message       ] ["p2p.mainnet.eosrio.io:9876 - 259659b" - 6 187.16.79.235:9876] Local network version: 9
info  2024-09-27T13:53:11.588 net-1     net_plugin.cpp:2331           recv_handshake       ] ["p2p.mainnet.eosrio.io:9876 - 259659b" - 6 187.16.79.235:9876] handshake lib 396440434, fhead 396440436, id fb0598a488a18012.. sync 1, fhead 396158354, lib 396158153
[...]

PR #920, opened 2024-10-10
Move get_info off main thread and run in http thread

linh2931:
`/v1/chain/get_info` is a top 1 or 2 used RPC endpoint (https://github.com/AntelopeIO/leap/issues/1212). It is currently processed on the `main` thread. Moving it off the main thread will reduce load on the `main` thread, and will shorten response time when the `main` thread is stuck (https://github.com/AntelopeIO/spring/issues/830).

This PR moves `get_info` off the `main` thread and runs it on `http` thread.  The data required by get_info is cached on `accept_block` signal  and is used to service `get_info` when needed.

Notably, the implementation is mutex free such that the `main` thread is never blocked by `get_info`.

Resolves https://github.com/AntelopeIO/spring/issues/527

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1078,6 +1078,10 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
             _trx_finality_status_processing->signal_irreversible_block(block, id);
          }
 
+         if (_get_info_db) {
+            _get_info_db->on_irreversible_block();
+         }
+
          irreversible_block_channel.publish( priority::low, t );
       } );
       
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1079,7 +1079,7 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
          }
 
          if (_get_info_db) {
-            _get_info_db->on_irreversible_block();
+            _get_info_db->on_irreversible_block(block, id);
          }
 
          irreversible_block_channel.publish( priority::low, t );
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -193,6 +193,7 @@ public:
    std::optional<scoped_connection>                                   applied_transaction_connection;
    std::optional<scoped_connection>                                   block_start_connection;
 
+   std::optional<chain_apis::get_info_db>                             _get_info_db;
    std::optional<chain_apis::account_query_db>                        _account_query_db;
    std::optional<chain_apis::trx_retry_db>                            _trx_retry_db;
    chain_apis::trx_finality_status_processing_ptr                     _trx_finality_status_processing;
@@ -976,12 +977,14 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
       }
 
       // only enable last tracked votes if chain_api_plugin enabled, if no http endpoint, no reason to track
-      bool last_tracked_votes_enabled = false;
+      bool chain_api_plugin_configured = false;
       if (options.count("plugin")) {
          const auto& v = options.at("plugin").as<std::vector<std::string>>();
-         last_tracked_votes_enabled = std::ranges::any_of(v, [](const std::string& p) { return p.find("eosio::chain_api_plugin") != std::string::npos; });
+         chain_api_plugin_configured = std::ranges::any_of(v, [](const std::string& p) { return p.find("eosio::chain_api_plugin") != std::string::npos; });
       }
-      _last_tracked_votes.emplace(*chain, last_tracked_votes_enabled);
+      _last_tracked_votes.emplace(*chain, chain_api_plugin_configured);
+
+      _get_info_db.emplace(*chain, chain_api_plugin_configured);
 
       // initialize deep mind logging
       if ( options.at( "deep-mind" ).as<bool>() ) {
@@ -1055,6 +1058,10 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
 
          if (_last_tracked_votes) {
             _last_tracked_votes->on_accepted_block(block, id);
+         }
+
+         if (_get_info_db) {
+            _get_info_db->on_accepted_block();
          }
 
          accepted_block_channel.publish( priority::high, t );
@@ -1202,9 +1209,8 @@ chain_apis::read_write chain_plugin::get_read_write_api(const fc::microseconds& 
 }
 
 chain_apis::read_only chain_plugin::get_read_only_api(const fc::microseconds& http_max_response_time) const {
-   return chain_apis::read_only(chain(), my->_account_query_db, my->_last_tracked_votes, get_abi_serializer_max_time(), http_max_response_time, my->_trx_finality_status_processing.get());
+   return chain_apis::read_only(chain(), my->_get_info_db, my->_account_query_db, my->_last_tracked_votes, get_abi_serializer_max_time(), http_max_response_time, my->_trx_finality_status_processing.get());
 }
-
 
 void chain_plugin::accept_transaction(const chain::packed_transaction_ptr& trx, next_function<chain::transaction_trace_ptr> next) {
    my->incoming_transaction_async_method(trx, false, transaction_metadata::trx_type::input, false, std::move(next));
@@ -1287,37 +1293,12 @@ namespace chain_apis {
 
 const string read_only::KEYi64 = "i64";
 
-read_only::get_info_results read_only::get_info(const read_only::get_info_params&, const fc::time_point&) const {
-   const auto& rm = db.get_resource_limits_manager();
+get_info_db::get_info_results read_only::get_info(const read_only::get_info_params&, const fc::time_point&) const {
+   EOS_ASSERT(gidb, plugin_config_exception, "get_info being accessed when not enabled");
 
-   auto head_id = db.head().id();
-   auto lib_id = db.last_irreversible_block_id();
-   auto fhead_id = db.fork_db_head().id();
-
-   return {
-      itoh(static_cast<uint32_t>(app().version())),
-      db.get_chain_id(),
-      block_header::num_from_id(head_id),
-      block_header::num_from_id(lib_id),
-      lib_id,
-      head_id,
-      db.head().block_time(),
-      db.head().producer(),
-      rm.get_virtual_block_cpu_limit(),
-      rm.get_virtual_block_net_limit(),
-      rm.get_block_cpu_limit(),
-      rm.get_block_net_limit(),
-      //std::bitset<64>(db.get_dynamic_global_properties().recent_slots_filled).to_string(),
-      //__builtin_popcountll(db.get_dynamic_global_properties().recent_slots_filled) / 64.0,
-      app().version_string(),
-      block_header::num_from_id(fhead_id),
-      fhead_id,
-      app().full_version_string(),
-      rm.get_total_cpu_weight(),
-      rm.get_total_net_weight(),
-      db.earliest_available_block_num(),
-      db.last_irreversible_block_time()
-   };
+   // To be able to run get_info on an http thread, get_info results are stored
+   // in get_info_db and updated whenever accepted_block signal is received.
+   return gidb->get_info();
 }
 
 read_only::get_transaction_status_results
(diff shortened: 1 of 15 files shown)


plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,89 @@
...
+            if (!get_info_enabled) {
+               return;
+            }
+
+            std::unique_lock write_lock(rw_mutex);
+
+            const auto& rm       = controller.get_resource_limits_manager();
+            auto        head_id  = controller.head().id();
+            auto        lib_id   = controller.last_irreversible_block_id();
+            auto        fhead_id = controller.fork_db_head().id();
```
heifner, line 41: `fork_db` is not available during replay. `on_accepted_block` can be called during replay. 
Should probably check `controller.fork_db_has_root()` and default forkdb values when forkdb not available.

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,89 @@
...
+   struct get_info_db_impl {
+      get_info_db_impl(const chain::controller& controller, bool get_info_enabled)
+         : controller(controller)
+         , get_info_enabled(get_info_enabled) {}
+
+      // A handle to the controller.
+      const chain::controller& controller;
+
+      // Indication whether get_info RPC is enabled.
+      bool get_info_enabled = false;
```
heifner, line 23: This can be `const` and should be since accessed from multiple threads.

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+      // Cache to store the current get_info results in respect to latest accepted block.
+      get_info_db::get_info_results  cached_results;
+      mutable std::shared_mutex      rw_mutex;
```
heifner, line 27: Style: I think having the mutex above what it protects is clearer.

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,89 @@
...
+            std::unique_lock write_lock(rw_mutex);
+
+            const auto& rm       = controller.get_resource_limits_manager();
+            auto        head_id  = controller.head().id();
+            auto        lib_id   = controller.last_irreversible_block_id();
+            auto        fhead_id = controller.fork_db_head().id();
+
+            // cache get_info results
+            cached_results = get_info_db::get_info_results {
+               itoh(static_cast<uint32_t>(app().version())),
```
heifner, line 45: I doubt it makes much difference, but some of these values never change and could be only set on the first call.

plugins/chain_plugin/include/eosio/chain_plugin/get_info_db.hpp
```diff
@@ -0,0 +1,80 @@
...
+   }; // get_info_db
+
+   template<typename I>
+   std::string itoh(I n, size_t hlen = sizeof(I)<<1) {
+       static const char* digits = "0123456789abcdef";
+       std::string r(hlen, '0');
+       for(size_t i = 0, j = (hlen - 1) * 4 ; i < hlen; ++i, j -= 4)
+           r[i] = digits[(n>>j) & 0x0f];
+       return r;
+   }
```
heifner, line 72: Consider moving to `fc/crypto/hex.hpp`

heifner (changes requested):

plugins/chain_plugin/get_info_db.cpp
linh2931: Thanks.

greg7mdp:
I think it could be implemented in a simpler way. 

Instead of having the chain_plugin query the controller for all the individual members, why not have the controller allocate a new struct of `get_info_results` at each `accepted_block` (using std::make_shared) and, when populated, store the result in controller in a `std::atomic<std::shared_ptr>`.

Then the `chain_plugin` can simply get this pointer and use it at each get_info (no mutex necessary).

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,109 @@
...
+      const std::string   full_version_string;
+
+      // Called on accepted_block signal.
+      void on_accepted_block() {
+         try {
+            if (!get_info_enabled) {
+               return;
+            }
+
+            std::unique_lock write_lock(rw_mutex);
```
heifner, line 47: This can be moved down right before the `cached_results = `

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,109 @@
...
+               rm.get_virtual_block_net_limit(),
+               rm.get_block_cpu_limit(),
+               rm.get_block_net_limit(),
+               version_string,
+               chain::block_header::num_from_id(fhead_id),
+               fhead_id,
+               full_version_string,
+               rm.get_total_cpu_weight(),
+               rm.get_total_net_weight(),
+               controller.earliest_available_block_num(),
```
heifner, line 83: If we keep the mutex, this calls should be moved above the mutex.

linh2931:
Will work on mutex locations tomorrow. Hopefully now all tests can pass without modifying them.

linh2931:
> I think it could be implemented in a simpler way.
> 
> Instead of having the chain_plugin query the controller for all the individual members, why not have the controller allocate a new struct of `get_info_results` at each `accepted_block` (using std::make_shared) and, when populated, store the result in controller in a `std::atomic<std::shared_ptr>`.
> 
> Then the `chain_plugin` can simply get this pointer and use it at each get_info (no mutex necessary).

Thanks. New implementation makes it mutex free while doing it in chain_plugin

plugins/chain_plugin/get_info_db.cpp
linh2931: Thanks. New implementation does not use mutex any more.

plugins/chain_plugin/get_info_db.cpp
linh2931: Mutex is no longer used.

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,159 @@
...
+      void store_info(const chain::signed_block_ptr& block, const block_id_type& id) {
+         std::shared_ptr<get_info_db::get_info_results> info = std::make_shared<get_info_db::get_info_results>();
+
+         store_info_common(info);
+
+         info->last_irreversible_block_id   = id;
+         info->last_irreversible_block_num  = block_header::num_from_id(info->last_irreversible_block_id);
+         info->last_irreversible_block_time = block->timestamp;
+
+         std::atomic_store(&info_cache, info);  // replace current cache safely
```
greg7mdp, line 139: This is not lock free. Most likely worse than using `std::atomic<std::shared_ptr>`.

greg7mdp:
> while doing it in chain_plugin

Why? I think it is better to do it in controller, as it makes the information available to whomever needs it (could be prometheus for example), not just the chain plugin.

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,159 @@
+#include <eosio/chain_plugin/get_info_db.hpp>
+#include <eosio/chain/resource_limits.hpp>
+#include <eosio/chain/application.hpp>
+
+#include <shared_mutex>
```
heifner, line 5: Not needed.

plugins/chain_plugin/get_info_db.cpp
heifner: Should be similar, I think. It is deprecated, but since we don't have `std::atomic<std::shared_ptr>` probably the best. Certainly seems better to me than the boost atomic shared ptr with its spin lock.
https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,159 @@
...
+         info->total_cpu_weight        = rm.get_total_cpu_weight();
+         info->total_net_weight        = rm.get_total_net_weight();
+      }
+
+      void store_info() {
+         std::shared_ptr<get_info_db::get_info_results> info = std::make_shared<get_info_db::get_info_results>();
+
+         store_info_common(info);
+
+         if (controller.fork_db_has_root()) {
```
heifner, line 121: Grab the `root` once like with `controller.head()` to avoid multiple mutex locks in fork db.

plugins/chain_plugin/include/eosio/chain_plugin/get_info_db.hpp
```diff
@@ -0,0 +1,79 @@
...
+         std::optional<uint32_t>              fork_db_head_block_num;
+         std::optional<chain::block_id_type>  fork_db_head_block_id;
+         std::optional<std::string>           server_full_version_string;
+         std::optional<uint64_t>              total_cpu_weight;
+         std::optional<uint64_t>              total_net_weight;
+         std::optional<uint32_t>              earliest_available_block_num;
+         std::optional<fc::time_point>        last_irreversible_block_time;
+
+         // Returns true if the struct contains full data for using.
+         bool contains_full_data() {
```
heifner, line 55: Can be `const`

plugins/chain_plugin/get_info_db.cpp
```diff
@@ -0,0 +1,159 @@
...
+         if (head.is_valid()) {
+            info->head_block_id       = head.id();
+            info->head_block_num      = block_header::num_from_id(info->head_block_id);
+            info->head_block_time     = head.block_time();
+            info->head_block_producer = head.producer();
+         }
+
+         // fork_db part
+         if (controller.fork_db_has_root()) {
+            info->fork_db_head_block_id        = controller.fork_db_head().id();
```
heifner, line 101: You can use `fork_db_head()` and then check if valid to avoid an extra mutex lock in fork db.

heifner (changes requested):

plugins/chain_plugin/get_info_db.cpp
linh2931: I agree. It is simpler and is locked shorter than mutex.

Will add an issue to remind us to change to atomic shared_ptr once it is supported by compilers.

linh2931:
> information

I think it is better separate concern doing it in chain_plugin; let controller just emit signals and its users decide what to do with the signals. All the data required by get_info is available with currently controller APIs. There is no need to add another one.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -975,12 +976,14 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
       }
 
       // only enable last tracked votes if chain_api_plugin enabled, if no http endpoint, no reason to track
-      bool last_tracked_votes_enabled = false;
+      bool chain_api_plugin_configured = false;
       if (options.count("plugin")) {
          const auto& v = options.at("plugin").as<std::vector<std::string>>();
-         last_tracked_votes_enabled = std::ranges::any_of(v, [](const std::string& p) { return p.find("eosio::chain_api_plugin") != std::string::npos; });
+         chain_api_plugin_configured = std::ranges::any_of(v, [](const std::string& p) { return p.find("eosio::chain_api_plugin") != std::string::npos; });
```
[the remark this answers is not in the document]
linh2931: It is most likely configured in BP nodes but not necessarily in other nodes.

plugins/chain_plugin/get_info_db.cpp
linh2931: `earliest_available_block_num()` requires `fork_db_root_block_num()` which in turn requires fork_db_has_root. Checking if the returned head of fork_db_head() is valid does not reduce the total number of locks in this code section.

heifner: `fork_db_head()` returning a valid `block_handle` means `fork_db_root` is available.

plugins/chain_plugin/get_info_db.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/920/commits/ee913fa8b0157c1dde7432c0356267fbc926abf1

plugins/chain_plugin/get_info_db.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/920/commits/20e31401e6cf5664f6e55df94d212768202918ae

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Move `get_info` off main thread and run in http thread.
Note:end

PR #921, opened 2024-10-10
[1.0.3] Limit apply blocks

### LINKED ISSUE #284: Limit apply blocks time in controller

heifner:

Resolves #284

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4472,7 +4472,7 @@ struct controller_impl {
                   }
                   // Break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run. There is a post
                   // for every incoming blocks; enough posted tasks to apply all blocks queued to the fork db.
-                  if (!replaying && fc::time_point::now() - start > fc::microseconds(500))
+                  if (!replaying && fc::time_point::now() - start > fc::milliseconds(500))
                      break;
 
                } catch ( const std::bad_alloc& ) {
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1689,8 +1689,6 @@ struct controller_impl {
    }
 
    void replay(startup_t startup) {
-      replaying = true;
-
       bool replay_block_log_needed = should_replay_block_log();
 
       auto blog_head = blog.head();
@@ -1837,8 +1835,6 @@ struct controller_impl {
 
       };
       fork_db.apply<void>(replay_fork_db);
-
-      replaying = false;
 
       if( except_ptr ) {
          std::rethrow_exception( except_ptr );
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -1988,6 +1984,7 @@ struct controller_impl {
          ilog( "chain database started with hash: ${hash}", ("hash", calculate_integrity_hash()) );
       okay_to_print_integrity_hash_on_stop = true;
 
+      replaying = true;
       replay( startup ); // replay any irreversible and reversible blocks ahead of current head
 
       if( check_shutdown() ) return;
```
linh2931, line 1990: Probably no need to set it back to false as node is going to shut down.

libraries/chain/controller.cpp
```diff
@@ -2034,6 +2031,8 @@ struct controller_impl {
 
       fork_db.apply<void>(finish_init);
 
+      replaying = false;
```
linh2931, line 2034: Should `replaying` be scoped so it will always be reset to `false` even exceptions happen above?

libraries/chain/controller.cpp
```diff
@@ -3630,8 +3629,6 @@ struct controller_impl {
 
    template<typename BSP>
    void log_applied(controller::block_report& br, const BSP& bsp) const {
-      if (replaying) // fork_db_root_block_num not available during replay
```
linh2931, line 3633: Why is this removed?

libraries/chain/controller.cpp
```diff
@@ -4472,6 +4470,11 @@ struct controller_impl {
                      shutdown();
                      break;
                   }
+                  // Break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run. There is a post
+                  // for every incoming blocks; enough posted tasks to apply all blocks queued to the fork db.
+                  if (!replaying && fc::time_point::now() - start > fc::milliseconds(500))
+                     break;
```
linh2931, line 4476: Should break or return? Not sure some of the code below still applies, like the ilog for `"successfully switched fork...`.

libraries/chain/controller.cpp
heifner: Right, any exception or if we shutdown there is no need to reset it. Although, maybe better to be pedantic and and put it in a scope exit to reset.

libraries/chain/controller.cpp
heifner: Because it is not needed anymore. It was there before because we were logging fork_db_root_block_num and fork_db is not available during replay. However, we are no longer logging fork_db_root_block_num, so it is not needed. Also I noticed that we were not logging every 1000 blocks during replay. This restores the behavior of logging every 1000 blocks during replay.

libraries/chain/controller.cpp
heifner: Thanks. This should check `!switch_fork`. I think better to `break`. Currently nothing below will execute, but logically I think it makes more sense to `break`.

greg7mdp (approved):

greg7mdp (approved):

linh2931 (approved):
Because this goes to release 1.0, it might be better to have a test to break in the middle of the loop and resume.

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Limit Apply Blocks.
Note:end

PR #922, opened 2024-10-10
Limit apply blocks

### LINKED ISSUE #284: Limit apply blocks time in controller

heifner:

Note this takes a different approach for `main` than #921 does for `release/1.0` because `main` has the optimization of not posting a `handler_id::process_incoming_block` if one already exists.

Resolves #284

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1701,8 +1701,6 @@ struct controller_impl {
    }
 
    void replay(startup_t startup) {
-      replaying = true;
-
       bool replay_block_log_needed = should_replay_block_log();
 
       auto blog_head = blog.head();
@@ -1839,8 +1837,6 @@ struct controller_impl {
          }
       };
       fork_db.apply<void>(replay_fork_db);
-
-      replaying = false;
 
       if( except_ptr ) {
          std::rethrow_exception( except_ptr );
@@ -1990,6 +1986,7 @@ struct controller_impl {
          ilog( "chain database started with hash: ${hash}", ("hash", calculate_integrity_hash()) );
       okay_to_print_integrity_hash_on_stop = true;
 
+      fc::scoped_set_value r(replaying, true);
       replay( startup ); // replay any irreversible and reversible blocks ahead of current head
 
       if( check_shutdown() ) return;
@@ -4354,6 +4351,7 @@ struct controller_impl {
             }
          }
 
+         auto start = fc::time_point::now();
          for( auto ritr = new_head_branch.rbegin(); ritr != new_head_branch.rend(); ++ritr ) {
             auto except = std::exception_ptr{};
             const auto& bsp = *ritr;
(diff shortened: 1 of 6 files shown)


libraries/chain/controller.cpp
```diff
@@ -4356,15 +4358,24 @@ struct controller_impl {
...
-                  shutdown();
-                  break;
+               if (!switch_fork) { // always complete a switch fork
+                  if (!applied || check_shutdown()) {
+                     shutdown();
+                     break; // result should be complete since we are shutting down
+                  }
+                  // Break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run. User expected
+                  // to call apply_blocks again if this returns incomplete.
+                  if (!replaying && fc::time_point::now() - start > fc::milliseconds(500)) {
```
greg7mdp, line 4375: do we need this `!replaying` check since the callers of `maybe_apply_blocks` now call it in a while loop?

libraries/chain/controller.cpp
```diff
@@ -2013,7 +2011,9 @@ struct controller_impl {
                // See comment below about pause-at-block for why `|| conf.num_configured_p2p_peers > 0`
                if (chain_head_is_root || conf.num_configured_p2p_peers > 0) {
                   ilog("applying branch from fork database ending with block: ${id}", ("id", pending_head->id()));
-                  maybe_apply_blocks(forked_callback_t{}, trx_meta_cache_lookup{});
+                  // applies all blocks up to forkdb head from forkdb, shouldn't return incomplete, but if it does loop until complete
+                  while (maybe_apply_blocks(forked_callback_t{}, trx_meta_cache_lookup{}) == controller::apply_blocks_result::incomplete)
+                     ;
```
greg7mdp, line 2016: should we `check_shutdown() ` here?

heifner: If we exit `maybe_apply_blocks` because of shutdown then `maybe_apply_blocks` returns `complete`.

libraries/chain/controller.cpp
heifner: If replaying there is no reason to come up for air. Nothing else can run.

greg7mdp: So can it be interrupted wit Ctrl-C?

heifner: Yes. I verfied that it works.

greg7mdp: How does it happen? Just curious I don't see it.

greg7mdp: > How does it happen? Just curious I don't see it. I see the ` check_shutdown()` at line 4369, but won't we have the issue that the main thread being busy doesn't process the signal?

Oh I get it, it is because the replay is called by `startup()`, which is called by `chain_plugin_impl::plugin_startup()` where we still have the separate thread catching signals.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: Limit apply blocks.
Note:end

PR #924, opened 2024-10-11
Update `ordered_diff` tests to check that the diff result is not changing (as it is part of consensus)

### LINKED ISSUE #701: Add additional ordered_diff tests
None of the existing `fc::ordered_diff` tests assert anything about the resulting `diff_result` values. Since `diff_result` is part of consensus, we should have some additional tests that verify the `diff_result` values do not change.

### LINKED ISSUE #743: Create dedicated class for QC verification (instead of using the same as for aggregation: `aggregating_qc_t`)

greg7mdp:
Resolves #701.

This PR adds a `checker` class which has a `Print` function to output the diff result in a form that can be used as an initializer. This was useful to generate the `diff_result` initializers we want to compare with.

### DIFF
--- a/libraries/libfc/include/fc/container/ordered_diff.hpp
+++ b/libraries/libfc/include/fc/container/ordered_diff.hpp
@@ -39,6 +39,8 @@ public:
    struct diff_result {
       Container<SizeType>                remove_indexes;
       Container<std::pair<SizeType, T>>  insert_indexes;
+
+      bool operator==(const diff_result&) const = default;
    };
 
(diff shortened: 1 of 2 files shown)


heifner (approved):

linh2931 (approved):
Please update the PR description. This one does not resolve #743, but #701.

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Update `ordered_diff` tests to check the diff result is not changing.
Note:end

PR #926, opened 2024-10-11
[1.0.3 -> main] Limit apply blocks

### LINKED ISSUE #284: Limit apply blocks time in controller

heifner:

Merges `release/1.0` into `main` including #921 
Empty diff because feature already added by #922

Resolves #284


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Limit apply blocks.
Note:end

PR #927, opened 2024-10-11
Split verify_qc and add a new exception type invalid_qc

### LINKED ISSUE #769: Split verify_qc and add a new exception type invalid_qc
In addition to expensive signature verification, currently `verify_qc` checks weighs and vote formats. It can be split into 2 functions, `verify_qc_signatures` and `verify_qc_basics`, such that `verify_qc_basics` can be called in replay when `--force-all-checks` is not specified. https://github.com/AntelopeIO/spring/pull/736 can just post `verify_qc_signatures` to a thread pool.

`verify_qc` uses `invalid_qc_claim` for all the exceptions it throws. This is not accurate. A new exception `invalid_qc` should be added.

greg7mdp:
Resolves #769.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4278,12 +4278,13 @@ struct controller_impl {
 
                if constexpr (std::is_same_v<typename std::decay_t<BSP>, block_state_ptr>) {
                   // do basic checks always (excluding signature verification)
-                  if (qc)
+                  if (qc) {
                      head->verify_qc_basic(*qc);
 
-                  if (conf.force_all_checks && qc) {
-                     // verify signatures only if `conf.force_all_checks`
-                     head->verify_qc_signatures(*qc);
+                     if (conf.force_all_checks) {
+                        // verify signatures only if `conf.force_all_checks`
+                        head->verify_qc_signatures(*qc);
+                     }
                   }
                }
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4112,11 +4112,6 @@ struct controller_impl {
       return {};
    }
 
-   // This verifies BLS signatures and is expensive.
-   void verify_qc( const block_state& prev, const qc_t& qc ) {
-      prev.verify_qc(qc);
-   }
-
    // thread safe, expected to be called from thread other than the main thread
    // tuple<bool best_head, block_handle new_block_handle>
    template<typename ForkDB, typename BS>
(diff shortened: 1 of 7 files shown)


libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -257,6 +257,10 @@ namespace eosio { namespace chain {
                                     3030013, "Block includes an ill-formed additional block signature extension" )
       FC_DECLARE_DERIVED_EXCEPTION( invalid_qc_claim, block_validate_exception,
                                     3030014, "Block includes an invalid QC claim" )
+      FC_DECLARE_DERIVED_EXCEPTION( invalid_qc, block_validate_exception,
+                                    3030014, "Block includes an invalid QC" )
+      FC_DECLARE_DERIVED_EXCEPTION( invalid_qc_signature, block_validate_exception,
+                                    3030014, "Block includes a QC with invalid signature(s)" )
```
heifner, line 263: Change the exception #. Also the # for `invalid_qc`

libraries/chain/controller.cpp
```diff
@@ -4128,8 +4124,11 @@ struct controller_impl {
       log_and_drop_future<void> verify_qc_future;
       if constexpr (is_proper_savanna_block) {
          if (qc) {
-            verify_qc_future = post_async_task(thread_pool.get_executor(), [this, &qc, &prev] {
-               verify_qc(prev, *qc);
+            // do basic checks on provided qc immediately (excluding signature verification)
+            prev.verify_qc_basic(*qc);
```
heifner, line 4128: Seems like this should be done inside the posted task below. It is unlikely to ever fail. Accessing the same data on the same thread would be better than accessing it on two different threads.

heifner (changes requested):

libraries/chain/include/eosio/chain/exceptions.hpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/927/commits/d134711fe9ba076f5a14084e64c3be2e1e5318fd

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/927/commits/d4b6861248443223d09a273ace17fe5db9d633f3

linh2931 (approved):

heifner (approved):

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: Split `verify_qc` and add a new exception type `invalid_qc`.
Note:end

PR #928, opened 2024-10-14
Remove `finality` directories.

greg7mdp:
We had `finality` directories in `libraries/chain` and `libraries/chain/include/eosio/chain`, however these were inconsistently used and were not proven to be useful. This PR removes these subdirectories and moves the files one level up.


linh2931 (approved):
I don't have strong feeling on if or not we should have a separate directory for `finality`. We do have a separate `webassmbly` under `chain`.

heifner (approved):
Don't really see the point, but this is fine I guess.

greg7mdp:
@areg, what do you think?

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Remove unused finality directories under `libraries `.
Note:end

greg7mdp:
Didn't get a very enthusiastic response to this PR. 
I'm OK tabling it if someone prefers having the `finality` directories as today, if so just let me know. 
Otherwise I'll probably merge it later this evening.

Issue #929, opened 2024-10-14
appbase io_context support for coroutines 


heifner:
`appbase` executor currently does not support c++ 20 coroutines. Currently SHiP is using `app().get_io_service()` which by-passes the priority queue executor. `appbase` executor/priority-queue should be enhanced to allow for its use with coroutines.

bhazzard:
For future discussion... Greg pointed out this as an upcoming standard Execution library: https://en.cppreference.com/w/cpp/execution

bhazzard:
Adopting co-routines and fixing some of our executor/priority-queue issues could be fertile ground for optimization and may have a positive impact on ability to process transactions under load.

PR #930, opened 2024-10-14
Plugin shutdown

### LINKED ISSUE #842: cleanup and define consistent pattern for handling nodeos plugin shutdown
My impression is that we do not have a good handle on the threading semantics during shutdown of nodeos. The lingering issues we've seen such as AntelopeIO/leap#646 seem to be good evidence to back this assertion up.

I've long wondered if we really need a two-stage shutdown: first we get all the various plugin threads to stop -- thus, critically, ensuring no callbacks are being run anywhere -- and then destroy the plugins (and their members). Such a pattern is very asio-like and easy to reason on because it doesn't require any shutdown specific handling inside various callbacks such as "keep an object alive until shutdown is complete" or other shenanigans -- just rely on C++ to dtor all the objects naturally which is easy to reason about the safety if you're guaranteed no callbacks are in flight.

It turns out that appbase _already_ seems to define a two stage shutdown. All plugins get their `plugin_shutdown()` called on them before all plugins are destroyed. The shutdown & destruction happens in reverse order to startup (which honors plugins' defined dependencies).

Surprisingly (to me), many plugins are already pretty good about just doing a minimal amount of work (such as stopping threads) in their `plugin_shutdown()`. But some, like chain_plugin, are significant violators of this proposed pattern.
https://github.com/AntelopeIO/leap/blob/cab88df475fe63d654f8020c65a6d04a6f071fd5/plugins/chain_plugin/chain_plugin.cpp#L1110-L1121
In particular destroying `chain::controller` here at this stage of shutdown is rather alarming. It's almost even a smoking gun for shutdown grief if it weren't for chain_plugin fortuitously being the last plugin shutdown.

But more problematic is due to the lack of a cohesive and consistent pattern our code has become littered with what I'm going to call _shutdown sprinkles_. These are little bits of code sprinkled about that seem to be doing something to make shutdown safer: checking if the app is currently shutting down to short circuit something, trying to increase the lifetime of some shared_ptr somewhere; either via weak_ptrs or asio queuing, etc. In some cases it's not clear how such sprinkles could have any side effect at all! In other cases it appears overly conservative (likely due to lack in confidence how shutdown behaves, again due to lack of a consistent pattern). For some examples of shutdown sprinkles see the section below.

<details>
  <summary>shutdown sprinkle examples and commentary</summary>

### producer_plugin
producer_plugin's plugin_shutdown seems to have two great examples. First, consider the timer cancel
https://github.com/AntelopeIO/leap/blob/cab88df475fe63d654f8020c65a6d04a6f071fd5/plugins/producer_plugin/producer_plugin.cpp#L1043-L1054
Recall that by the time `plugin_shutdown()` is called, the main event loop has been stopped and drained/cleared. It's impossible for callbacks to be run on the main event loop once plugins' shutdown has started. `_timer` is created on the main event loop, and all of `_timer.async_wait()`s are also being handled on the main thread. It seems impossible for this `cancel()` to have any side effect because the main event loop will never be pumped again.

An important caveat: if the `async_wait()` had a callback wrapped to an executor on a different thread, there would still be an opportunity it could have a side effect. For example, if the `async_wait()` was wrapped to the `_thread_pool` executor.

Another interesting sprinkle is this one
https://github.com/AntelopeIO/leap/blob/cab88df475fe63d654f8020c65a6d04a6f071fd5/plugins/producer_plugin/producer_plugin.cpp#L1060
The comment appears to be inaccurate. Again, by this point the queue has already been drained/cleared and will never be pumped again. But that doesn't stop you from submitting work on to the queue still. And that appears to be what this line is doing: just wedging a reference to the shared_ptr that will cause the plugin_impl to live until after _all_ the plugin's have been both shutdown & destroyed. I assume this is related to the circular dependency of chain_plugin & producer_plugin.

### trace_api_plugin & http async handlers
trace_api_plugin makes uses of the ability to handle its HTTP RPC requests within http_plugin's thread pool.

trace_api_plugin requires http_plugin
https://github.com/AntelopeIO/leap/blob/cab88df475fe63d654f8020c65a6d04a6f071fd5/plugins/trace_api_plugin/include/eosio/trace_api/trace_api_plugin.hpp#L12
This guarantees that http_plugin is started before trace_api_plugin, but also that `trace_api_plugin::plugin_shutdown()` is before `http_plugin::plugin_shutdown()`, and that `trace_api_plugin::~trace_api_plugin()` is before `http_plugin::~http_plugin()`. Importantly this means that http_plugin's threads will be stopped before `trace_api_plugin::~trace_api_plugin()`. It should never be possible for an http async callback to enter trace_api_plugin after `trace_api_plugin::~trace_api_plugin()` has started. Yet, trace_api_plugin always tries to lock a weak_ptr on itself for every http request
https://github.com/AntelopeIO/leap/blob/cab88df475fe63d654f8020c65a6d04a6f071fd5/plugins/trace_api_plugin/trace_api_plugin.cpp#L298-L304
I suspect this is just being overly conservative.

### http_plugin
https://github.com/AntelopeIO/leap/blob/df6ffbb016282d8f4a9f76ab3db04e52816d3c12/plugins/http_plugin/http_plugin.cpp#L102-L109
This check may be something to do with the the previous appbase behavior which drained its queue before calling plugins' shutdown. Consider in the proposed pattern that once all plugin's shutdowns are completed that no callbacks can be executing, and no plugins have started their destruction. Such a sprinkle would be unnecessary as there is no chance a straggling late callback may access a destroyed object.

---

</details>

I'm reluctant to make definitive sweeping recommendations because a deep dive through the code may yield some surprising interactions that make these impossible to implement. But I'm going to try anyway.

* `plugin_shutdown()`s should only stop plugin specific threads, with few exceptions. Absolutely no destruction of any objects.
* Do not rely on shared_ptr ownership to keep alive objects to varying degrees during shutdown.
  * More generally, lean heavily on standard C++ destruction ordering.
* Aggressively remove any shutdown sprinkles.
* Remove any circular plugin dependencies as these without a doubt cause shutdown sprinkles (https://github.com/AntelopeIO/leap/issues/668 for a known one)
* Refactor appbase to not use a static for its lifecycle (https://github.com/AntelopeIO/leap/issues/650 is blocker)
* Reflect further on how exceptions during `plugin_shutdown()` and destruction are handled.
  * Can we make it a rule that exceptions during shutdown & dtor not be allowed to communicate error codes through main?
* Reflect further on how exceptions during `plugin_initialize()` & `plugin_startup()` are handled.
  * Some plugins, like producer_plugin, manually call `plugin_shutdown()` themselves when their `plugin_startup()` fails. Why? 
  * [greg: see https://github.com/AntelopeIO/appbase/issues/25]
* Clearly define and document appbase's semantics of init/startup/shutdown/dtor and their rules w.r.t. exceptions. Some of these you can discover just by reading the code but they need to be clearly documented. For example:
  * If a plugin throws during `plugin_startup()`, what other plugins get `plugin_shutdown()`?
  * If a plugin throws during `plugin_shutdown()`, what happens -- are other plugins still shutdown even after that exception occurs?

heifner:
Simplify `plugin_shutdown` logic.

- Plugins are responsible for stopping their thread pool in `plugin_shutdown`.
- Modifies `producer_plugin` to use `app().make_timer()` instead of using `app().get_io_service()`
- This PR does not remove `app().get_io_service()`, see #929 

Requires https://github.com/AntelopeIO/appbase/pull/34

Resolves #842


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -338,61 +338,46 @@ void state_history_plugin::plugin_initialize(const variables_map& options) {
...
-   // This is a temporary fix until https://github.com/AntelopeIO/spring/issues/842 can be worked.
-   // Drain the io_service of anything that could be referencing state_history_plugin
-   if (app().executor().get_io_service().stopped()) {
-      app().executor().get_io_service().restart();
-      while (app().executor().get_io_service().poll())
-         ;
-      // clear priority queue of anything pushed by poll(), see application_base exec()
-      app().executor().clear();
-   }
+   fc_dlog(_log, "exit shutdown");
```
spoonincode, line 380: This is inconsistent with the other changes where the 'exit shutdown' is printed at info level, for example
https://github.com/AntelopeIO/spring/blob/2d8b042e111cb04538af8fb1abe4181e6583dbc8/plugins/resource_monitor_plugin/resource_monitor_plugin.cpp#L119
and
https://github.com/AntelopeIO/spring/blob/2d8b042e111cb04538af8fb1abe4181e6583dbc8/plugins/net_plugin/net_plugin.cpp#L4432

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3150,10 +3150,8 @@ namespace eosio {
    }
 
    void net_plugin_impl::plugin_shutdown() {
-         in_shutdown = true;
-
-         connections.close_all();
-         thread_pool.stop();
+      in_shutdown = true;
```
spoonincode, line 3153: What is the purpose of `in_shutdown`? From a quick scan, looks like it could never be `true` anywhere we check for it.

spoonincode: okay I see how it can be `true`, but is this necessary?

heifner: No, I don't think it is. I'll remove it.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -687,6 +687,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    std::map<chain::public_key_type, signature_provider_type> _signature_providers;
    chain::bls_pub_priv_key_map_t                     _finalizer_keys; // public, private
    std::set<chain::account_name>                     _producers;
+   named_thread_pool<struct prod>                    _prod_thread_pool; // for timers
```
spoonincode, line 690: Isn't this premature before understanding the design of #929[^1]? imo this just adds unnecessary timer accuracy error -- as long we have an io_service running on the main thread, main thread timers should expire there. That seems preferable even if we don't expose the io_service (we could hide the io_service from being used for anything other than timers via a `appbase::make_timer()` sort of pattern). Otherwise we've needlessly made these timers require waking 2 threads instead of 1.

[^1]: strictly speaking not even #929 since refactoring the priority queue and its executor doesn't imply removing the main thread io_context; that could be its own issue

heifner: I'll revert this and go with a `make_timer` approach.

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Simplify `plugin_shutdown` logic.
Note:end

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4421,29 +4408,17 @@ namespace eosio {
...
-         fc_ilog( logger, "shutdown.." );
-
-         my->plugin_shutdown();
-         app().executor().post( 0, [me = my](){} ); // keep my pointer alive until queue is drained
-         fc_ilog( logger, "exit shutdown" );
-      }
-      FC_CAPTURE_AND_RETHROW()
+      fc_dlog( logger, "shutdown.." );
+      my->plugin_shutdown();
+      fc_ilog( logger, "exit shutdown" );
```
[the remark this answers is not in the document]
heifner: `net_plugin` is redundant.
```
info  2024-10-16T12:12:03.410 nodeos    net_plugin.cpp:4421           plugin_shutdown      ] exit shutdown
```

libraries/custom_appbase/include/eosio/chain/application.hpp
```diff
@@ -155,7 +155,7 @@ class priority_queue_executor {
    // members are ordered taking into account that the last one is destructed first
 private:
    std::thread::id                    main_thread_id_{ std::this_thread::get_id() };
-   boost::asio::io_service            io_serv_;
+   boost::asio::io_context            io_serv_;
```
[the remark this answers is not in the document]
greg7mdp: Other API like:
```
void application_base::setup_signal_handling_on_ios(boost::asio::io_service& ios
void application_base::startup(boost::asio::io_service& io_serv)
```

should also be updated to take an `io_context&`.

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: depends how you look at the log, zigzag fashion or more linearly, but OK.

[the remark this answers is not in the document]
spoonincode: tbh I'm not really sure how useful these shutdown logs are; I almost suggested downgrading them to debug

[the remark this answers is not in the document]
heifner: I think they only existed because our shutdown logic was so suspect and we wanted these hints of where on shutdown things may have gone wrong. With this PR, changing them all to debug I think makes sense.

spoonincode:
still reviewing

libraries/libfc/include/fc/log/appender.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 #include <memory>
 
-namespace boost { namespace asio { class io_context; typedef io_context io_service; } }
+namespace boost { namespace asio { class io_context; } }
```
spoonincode, line 4: What is this for? I feel like this line was originally some sort of boost backwards compatibility thing.

heifner: Not needed. Removed.

spoonincode (approved):

spoonincode:
possibly unused?
https://github.com/AntelopeIO/spring/blob/3828acdfd14ec415b972c9fc89f7a89a331e3108/plugins/net_plugin/net_plugin.cpp#L4685

greg7mdp (approved):

PR #937, opened 2024-10-15
P2P: Use net threads for connecting and timers

### LINKED ISSUE #817: P2P: Do not burden the main thread with the connection monitor
Currently `net_plugin` launches the connection monitor and keep alive timer on the main thread:
https://github.com/AntelopeIO/spring/blob/a5f0a7f9704385196f66cc1ca1a131cc29d47492/plugins/net_plugin/net_plugin.cpp#L4529-L4530

These can and should be started on the `net_plugin` thread pool.

heifner:
No reason to use the main application thread for the connecting or for the timers.
Also `create_listener` on the main thread at startup. This allows for failures on listen to exit before startup is complete and causes nodeos to return an error code on failure instead of returning a successful exit.

Resolves #817


plugins/net_plugin/net_plugin.cpp
```diff
@@ -4377,23 +4377,23 @@ namespace eosio {
       for(auto listen_itr = listen_addresses.begin(), p2p_iter = p2p_addresses.begin();
           listen_itr != listen_addresses.end();
           ++listen_itr, ++p2p_iter) {
-         app().executor().post(priority::highest, [my=shared_from_this(), address = std::move(*listen_itr), p2p_addr = *p2p_iter](){
```
linh2931, line 4380: Those two used highest priority; any particular reasons for that? Any side effect after moving them to the net thread pool as they will lose highest priority?

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: Use net threads for listener and timers.
Note:end

plugins/net_plugin/net_plugin.cpp
heifner: They were highest priority so they would run before anything else on the main thread. Now they are run in the net thread pool and should also run right away.

linh2931 (approved):

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4364,23 +4364,23 @@ namespace eosio {
...
+         boost::asio::post(thread_pool.get_executor(), [this, address = std::move(*listen_itr), p2p_addr = *p2p_iter]() {
```
spoonincode, line 4367: In the original code the main thread was posting to the app thread presumably to delay the listeners from being started until the main thread event loop starts. With this change, that is no longer the case. The listeners (and also the timers and outgoing connections) will be started immediately. Does that change anything? Does it affect anything wrt replay?

spoonincode (changes requested):
see question

spoonincode:
The PR title and description implies the listener is running on the main thread but afaict that isn't the case, only the first `async_accept()` is called on the main thread and I'm not really sure how that can be a problem.

You mentioned wanting to fix something similar for http & ship but I think if you really want to make the first `async_accept()` call be called from the "right thread", the change should be internal to `fc::listener` to do a dispatch or post before calling that first `async_accept()`

plugins/net_plugin/net_plugin.cpp
heifner: `net_plugin` APPBASE_PLUGIN_REQUIRES `chain_plugin` and `producer_plugin`. Replay will be finished before `net_plugin::plugin_startup` is called. There should be no issue with accepting trxs and blocks before the main thread is started.

heifner:
> You mentioned wanting to fix something similar for http & ship but I think if you really want to make the first `async_accept()` call be called from the "right thread", the change should be internal to `fc::listener` to do a dispatch or post before calling that first `async_accept()`

Thanks. I like this idea. Then the `boost::asio::post` in this PR can be removed and just call `fc::create_listener` directly from `plugin_startup`.

plugins/net_plugin/net_plugin.cpp
spoonincode: What was the purpose of the main thread to main thread post that was previously there? Delaying startup until main thread event loop is running is the only rationale I'm coming up with

heifner: Speed up startup maybe. It might have been done originally because of some startup issue, but doesn't seem necessary.

spoonincode (approved):

plugins/net_plugin/net_plugin.cpp
spoonincode: Adding that main thread to main thread post was described expressly to fix replay: 925007b (from eosnetworkfoundation/mandel#381).

spoonincode:
Yeah something I like about removing the post is that I'd much prefer `plugin_startup()` to fail when unable to bind.

plugins/net_plugin/net_plugin.cpp
heifner: I'm not sure what I was thinking there. The fix there is the reconnect on sync timeout. I've testing the current version of this PR with create_listener on the main thread at startup. I tried a number of different replays and found no issue with listening or connecting after the replay finishes.

heifner:
> Yeah something I like about removing the post is that I'd much prefer `plugin_startup()` to fail when unable to bind.

This is much nicer and causes nodeos to return an error code when it is unable to listen to the provided address(es).

linh2931 (approved):
You might want to update the PR description.

heifner:
> You might want to update the PR description.

I did update it. Anything in particular I should add?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4361,45 +4361,42 @@ namespace eosio {
       incoming_transaction_ack_subscription = app().get_channel<compat::channels::transaction_ack>().subscribe(
             [this](auto&& t) { transaction_ack(std::forward<decltype(t)>(t)); });
 
+      const boost::posix_time::milliseconds accept_timeout(100);
+      std::string extra_listening_log_info = ", max clients is " + std::to_string(connections.get_max_client_count());
       for(auto listen_itr = listen_addresses.begin(), p2p_iter = p2p_addresses.begin();
-          listen_itr != listen_addresses.end();
+         listen_itr != listen_addresses.end();
           ++listen_itr, ++p2p_iter) {
```
greg7mdp, line 4368: indentation

linh2931:
> > You might want to update the PR description.
> 
> I did update it. Anything in particular I should add?

Oh, I was looking at an old version of the description. Current one is good.

PR #938, opened 2024-10-15
Update `raw.hpp` so that we can serialize and deserialize `const` types

### LINKED ISSUE #796: Update `raw.hpp` so that we can serialize and deserialize `const` types
Some types, such as `block_state_legacy::action_mroot_savanna`, are only set in the constructor and never modified. So they are logically const and could be accessed from multiple threads safely. 

Using this would allow us to remove some friend declarations (we use `private` as a way to prevent unchecked access to non thread-safe members).

Ideally we would make the member type const. 
So for `block_state_legacy::action_mroot_savanna` the type would be:
`const std::optional<digest_type>` instead of 
`std::optional<digest_type>`.

Again, the serialization is what trips us up. I feel that the unpack functions in `raw.hpp` should accept to unpack into const types (logically deserializing to recreate a value which is never modified after creation). Right now we cannot mark anything that needs to be serialized const.

So for `optional<T>` the signature should be: `void unpack( Stream& s, const std::optional<T>& v )`

even if it requires a `const_cast` internally.

greg7mdp:
Resolves #796.

Some types are only set in the constructor and never modified. So they are logically const and could be accessed from multiple threads safely.

Using this would allow us to remove some friend declarations (we use private as a way to prevent unchecked access to non thread-safe members).

Ideally we would make the member type const.
The serialization is what trips us up. I feel that the unpack functions in raw.hpp should accept to unpack const members of structs (logically deserializing to recreate a value which is never modified after creation). Right now we cannot mark non-funadamental types that needs to be serialized const.

Also added a test.

### DIFF
--- a/libraries/chain/include/eosio/chain/finality/finalizer.hpp
+++ b/libraries/chain/include/eosio/chain/finality/finalizer.hpp
@@ -189,7 +189,6 @@ namespace eosio::chain {
 
       // for testing purposes only, not thread safe
       const fsi_t& get_fsi(const bls_public_key& k) const { return finalizers.at(k).fsi; }
-
       void set_fsi(const bls_public_key& k, const fsi_t& fsi) {
          if (auto it = finalizers.find(k); it != finalizers.end())
             it->second.fsi = fsi;
--- a/libraries/chain/include/eosio/chain/finality/finalizer.hpp
+++ b/libraries/chain/include/eosio/chain/finality/finalizer.hpp
@@ -59,13 +59,18 @@ namespace eosio::chain {
          bool          monotony_check {false};
       };
 
-      bls_private_key               priv_key;
+      const bls_private_key         priv_key;
       finalizer_safety_information  fsi;
 
       vote_result  decide_vote(const block_state_ptr& bsp);
       vote_message_ptr maybe_vote(const bls_public_key& pub_key, const block_state_ptr& bsp, const digest_type& digest);
+
       // finalizer has voted strong, update fsi if it does not already contain vote or better
       bool maybe_update_fsi(const block_state_ptr& bsp);
+
+      finalizer(const bls_private_key& priv_key, const finalizer_safety_information& fsi)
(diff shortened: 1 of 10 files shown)


heifner (changes requested):
Since we have the time. What about making a pass through the code and making what can be const, `const`.

greg7mdp:
> Since we have the time. What about making a pass through the code and making what can be const, `const`.

I started doing that, but it does require a lot of changes through the code. For example most of the `block_header_state` should be const. But typically we create the empty struct and assign members after the fact (see [here](https://github.com/AntelopeIO/spring/blob/8cc763f3c331aff2287940669c3ca1d60dab2285/libraries/chain/block_state.cpp#L136-L160) for example). Also the `finish_next` function.  

Probably if we want to do this I should create another issue.

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Update `raw.hpp` so that we can serialize and deserialize `const` types.
Note:end

libraries/chain/include/eosio/chain/finality/finalizer.hpp
```diff
@@ -184,7 +189,13 @@ namespace eosio::chain {
 
       // for testing purposes only, not thread safe
       const fsi_t& get_fsi(const bls_public_key& k) const { return finalizers.at(k).fsi; }
-      void         set_fsi(const bls_public_key& k, const fsi_t& fsi) { finalizers[k].fsi = fsi; }
+
```
heifner, line 192: I think remove this blank line or add another comment saying this is only for testing purposes and not thread safe.

heifner (approved):

libraries/chain/include/eosio/chain/finality/finalizer.hpp
greg7mdp: removed!

heifner (approved):

linh2931 (approved):

Issue #515, opened 2024-08-09
Disable tracked_votes processing if http category chain_ro not enabled


heifner:
The `chain_api_plugin` does extra processing to support `v1/chain/get_finalizer_info` http endpoint. If the http `chain_ro` category is not enabled then no reason to do the extra processing.

For example, if running a block producer node where it is desirable to run with `chain_api_plugin` so that `v1/chain/get_info` http endpoint is available, but no need to provide `v1/chain/get_finalizer_info`, the following configuration could be used.
```
http-server-address   = http-category-address
http-category-address = producer_ro,/tmp/absolute_unix_path.sock
http-category-address = producer_rw,/tmp/absolute_unix_path.sock
```

See https://github.com/AntelopeIO/leap/pull/1137 for documentation on http categories.

PR #939, opened 2024-10-15
Enable tracked_votes processing only if http chain_ro category is configured

linh2931:
The chain_api_plugin does extra processing to support `v1/chain/get_finalizer_info` http endpoint. If the http `chain_ro` category is not enabled then there is no need to do the extra processing.

Resolves https://github.com/AntelopeIO/spring/issues/515

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -975,14 +975,15 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
          }
       }
 
-      // only enable last tracked votes if chain_api_plugin enabled, if no http endpoint, no reason to track
+      _last_tracked_votes.emplace(*chain);
+
       bool chain_api_plugin_configured = false;
       if (options.count("plugin")) {
          const auto& v = options.at("plugin").as<std::vector<std::string>>();
          chain_api_plugin_configured = std::ranges::any_of(v, [](const std::string& p) { return p.find("eosio::chain_api_plugin") != std::string::npos; });
       }
-      _last_tracked_votes.emplace(*chain, chain_api_plugin_configured);
 
+      // only enable _get_info_db if chain_api_plugin enabled.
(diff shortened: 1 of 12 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -975,14 +975,39 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
...
+            }
+         }
+      }
+
+      // only enable last tracked votes if chain_api_plugin and chain_ro are enabled.
+      bool tracking_enabled = chain_api_plugin_configured && chain_ro_category_configured;
+      _last_tracked_votes.emplace(*chain, tracking_enabled);
+
+      // only enable last tracked votes if chain_api_plugin enabled.
       _get_info_db.emplace(*chain, chain_api_plugin_configured);
```
heifner, line 1011: `chain_apis::read_only` has a `last_tracked_votes`. Seems like `chain_api_plugin::plugin_startup` could ask `http_plugin` if `chain_ro` category is enabled and then inform `last_tracked_votes`.

heifner (changes requested):

plugins/chain_plugin/chain_plugin.cpp
linh2931: Thanks! This is a good idea. 

I was trying to query `http_plugin` from here but did not like introducing dependency of `chain_plugin` on `http_plugin`. I will make the changes.

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Enable tracked_votes processing only if http chain_ro category is configured.
Note:end

plugins/chain_plugin/chain_plugin.cpp
linh2931: Although more files are touched, the new approach hides the details of how http categories are intialized.

heifner (approved):

PR #943, opened 2024-10-16
HTTP: Warning and test fix

### LINKED ISSUE #683: Test failure: http_plugin_unit_tests
https://github.com/AntelopeIO/spring/actions/runs/10670275150/job/29574368045
```
info  2024-09-02T16:20:00.531 http_plug listener.hpp:131              log_listening        ] start listening on 127.0.0.1:8892 resolved from 127.0.0.1:8892 for API categories: all
info  2024-09-02T16:20:00.540 http_plug http_plugin.cpp:524           log_add_handler      ] add node api url: /doit on 127.0.0.1:8892 
"hello"
"world!"
"yes"
"yes"
/__w/spring/spring/plugins/http_plugin/tests/unit_tests.cpp(715): fatal error: in "requests_in_flight": critical check r[boost::beast::http::status::ok] == 8u has failed [7 != 8]
info  2024-09-02T16:20:00.543 http_plug http_plugin.cpp:515           plugin_shutdown      ] exit shutdown

*** 1 failure is detected in the test module "http_plugin unit tests"
```

heifner:
- Fixes compiler warning about `this` not used by lambda
- Add additional wait for no bytes in flight.
  - On my machine, it takes awhile for the http_plugin to react to the close of the sockets and clear out the bytes in flight.
  - Not sure why this requires such a time on my machine. Maybe WSL2 connections are really slow.

Resolves #683


plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -606,6 +606,8 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
          req.set(http::field::host, "127.0.0.1:8891");
          boost::beast::http::write(s, req);
       }
+      //make sure got to the point http threads had responses queued
```
spoonincode, line 609: why does it matter if the response is queued or not at this point?

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -606,6 +606,8 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
...
+      std::this_thread::sleep_for(std::chrono::seconds(5));
```
linh2931, line 610: Why here sleeping 5 seconds but below sleeping 1 second?

plugins/http_plugin/tests/unit_tests.cpp
heifner: This avoids `boost::beast::http::status::service_unavailable` on my machine

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -642,13 +644,14 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
...
    send_4mb_requests(32u);
-   //make sure got to the point http threads had responses queued
-   std::this_thread::sleep_for(std::chrono::seconds(1));
    //now rip these connections out before the responses are completely sent
    connections.clear();
    //send some requests that should work still
    send_4mb_requests(8u);
    r = drain_http_replies();
+   for (const auto& e : r) {
+      ilog( "response: ${r}, count: ${c}", ("r", std::string(obsolete_reason(e.first)))("c", e.second));
```
linh2931, line 653: I almost thought the `r` in `${r}` was the `r` in `r = drain_http_replies()`.

plugins/http_plugin/tests/unit_tests.cpp
heifner: Because the 1s is not for 4mb messages.

plugins/http_plugin/tests/unit_tests.cpp
heifner: Changed

spoonincode (changes requested):
I don't think we should be adding sleeps to tests

plugins/http_plugin/tests/unit_tests.cpp
spoonincode: maybe after the `boost::beast::http::write(s, req)` try doing a `s.wait(wait_read)`, that will guarantee that the response is in flight. just a guess though

greg7mdp (approved):

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -642,13 +644,14 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
 
    //load up some more requests that exceed max
    send_4mb_requests(32u);
-   //make sure got to the point http threads had responses queued
-   std::this_thread::sleep_for(std::chrono::seconds(1));
    //now rip these connections out before the responses are completely sent
    connections.clear();
```
spoonincode, line 648: If I understood your description right, this is the area that is problematic. Between this line and the line below the http plugin needs to have cleaned up all its ownstanding async_writes. Just closing the socket here does indeed send the FIN but the http threads need time to wake up and handle that error on the async_write to then clear out what's in flight.

I wonder if what might work is to `shutdown` each connection and then `async_read` on each one until getting confirmation the remote side disconnected via an error code.

heifner: That would likely work. I added a check for `bytes_in_flight` since that is really what we want to test anyway. We want to verify that bytes in flight goes to 0 once connections are closed.

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -624,6 +623,12 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
       return count_of_status_replies;
    };
 
+   auto wait_for_no_bytes_in_flight = [&](uint16_t max = std::numeric_limits<uint16_t>::max()) {
```
greg7mdp, line 626: `uint16_t max = std::numeric_limits<uint16_t>::max()` is kinda weird.  Why not use `size_t` and an constant like `10000`?

heifner: Just following the pattern of `drain_http_replies`.

ericpassmore:
Note:start
category: System Stability
component: HTTP
summary: Fix warnings and test.
Note:end

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -694,12 +704,19 @@ BOOST_FIXTURE_TEST_CASE(requests_in_flight, http_plugin_test_fixture) {
...
+      BOOST_CHECK(max > 0);
+   };
+
 
    //8 requests to start with
    send_requests(8u);
    std::unordered_map<boost::beast::http::status, size_t> r = scan_http_replies();
    BOOST_REQUIRE_EQUAL(r[boost::beast::http::status::ok], 8u);
    connections.clear();
+   wait_for_no_bytes_in_flight();
```
spoonincode, line 719: this test case has to do with requests in flight, not bytes, so this feels mismatched but maybe it still serves enough of its purpose?

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -604,6 +604,7 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
          boost::beast::http::request<boost::beast::http::empty_body> req(boost::beast::http::verb::get, "/4megabyte", 11);
          req.keep_alive(true);
          req.set(http::field::host, "127.0.0.1:8891");
+         s.wait(boost::asio::ip::tcp::socket::wait_write);
```
spoonincode, line 607: What is the purpose of this and the `wait_read` below? Why would the socket not be writable here? And if it isn't writable why does it matter before a sync write is issued?

heifner: Not needed. I'll remove.

plugins/http_plugin/tests/unit_tests.cpp
heifner: A requests in flight would be better and useful to verify in both cases. I'll add it.

linh2931 (approved):

greg7mdp (approved):

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -622,6 +622,12 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
       return count_of_status_replies;
    };
 
+   auto wait_for_no_bytes_in_flight = [&](uint16_t max = std::numeric_limits<uint16_t>::max()) {
+      while (http_plugin->bytes_in_flight() > 0 && http_plugin->requests_in_flight() > 0 && --max)
```
spoonincode, line 626: When waiting for no bytes in flight why do we need to look at requests in flight too?

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -646,9 +652,13 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
    std::this_thread::sleep_for(std::chrono::seconds(1));
```
spoonincode, line 652: Can we get rid of this final sleep? It seems like waiting for each connection to indicate it's readable would suffice. But if we need to make another busy loop maybe we could wait to see requests_in_flight is 32

plugins/http_plugin/tests/unit_tests.cpp
heifner: No "need", just thought it was good to also verify `requests_in_flight` appropriately hits 0.

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -623,7 +623,7 @@ BOOST_FIXTURE_TEST_CASE(bytes_in_flight, http_plugin_test_fixture) {
    };
 
    auto wait_for_no_bytes_in_flight = [&](uint16_t max = std::numeric_limits<uint16_t>::max()) {
-      while (http_plugin->bytes_in_flight() > 0 && http_plugin->requests_in_flight() > 0 && --max)
```
greg7mdp: Seems like this shouldn't make a difference as they should both either be 0 or both non-zero.

spoonincode: I'm confused by that too. And I don't understand the discrepancy why `wait_for_no_bytes_in_flight` looks at both while `wait_for_no_requests_in_flight` looks at only one. If the intent is to look at both then a common `wait_for_nothing_in_flight` could be made that looks at both.

spoonincode (approved):

plugins/http_plugin/tests/unit_tests.cpp
heifner: In practice the requests in flight also matter. The issue is that bytes in flight can reach 0 before all the requests are done (or started).

`bytes_in_flight` is only incremented on sending the response. We do not, currently at least, include the request.
https://github.com/AntelopeIO/spring/blob/2b43faf02056ed303768708ea9e4152cadc1a56b/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp#L466-L468

Therefore, requests can be in progress, but `bytes_in_flight` can be 0. So we need to wait until all requests are complete to know that all `bytes_in_flight` are accounted for.

greg7mdp: So could we just wait for `requests_in_flight()` to be 0, and then assert that `bytes_in_flight()` is 0 as well?

heifner: Sure, that sounds better

greg7mdp (approved):

Issue #935, opened 2024-10-15
[1.1] get_info in irreversible mode (nodeos_irreversible_mode_if_lr_test)


heifner:
Before #920, a node running in `irreversible` mode would always report the same value for:
```
"head_block_num":456
"last_irreversible_block_num":456
```

Now it is possible for a node running in `irreversible` mode to return different values:
```
"head_block_num":457
"last_irreversible_block_num":456
```

See test failure: https://github.com/AntelopeIO/spring/actions/runs/11349047274/job/31565035139

This is possible now because `accepted_block` is signaled before `irreversible_block` and a `get_info` call can come in between those two signals. Before since the `get_info` ran on the main thread, it could not execute between the two signals.

linh2931:
https://github.com/AntelopeIO/spring/actions/runs/11367787258/job/31622083125?pr=941#step:4:1332

PR #944, opened 2024-10-16
Do not store get_info results on accepted_block signal when read_mode is IRREVERSIBLE

linh2931:
In `IRREVERSIBLE` read mode, it is expected `get_info` to return the same `head_block_num` and `last_irreversible_block_num`.  But a `get_info` request can come between `accepted_block` signal and `irreversible_block` signal, which results in inconsistencies.

A solution is not storing `get_info` results on `accepted_block` signal, but storing only on `irreversible_block` signal.

Resolves https://github.com/AntelopeIO/spring/issues/935

### DIFF
--- a/plugins/chain_plugin/get_info_db.cpp
+++ b/plugins/chain_plugin/get_info_db.cpp
@@ -26,6 +26,16 @@ namespace eosio::chain_apis {
                return;
             }
 
+            // In IRREVERSIBLE mode, it is expected get_info to return the same
+            // head_block_num and last_irreversible_block_num.
+            // But a get_info request can come between accepted_block signal and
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

greg7mdp:
I think instead of having these two returns, it would be more straightforward to have:
```
if (get_info_enabled && !(controller.get_read_mode() == db_read_mode::IRREVERSIBLE))
   store_info();
```

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: System Compatibility
component: Internal
summary: Ensure head block and last irreversible block are in sync when in IRREVERSIBLE mode.
Note:end

PR #945, opened 2024-10-17
use boringssl's upstream directly; removes need for our fork and committed generated files

spoonincode:
When we added boringssl the only way to integrate it without requiring golang was to commit generated files in to the repository. This is quite the horror to me because it makes it difficult for anyone (including ourselves) to check that the generated files haven't been tampered with by whoever committed them. To try and mitigate that, the repo we store the generated files in (https://github.com/AntelopeIO/boringssl-build) contains a workflow to double check the committed files by generating them itself using golang etc. Still very undesirable though.

A handful of months ago boringssl modified their build structure so that golang was no longer required to use the cmake files. And, astonishingly, it Just Works even in our use case.

This is a big deal. It makes it trivial for anyone (including ourselves) to verify that we're using upstream verbatim. I think it's worth doing the upgrade.

I eyeballed the commit titles (and it some cases the commit contents) from our prior pinned commit of a1843d6. Didn't see anything alarming (anything that could affect our usage). Certainly could be worth another set of eyeballs skimming over the commit title history (https://boringssl.googlesource.com/boringssl/+log).

Didn't see any differences in our benchmark application (the merkle test is a bit persnickety for me to get consistent results though unless I increase the number of runs substantially, since these otherwise run very few runs)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,6 +37,6 @@
 [submodule "libraries/boost"]
 	path = libraries/boost
 	url = https://github.com/boostorg/boost.git
-[submodule "libraries/libfc/libraries/boringssl/boringssl"]
-	path = libraries/libfc/libraries/boringssl/boringssl
-	url = https://github.com/AntelopeIO/boringssl-build
+[submodule "libraries/libfc/libraries/boringssl/bssl"]
+	path = libraries/libfc/libraries/boringssl/bssl
+	url = https://boringssl.googlesource.com/boringssl
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,7 +238,7 @@ configure_file(libraries/eos-vm/LICENSE                                    licen
 configure_file(libraries/prometheus/prometheus-cpp/LICENSE                 licenses/spring/LICENSE.prom      COPYONLY)
 configure_file(programs/cleos/LICENSE.CLI11                                licenses/spring/LICENSE.CLI11     COPYONLY)
 configure_file(libraries/libfc/libraries/bls12-381/LICENSE                 licenses/spring/LICENSE.bls12-381 COPYONLY)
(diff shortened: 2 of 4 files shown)


.gitmodules
```diff
@@ -37,6 +37,6 @@
 [submodule "libraries/boost"]
 	path = libraries/boost
 	url = https://github.com/boostorg/boost.git
-[submodule "libraries/libfc/libraries/boringssl/boringssl"]
-	path = libraries/libfc/libraries/boringssl/boringssl
-	url = https://github.com/AntelopeIO/boringssl-build
+[submodule "libraries/libfc/libraries/boringssl/bssl"]
+	path = libraries/libfc/libraries/boringssl/bssl
```
linh2931, line 41: Wondering why two levels `boringssl/bssl` in the path, not just `boringssl` or `bssl`?

linh2931 (approved):
Scanned the commit history list. It seems fine.

greg7mdp (approved):

.gitmodules
spoonincode: We still need our "wrapper" CMakeLists.txt that is in the `libraries/libfc/libraries/boringssl` directory, to then include upstream. All that could have just been placed further up in `libraries/libfc/libraries` but it's kinda long. In other words we still can't _just_ directly use the root of upstream's CMakeLists because we need to disable Werror, add install instructions, etc

I changed from `boringssl` to `bssl` to reduce grief when we switch between 1.0 & main because I think in the past we've had some annoyances with git in this regard... but maybe that was only when desubmoduling -- which this is not. So maybe unnecessary.

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Pull boringssl submodule directly from upstream source.
Note:end

PR #949, opened 2024-10-17
HTTP: plugin_startup throw on listener error

heifner:
- `http_plugin::plugin_startup`.
  - The removed comment about not wanting to start listening until replay has finished is not a concern because `chain_plugin` is explicitly listed as a auto-start plugin in nodeos main: `app->initialize<chain_plugin, net_plugin, producer_plugin, resource_monitor_plugin>` and since `chain_plugin` does not depend on `http_plugin` then `chain_plugin` will be loaded first.
  - Add a better comment about why the post is needed.
  - Orginally this code was added because `keosd` would start listening for http calls before the `wallet_api_plugin` was loaded.
  - `throw` instead of `app().quit()` since we want `nodeos` to exit with an error code.
- Move http unittest from `tests` directory into the `http_plugin/tests` directory.

Related to #842


greg7mdp (approved):

plugins/http_plugin/http_plugin.cpp
```diff
@@ -477,9 +477,11 @@ namespace eosio {
...
 
    void http_plugin::plugin_startup() {
-      app().executor().post(appbase::priority::high, [this] ()
-      {
-         // The reason we post here is because we want blockchain replay to happen before we start listening.
+      // post here because *_api_plugins that register api handlers depend on http_plugin. Since they depend on the
+      // http_plugin, this plugin_startup is called before any *_api_plugin::plugin_startup. The post avoid situation
+      // where the application is running and accepting http requests, but it doesn't have the request api handler
+      // registered yet.
+      app().executor().post(appbase::priority::high, exec_queue::read_write, [this]() {
```
linh2931, line 484: Like this comment.

linh2931 (approved):

ericpassmore:
Note:start
category: Chores
component: HTTP
summary: Code cleanup on HTTP plugin.
Note:end

spoonincode (approved):
Disappointed we can't immediately fail in startup, but I know that's more complex to pull off here

PR #950, opened 2024-10-18
P2P: Check for is_quitting() for faster shutdown

heifner:
For large `sync-fetch-span`, shutdown can be slow as there can be many queued up blocks to process on the dispatcher strand. Check for `app().is_quitting()` for faster shutdown. This provides a noticable improvement in response to ctrl-c shutdown request.

Related to #922

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3694,6 +3694,8 @@ namespace eosio {
       // post to dispatcher strand so that we don't have multiple threads validating the block header
       peer_dlog(this, "posting block ${n} to dispatcher strand", ("n", ptr->block_num()));
       my_impl->dispatcher.strand.post([id, c{shared_from_this()}, ptr{std::move(ptr)}, cid=connection_id]() mutable {
+         if (app().is_quiting()) // large sync span can have many of these queued up, exit quickly
+            return;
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3694,6 +3694,8 @@ namespace eosio {
       // post to dispatcher strand so that we don't have multiple threads validating the block header
       peer_dlog(this, "posting block ${n} to dispatcher strand", ("n", ptr->block_num()));
       my_impl->dispatcher.strand.post([id, c{shared_from_this()}, ptr{std::move(ptr)}, cid=connection_id]() mutable {
+         if (app().is_quiting()) // large sync span can have many of these queued up, exit quickly
```
linh2931, line 3697: Why not check this before `post`?

heifner: The only thing that happens before the post is pulling the block off the wire. The expensive part is done inside this post. So this implicit queue can become large. The check needs to be inside since that is what is queued up.

linh2931 (approved):

ericpassmore:
Note:start
category: Enhancement
component: Internal
summary: Speed up shutdown when configured with large sync span.
Note:end

PR #951, opened 2024-10-18
Use constructor to construct `action` instead of assigning members after the fact.

greg7mdp:
This is an experiment attempting to make two members of `action_base` const. 

We may not want to merge it as it may cause other repo tests to fail.

However, we could merge the first commit, which does not change `action_base`, but updates or code to construct `action` using the constructor, instead of declaring a default constructed object and assigning the members after the fact.

### DIFF
--- a/unittests/currency_tests.cpp
+++ b/unittests/currency_tests.cpp
@@ -25,11 +25,8 @@ class currency_tester : public T {
       auto push_action(const account_name& signer, const action_name &name, const variant_object &data ) {
          string action_type_name = abi_ser.get_action_type(name);
 
-         action act;
-         act.account = "eosio.token"_n;
-         act.name = name;
-         act.authorization = vector<permission_level>{{signer, config::active_name}};
-         act.data = abi_ser.variant_to_binary(action_type_name, data, abi_serializer::create_yield_function( T::abi_serializer_max_time ));
+         action act(vector<permission_level>{{signer, config::active_name}}, "eosio.token"_n, name,
+                    abi_ser.variant_to_binary(action_type_name, data, abi_serializer::create_yield_function( T::abi_serializer_max_time )));
 
          signed_transaction trx;
          trx.actions.emplace_back(std::move(act));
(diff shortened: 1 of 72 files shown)


heifner (approved):

unittests/wasm-spec-tests/generated-tests/address.cpp
```diff
@@ -11,10 +11,7 @@ BOOST_DATA_TEST_CASE(address_0_check_throw, boost::unit_test::data::xrange(0,11)
    tester.set_code("wasmtest"_n, wasm_address_0);
    tester.produce_block();
 
-   action test;
-   test.account = "wasmtest"_n;
-   test.name = account_name((uint64_t)index);
-   test.authorization = {{"wasmtest"_n, config::active_name}};
+   action test({{"wasmtest"_n, config::active_name}}, "wasmtest"_n, account_name((uint64_t)index), {});
```
spoonincode, line 14: I guess we found no way to regenerate these files properly? I think if anything should update the generator,
https://github.com/AntelopeIO/spring/blob/f45de1f9f0d034855efdd3d8fdf29c77ec43759d/unittests/wasm-spec-tests/generator/eosio_wasm_spec_test_generator.cpp#L39-L43

btw it sure looks like AntelopeIO/wasm-spec-tests@026fb4b found a way to regenerate.

unittests/currency_tests.cpp
```diff
@@ -25,11 +25,8 @@ class currency_tester : public T {
       auto push_action(const account_name& signer, const action_name &name, const variant_object &data ) {
          string action_type_name = abi_ser.get_action_type(name);
 
-         action act;
-         act.account = "eosio.token"_n;
-         act.name = name;
-         act.authorization = vector<permission_level>{{signer, config::active_name}};
-         act.data = abi_ser.variant_to_binary(action_type_name, data, abi_serializer::create_yield_function( T::abi_serializer_max_time ));
+         action act(vector<permission_level>{{signer, config::active_name}}, "eosio.token"_n, name,
+                    abi_ser.variant_to_binary(action_type_name, data, abi_serializer::create_yield_function( T::abi_serializer_max_time )));
```
linh2931, line 30: I feel the original way of initializing individual fields explicitly clearer and more readable. `action` is a data class and used most locally. It is probably appropriate to be a `struct` without const.

greg7mdp: I very much disagree. Because `action` has constructors, it implies that the object should be constructed using the provided constructors. 

The direct member initialization forgoes any hope to ensure that objects are always constructed in a consistent fashion, and that some internal members possibly dependent on the constructors parameters are always initialized correctly.

greg7mdp: https://github.com/AntelopeIO/spring/pull/951/commits/b158e5b2b187b226e59e58eb64cb735a21f8d696

unittests/wasm-spec-tests/generated-tests/address.cpp
greg7mdp: Updated the generator https://github.com/AntelopeIO/spring/pull/951/commits/b158e5b2b187b226e59e58eb64cb735a21f8d696.

heifner (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Improve constructor by using action instead of assigning members after the fact.
Note:end

unittests/currency_tests.cpp
greg7mdp: @arhag do you have an opinion on whether it is worthwhile to merge this PR?

PR #952, opened 2024-10-18
TestHarness: Remove hardcoded version checks to support Leap v3.1 and below.

### LINKED ISSUE #48: `TestHarness` version roll back ambiguity
`TestHarness` configures itself differently based on what `nodeos --full-version` returns. Its behavior is altered if the version begins with `v2` or `v3.1` with the assumption if it's not one of those versions it must be the "latest". So for spring v1 this code still does what is expected ("latest"), but spring v2 will not follow the expected path.
https://github.com/AntelopeIO/spring/blob/bef48a2287f3cff0421f09b40cc4a7bae48317d9/tests/TestHarness/Node.py#L71-L91

greg7mdp:
Resolves #48


spoonincode (approved):
I noticed one more instance of something old-version related,
https://github.com/AntelopeIO/spring/blob/f881e8adb66827694d747cfbc023bc00c9f45690/tests/TestHarness/Cluster.py#L468-L472
which gets plumbed through to 
https://github.com/AntelopeIO/spring/blob/f881e8adb66827694d747cfbc023bc00c9f45690/tests/TestHarness/launcher.py#L210
and
https://github.com/AntelopeIO/spring/blob/f881e8adb66827694d747cfbc023bc00c9f45690/tests/TestHarness/launcher.py#L583-L588

I suspect these can be removed too, but I don't know enough about the test harness to be confident

greg7mdp:
> I noticed one more instance of something old-version related,

From my reading of the code, I think this code is currently active for any Spring/Leap version, including Spring 1.0, so I feel the safa alternative is to let that code remain. I don't know enough  to say whether adding this flag is useful or not though.

tests/TestHarness/Node.py
```diff
@@ -82,26 +82,13 @@ def __init__(self, host, port, nodeId: int, data_dir: Path, config_dir: Path, cm
...
-            if 'v3.1' in self.nodeosVers:
-                self.cleosLimit = ""
-            else:
-                self.cleosLimit = "--time-limit 999"
+        self.fetchTransactionCommand = lambda: "get transaction_trace"
+        self.fetchTransactionFromTrace = lambda trx: trx['id']
+        self.fetchBlock = lambda blockNum: self.processUrllibRequest("trace_api", "get_block", {"block_num":blockNum}, silentErrors=False, exitOnError=True)
+        self.fetchKeyCommand = lambda: "[transaction][transaction_header][ref_block_num]"
+        self.fetchRefBlock = lambda trans: trans["block_num"]
+        self.fetchHeadBlock = lambda node, headBlock: node.processUrllibRequest("chain", "get_block_info", {"block_num":headBlock}, silentErrors=False, exitOnError=True)
```
heifner, line 90: Seems like we could remove the lambda altogether. Not need for this indirection.

heifner:
> From my reading of the code, I think this code is currently active for any Spring/Leap version, including Spring 1.0, so I feel the safa alternative is to let that code remain. I don't know enough to say whether adding this flag is useful or not though.

The `history_api_plugin` does not exist in Spring. All references to it can be removed.

heifner (changes requested):

tests/TestHarness/Node.py
heifner: I meant we could remove all of these lambdas. You will need to update the usage of the lambdas with the actual code.

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Cleanup tests, removed tests for Leap 3.1 and below. Removed tests for `history_api_plugin` which does not exist in Spring.
Note:end

tests/TestHarness/Node.py
greg7mdp: https://github.com/AntelopeIO/spring/pull/952/commits/bd5c628b9b6a06f857a084b08efcf0debdf76149

heifner (approved):

heifner (approved):

PR #953, opened 2024-10-18
PerformanceHarness: Remove hardcoded version checks to support older Leap versions

### LINKED ISSUE #49: `PerformanceHarness` version roll back ambiguity
`PerformanceHarness` configures itself differently based on what `nodeos --full-version` returns. Its behavior is altered if the version begins with `v2` or `v4` with the assumption if it's not one of those versions it must be the "latest". So for spring v1 this code still does what is expected ("latest"), but spring v2 will not follow the expected path. Examples,
https://github.com/AntelopeIO/spring/blob/bef48a2287f3cff0421f09b40cc4a7bae48317d9/tests/PerformanceHarness/performance_test_basic.py#L128-L131
https://github.com/AntelopeIO/spring/blob/bef48a2287f3cff0421f09b40cc4a7bae48317d9/tests/PerformanceHarness/performance_test_basic.py#L144-L147

greg7mdp:
Resolves #49.

### DIFF
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -125,12 +125,9 @@ class PerformanceTestBasic:
             def configureValidationNodes():
                 validationNodeSpecificNodeosStr = ""
                 validationNodeSpecificNodeosStr += '--p2p-accept-transactions false '
-                if "v2" in self.nodeosVers:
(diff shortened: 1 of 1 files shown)


ericpassmore:
Note:start
category: Tests
component: Internal
summary: Clean up Performance Harness tests, removed logic for versions of Leap 2.x.
Note:end

ericpassmore (approved):

PR #957, opened 2024-10-21
[1.0.3] Prometheus: Update speculative block metrics for produced blocks

### LINKED ISSUE #956: nodeos_block_num metric flapping on BP node
#### Describe the bug
`nodeos_block_num` Prometheus metric is showing **0** when block producer is producing a block itself. 
It generates false-positive alerts and renders Grafana charts unreadable. Non-producer nodes are not affected.

#### Expected behavior
Show real block number even if node is currently producing without flapping.

#### Screenshots

Left side - producer logs, right side is showing `while true curl | grep nodeos_block_num`

#### Specs / Additional context
OS - Ubuntu 22.04 LTS
RAM - 256 GB
CPU - AMD Ryzen Threadripper 7960X

We observe this issue **starting from** 
```   
v1.0.0-a8159feae6f7f0d89dc0f990682a3b09635a3e1f
```
it **persists** on current version
```
v1.0.1-574650744460373f635d48cac9aa6dee67dcbfdb
```

heifner:
Update the speculative block metrics for produced blocks.

Resolves #956

### DIFF
--- a/plugins/prometheus_plugin/metrics.hpp
+++ b/plugins/prometheus_plugin/metrics.hpp
@@ -259,6 +259,7 @@ struct catalog_type {
       net_usage_us_produced_block.Increment(metrics.net_usage_us);
 
       update(produced_metrics, metrics);
+      update(speculative_metrics, metrics);
 
       last_irreversible.Set(metrics.last_irreversible);
       head_block_num.Set(metrics.head_block_num);


greg7mdp (approved):

plugins/prometheus_plugin/metrics.hpp
```diff
@@ -259,6 +259,7 @@ struct catalog_type {
       net_usage_us_produced_block.Increment(metrics.net_usage_us);
 
       update(produced_metrics, metrics);
+      update(speculative_metrics, metrics);
```
spoonincode, line 262: it's okay that this increments speculative `num_blocks_created` right?

heifner: Yes, I think so. If we want to report produced blocks also as speculative, seems we should report all the values.

linh2931 (approved):

ericpassmore:
Note:start
category: Logging
component: Prometheus
summary: Update speculative block metrics for when producing blocks.
Note:end

PR #963, opened 2024-10-21
Add bn256 tests

### LINKED ISSUE #962: bn256 unit tests are not run as part of spring's build
The unit tests from the bn256 library that leap uses are not run as part of leap's build. While the bn256 repo's CI does run these tests, the compilers (and flags) used in that CI may not be the same as we (or individual users) use for building leap. We really should add these tests as part of leap's build.

heifner:
Resolves #962

### DIFF
--- a/libraries/libfc/CMakeLists.txt
+++ b/libraries/libfc/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory( secp256k1 )
-add_subdirectory( libraries/bn256/src )
+set(BN256_ENABLE_TEST ON CACHE BOOL "enable BN256 tests")
+add_subdirectory( libraries/bn256 )
 add_subdirectory( libraries/bls12-381 )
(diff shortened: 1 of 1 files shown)


spoonincode (changes requested):
bn256 removed its local span when it was updated to require c++20, but the span reference lives on in a few places like
https://github.com/AntelopeIO/bn256/blob/b5adbb76d456e85385296a1388ba18f6f07d9d9e/CMakeLists.txt#L28
and
https://github.com/AntelopeIO/bn256/blob/b5adbb76d456e85385296a1388ba18f6f07d9d9e/src/CMakeLists.txt#L22
the former being what is causing your immediate failure

But I think including the root CMakeLists is going to cause bn256 to become part of the `make install` target which we want to avoid. That's going to be more invasive to get right

heifner:
@spoonincode thanks for looking into this. I'm going to close this and you or someone else can pick it up later.

Issue #942, opened 2024-10-16
Test failure: separate_prod_fin_test


linh2931:
https://github.com/AntelopeIO/spring/actions/runs/11367823623/job/31622189569#step:4:859

```
Traceback (most recent call last):
  File "/__w/spring/spring/build/tests/separate_prod_fin_test.py", line 50, in <module>
    if cluster.launch(pnodes=pnodes, totalNodes=total_nodes, totalProducers=pnodes,
  File "/__w/spring/spring/build/tests/TestHarness/Cluster.py", line 536, in launch
    if not self.bootstrap(launcher, self.biosNode, self.startedNodesCount, prodCount + sharedProducers, totalProducers, pfSetupPolicy, onlyBios, onlySetProds, loadSystemContract, activateIF, biosFinalizer, signatureProviderForNonProducer):
  File "/__w/spring/spring/build/tests/TestHarness/Cluster.py", line 1294, in bootstrap
    trans=biosNode.publishContract(eosioAccount, contractDir, wasmFile, abiFile, waitForTransBlock=True)
  File "/__w/spring/spring/build/tests/TestHarness/transactions.py", line 171, in publishContract
    trans=Utils.runCmdReturnJson(cmd, trace=False, silentErrors=shouldFail)
  File "/__w/spring/spring/build/tests/TestHarness/testUtils.py", line 327, in runCmdReturnJson
    return Utils.runCmdArrReturnJson(cmdArr, trace=trace, silentErrors=silentErrors)
  File "/__w/spring/spring/build/tests/TestHarness/testUtils.py", line 311, in runCmdArrReturnJson
    return Utils.toJson(retStr, trace, silentErrors)
  File "/__w/spring/spring/build/tests/TestHarness/testUtils.py", line 297, in toJson
    raise TypeError(msg)
TypeError: Received empty JSON response
```

linh2931:
Looks like the retrying here not working https://github.com/AntelopeIO/spring/blob/d3d51ff744d86ac097d64e0b715cb958d09adb0e/tests/TestHarness/transactions.py#L164

heifner:
Certainly looks like a bug in trace_api_plugin get_trx_block_number.

The block 125 that originally had the eosio system contract in it was forked out. It was later put into block 137. That block 137 did quickly become final in block 139. 

However, get_trx_block_number keeps reporting that the trx is in block 125.

[...]

PR #966, opened 2024-10-22
[1.0.3] Trace_API: Fix get_transaction_trace endpoint fails to return transaction trace if the initial block including the transaction forks out

linh2931:
Trace_API's `get_transaction_trace` is supposed to return the transaction trace of a transaction. But the current implementation only finds any block whose block number matches the first block proposing the transaction; this block might not contains the transaction due to forking.

This PR
* adds a test reproducing the problem and a test for normal case.
* fixes the problem by not updating `trx_block_num` in `get_trx_block_number` if the block does not include the transaction.

Resolves https://github.com/AntelopeIO/spring/issues/942

### DIFF
--- a/plugins/trace_api_plugin/store_provider.cpp
+++ b/plugins/trace_api_plugin/store_provider.cpp
@@ -136,7 +136,7 @@ namespace eosio::trace_api {
             } else if (std::holds_alternative<lib_entry_v0>(entry)) {
                auto lib = std::get<lib_entry_v0>(entry).lib;
                if (!trx_block_nums.empty() && lib >= *(--trx_block_nums.end())) {
-                  return *(--trx_block_nums.end());
+                  return *(--trx_block_nums.end()); // *(--trx_block_nums.end()) is the block with highest block number which is final
                }
             } else {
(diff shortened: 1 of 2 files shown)


plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -98,6 +98,43 @@ namespace eosio::trace_api {
...
+            has_trx    = (itr != trxs.end());
+         } else {
+            FC_ASSERT( false, "block_trace_v2 should contain transaction_trace_v2 or transaction_trace_v3 transactions" );;
+         }
+      } else {
+         FC_ASSERT( false, "block_trace should be a block_trace_v0, block_trace_v1, or a block_trace_v3" );
+      }
+
+      return has_trx;
+   }
```
spoonincode, line 136: I can't really tell for sure, but would a `visit()` + overloaded lambda be more concise or cleaner?

linh2931: The function is no longer needed.

spoonincode (approved):

ericpassmore:
Note:start
category: System Stability
component: Trace_API
summary: Properly handle case when initial block forks out and return transaction traces.
Note:end

PR #965, opened 2024-10-21
[1.0.4] skip EOS VM interpreter parallel unit_tests and spec_tests on sanitizer builds

spoonincode:
EOS VM interpreter is slow. Sanitizer builds are slow. The two combined is _really_ slow and we've fought CI test timeouts on and off with that combo.

EOS VM interpreter isn't useful for production usage so maybe just go ahead and disable the `unit_test` and `wasm_spec_test` runs with EOS VM interpreter for ASAN & UBSAN builds in CI.

Will resolve #958.

Issue #958, opened 2024-10-21
Test failure: api_part2_unit_test_eos-vm


heifner:
`api_part2_unit_test_eos-vm .........................***Timeout 480.17 sec`

https://github.com/AntelopeIO/spring/actions/runs/11443843504/job/31838074563

heifner:
Resolved by https://github.com/AntelopeIO/spring/pull/1018

PR #967, opened 2024-10-22
Replace `static_cast` of forwarding references with `std::forward`

greg7mdp:
When stepping through eos-vm code, I noticed some incorrect `static_cast` and `std::move` usages, which don't seem to hide bugs because these functions are called with references which are afterwards unused. Still this seems worthwhile fixing imo.

I linked this PR to issue #958, since these changes came to mind when investigating that issue, but this PR doesn't resolve the issue (that would be PR #965 )

### DIFF
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
@@ -30,7 +30,7 @@
 
 extern "C" int arch_prctl(int code, unsigned long* addr);
 
-namespace eosio { namespace chain { namespace eosvmoc {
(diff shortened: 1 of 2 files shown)


heifner (changes requested):
Need an eos-vm PR for the eos-vm changes.

greg7mdp:
> Need an eos-vm PR for the eos-vm changes.

Yes, thanks. https://github.com/AntelopeIO/eos-vm/pull/27

heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Replace `static_cast` of forwarding references with `std::forward`.
Note:end

PR #970, opened 2024-10-23
[1.0.3 -> main] Trace_API: Fix get_transaction_trace endpoint fails to return transaction trace if the initial block including the transaction forks

linh2931:
Forward https://github.com/AntelopeIO/spring/pull/966

Trace_API's `get_transaction_trace` is supposed to return the transaction trace of a transaction. But the current implementation only finds any block whose block number matches the first block proposing the transaction; this block might not contains the transaction due to forking.

This PR
* adds a test reproducing the problem and a test for normal case.
* fixes the problem by not updating `trx_block_num` in `get_trx_block_number` if the block does not include the transaction.

Resolves https://github.com/AntelopeIO/spring/issues/942


heifner (approved):

ericpassmore:
Note:start
category: System Stability
component: Trace_API
summary: Properly handle case when initial block forks out and return transaction traces.
Note:end

PR #976, opened 2024-10-25
Handle exceptions in controller signals

### LINKED ISSUE #946: TEST: Make sure nodeos shuts down on `controller_emit_signal_exception`s
Related to: https://github.com/AntelopeIO/spring/issues/180

Be consistent in how resource failures cause shutdown:
- A exception derived from `controller_emit_signal_exception` should be thrown from the plugin
- Controller should handle calling app().quit()

Add tests that ensure that all of the exceptions derived from `controller_emit_signal_exception` result in shutdown

heifner:
Adds new integration test that thows in all controller signals. It verifies that non-`controller_emit_signal_exception` exceptions are caught and dropped and that `controller_emit_signal_exception` shuts down the node in a state where it can be restarted. `test_control_plugin` updated with `/v1/test_control/throw_on` endpoint for allowing external triggers of exceptions on signals.

Includes a fix for reverting `chain_head` if `accepted_block` throws. This caused nodeos not to be able to restart as the `chain_head` did not match the chainbase state.
Includes a fix for `applied_transaction` throwing and shutting down nodeos.
Includes a fix for `vote_signal` throwing and shutting down nodeos.

Looking for opinions on if this PR should be backported to `release/1.0`

Resolves #946

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -928,6 +928,39 @@ struct controller_impl {
              // Read-write tasks are not being executed.
    };
 
+   /**
+    *  Plugins / observers listening to signals emitted might trigger
+    *  errors and throw exceptions. Unless those exceptions are caught it could impact consensus and/or
+    *  cause a node to fork.
+    *
+    *  If it is ever desirable to let a signal handler bubble an exception out of this method
+    *  a full audit of its uses needs to be undertaken.
+    *
+    */
+   template<typename Signal, typename Arg>
+   void emit( const Signal& s, Arg&& a, const char* file, uint32_t line ) {
+      try {
+         s( std::forward<Arg>( a ));
+      } catch (std::bad_alloc& e) {
+         wlog( "${f}:${l} std::bad_alloc: ${w}", ("f", file)("l", line)("w", e.what()) );
+         throw e;
+      } catch (boost::interprocess::bad_alloc& e) {
+         wlog( "${f}:${l} boost::interprocess::bad alloc: ${w}", ("f", file)("l", line)("w", e.what()) );
+         throw e;
+      } catch ( controller_emit_signal_exception& e ) {
+         wlog( "${f}:${l} controller_emit_signal_exception: ${details}", ("f", file)("l", line)("details", e.to_detail_string()) );
+         wlog( "Shutting down due to controller_emit_signal_exception" );
+         shutdown();
+         throw e;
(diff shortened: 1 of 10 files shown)


libraries/chain/controller.cpp
```diff
@@ -928,6 +928,39 @@ struct controller_impl {
...
+         wlog( "${f}:${l} std::bad_alloc: ${w}", ("f", file)("l", line)("w", e.what()) );
+         throw e;
+      } catch (boost::interprocess::bad_alloc& e) {
+         wlog( "${f}:${l} boost::interprocess::bad alloc: ${w}", ("f", file)("l", line)("w", e.what()) );
+         throw e;
+      } catch ( controller_emit_signal_exception& e ) {
+         wlog( "${f}:${l} controller_emit_signal_exception: ${details}", ("f", file)("l", line)("details", e.to_detail_string()) );
+         wlog( "Shutting down due to controller_emit_signal_exception" );
+         shutdown();
+         throw e;
```
linh2931, line 954: Is this `throw` redundant as `shutdown` has already started?

libraries/chain/controller.cpp
```diff
@@ -928,6 +928,39 @@ struct controller_impl {
...
    };
 
+   /**
+    *  Plugins / observers listening to signals emitted might trigger
+    *  errors and throw exceptions. Unless those exceptions are caught it could impact consensus and/or
+    *  cause a node to fork.
+    *
+    *  If it is ever desirable to let a signal handler bubble an exception out of this method
+    *  a full audit of its uses needs to be undertaken.
+    *
```
linh2931, line 938: Add comments why some exceptions are re-thrown and some are silenced?

linh2931 (commented):
Non-test-related changes in this PR are enhancing exception handling. Seems safe to be put in Rel 1.0.3.

libraries/chain/controller.cpp
heifner: You still need the throw to stop processing and abort block.

libraries/chain/controller.cpp
heifner: Updated comment.

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Handle exceptions in controller signals.
Note:end

Issue #947, opened 2024-10-17
Ensure that plugin recovers when restarted with corrupted SHiP log


bhazzard:
Related to https://github.com/AntelopeIO/spring/issues/180

Add a test that uses a suite of corrupt SHiP logs to test that plugin recovers when restarted in a variety of log corruption scenarios.... and make it pass.

PR #983, opened 2024-10-29
SHiP:  tests for restarting from corrupted log and index files

linh2931:
Will mark it ready for review after issues for failed tests are created (The PR and issues depend on each other).

Added a suite of tests to verify SHiP recovers when restarted in a variety of log and index corruption scenarios.

Resolves https://github.com/AntelopeIO/spring/issues/947


tests/ship_restart_test.py
```diff
@@ -0,0 +1,304 @@
...
+prodNodeId = 0
+shipNodeId = 1
+
+origStateHistoryLog   = ""
+stateHistoryLog       = ""
+origStateHistoryIndex = ""
+stateHistoryIndex     = ""
+
+# Returns True if file1 and file2 contain the same content
+def equalFiles(file1, file2):
```
heifner, line 43: Can you use `Utils.compareFiles` instead?

heifner (changes requested):

tests/ship_restart_test.py
linh2931: Thanks. I did not notice that function. I just need to add an extra argument for file mode (the original version does not work for binary file comparison).

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: SHiP
summary: Added a suite of tests to verify SHiP recovers when restarted in a variety of log and index corruption scenarios.
Note:end

tests/ship_restart_test.py
```diff
@@ -0,0 +1,285 @@
...
+    assert not isRelaunchSuccess, "SHiP node should have failed to relaunch"
+
+try:
+    TestHelper.printSystemInfo("BEGIN")
+
+    cluster.setWalletMgr(walletMgr)
+    Print("Stand up cluster")
+
+    specificExtraNodeosArgs={}
+    specificExtraNodeosArgs[shipNodeId]="--plugin eosio::state_history_plugin --trace-history --chain-state-history --finality-data-history --state-history-stride 200 --plugin eosio::net_api_plugin --plugin eosio::producer_api_plugin"
```
spoonincode, line 62: Does the chain ever make it past block 200 to exercise anything other then the simple "flat log" case? But I suppose there is probably not much that can be done for a corrupted non-head split log anyways -- fixing it just makes a hole that isn't allowed.

I do wonder if we should do some additional corruption tests with pruned logs. But those are even less repairable.

spoonincode (approved):

tests/ship_restart_test.py
linh2931: > Does the chain ever make it past block 200 to exercise anything other then the simple "flat log" case? But I suppose there is probably not much that can be done for a corrupted non-head split log anyways -- fixing it just makes a hole that isn't allowed.
> 
> I do wonder if we should do some additional corruption tests with pruned logs. But those are even less repairable.

Thanks.  I created https://github.com/AntelopeIO/spring/issues/1011 to track additional tests for corrupted pruned and split tests. I think they should be done as a part of the bigger initiative to revisit how corrupted logs are handled (repairing at the startup or offline via `spring-util`.

I am merging the PR in current form so we have some basic tests.

PR #993, opened 2024-10-30
Support interrupt transaction

### LINKED ISSUE #876: Handle signals on separate thread
Currently during application startup `application_base::startup`, `startup_thread` is used to handle signals until all `plugin->startup()` has completed. This allows for example, ctrl-c to interrupt replay of the blocklog during `chain_plugin` startup.

Once startup is complete, the signal handling is moved to the main thread. This means that a ctrl-c or SIGTERM is not processed until the `application_base::exec()` calls `io_serv.poll()` on the main thread. If the main thread is busy, for example applying a large range of blocks during syncing, then shutdown has to wait until the current task completes. It requires the main thread can call `poll()` to process the signal to set the `is_quitting` atomic flag. 

Instead we could continue to use a seperate thread for signal processing (except for SIGHUP which we want on the main thread since it resets all the loggers). This thread then can immediately set the `is_quitting` flag allowing the application to immediately exit.

### LINKED ISSUE #985: OC Whitelist: ctrl-c interrupt of executing trx
Add ctrl-c interrupt of executing trx. Execute queued compiles up to some limit (maybe 10 seconds) before exiting. On restart the contract would be compiled with oc and the block should be able to be validated.

heifner:
Add support for interrupting a transaction in `apply_block` via `ctrl-c` or `SIGINT/SIGTERM/SIGPIPE`.

Requires https://github.com/AntelopeIO/appbase/pull/36

Resolves #985 
Resolves #876


libraries/chain/platform_timer_posix.cpp
```diff
@@ -71,6 +71,11 @@ void platform_timer::start(fc::time_point tp) {
    }
 }
 
+void platform_timer::expire_now() {
+   expired = 1;
+   call_expiration_callback();
+}
+
```
spoonincode, line 78: Feels like this may introduce a possibility of the expiration callback being called twice? I am not sure how much that matters. Do we need to test_and_set both here and in the normal expiry path to guard against that?

heifner: I was under the impression that it didn't matter to be called twice. Not sure how to exactly verify that. Maybe safest to do the test_and_set.

libraries/chain/platform_timer_asio_fallback.cpp
```diff
@@ -57,39 +57,48 @@ platform_timer::~platform_timer() {
...
       my->timer->expires_after(std::chrono::microseconds(x.count()));
       my->timer->async_wait([this](const boost::system::error_code& ec) {
          if(ec)
             return;
-         expired = 1;
-         call_expiration_callback();
+         bool expected = false;
+         if (expired.compare_exchange_strong(expected, true)) {
+            call_expiration_callback();
+         }
```
greg7mdp, line 84: call `expire_now()` here?

libraries/chain/platform_timer_asio_fallback.cpp
```diff
@@ -57,39 +57,48 @@ platform_timer::~platform_timer() {
...
 #if 0
       std::promise<void> p;
       auto f = p.get_future();
       checktime_ios->post([&p,this]() {
-         expired = 0;
+         expired = false;
          p.set_value();
       });
       f.get();
 #endif
```
greg7mdp, line 75: Should we remove this dead code?

libraries/chain/platform_timer_kqueue.cpp
```diff
@@ -58,8 +58,10 @@ platform_timer::platform_timer() {
 
             if(c == 1 && anEvent.filter == EVFILT_TIMER) {
                platform_timer* self = (platform_timer*)anEvent.udata;
-               self->expired = 1;
-               self->call_expiration_callback();
+               bool expected = false;
+               if (self->expired.compare_exchange_strong(expected, true)) {
+                  self->call_expiration_callback();
+               }
```
greg7mdp, line 64: call `expire_now()` here?

libraries/chain/platform_timer_posix.cpp
```diff
@@ -19,18 +21,20 @@ struct platform_timer::impl {
 
    static void sig_handler(int, siginfo_t* si, void*) {
       platform_timer* self = (platform_timer*)si->si_value.sival_ptr;
-      self->expired = 1;
-      self->call_expiration_callback();
+      bool expected = false;
+      if (self->expired.compare_exchange_strong(expected, true)) {
+         self->call_expiration_callback();
+      }
```
greg7mdp, line 27:
```suggestion
      self->expire_now();
```

libraries/chain/transaction_context.cpp
```diff
@@ -459,7 +459,10 @@ namespace eosio::chain {
          return;
 
       auto now = fc::time_point::now();
-      if( explicit_billed_cpu_time || deadline_exception_code == deadline_exception::code_value ) {
+      if (explicit_billed_cpu_time && block_deadline > now) {
+         EOS_THROW( interrupt_exception, "interrupt signaled, ran ${bt}us, start ${s}",
+                    ("bt", now - pseudo_start)("s", start) );
+      } else if( explicit_billed_cpu_time || deadline_exception_code == deadline_exception::code_value ) {
          EOS_THROW( deadline_exception, "deadline exceeded ${billing_timer}us",
                      ("billing_timer", now - pseudo_start)("now", now)("deadline", _deadline)("start", start) );
```
greg7mdp, line 467:
```suggestion
                     ("billing_timer", now - pseudo_start) );
```

libraries/chain/controller.cpp
```diff
@@ -3780,6 +3783,9 @@ struct controller_impl {
                }
             }
 
+            applying_block = true;
+            auto apply = fc::make_scoped_exit([&](){ applying_block = false; });
```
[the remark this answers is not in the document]
heifner: Because `appling_block` is `std::atomic` which does not provide a move constructor.

libraries/chain/transaction_context.cpp
heifner: Why? That data is still in the exception if you print out the details.

greg7mdp: I didn't see (and still don't) where "now", "deadline" are "start" are referenced, so I thought they were unused.

heifner: They are not referenced in the string. However, non-referenced captured values are printed when details of exceptions are requested.

See https://github.com/AntelopeIO/spring/blob/bb4d159218eb5fb6603f0c82702800c7bd4dae90/libraries/libfc/src/exception.cpp#L174-L174

greg7mdp: Oh, cool, I didn't know that!

greg7mdp (approved):

libraries/chain/platform_timer_posix.cpp
heifner: Done.

libraries/chain/platform_timer_posix.cpp
```diff
@@ -55,19 +56,26 @@ platform_timer::~platform_timer() {
...
+      expired = false;
       if(timer_settime(my->timerid, 0, &enable, NULL) != 0)
-         expired = 1;
+         expired = true;
+   }
+}
+
+void platform_timer::expire_now() {
+   bool expected = false;
+   if (expired.compare_exchange_strong(expected, true)) {
```
spoonincode, line 77: I guess you can save a line of code using atomic_flag, but it doesn't really matter

spoonincode (approved):

ericpassmore:
Support interrupt transaction
Note:start
category: System Stability
component: Internal
summary: Support interrupt transaction, brings along app base submodule update.
Note:end

PR #1016, opened 2024-11-06
workaround cdt's small const memcpy host function calls in EOS VM OC

spoonincode:
AntelopeIO/cdt#102 is a long time thorn and while obviously it would be ideal to fix cdt to not generate these host function calls in the first place, it's not clear when that will happen and regardless there will be many contracts that may never be recompiled and years of historical blocks that will have these calls forever.

This implements a workaround in EOS VM OC that identifies small constant size memcpy host function calls during code generation and replaces them with simple memory loads and stores plus a small call to a native function that verifies the memcpy ranges do not overlap as required by protocol rules.

As a simple example, a contract such as,
```wat
(module
 (import "env" "memcpy" (func $memcpy (param i32 i32 i32) (result i32)))
 (export "apply" (func $apply))
 (memory 1)
 (func $apply (param $0 i64) (param $1 i64) (param $2 i64)
    (drop (call $memcpy (i32.const 4) (i32.const 200) (i32.const 8)))
 )
)
```
Will generate machine code (annotated by me)
```asm
    pushq	%rax
    decl	%gs:-18888     ;;decrement depth count and jump if zero
    je  	51
    movq	%gs:200, %rax  ;;load 8 bytes from address 200 in to register
    movq	%rax, %gs:4    ;;store 8 bytes to address 4 from register
    movl	$4, %edi       ;;prepare the 3 parameters to check_memcpy_params
    movl	$200, %esi
    movl	$8, %edx
    callq	*%gs:-21056    ;;call check_memcpy_params
    incl	%gs:-18888     ;;increment depth count
    popq	%rax
    retq
    callq	*%gs:-18992    ;;call depth_assert
```
(in practice the source and destination addresses are typically not constants)

For some tested block ranges starting at 346025446, 348083350, 371435719, and 396859576 replay performance increases within a range of 3.5% to 6%. This is a bit lower than I expected -- I was expecting consistently north of 5% -- the reason for missing my expectation is that there are still many memcpy host function calls that are not optimized out, and the overlap check can still consume significant amount of CPU.

A run of the LLVM exhaustive workflow is at,
https://github.com/AntelopeIO/spring/actions/runs/11711214896

### DIFF
--- a/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic.hpp
+++ b/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic.hpp
@@ -25,4 +25,7 @@ using intrinsic_map_t = std::map<std::string, intrinsic_entry>;
 
 const intrinsic_map_t& get_intrinsic_map();
 
+static constexpr unsigned minimum_const_memcpy_intrinsic_to_optimize = 1;
+static constexpr unsigned maximum_const_memcpy_intrinsic_to_optimize = 128;
+
 }}}
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/LLVMEmitIR.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/LLVMEmitIR.cpp
@@ -697,12 +697,14 @@ namespace LLVMJIT
 			llvm::Value* callee;
 			const FunctionType* calleeType;
 			bool isExit = false;
+			bool isMemcpy = false;
 			if(imm.functionIndex < moduleContext.importedFunctionOffsets.size())
 			{
 				calleeType = module.types[module.functions.imports[imm.functionIndex].type.index];
 				llvm::Value* ic = irBuilder.CreateLoad( emitLiteralPointer((void*)(OFFSET_OF_FIRST_INTRINSIC-moduleContext.importedFunctionOffsets[imm.functionIndex]*8), llvmI64Type->getPointerTo(256)) );
 				callee = irBuilder.CreateIntToPtr(ic, asLLVMType(calleeType)->getPointerTo());
 				isExit = module.functions.imports[imm.functionIndex].moduleName == "env" && module.functions.imports[imm.functionIndex].exportName == "eosio_exit";
+				isMemcpy = module.functions.imports[imm.functionIndex].moduleName == "env" && module.functions.imports[imm.functionIndex].exportName == "memcpy";
 			}
 			else
 			{
@@ -714,6 +716,28 @@ namespace LLVMJIT
 			// Pop the call arguments from the operand stack.
 			auto llvmArgs = (llvm::Value**)alloca(sizeof(llvm::Value*) * calleeType->parameters.size());
 			popMultiple(llvmArgs,calleeType->parameters.size());
+
+			//convert small constant sized memcpy host function calls to a load+store (plus small call to validate non-overlap rule)
+			if(isMemcpy) {
+				assert(calleeType->parameters.size() == 3);
+				if(llvm::ConstantInt* const_memcpy_sz = llvm::dyn_cast<llvm::ConstantInt>(llvmArgs[2]);
+				     const_memcpy_sz &&
+					  const_memcpy_sz->getZExtValue() >= minimum_const_memcpy_intrinsic_to_optimize &&
+					  const_memcpy_sz->getZExtValue() <= maximum_const_memcpy_intrinsic_to_optimize) {
+					const unsigned sz_value = const_memcpy_sz->getZExtValue();
+					llvm::IntegerType* type_of_memcpy_width = llvm::Type::getIntNTy(context, sz_value*8);
+
+					llvm::Value* load_pointer = coerceByteIndexToPointer(llvmArgs[1],0,type_of_memcpy_width);
+					llvm::Value* store_pointer = coerceByteIndexToPointer(llvmArgs[0],0,type_of_memcpy_width);
+					irBuilder.CreateStore(irBuilder.CreateLoad(load_pointer), store_pointer, true);
+
+					emitRuntimeIntrinsic("eosvmoc_internal.check_memcpy_params",
+					                     FunctionType::get(ResultType::none,{ValueType::i32,ValueType::i32,ValueType::i32}),
+						                  {llvmArgs[0],llvmArgs[1],llvmArgs[2]});
+					push(llvmArgs[0]);
+					return;
+				}
+			}
 
 			// Call the function.
 			auto result = createCall(callee,llvm::ArrayRef<llvm::Value*>(llvmArgs,calleeType->parameters.size()));
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
@@ -122,6 +122,23 @@ DEFINE_EOSVMOC_TRAP_INTRINSIC(eosvmoc_internal,unreachable) {
    throw_internal_exception(wasm_execution_error(FC_LOG_MESSAGE(error, "Unreachable reached")));
 }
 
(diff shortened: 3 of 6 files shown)


libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic.hpp
```diff
@@ -25,4 +25,7 @@ using intrinsic_map_t = std::map<std::string, intrinsic_entry>;
 
 const intrinsic_map_t& get_intrinsic_map();
 
+static constexpr unsigned minimum_const_memcpy_intrinsic_to_optimize = 1;
+static constexpr unsigned maximum_const_memcpy_intrinsic_to_optimize = 128;
```
spoonincode, line 29: 128 is a bit of a dice roll. I don't have data to know where the exact threshold is between this local copy being faster vs slower than the full call off to memcpy. I wouldn't want to go higher than this anyways though, since it bloats the code a little more.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic.hpp
```diff
@@ -25,4 +25,7 @@ using intrinsic_map_t = std::map<std::string, intrinsic_entry>;
...
+
```
spoonincode, line 30: This is a weird place for this, but header files for OC are a little weird because of some parts still are compiled with C++17 -- need to place these in a header file that is consumed by the C++17 code.

libraries/chain/webassembly/runtimes/eos-vm-oc/LLVMEmitIR.cpp
```diff
@@ -715,6 +717,28 @@ namespace LLVMJIT
...
 
+			//convert small constant sized memcpy host function calls to a load+store (plus small call to validate non-overlap rule)
+			if(isMemcpy) {
+				assert(calleeType->parameters.size() == 3);
+				if(llvm::ConstantInt* const_memcpy_sz = llvm::dyn_cast<llvm::ConstantInt>(llvmArgs[2]);
+				     const_memcpy_sz &&
+					  const_memcpy_sz->getZExtValue() >= minimum_const_memcpy_intrinsic_to_optimize &&
+					  const_memcpy_sz->getZExtValue() <= maximum_const_memcpy_intrinsic_to_optimize) {
+					const unsigned sz_value = const_memcpy_sz->getZExtValue();
+					llvm::IntegerType* type_of_memcpy_width = llvm::Type::getIntNTy(context, sz_value*8);
```
spoonincode, line 728: The use of a non-power-of-2 type size here is rather unorthodox, though one would have to imagine something like `_BitInt` from C23 would do exactly this. Unit tests exhaustively explore every size (1 through 128). An oddball size does generate mostly expected machine code (loads and stores are maybe not in "natural" order one would expect but that doesn't matter of course), for example for size 13:
```asm
		                pushq	%rax
		                decl	%gs:-18888
		                je	83
		                movl	%gs:208, %eax  ;;load 4 bytes (208-211)
		                movq	%gs:200, %rcx  ;;load 8 bytes (200-207)
		                movb	%gs:212, %dl   ;;load 1 byte  (212)
		                movb	%dl, %gs:16    ;;store 1 byte  (16)
		                movq	%rcx, %gs:4    ;;store 8 bytes (4-11)
		                movl	%eax, %gs:12   ;;store 4 bytes (12-15)
		                movl	$4, %edi
		                movl	$200, %esi
		                movl	$13, %edx
		                callq	*%gs:-21056
		                incl	%gs:-18888
		                popq	%rax
		                retq
		                callq	*%gs:-18992
```

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
```diff
@@ -102,25 +103,42 @@ static void throw_internal_exception(const char* const s) {
...
 }
 
+static void eos_vm_oc_check_memcpy_params(int32_t dest, int32_t src, int32_t length) {
+   //make sure dest & src are zexted when converted from signed 32-bit to signed ptrdiff_t; length should always be small but do it too
+   const unsigned udest = dest;
+   const unsigned usrc = src;
+   const unsigned ulength = length;
+
+   //this must remain the same behavior as the memcpy host function
+   if((size_t)(std::abs((ptrdiff_t)udest - (ptrdiff_t)usrc)) >= ulength)
```
spoonincode, line 132: This check compiles down to about ~11 instructions but can still consume a substantial amount of CPU (beyond 1% in some cases). Perhaps something more clever can shave off another 0.5%.

linh2931: > This check compiles down to about ~11 instructions but can still consume a substantial amount of CPU (beyond 1% in some cases). Perhaps something more clever can shave off another 0.5%.

Did you try plain `if` comparisons without using `std::abs`?

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: EOS EVM
summary: Improve EOS EVM by working around CDT's memcpy function. 
Note:end

heifner (approved):

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
spoonincode: Both of these interesting suggestions and worth trying. Though, I'd suggest we keep the existing impl initially as it's most like what already exists (and has already been performance tested at the moment).

I'm also curious what would happen if the check was moved completely inline. The obviously bad aspect of this is that it's going to put a lot of pressure on the branch prediction cache. But.. there is a non-trivial amount of work just to get the values in to the right registers to pass off to a standard ABI function; then again maybe these get pipelined well. Plenty to experiment with here.

spoonincode:
exhaustive LLVM run here,
https://github.com/AntelopeIO/spring/actions/runs/13105241743

chicken dance pending

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/eos-vm-oc.hpp
```diff
@@ -44,7 +44,7 @@ enum eosvmoc_exitcode : int {
    EOSVMOC_EXIT_EXCEPTION
 };
 
-static constexpr uint8_t current_codegen_version = 1;
+static constexpr uint8_t current_codegen_version = 2;
```
spoonincode, line 47: This small change will effectively purge a pre-1.1 cache since on load,
https://github.com/AntelopeIO/spring/blob/39844283ed44e7d8764ac5ba25bda4cc45d95584/libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp#L345-L354

libraries/libfc/include/fc/crypto/sha256.hpp
```diff
@@ -147,5 +147,12 @@ namespace boost
        }
     };
 }
+
+namespace fc {
+   inline size_t hash_value(const fc::sha256& s) {
+      return boost::hash<fc::sha256>()(s);
```
greg7mdp, line 153: I would have thought that the version taking one of the `uint64_t` (like `return v._hash[3];`) would be just as good as rehashing the four values.

greg7mdp (approved):

spoonincode:
Replay test completed successfully

libraries/libfc/include/fc/crypto/sha256.hpp
spoonincode: Isn't this just calling that version? It appears to be due to `template<> struct hash<fc::sha256>` not being `const`, whereas this one is `const`. afaict both const & non-const are needed. I suggest we clean it up in a separate PR

greg7mdp: Yes, sorry, I missed that we defined it for the boost namespace.

greg7mdp (approved):

PR #1024, opened 2024-11-12
add unit test for `find_secondary` host functions

### LINKED ISSUE #1019: unit test exercising secondary key find
afaict there is no test for this; and cdt never touches it either. Really ought to have a basic test for it.

spoonincode:
Which key should the `find_secondary` host function return when there are duplicate keys? cdt avoids this question entirely by never using these host functions and instead always doing a `lowerbound` (it's possible the `find_secondary` host function has never once been used on EOS! Unless a rare contract built with something other then cdt used it of course)

The host function ultimately performs a "partial key lookup" on the table (this then goes through to chainbase's `find()`)
https://github.com/AntelopeIO/spring/blob/f871bf2c6356cc6273473aa2ab0ac8ddcf7f4f21/libraries/chain/include/eosio/chain/apply_context.hpp#L291-L297

In leap 3.1+ & spring, the underlying data structure (boost intrusive avltree) does specify that `find()` is treated like `lowerbound()`. And even looking through the code makes it clear that `find()` simply calls `lowerbound()`.

EOSIO 2.0 and earlier it is not so clear: the documentation for the data structure in use (boost multindex directly) simply says it finds "a" matching key. But eyeballing the code suggests it does effectively behave like lowerbound. And empirically the test code still passes on 2.0.

This test confirms that the `find_secondary` host functions behave like lowerbound: lowest primary key is returned.

Closes #1019


unittests/test-contracts/db_find_secondary_test/db_find_secondary_test.wast
```diff
@@ -0,0 +1,120 @@
...
+      (if (i64.ne (i64.load (i32.const 1024)) (i64.const 1)) (then unreachable))
+      (drop (call $db_idx_long_double_find_secondary (local.get $receiver) (local.get $receiver) (local.get $receiver) (i32.const 48) (i32.const 1024)))
+      (if (i64.ne (i64.load (i32.const 1024)) (i64.const 1)) (then unreachable))
+   )
+
+   ;;secondary keys to be used
+   (data (i32.const 16) "\00\00\00\00\01\02\03\04AbcdefghABCDEFGH01234567")
+   (data (i32.const 48) "\00\00\00\00\01\02\03\05abcdefghABCDEFGH99999999")
+   (data (i32.const 80) "\00\00\00\00\01\02\03\06aBcdefghABCDEFGH00000000")
+)
```
spoonincode, line 120: hm maybe I should have just done this in a .cpp contract, but I would have had to use the `internal_use_do_not_use` functions which I wanted to avoid.

unittests/database_tests.cpp
```diff
@@ -87,4 +88,15 @@ BOOST_AUTO_TEST_SUITE(database_tests)
       } FC_LOG_AND_RETHROW()
    }
 
+   //test that find_secondary_key behaves like lowerbound
+   BOOST_AUTO_TEST_CASE_TEMPLATE( find_seconary_key, T, validating_testers ) {
```
heifner, line 92: Seems like `api_tests.cpp` `db_tests` section would be a better place. Even better would to pull out all multi index db tests into a separate file and include this.

unittests/test-contracts/db_find_secondary_test/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/db_find_secondary_test.wasm ${CMAKE_CURRENT_BINARY_DIR}/db_find_secondary_test.wasm COPYONLY )
+configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/db_find_secondary_test.abi  ${CMAKE_CURRENT_BINARY_DIR}/db_find_secondary_test.abi  COPYONLY )
```
heifner, line 2: Can we set this up like the other test directories where it compiles if `EOSIO_COMPILE_TEST_CONTRACTS` is set?

heifner (changes requested):

unittests/test-contracts/db_find_secondary_test/CMakeLists.txt
spoonincode: Yeah that crossed my mind... I was sloppy about that over here on this one too
https://github.com/AntelopeIO/spring/blob/f871bf2c6356cc6273473aa2ab0ac8ddcf7f4f21/unittests/contracts/jumborow/CMakeLists.txt#L1-L2

spoonincode: Will do what's in #1026 here if that one is approved

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Host Functions
summary: Add unit test for `find_secondary` host functions.
Note:end

PR #1031, opened 2024-11-14
Remove fc dynamic_rethrow_exception

### LINKED ISSUE #980: fc exception handling
`fc::exception` has support for `dynamic_rethrow_exception` which is documented as:
```
   /**
    * Rethrows the exception restoring the proper type based upon
    * the error code.  This is used to propagate exception types
    * across conversions to/from JSON
    */
```
We use this in a few places:
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/plugins/http_plugin/include/eosio/http_plugin/macros.hpp#L16-L16
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/chain/snapshot_scheduler.cpp#L136-L136

This requires:
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/libfc/src/exception.cpp#L10-L28

Which we do NOT do for any of the chain exceptions in: 
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/chain/include/eosio/chain/exceptions.hpp#L185-L185

Note we add an `error_code` to `chain_exception` which all chain exceptions derive from. This is used here:
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/chain/controller.cpp#L5858-L5866

But even if we did register them, they would slice off the `error_code` on rethrow:
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/libfc/include/fc/exception/exception.hpp#L221-L224

When exceptions are placed `action_trace` and `transaction_trace` they are sliced to a `fc::exception` losing this `error_code`:
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/chain/include/eosio/chain/trace.hpp#L45-L45

When an exception is placed in the `action_trace` or `transaction_trace` and is rethrown, it is always rethrown as a `fc::exception`. As commented here: 
https://github.com/AntelopeIO/spring/blob/8169da96e5771d820a3e80f3ade60b4054005538/libraries/testing/tester.cpp#L485-L486

Changing `action_trace` and `transaction_trace` to use an `fc::exception_ptr` is an API breaking change.
Changing `fc::exception` to have an `error_code` is also an API breaking change. Since this is an addition, maybe this would cause no harm. Not sure if that affects consensus.

heifner:
Remove `fc::exception::dynamic_rethrow_exception`. To work correctly `dynamic_rethrow_exception` requires registration via `FC_REGISTER_EXCEPTIONS` which was never done for any of the `eosio::chain` defined exceptions.

Since this feature never has worked as expected for `eosio::chain` defined exceptions, remove it to avoid confusion.

Resolves #980

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/macros.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/macros.hpp
@@ -13,7 +13,7 @@
            (const chain::next_function_variant<call_result>& result) mutable {                                  \
               if (std::holds_alternative<fc::exception_ptr>(result)) {                                          \
                  try {                                                                                          \
-                    std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();                           \
+                    throw *std::get<fc::exception_ptr>(result);                                                 \
                  } catch (...) {                                                                                \
                     http_plugin::handle_exception(#api_name, #call_name, body, cb);                             \
                  }                                                                                              \
@@ -28,7 +28,7 @@
                     chain::t_or_exception<call_result> result = http_fwd();                                     \
                     if (std::holds_alternative<fc::exception_ptr>(result)) {                                    \
                        try {                                                                                    \
-                          std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();                     \
+                          throw *std::get<fc::exception_ptr>(result);                                           \
                        } catch (...) {                                                                          \
                           http_plugin::handle_exception(#api_name, #call_name, body, cb);                       \
                        }                                                                                        \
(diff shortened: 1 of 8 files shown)


plugins/http_plugin/include/eosio/http_plugin/macros.hpp
```diff
@@ -13,7 +13,7 @@
            (const chain::next_function_variant<call_result>& result) mutable {                                  \
               if (std::holds_alternative<fc::exception_ptr>(result)) {                                          \
```
greg7mdp, line 14: Not related to your change, but instead of these `std::holds_alternative<.` and `std::get<>`, it would be cleaner to use `std::visit` with lambdas.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Improved support for `eosio::chain` defined exceptions by removing dynamic-rethrow-exception
Note:end

Issue #1015, opened 2024-11-06
Test failure: nodeos_startup_catchup_if_lr_test


heifner:
https://github.com/AntelopeIO/spring/actions/runs/11707186905/job/32607464282

```
2024-11-06T16:28:25.027786 Node10 has 509 unlinkable_block in /__w/spring/spring/build/TestLogs/nodeos_startup_catchup8/node_10
2024-11-06T16:28:25.027940  ERROR: Node10 has 509 which is more than the configured allowed 500 unlinkable blocks: /__w/spring/spring/build/TestLogs/nodeos_startup_catchup8/node_10.
```

This is different than #998

bhazzard:
When looking into this test, consider whether any changes would need to be backported to 1.0.x.

linh2931:
Two questions, @heifner:
1. `numUnlinkable` is the total number of unlinkable blocks in all the logging files. Should it be the number of unlinkable blocks in only the latest log file?
2. For the test to be less flaky, should we check whether or not the unlinkable block numbers are consecutively incremented (no gaps), instead of checking against a fixed number of unlinkable like 500?

linh2931:
> If you have verified that the unlinkable blocks are expected, then I think counting consecutive blocks less than fetch-span as one is a good choice. If you make that change, seems like option 1 doesn't matter.

Thanks Kevin. Will do that as discussed in the meeting and fix it in 1.0.x.

linh2931:
The test is broken in `Rel 1.0.3` and flaky in `main`. It counts the number of `unlinkable_block` occurrences in logging files and make sure it is not greater than `500`.

But in `Rel 1.0.3`, `unlinkable_block` (part of `unlinkable_block_exception`) shows up 3 times per unlinkable block, for example:
```
warn  2024-11-14T13:48:06.038 nodeos    controller.cpp:4334           push_block           ] 3030001 unlinkable_block_exception: Unlinkable block
info  2024-11-14T13:48:06.038 nodeos    producer_plugin.cpp:938       operator()           ] Exception on block 130: 3030001 unlinkable_block_exception: Unlinkable block
info  2024-11-14T13:48:06.038 nodeos    net_plugin.cpp:3870           process_signed_block ] unlinkable_block_exception connection - 1: #130 1d74c43582d10251...: Unlinkable block (3030001)
```
While in `main`, only one  occurrence of `unlinkable_block` per unlinkable block:
[...]

PR #1033, opened 2024-11-14
[1.0.4] Fix nodeos_startup_catchup test

linh2931:
The test is broken in `Rel 1.0.3` and flaky in `main`. The counts the number of `unlinkable_block` occurrences in logging files and make sure it is not greater than `500`.

But in `Rel 1.0.3`, `unlinkable_block` (part of `unlinkable_block_exception`) shows up 3 times per unlinkable block, for example:
```
warn  2024-11-14T13:48:06.038 nodeos    controller.cpp:4334           push_block           ] 3030001 unlinkable_block_exception: Unlinkable block
info  2024-11-14T13:48:06.038 nodeos    producer_plugin.cpp:938       operator()           ] Exception on block 130: 3030001 unlinkable_block_exception: Unlinkable block
info  2024-11-14T13:48:06.038 nodeos    net_plugin.cpp:3870           process_signed_block ] unlinkable_block_exception connection - 1: #130 1d74c43582d10251...: Unlinkable block (3030001)
```

This PR searches for `unlinkable_block_exception connection - \d+: #(\d+)`, extracts the unlinkable block number, and verifies that
1. unlinkable blocks are consecutive
2. the number of unlinkable blocks is less than sync fetch span

Resolves https://github.com/AntelopeIO/spring/issues/1015

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -705,15 +705,39 @@ class Node(Transactions):
                         lines.append(line)
         return lines
 
-    def countInLog(self, searchStr) -> int:
+    # Verfify that in during synching, unlinkable blocks are expected if
+    # 1. they are consecutive,
+    # 2. the number of unlinkable blocks is less than sync fetch span
+    def verifyUnlinkableBlocksExpected(self, syncFetchSpan) -> bool:
         dataDir=Utils.getNodeDataDir(self.nodeId)
         files=Node.findStderrFiles(dataDir)
-        count = 0
+
+        # A sample of unique line of unlinkable_block in logging file looks like:
+        # info  2024-11-14T13:48:06.038 nodeos    net_plugin.cpp:3870           process_signed_block ] unlinkable_block_exception connection - 1: #130 1d74c43582d10251...: Unlinkable block (3030001)
+        pattern = re.compile(r"unlinkable_block_exception connection - \d+: #(\d+)")
+
         for file in files:
+            blocks = []
             with open(file, 'r') as f:
-                contents = f.read()
-                count += contents.count(searchStr)
-        return count
+                for line in f:
+                    match = pattern.search(line)
+                    if match:
+                        try:
+                            blockNum = int(match.group(1))
+                            blocks.append(blockNum)
+                        except ValueError:
+                            Utils.Print(f"unlinkable block number cannot be converted into integer: in {line.strip()} of {f}")
+                            return False
+                # Check if the unlinkable blocks are consecutive
+                for i in range(1, len(blocks)):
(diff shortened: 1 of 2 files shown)


tests/TestHarness/Node.py
```diff
@@ -705,15 +705,43 @@ def linesInLog(self, searchStr):
...
+                    if blocks[i] == blocks[i - 1] or blocks[i] == blocks[i - 1] + 1: # look for consecutive blocks, including duplicate
+                        if blocks[i] == blocks[i - 1] + 1: # excluding duplicate
+                            ++numConsecutiveUnlinkableBlocks
+                    else: # start a new group of consecutive blocks
+                        if numConsecutiveUnlinkableBlocks > syncFetchSpan:
+                            Utils.Print(f"the number of a group of unlinkable blocks {numConsecutiveUnlinkableBlocks} greater than syncFetchSpan {syncFetchSpan} in {f}")
+                            return False
+                        numConsecutiveUnlinkableBlocks = 1
+        if numConsecutiveUnlinkableBlocks > syncFetchSpan:
+            Utils.Print(f"the number of a group of unlinkable blocks {numConsecutiveUnlinkableBlocks} greater than syncFetchSpan {syncFetchSpan} in {f}")
```
greg7mdp, line 741: This seems like a somewhat fragile error detection mechanism, because if the order was not as expected (suppose the log shows unlinkable blocks in alternate order  2, 4, 3, 5, 7, 6, 8, 7, ...), you could have a chain of consecutive unlinkable blocks greater than sync fetch span and not detect it. I am not sure whether it is a concern or not.

If this was deemed useful to fix, I think I would use two dictionaries, one with the last block number of every sequence as a key and the length of the sequence as the mapped value, and the other same with the key being the start block # of the sequence.

greg7mdp: Or simpler (and slower but who cares) create a list of all unlinkable blocks, sort it, and count longest sequence.

linh2931:
> Can you explain why the number of unlinkable blocks is bounded by `sync fetch span` in order not to be an error? It doesn't seem obvious to me.

The idea is, it is inevitable to have some unlinkable blocks; the worst case is all blocks in a batch are temporarily unlinkable.

tests/TestHarness/Node.py
linh2931: https://github.com/AntelopeIO/spring/pull/1033/commits/150915f88e32f1b1448c489f9e5c99027927d4c2

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Fix `nodeos_startup_catchup` tests.
Note:end

PR #1034, opened 2024-11-15
Do not reject future blocks on speculative nodes

### LINKED ISSUE #239: Block from future is ignored
On the spring beta1 testnet, local node synced up:

```
info  2024-06-05T12:51:28.342 nodeos.be net_plugin.cpp:3902           process_signed_block ] block_validate_exception connection - 6: #4782125 4af601b35d70b371...: Block is from the future (3030007)
received a block from the future, ignoring it: 0048f82d4af601b35d70b371b28cb24dad93753400183c37f5023f70d586e68b
```
Sends handshake and then gets the block again and it applies it with no issue. Seems we are a bit too aggressive about what we consider the future.

heifner:
On speculative, non-block-producer, nodes do not reject blocks from the future. This allows a speculative node to have a clock skew without it interfering with maintaining synchronization. Before, this change a clock skew of 7 seconds would cause a speculative node to reject a block. 

Note block producers still need to keep their clocks synchronized so they produce on time. Block producers will continue to reject blocks with timestamps >= 7 seconds in the future.

Resolves #239

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -829,6 +829,10 @@ public:
       return _implicit_pause_vote_tracker.check_pause_status(fc::time_point::now()).should_pause();
    }
 
+   bool is_configured_producer() const {
+      return !_producers.empty();
+   }
+
    void on_accepted_block(const signed_block_ptr& block, const block_id_type& id) {
       auto& chain  = chain_plug->chain();
       auto  before = _unapplied_transactions.size();
@@ -1312,7 +1316,7 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
 
    chain::controller& chain = chain_plug->chain();
 
-   chain.set_producer_node(!_producers.empty());
+   chain.set_producer_node(is_configured_producer());
 
    if (options.count("signature-provider")) {
       const std::vector<std::string> key_spec_pairs = options["signature-provider"].as<std::vector<std::string>>();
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1963,8 +1963,10 @@ producer_plugin::get_unapplied_transactions_result producer_plugin::get_unapplie
 
 block_timestamp_type producer_plugin_impl::calculate_pending_block_time() const {
    const chain::controller& chain = chain_plug->chain();
-   const fc::time_point     now   = fc::time_point::now();
-   const fc::time_point     base  = std::max<fc::time_point>(now, chain.head().block_time());
+   // on speculative nodes, always use next block time. On producers, honor current clock time
+   const fc::time_point base  = _producers.empty()
```
linh2931, line 1967: Not sure whether or not `chain.is_producer_node()` more clear than `_producers.empty()`.

heifner: The `producer_plugin` uses `_producers.empty()` in a number of places for this concept. I could wrap that in a `producer_plugin` method if desired, but I don't think we should reach into controller for that.

linh2931: It is better to wrap that in a method in producer_plugin and modify other places. That makes the concept more clearer. But the change is not necessary for the PR.

heifner: https://github.com/AntelopeIO/spring/pull/1034/commits/9f3c603e1ab24730beb764d63aca8b8ff57729eb

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Enhance robustness of speculative block mode by allowing for future blocks. Future blocks commonly from clock-skew.
Note:end

PR #1036, opened 2024-11-15
Support speculative trx execution in irreversible mode

### LINKED ISSUE #1030: Allow speculative trx execution in irreversible mode
With the introduction of Savanna instant finality, there is now no longer a reason to prevent speculative trx execution on irreversible nodes. Allowing speculative execution on irreversible nodes would allow a local node to be used for multiple purposes. 

Particularly exchanges may find this useful. It would allow them to run in irreversible mode, but still submit their trxs to the chain without the need to run a separate node or use a public API.

heifner:
With Savanna instant finality, there is now no longer a good reason to prevent speculative trx execution on irreversible nodes; so remove the restriction.

Changed the distributed transaction test to always run with speculative, head, and irreversible mode nodes.

Note to disable speculative trx execution on irreversible nodes (or any node), it now must be specified via `api-accept-transactions = false` and `p2p-accept-transactions = false`.

Modify syncing heuristic in `producer_plugin` to 5 minutes to match what is used in controller.

Resolves #1030

### DIFF
--- a/libraries/chain/fork_database.cpp
+++ b/libraries/chain/fork_database.cpp
@@ -192,10 +192,11 @@ namespace eosio::chain {
 
    template<class BSP>
    void fork_database_impl<BSP>::reset_root_impl( const bsp_t& root_bsp ) {
-      index.clear();
       assert(root_bsp);
       root = root_bsp;
       root->set_valid(true);
+      pending_savanna_lib_id = block_id_type{};
+      index.clear();
    }
 
    template<class BSP>
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -2095,7 +2095,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       // Determine if we are syncing: if we have recently started an old block then assume we are syncing
       if (last_start_block_time < now + fc::microseconds(config::block_interval_us)) {
          auto head_block_age = now - chain.head().block_time();
-         if (head_block_age > fc::seconds(5))
+         if (head_block_age > fc::minutes(5))
             return start_block_result::waiting_for_block; // if syncing no need to create a block just to immediately abort it
       }
       last_start_block_time = now;
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -834,6 +834,10 @@ public:
       return !_producers.empty();
    }
 
+   bool irreversible_mode() const {
+      return _db_read_mode == db_read_mode::IRREVERSIBLE;
+   }
+
    void on_accepted_block(const signed_block_ptr& block, const block_id_type& id) {
       auto& chain  = chain_plug->chain();
       auto  before = _unapplied_transactions.size();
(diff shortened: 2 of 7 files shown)


greg7mdp (approved):

tests/distributed-transactions-test.py
```diff
@@ -64,11 +63,13 @@
...
 
         Print("Stand up cluster")
-        extraNodeosArgs = ""
-        if speculative:
-           extraNodeosArgs = " --read-mode speculative "
+        specificExtraNodeosArgs = {}
+        specificExtraNodeosArgs[total_nodes-1] = f' --read-mode irreversible '
+        specificExtraNodeosArgs[total_nodes-2] = f' --read-mode irreversible '
+        specificExtraNodeosArgs[total_nodes-3] = f' --read-mode speculative '
+        specificExtraNodeosArgs[total_nodes-4] = f' --read-mode speculative '
```
linh2931, line 70: To reduce testing time and more clear, can they just be `specificExtraNodeosArgs[total_nodes-1] = f' --read-mode head '
        specificExtraNodeosArgs[total_nodes-2] = f' --read-mode irreversible '
        specificExtraNodeosArgs[total_nodes-3] = f' --read-mode speculative '

libraries/chain/fork_database.cpp
```diff
@@ -192,10 +192,11 @@ namespace eosio::chain {
 
    template<class BSP>
    void fork_database_impl<BSP>::reset_root_impl( const bsp_t& root_bsp ) {
-      index.clear();
       assert(root_bsp);
       root = root_bsp;
       root->set_valid(true);
+      pending_savanna_lib_id = block_id_type{};
+      index.clear();
    }
```
heifner, line 201: The reset of `pending_savanna_lib_id` is needed as starting in irreversible mode can reset the forkdb root to head. If the node is started again then it would try and advance LIB and segfault because the fork db didn't have anything in the index which means `forkdb.head()` returns nullptr. I attempted to reproduce this in 1.0.x, but couldn't find a way to trigger it. Not sure if we should backport it to 1.0.x or not.

tests/distributed-transactions-test.py
heifner: https://github.com/AntelopeIO/spring/pull/1036/commits/72f673a775472bde42798d9bf460482019ee8743

libraries/chain/fork_database.cpp
linh2931: Maybe not backport for now as it is a rare case and 1.1.x is not far away.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1587,6 +1594,10 @@ void producer_plugin_impl::plugin_startup() {
          _irreversible_block_time = fc::time_point::maximum();
       }
 
+      if (!_is_savanna_active && irreversible_mode() && chain_plug->accept_transactions()) {
```
linh2931, line 1597: Is this only for tests? All nodes in the field using this version of software (Rel 1.1.x) should have enabled Savanna. This is probably not needed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1587,6 +1594,10 @@ void producer_plugin_impl::plugin_startup() {
...
+         wlog("Accepting speculative transaction execution not recommended in read-mode=irreversible");
```
linh2931, line 1598: If this check is needed, add something indicating nodeos is in Legacy.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2081,7 +2095,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       // Determine if we are syncing: if we have recently started an old block then assume we are syncing
       if (last_start_block_time < now + fc::microseconds(config::block_interval_us)) {
          auto head_block_age = now - chain.head().block_time();
-         if (head_block_age > fc::seconds(5))
+         if (head_block_age > fc::minutes(5))
```
linh2931, line 2098: This is a big change in terms of the value. Any potential side effects?

plugins/producer_plugin/producer_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1036/commits/792f4778182d9c80ac7cfe2995b05e6d060b8776

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Support speculative transaction execution in irreversible mode.
Note:end

PR #1041, opened 2024-11-20
Add live graphic visualization of memory allocated/freed while nodeos is running

greg7mdp:
This change is mostly done in two separate parts:

## Boost interprocess memory occupancy tracking

The Boost interprocess library (which is used to manage memory within the shared memory segment allocated by chainbase) can use multiple memory allocation managers. 

Chainbase, and therefore nodeos, use the `rbtree_best_fit` algorithm which keeps track of free areas using a rbtree (the nodes of the rbtree are intrusively stored in the free areas themselves, so there is no additional memory cost for this tree, besides the fact that it requires a minimum allocation size).

This PR adds a new member in the memory segment header managed by the `interprocess` library, of the type:
```cpp
using occupancy_array_t = std::vector<uint8_t>; // one byte describes the occupancy percentage of each page of the segment
```
and updates this occupancy vector:
- when a new function `bool initialize_occupancy();` is called
- when memory is allocated or freed

As we cannot add these changes in the original boost repo, this PR links to a private repo which includes the changes in the `interprocess` library. These changes are also included in the following file:

[ip.txt](https://github.com/user-attachments/files/17835958/ip.txt)

## Use the occupancy vector maintained by `Boost interprocess`  to update a texture and visualize it on-screen

The visualization is implemented using `OpenGL`. This required adding some dependencies to chainbase:
1. some `glad` header files describing the OpenGL interface used. These were generated from https://gen.glad.sh/, specifying `gl 4.6` and `glx 1.4`, `compat profile`, `header only`, `GL_ARB_direct_state_access` extension.
2. the [glm](https://github.com/g-truc/glm) library, used as-is
3. the [glfw](https://github.com/glfw/glfw) library, with this trivial change: 
[glfw.txt](https://github.com/user-attachments/files/17836101/glfw.txt)

The visualization is implemented in the new files `mem_visualizer[.hpp, .cpp]`, and takes place on a separate thread.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -36,7 +36,7 @@
 	url = https://github.com/AntelopeIO/bls12-381
 [submodule "libraries/boost"]
 	path = libraries/boost
-	url = https://github.com/boostorg/boost.git
+	url = https://github.com/greg7mdp/boost.git
 [submodule "libraries/libfc/libraries/boringssl/bssl"]
 	path = libraries/libfc/libraries/boringssl/bssl
 	url = https://boringssl.googlesource.com/boringssl


ericpassmore:
Note:start
category: Tooling
component: Internal
summary: Added openGL tooling to visualize memory allocation in real time. 
Note:end

greg7mdp:
There are still build issues when testing reference contracts using libtester, but since we don't plan on merging this PR there is no point in wasting time fixing these. 

Branch can be built and used for memory usage visualization, but will not be merged. Closing the issue as completed.

PR #1044, opened 2024-11-25
[1.0.4] Test fix: Allow for block to not have qc_extension

### LINKED ISSUE #1023: Test failure: separate_prod_fin_test
https://github.com/AntelopeIO/spring/actions/runs/11788970348/job/32837287849
```
  File "/__w/spring/spring/build/tests/separate_prod_fin_test.py", line 75, in <module>
    qc_ext = block["qc_extension"]
KeyError: 'qc_extension'
```
Need to update test to check that `qc_extension` is in the `block` before trying to access it.

The test does try to look for it in other blocks, but fails to check that it is there before trying to access it.
```python
    while i < 3: # slow ci/cd might not get all votes in time, try a few blocks
        # LIB has advanced, which indicate at least 2 of non-producer finalizers have voted.
        block_num = cluster.getNode(1).getIrreversibleBlockNum()
        block = cluster.getNode(1).getBlock(block_num)
        qc_ext = block["qc_extension"]
        Print(f'{qc_ext}')
        # "11111" is the representation of a bitset showing which finalizers have voted (we have five total finalizers)
        if qc_ext["qc"]["active_policy_sig"]["strong_votes"] == "11111":
            break
        cluster.getNode(1).waitForLibToAdvance()
        i += 1
    assert i < 3, "Not all finalizers voted"

```

heifner:
It is possible for a block to not contain a `qc_extension` if the qc claim is the same as the previous block.

Resolves #1023

### DIFF
--- a/tests/separate_prod_fin_test.py
+++ b/tests/separate_prod_fin_test.py
@@ -72,11 +72,12 @@ try:
         # LIB has advanced, which indicate at least 2 of non-producer finalizers have voted.
         block_num = cluster.getNode(1).getIrreversibleBlockNum()
         block = cluster.getNode(1).getBlock(block_num)
-        qc_ext = block["qc_extension"]
-        Print(f'{qc_ext}')
-        # "11111" is the representation of a bitset showing which finalizers have voted (we have five total finalizers)
-        if qc_ext["qc"]["active_policy_sig"]["strong_votes"] == "11111":
-            break
+        if "qc_extension" in block:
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Allow for block to not have qc_extension
Note:end

PR #1047, opened 2024-11-26
Interrupt block validation on new best head

### LINKED ISSUE #1039: Poison block handling: Interrupt block validation to produce block
* For BP nodes, interrupt a block validation when scheduled to produce
* All nodes including BP nodes, interrupt a block validation when a new best-head* at the validation block height is received from the network.

> Note new best-head here means that a fork switch is required. If the new best head is a descendant of the current block being validated then validation should not be interrupted. Only when the new block is on an alternative fork should validation be interrupted.

A BP will always produce on schedule with this approach. The produced block should nominally be a new best-head on all other nodes since it will have a higher block timestamp. When the newly produced block comes into nodeos, it will determine it is a new best-head, interrupt validation of the current block, and switch over to validating the new fork which includes the block. This provides a natural recovery when a node is stuck validating a block.

This approach brings back the old drop-late-blocks functionality as a BP will begin producing blocks immediately when scheduled to do so. If the BP is in the middle of validating a block from the previous BP, it will interrupt that validation and produce on the previous block. We could potentially provide a bit of time for validation to potentially complete before interruption. However, it is simpler to just interrupt immediately. BPs already have a mechanism for making sure their blocks are received by the next BP in time.

This also provides faster switching to a better fork under normal conditions.

In all cases interrupted blocks will not be removed from the fork-database. In nominal cases, the interrupted block will be purged from the fork-database when irreversible moves so that the branch is purged. However, the block could potentially be used again if a fork switch to its branch is received from the network. In that situation the node will begin validating the block no matter how long it takes; or until it receives yet another better block to validate.

### LINKED ISSUE #1040: Poison block handling: Interrupt of block should leave block in fork database
Depends on #1039 

Modify the behavior of ctrl-c/signal interrupt blocks #993 to exit nodeos without removing the block from the fork database.

heifner:
Provide a mechanism to auto recover from "poison" blocks. Poison blocks are blocks with transactions which take an unexpectedly long time to execute. Poison block validation will automatically be interrupted when a better block is received from the network. 

If a BP node, always produce when scheduled to do so. Interrupting any block validation in progress. This restores the old behavior of "drop late blocks". When a producer is scheduled to produce it will interrupt any validation in progress and produce on top of the previous block.

- For BP nodes, interrupt block validation when scheduled to produce
  - Moved `producer_plugin` timer off the main thread executor and onto its own thread.
- Interrupt block validation when a new best head* is received from the network.
  - Do not apply blocks from the fork database on startup. Blocks in the fork database have not been validated. If one of them would happen to be a poison block then a node would be stuck attempting to validate the block. By allowing network blocks to begin flowing into the node immediately on startup, a new best head can be received and interrupt block validation of blocks being processed out of the fork database.
- Ctrl-c/signal interrupt now leaves block in the fork database instead of purging it and its descendants.
  - The block and its descendants can be left in the fork database because a node now auto recovers when a new block is received from the network. Or is interrupted if a BP and is scheduled to produce. 

* New best head means that a fork switch is required. If the new best head is a descendant of the current block being validated then validation is not interrupted. Only when the new block is on an alternative fork is validation be interrupted.

Resolves #1039 
Resolves #1040

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2038,21 +2038,6 @@ struct controller_impl {
       // Furthermore, fork_db.root()->block_num() <= lib_num.
       // Also, even though blog.head() may still be nullptr, blog.first_block_num() is guaranteed to be lib_num + 1.
 
-      auto finish_init = [&](auto& fork_db) {
-         if( read_mode != db_read_mode::IRREVERSIBLE ) {
-            auto pending_head = fork_db.head();
-            if ( pending_head && pending_head->id() != chain_head.id() ) {
-               // chain_head equal to root means that read_mode was changed from irreversible mode to head/speculative
-               bool chain_head_is_root = chain_head.id() == fork_db.root()->id();
-               if (chain_head_is_root) {
-                  ilog( "read_mode has changed from irreversible" );
-               }
-            }
-         }
-      };
-
-      fork_db_.apply<void>(finish_init);
-
       // At Leap startup, we want to provide to our local finalizers the correct safety information
       // to use if they don't already have one.
       // If we start at a block prior to the IF transition, that information will be provided when
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1870,12 +1870,8 @@ struct controller_impl {
             fork_db_reset_root_to_chain_head();
          } else if( !except_ptr && !check_shutdown() && !irreversible_mode() ) {
             if (auto fork_db_head = fork_db.head()) {
-               // applies all blocks up to fork_db head from fork_db, shouldn't return incomplete, but if it does loop until complete
-               ilog("applying ${n} fork database blocks from ${ch} to ${fh}",
+               ilog("fork database contains ${n} blocks after head from ${ch} to ${fh}",
                     ("n", fork_db_head->block_num() - chain_head.block_num())("ch", chain_head.block_num())("fh", fork_db_head->block_num()));
-               while (maybe_apply_blocks(forked_callback_t{}, trx_meta_cache_lookup{}) == controller::apply_blocks_result::incomplete)
-                  ;
-               ilog( "reversible blocks replayed to ${bn} : ${id}", ("bn", fork_db_head->block_num())("id", fork_db_head->id()) );
             }
          }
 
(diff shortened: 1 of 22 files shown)


greg7mdp (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Interrupt block validation on new best head, enables auto recovery from poison blocks.
Note:end

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -12,6 +12,7 @@ namespace eosio::chain {
    using block_branch_t = std::vector<signed_block_ptr>;
    enum class ignore_duplicate_t { no, yes };
    enum class include_root_t { no, yes };
+   enum class fork_db_add_t { failure, duplicate, added, appended_to_head, fork_switch };
```
linh2931, line 15: Should comments below (lines 71 to 74) be moved here?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2938,8 +2998,12 @@ void producer_plugin_impl::produce_block() {
    _time_tracker.clear();
 }
 
-void producer_plugin::received_block(uint32_t block_num) {
+void producer_plugin::received_block(uint32_t block_num, chain::fork_db_add_t fork_db_add_result) {
    my->_received_block = block_num;
+   if (fork_db_add_result == fork_db_add_t::fork_switch) {
```
linh2931, line 3003: Add a comment saying fork_db_add_t::fork_switch means head block of best fork (different from the current branch) is received. Or just pass in a parameter of new_best_fork instead of fork_db_add_result

libraries/chain/controller.cpp
```diff
@@ -2049,30 +2045,9 @@ struct controller_impl {
...
-                  ilog( "read_mode has changed from irreversible: applying best branch from fork database" );
-               }
-
-               // See comment below about pause-at-block for why `|| conf.num_configured_p2p_peers > 0`
-               if (chain_head_is_root || conf.num_configured_p2p_peers > 0) {
-                  ilog("applying branch from fork database ending with block: ${id}", ("id", pending_head->id()));
-                  // applies all blocks up to forkdb head from forkdb, shouldn't return incomplete, but if it does loop until complete
-                  while (maybe_apply_blocks(forked_callback_t{}, trx_meta_cache_lookup{}) == controller::apply_blocks_result::incomplete)
-                     ;
+                  ilog( "read_mode has changed from irreversible" );
```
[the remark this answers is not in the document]
heifner: Actually, I don't think we can definitively say that is the case. Seems like we should just remove this as I don't think there is a 100% accurate way to tell.

libraries/chain/fork_database.cpp
```diff
@@ -278,15 +278,25 @@ namespace eosio::chain {
...
-      return inserted.second && n == head_impl(include_root_t::no);
+      if (!inserted.second)
+         return fork_db_add_t::duplicate;
+      const bool new_head = n == head_impl(include_root_t::no);
+      if (new_head && n->previous() == prev_head->id())
+         return fork_db_add_t::appended_to_head;
+      if (new_head)
+         return fork_db_add_t::fork_switch;
+
+      return fork_db_add_t::added;
```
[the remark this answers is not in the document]
heifner: No, `added` means added to some branch and didn't become the new best head. It could be added anywhere on any branch.

plugins/producer_plugin/producer_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1047/commits/1efb62c7899f3997fe4b4836b0f12939fb63da47

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1047/commits/5e16cd5e29076e72ef804c083bc95c5ea7ba6a51

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1047/commits/7740940031243082098cdc67f48a3a9440047efb

linh2931 (approved):

PR #1057, opened 2024-12-06
Enable `/v1/producer/pause_at_block`

### LINKED ISSUE #570: Add pause-at-block option

heifner:
Enable new `/v1/producer/pause_at_block` endpoint for pausing a node at a particular block number. Use `/v1/producer/resume` to resume after pause.

See #1056 for most of the functionality of the implementation.

Resolves #570


tests/pause_at_block_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+    assert prodNode.waitForLibToAdvance(), "LIB did not advance with paused nodes"
+    # blockNum -1 since waitForBlock uses `> blockNum` instead of `>=`
+    assert prodNode2.waitForBlock(blockNum-1), f"Block {blockNum} did not arrive after pausing"
+
+    Utils.Print(f"Verify paused at block {blockNum}")
+    assert prodNode2.getHeadBlockNum() == blockNum, "Prod Node_01 did not pause at block"
+    assert headNode.getHeadBlockNum() == blockNum, "Head Node_02 did not pause at block"
+    assert specNode.getHeadBlockNum() == blockNum, "Speculative Node_03 did not pause at block"
+    assert irrvNode.getHeadBlockNum() == blockNum, "Irreversible Node_04 did not pause at block"
```
greg7mdp, line 73: Regarding the comments, above we have `headNode = cluster.getNode(3)`. Wouldn't that make it `Node_03`?

tests/pause_at_block_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+    blockNum += 5
+
+    Utils.Print(f"Pausing at block {blockNum}")
+    prodNode2.processUrllibRequest("producer", "pause_at_block", {"block_num":blockNum}),
+    headNode.processUrllibRequest("producer", "pause_at_block", {"block_num":blockNum}),
+    specNode.processUrllibRequest("producer", "pause_at_block", {"block_num":blockNum}),
+    irrvNode.processUrllibRequest("producer", "pause_at_block", {"block_num":blockNum}),
+
+    assert prodNode.waitForLibToAdvance(), "LIB did not advance with paused nodes"
```
greg7mdp, line 65: Is it a good practice to use `assert` for checking stuff - see issue #332.

greg7mdp (approved):
Nice test!

tests/pause_at_block_test.py
heifner: Thanks. I added another node and didn't update the comments.

heifner: https://github.com/AntelopeIO/spring/pull/1057/commits/836f87d4374b014e1c2e017c828008870af43a13

ericpassmore:
Note:start
category: Tests
component: Plugins
summary: Test for producer_plugin pause_at_block 
Note:end

PR #1061, opened 2024-12-10
[1.0.4] Fast shutdown in irreversible mode

### LINKED ISSUE #1053: Slow shutdown in irreversible mode
When syncing in irreversible mode, nodeos takes a long time to shutdown:

```
info  2024-12-05T15:13:25.500 nodeos    controller.cpp:3533           log_applied          ] Received block c12f4a344a37c6a2... #405146000 @ 2024-11-17T04:34:16.500 signed by eoscannonchn [trxs: 1, lib: 405145998, net: 2552, cpu: 6915 us, elapsed: 18638 us, applying time: 19954 us, latency: 1593548999 ms]
^Cinfo  2024-12-05T15:13:30.712 nodeos    main.cpp:193                  operator()           ] appbase quit called
info  2024-12-05T15:13:30.712 nodeos    controller.cpp:4513           interrupt_apply_bloc ] Interrupting apply block
info  2024-12-05T15:13:30.786 nodeos    controller.cpp:3787           apply_block          ] Interrupt of trx id: 587b30712a81f12b7242edd8094064ee839b30db312fc0cc516ec01a736d275e
info  2024-12-05T15:14:15.187 nodeos    controller.cpp:3533           log_applied          ] Received block 175302ddbd2bc49f... #405147000 @ 2024-11-17T04:42:36.500 signed by eoscannonchn [trxs: 4, lib: 405146998, net: 680, cpu: 5079 us, elapsed: 4373 us, applying time: 5759 us, latency: 1593098687 ms]
info  2024-12-05T15:14:35.592 nodeos    main.cpp:193                  operator()           ] appbase quit called
debug 2024-12-05T15:14:35.593 nodeos    producer_plugin.cpp:1670      plugin_shutdown      ] exit shutdown
info  2024-12-05T15:14:35.612 nodeos    block_handle.cpp:21           write                ] Writing chain_head block 405147675 1826101bf0b1daf5c14e79c90c518086216bbaa3d7662adb4b70297effa7eab2
info  2024-12-05T15:14:35.613 nodeos    fork_database.cpp:671         close                ] Persisting to fork_database file: /home/heifner/ext/leap/cmake-build-debug/bin/main-d/blocks/reversible/fork_db.dat
info  2024-12-05T15:14:35.613 nodeos    fork_database.cpp:167         close_impl           ] Writing fork_database 2 blocks with root 405147675:1826101bf0b1daf5c14e79c90c518086216bbaa3d7662adb4b70297effa7eab2 and head 405147677:1826101d060c851fe2da88359796ec9aeea688f8656aa451ecdb7a9461fdc37b
info  2024-12-05T15:14:49.966 nodeos    main.cpp:165                  operator()           ] nodeos version v1.1.0-dev v1.1.0-dev-8a47ac57588a1c42987206cc0831e8f828f0ab01-dirty
info  2024-12-05T15:14:49.966 nodeos    main.cpp:69                   log_non_default_opti ] Non-default options: data-dir = main-d, config-dir = mainconfig, read-mode = irreversible, chain-state-db-size-mb = 125000, p2p-peer-address = peer.main.alohaeos.com:9876, p2p-peer-address = p2p.eosargentina.io:9876, p2p-peer-address = eos.defibox.xyz:9876, p2p-peer-address = p2p.eos.cryptolions.io:9876, p2p-peer-address = p2p.donates2eden.io:9876, p2p-peer-address = p2p.eos42.io:9876, p2p-peer-address = mainnet.eosamsterdam.net:9876, p2p-peer-address = mainnet.eosarabia.net:3571, p2p-peer-address = p2p2.eoseoul.io:30333, p2p-peer-address = p2p.eosflare.io:9876, p2p-peer-address = p2p.bitmars.one:8080, p2p-peer-address = p2p.eos.detroitledger.tech:1337, p2p-peer-address = peer.eosio.sg:9876, p2p-peer-address = eos.seed.eosnation.io:9876, p2p-peer-address = peer1.eosphere.io:9876, p2p-peer-address = p2p.eossweden.org:9876, p2p-peer-address = eos.p2p.eosusa.io:9882, http-max-response-time-ms = 2500, api-accept-transactions = false, net-threads = 6, chain-threads = 4, read-only-threads = 0, verbose-http-errors = true, abi-serializer-max-time-ms = 2000, block-log-retain-blocks = 0, p2p-listen-endpoint = 0.0.0.0:9999, http-server-address = 127.0.0.1:7777, plugin = eosio::http_plugin, plugin = eosio::chain_api_plugin, plugin = eosio::net_api_plugin, plugin = eosio::producer_api_plugin, plugin = eosio::db_size_api_plugin
info  2024-12-05T15:14:49.967 nodeos    main.cpp:253                  main                 ] nodeos successfully exiting
```

heifner:
Faster shutdown in response to ctrl-c/SIGINT/SIGTERM/SIGPIPE especially when syncing in irreversible mode.

Resolves #1053

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1308,8 +1308,6 @@ struct controller_impl {
    // When in IRREVERSIBLE mode fork_db blocks are applied and marked valid when they become irreversible
    template<typename ForkDB, typename BSP>
    bool apply_irreversible_block(ForkDB& forkdb, const BSP& bsp) {
-      if (read_mode != db_read_mode::IRREVERSIBLE)
-         return true; // ignore
       controller::block_report br;
       if constexpr (std::is_same_v<block_state_legacy_ptr, std::decay_t<decltype(bsp)>>) {
          // before transition to savanna
@@ -1516,7 +1514,7 @@ struct controller_impl {
                // can then move forward on the then validated blocks.
                // In irreversible mode, break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run. There is a post
                // for every incoming blocks; enough posted tasks to apply all blocks queued to the fork db.
-               if (read_mode == db_read_mode::IRREVERSIBLE) {
+               if (irreversible_mode()) {
                   if (!replaying && fc::time_point::now() - start > fc::milliseconds(500))
                      return false;
                   return true;
@@ -1524,24 +1522,29 @@ struct controller_impl {
                return bsp->is_valid();
             };
 
-            std::vector<std::future<std::vector<char>>> v;
-            v.reserve( branch.size() );
-            for( auto bitr = branch.rbegin(); bitr != branch.rend() && should_process(*bitr); ++bitr ) {
-
-               v.emplace_back( post_async_task( thread_pool.get_executor(), [b=(*bitr)->block]() { return fc::raw::pack(*b); } ) );
+            using packed_block_future = std::future<std::vector<char>>;
+            std::vector<packed_block_future> v;
+            if (!irreversible_mode()) {
+               v.reserve( branch.size() );
+               for( auto bitr = branch.rbegin(); bitr != branch.rend() && should_process(*bitr); ++bitr ) {
+                  v.emplace_back( post_async_task( thread_pool.get_executor(), [b=(*bitr)->block]() { return fc::raw::pack(*b); } ) );
+               }
             }
             auto it = v.begin();
 
             for( auto bitr = branch.rbegin(); bitr != branch.rend() && should_process(*bitr); ++bitr ) {
-               if (!apply_irreversible_block(forkdb, *bitr))
-                  break;
+               packed_block_future f;
+               if (irreversible_mode()) {
+                  f = post_async_task( thread_pool.get_executor(), [b=(*bitr)->block]() { return fc::raw::pack(*b); } );
+                  if (!apply_irreversible_block(forkdb, *bitr))
+                     break;
+               }
 
                emit( irreversible_block, std::tie((*bitr)->block, (*bitr)->id()), __FILE__, __LINE__ );
 
                // blog.append could fail due to failures like running out of space.
                // Do it before commit so that in case it throws, DB can be rolled back.
-               blog.append( (*bitr)->block, (*bitr)->id(), it->get() );
-               ++it;
+               blog.append( (*bitr)->block, (*bitr)->id(), irreversible_mode() ? f.get() : it++->get() );
 
                db.commit( (*bitr)->block_num() );
                root_id = (*bitr)->id();
@@ -1557,9 +1560,9 @@ struct controller_impl {
                   forkdb.advance_root(root_id);
                }
             } catch( const fc::exception& e2 ) {
-               wlog("Caught exception ${e2}, while processing exception ${e}", ("e2", e2.to_detail_string())("e", e.what()));
+               elog("Caught exception ${e2}, while processing exception ${e}", ("e2", e2.to_detail_string())("e", e.what()));
             } catch( const std::exception& e2 ) {
-               wlog("Caught exception ${e2}, while processing exception ${e}", ("e2", e2.what())("e", e.what()));
+               elog("Caught exception ${e2}, while processing exception ${e}", ("e2", e2.what())("e", e.what()));
             }
             throw;
          }
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1556,6 +1556,7 @@ struct controller_impl {
             }
          } catch( const std::exception& e ) {
             try {
+               elog("Caught exception while logging irreversible: ${e}", ("e", e.what()));
                if (root_id != forkdb.root()->id()) {
                   forkdb.advance_root(root_id);
                }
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -1512,7 +1514,14 @@ struct controller_impl {
                // irreversible. Instead, this moves irreversible as much as possible and allows the next maybe_switch_forks call to apply these
                // non-validated blocks. After the maybe_switch_forks call (before next produced block or on next received block), irreversible
                // can then move forward on the then validated blocks.
-               return read_mode == db_read_mode::IRREVERSIBLE || bsp->is_valid();
+               // In irreversible mode, break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run. There is a post
+               // for every incoming blocks; enough posted tasks to apply all blocks queued to the fork db.
+               if (read_mode == db_read_mode::IRREVERSIBLE) {
+                  if (!replaying && fc::time_point::now() - start > fc::milliseconds(500))
```
linh2931, line 1520: Shouldn't `start` be reset?

heifner: Once this returns `false`, the method exits.

linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -1512,7 +1514,14 @@ struct controller_impl {
...
+                     return false;
```
greg7mdp, line 1521: When we return false, we will stop processing the second loop below, but I believe that the `fc::raw::pack(*b);` tasks to serialize the blocks for the block log will still be executed, even though the future to get their result will be lost. Is this a performance issue? If it is maybe we could just have a single loop and post the serialization tasks as we apply the blocks?

heifner: In non-irreversible mode `apply_irreversible_block` does nothing.

greg7mdp: Thanks, makes sense.
It still makes me uneasy checking code in that possibly exit the loop leaving what could be hundreds of blocks queued for packing, and when we re-enter the loop immediately after we will queue again these hundreds of blocks for packing, and exit after processing a couple blocks (if they are full), and repeat.

If you have a thousand blocks, and process only a couple at a time, many blocks could be packed hundreds of times.

heifner: I agree. I'm working on a fix.

greg7mdp: Well, maybe I'm mistaken. We will exit based on time only in irreversible mode, and when in this mode `apply_irreversible_block` does nothing. Still I'm not sure if the rest (`emit`, `blog.append`, `db.commit`, and waiting for the first `pack` tasks) could take enough time to leave a large number of queued `pack` tasks.

greg7mdp: Also I wonder what happens if we get an exception while processing the 10th block of the branch. We have signaled `irreversible_block`, updated `db` and the blog, but not advanced forkdb's root. 

So would we enter `log_irreversible` and reapply the first 9 blocks which were already applied?

greg7mdp: I think the two `wlog` in the catch block should be `elog` because they would leave us in an incorrect state.

libraries/chain/controller.cpp
```diff
@@ -1512,27 +1512,39 @@ struct controller_impl {
...
+                     break;
+               }
 
                emit( irreversible_block, std::tie((*bitr)->block, (*bitr)->id()), __FILE__, __LINE__ );
 
                // blog.append could fail due to failures like running out of space.
                // Do it before commit so that in case it throws, DB can be rolled back.
-               blog.append( (*bitr)->block, (*bitr)->id(), it->get() );
-               ++it;
+               blog.append( (*bitr)->block, (*bitr)->id(), irreversible_mode() ? f.get() : it++->get() );
```
greg7mdp, line 1547: Hum, I think the issue could still occur. 

Even in non-irreversible mode (where we queue all the `pack` tasks), the rest of the second loop (`blog.append()` for example) could take some time (suppose the system is swapping because the memory is over-committed) causing us to exit the second loop leaving many of the `pack` tasks queued, only to be re-enqueued immediately.

heifner: It will never exit the loop early in non-irreversible mode. There is no mechanism to interrupt the loop in non-irreversible mode.

greg7mdp: Oh yes, makes sense, I had missed that.
Is it possible still that we would get an exception? Maybe we should also add an `elog` before the throw at line `1567`?

greg7mdp (approved):

libraries/chain/controller.cpp
heifner: yes, we should log that exception. I'll add that.

greg7mdp: I honestly question whether the optimization of posting the `pack` to a thread pool is worth it, considering it makes the code harder to understand.

heifner: Certainly can be removed if/when #1062 is worked. When it was added it did make a measurable difference. That was with legacy consensus which typically moves LIB by 12 instead of 2. Regardless, seems like it would be useful during syncing until #1062 is complete.

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Fast shutdown in irreversible mode.
Note:end

PR #1063, opened 2024-12-11
On interrupt do not attempt to apply blocks

### LINKED ISSUE #1058: Test Failure: interrupt_trx_test (Timeout)
https://github.com/AntelopeIO/spring/actions/runs/12201296408/job/34039980125?pr=1057

Looks like it was caused by producer node shutting down too quickly. The LIB was seen to advance before the producer produced another block.

Need to wait until the producer has produced a block and LIB has moved before shutting down.

heifner:
On shutdown the `interrupt_trx_test.py` sends a SIGTERM to all processes. The test timed out #1058 because the node was stuck trying to validate a block with an infinite transaction in it. The SIGTERM interrupted the block apply but there was a queued up process block which caused it to attempt the `apply_block` again. The fix for this is actually in #1064 which added a `check_shutdown()` to `should_terminate()`. With the `check_shutdown()` in `should_terminate()`, the node will check the `app().is_quitting()` before attempting to apply a block. This prevents it from attempting to apply a block that was interrupted via SIGTERM which sets the `app().is_quitting()` flag.

Also added to the test a verification that head is advancing before verifying that LIB is advancing as a clearer indication that LIB is advancing after recovering from the interrupted block.

Resolves #1058

### DIFF
--- a/tests/interrupt_trx_test.py
+++ b/tests/interrupt_trx_test.py
@@ -109,6 +109,7 @@ try:
     trans=prodNode.pushMessage(contract, action, data, opts)
     assert trans and trans[0], "Failed to push doitslow action"
 
+    assert prodNode.waitForHeadToAdvance(blocksToAdvance=3), "prodNode not advancing head after doitforever action"
     assert prodNode.waitForLibToAdvance(), "prodNode did not advance lib after doitforever action"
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: On interrupt and shutdown do not attempt to apply blocks.
Note:end

PR #1084, opened 2024-12-19
Test: plugin_http_api_test correctly handle keeping logs when a test fails

### LINKED ISSUE #1079: Test failure: plugin_http_category_api_test
https://github.com/AntelopeIO/spring/actions/runs/12393958806/job/34597060951

heifner:
Fix `plugin_http_api_test.py` to keep logs when the test fails.
Note: `unittest.TestResult().wasSuccessful()` always returns `True` even when a test fails.

This doesn't really resolve #1079 but without this fix there are no logs to determine what went wrong. We can create a new issue if/when it fails again.

Resolves #1079

### DIFF
--- a/tests/plugin_http_api_test.py
+++ b/tests/plugin_http_api_test.py
@@ -1619,10 +1619,7 @@ class PluginHttpTest(unittest.TestCase):
 
     @classmethod
     def tearDownClass(self):
-        global keepLogs
         self.killNodes(self)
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Retain logs when plugin tests fail.
Note:end

PR #1086, opened 2024-12-20
Irreversible mode improvements

### LINKED ISSUE #1082: Test failure: transition_to_if
https://github.com/AntelopeIO/spring/actions/runs/12413237954/job/34655392902

```
  File "/__w/spring/spring/build/tests/transition_to_if.py", line 82, in <module>
    assert cluster.getNode(irreversibleNodeId).waitForLibToAdvance(), f"Lib stopped advancing on Node {irreversibleNodeId}, irreversible node"
AssertionError: Lib stopped advancing on Node 4, irreversible node
```

heifner:
- Return `apply_blocks` result in irreversible mode so incomplete apply is handled correctly.
- Handle interrupt in irreversible mode without logging an error.

Test failed because irreversible mode nodeos would not transition to Savanna. It would not transition because the `log_irreversible` was incomplete but nothing called `apply_blocks` again to continue the process.

Resolves #1082

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1584,12 +1584,6 @@ struct controller_impl {
                   result = apply_irreversible_block(fork_db, *bitr);
                   if (result != controller::apply_blocks_result::complete)
                      break;
-                  // In irreversible mode, break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run
-                  const bool more_blocks_to_process = bitr + 1 != branch.rend();
-                  if (!replaying && more_blocks_to_process && fc::time_point::now() - start > fc::milliseconds(500)) {
-                     result = controller::apply_blocks_result::incomplete;
-                     break;
-                  }
                }
 
                emit( irreversible_block, std::tie((*bitr)->block, (*bitr)->id()), __FILE__, __LINE__ );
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -1555,22 +1554,15 @@ struct controller_impl {
...
                // Only make irreversible blocks that have been validated. Blocks in the fork database may not be on our current best head
                // and therefore have not been validated.
                // An alternative more complex implementation would be to do a fork switch here and validate all blocks so they can be then made
                // irreversible. Instead, this moves irreversible as much as possible and allows the next maybe_switch_forks call to apply these
                // non-validated blocks. After the maybe_switch_forks call (before next produced block or on next received block), irreversible
                // can then move forward on the then validated blocks.
-               // In irreversible mode, break every ~500ms to allow other tasks (e.g. get_info, SHiP) opportunity to run. There is a post
-               // for every incoming blocks; enough posted tasks to apply all blocks queued to the fork db.
-               if (irreversible_mode()) {
-                  if (!replaying && fc::time_point::now() - start > fc::milliseconds(500))
```
heifner, line 1569: This timed `apply_block` is not needed anymore since `should_terminate` now checks the `app().quit()` flag.

linh2931 (approved):

libraries/chain/controller.cpp
greg7mdp: I thought we wanted to make sure that the main thread was not stuck doing something for a long time, to allow other tasks (e.g. get_info, SHiP) an opportunity to run (as the comment mentions), in addition to interrupt processing.

heifner: Good point. Probably best to add this back in.

heifner: https://github.com/AntelopeIO/spring/pull/1086/commits/2c0ca2e74e0ef89e116b1cc705a2a4547f52dc0b

greg7mdp (approved):

greg7mdp (commented):
I think that this would have been equivalent to leaving the check in `should_process`, setting `result` to `incomplete` before entering the loop, not updating it within the loop,  and setting it to complete after the loop if the iterator is at end().

heifner:
> I think that this would have been equivalent to leaving the check in `should_process`, setting `result` to `incomplete` before entering the loop, not updating it within the loop, and setting it to complete after the loop if the iterator is at end().

If you think that is clearer, I can change it.

greg7mdp:
> If you think that is clearer, I can change it.

Up to you what you think is best since you are the one maintaining this code generally. I'm fine either way.

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Improvement to Irreversible mode.
Note:end

PR #1090, opened 2025-01-02
P2P: Prioritize blocks, votes, and all messages over trxs

### LINKED ISSUE #1071: P2P: Prioritize blocks over trxs in connection buffer queue
Each connection has its own `buffer_queue` where output is queued for writing to the connection socket. Currently, this is a FIFO queue. When many trxs are following through a node, a block can be delayed since it has to wait for any trx ahead of it to be sent first. The `buffer_queue` could be made to be a priority queue. Alternatively, the `buffer_queue` already has a queue for "sync" blocks which are prioritized over everything else. The "sync" queue could be used for all blocks, not just sync blocks.

**Queue prioritization**
Preferred Option:
1. everything but transactions (blocks, votes, and more)
2. transactions

Another option:
1. everything but blocks, votes, and transactions
2. blocks & finality votes
3. transactions

heifner:
- Send all messages including blocks and votes over transactions. This prevents large trxs or large number of trxs from delaying block or vote propagation. 
- Includes a fix for `buffer_queue.out_callback()` called before `clear_out_queue()`. Long standing bug where the callbacks were never called because they were cleared before being called.
- Do not post `async_write` to connection strand as the call is already on the connection strand.
- Move serialization of votes off the main thread.

Resolves #1071

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -635,9 +635,9 @@ namespace eosio {
          _write_queue_size = 0;
       }
 
-      void clear_out_queue(boost::system::error_code ec, std::size_t w) {
+      void clear_out_queue(boost::system::error_code ec, std::size_t number_of_bytes_written) {
          fc::lock_guard g( _mtx );
-         out_callback( ec, w );
+         out_callback( ec, number_of_bytes_written );
          _out_queue.clear();
       }
 
@@ -704,9 +704,9 @@ namespace eosio {
          }
       }
 
-      void out_callback( boost::system::error_code ec, std::size_t w ) REQUIRES(_mtx) {
+      void out_callback( boost::system::error_code ec, std::size_t number_of_bytes_written ) REQUIRES(_mtx) {
          for( auto& m : _out_queue ) {
-            m.callback( ec, w );
+            m.callback( ec, number_of_bytes_written );
          }
       }
 
@@ -718,10 +718,10 @@ namespace eosio {
 
       alignas(hardware_destructive_interference_sz)
       mutable fc::mutex   _mtx;
-      uint32_t            _write_queue_size GUARDED_BY(_mtx) {0};
-      deque<queued_write> _write_queue      GUARDED_BY(_mtx);
-      deque<queued_write> _trx_write_queue  GUARDED_BY(_mtx); // trx_write_queue will be sent last
-      deque<queued_write> _out_queue        GUARDED_BY(_mtx);
+      uint32_t            _write_queue_size GUARDED_BY(_mtx) {0}; // size of _write_queue and _trx_write_queue
+      deque<queued_write> _write_queue      GUARDED_BY(_mtx); // queued messages, all messages except trxs
+      deque<queued_write> _trx_write_queue  GUARDED_BY(_mtx); // queued trx messages, trx_write_queue will be sent last
+      deque<queued_write> _out_queue        GUARDED_BY(_mtx); // currently being async_write
 
    }; // queued_buffer
 
@@ -1593,8 +1593,8 @@ namespace eosio {
       std::vector<boost::asio::const_buffer> bufs;
       buffer_queue.fill_out_buffer( bufs );
 
-      boost::asio::async_write( *c->socket, bufs,
-         boost::asio::bind_executor( c->strand, [c, socket=c->socket]( boost::system::error_code ec, std::size_t w ) {
+      boost::asio::async_write( *socket, bufs,
+         boost::asio::bind_executor( strand, [c, socket=socket]( boost::system::error_code ec, std::size_t w ) {
          try {
             peer_dlog(c, "async write complete");
             // May have closed connection and cleared buffer_queue
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -598,6 +598,14 @@ namespace eosio {
    // thread safe
    class queued_buffer : boost::noncopyable {
    public:
+      void reset() {
+         fc::lock_guard g( _mtx );
+         _write_queue.clear();
+         _trx_write_queue.clear();
+         _write_queue_size = 0;
+         _out_queue.clear();
+      }
+
       void clear_write_queue() {
          fc::lock_guard g( _mtx );
          _write_queue.clear();
@@ -605,11 +613,10 @@ namespace eosio {
          _write_queue_size = 0;
       }
 
-      void clear_out_queue() {
+      void clear_out_queue(boost::system::error_code ec, std::size_t w) {
          fc::lock_guard g( _mtx );
-         while ( !_out_queue.empty() ) {
-            _out_queue.pop_front();
-         }
+         out_callback( ec, w );
+         _out_queue.clear();
       }
 
       uint32_t write_queue_size() const {
@@ -662,13 +669,6 @@ namespace eosio {
          }
       }
 
-      void out_callback( boost::system::error_code ec, std::size_t w ) {
-         fc::lock_guard g( _mtx );
-         for( auto& m : _out_queue ) {
-            m.callback( ec, w );
-         }
-      }
-
    private:
       struct queued_write;
       void fill_out_buffer( std::vector<boost::asio::const_buffer>& bufs,
@@ -679,6 +679,12 @@ namespace eosio {
             _write_queue_size -= m.buff->size();
             _out_queue.emplace_back( m );
             w_queue.pop_front();
+         }
+      }
+
+      void out_callback( boost::system::error_code ec, std::size_t w ) REQUIRES(_mtx) {
+         for( auto& m : _out_queue ) {
+            m.callback( ec, w );
          }
       }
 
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -692,7 +698,7 @@ namespace eosio {
       mutable fc::mutex   _mtx;
       uint32_t            _write_queue_size GUARDED_BY(_mtx) {0};
       deque<queued_write> _write_queue      GUARDED_BY(_mtx);
-      deque<queued_write> _sync_write_queue GUARDED_BY(_mtx); // sync_write_queue will be sent first
+      deque<queued_write> _trx_write_queue GUARDED_BY(_mtx); // trx_write_queue will be sent last
```
greg7mdp, line 701: indentation

plugins/net_plugin/net_plugin.cpp
```diff
@@ -638,10 +645,10 @@ namespace eosio {
       // @param callback must not callback into queued_buffer
       bool add_write_queue( const std::shared_ptr<vector<char>>& buff,
                             std::function<void( boost::system::error_code, std::size_t )> callback,
-                            bool to_sync_queue ) {
+                            uint32_t net_message_which ) {
```
greg7mdp, line 648: I feel this `uint32_t net_message_which` could be replaced with `msg_type_t msg_type`, where `msg_type_t` is defined as:
```
   enum class msg_type_t {
      signed_block       = fc::get_index<net_message, signed_block>(),
      packed_transaction = fc::get_index<net_message, packed_transaction>(),
      vote_message       = fc::get_index<net_message, vote_message>()
   };
```

greg7mdp (commented):
> Move serialization of votes off the main thread.

Where is that done?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1564,55 +1571,54 @@ namespace eosio {
...
-               }
-               if (socket != c->socket ) { // different socket, c must have created a new socket, make sure previous is closed
-                  peer_ilog( c, "async write socket changed before callback");
-                  boost::system::error_code ec;
-                  socket->shutdown( tcp::socket::shutdown_both, ec );
-                  socket->close( ec );
-                  return;
-               }
+      boost::asio::async_write( *c->socket, bufs,
+         boost::asio::bind_executor( c->strand, [c, socket=c->socket]( boost::system::error_code ec, std::size_t w ) {
```
greg7mdp, line 1575: I think it could be slightly simplified to:
```suggestion
      boost::asio::async_write( *socket, bufs,
         boost::asio::bind_executor( strand, [c, socket]( boost::system::error_code ec, std::size_t w ) {
```

plugins/net_plugin/net_plugin.cpp
```diff
@@ -598,18 +598,25 @@ namespace eosio {
...
+         _out_queue.clear();
+      }
+
       void clear_write_queue() {
          fc::lock_guard g( _mtx );
          _write_queue.clear();
-         _sync_write_queue.clear();
+         _trx_write_queue.clear();
          _write_queue_size = 0;
       }
```
linh2931, line 614: `clear_write_queue()` and `reset()` have most of code in common.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -598,18 +598,25 @@ namespace eosio {
...
 
-      void clear_out_queue() {
+      void clear_out_queue(boost::system::error_code ec, std::size_t w) {
```
linh2931, line 616: What does parameter `w` mean?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -654,18 +661,11 @@ namespace eosio {
...
-         }
-      }
-
-      void out_callback( boost::system::error_code ec, std::size_t w ) {
-         fc::lock_guard g( _mtx );
-         for( auto& m : _out_queue ) {
-            m.callback( ec, w );
+         } else {
+            fill_out_buffer( bufs, _trx_write_queue );
+            assert(_trx_write_queue.empty() && _write_queue.empty() && _write_queue_size == 0);
```
linh2931, line 668: Why `_trx_write_queue.empty()`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -692,7 +698,7 @@ namespace eosio {
       mutable fc::mutex   _mtx;
       uint32_t            _write_queue_size GUARDED_BY(_mtx) {0};
       deque<queued_write> _write_queue      GUARDED_BY(_mtx);
```
linh2931, line 700: Add a comment about what are those 3 queues for.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3928,15 +3936,18 @@ namespace eosio {
    }
 
    void net_plugin_impl::bcast_vote_message( uint32_t exclude_peer, const chain::vote_message_ptr& msg ) {
-      buffer_factory buff_factory;
-      auto send_buffer = buff_factory.get_send_buffer( *msg );
```
heifner, line 3932: Move the serialization of vote_message off of main thread and move it down below into the net thread pool.

plugins/net_plugin/net_plugin.cpp
heifner: The mutex means we would need an internal private method. The code duplication here seemed to me to be clearer than that.

plugins/net_plugin/net_plugin.cpp
heifner: Replaced with `number_of_bytes_written`

plugins/net_plugin/net_plugin.cpp
heifner: `_write_queue.empty` because we would have gone into the `if` instead of the `else` if `_write_queue` not empty.
`_trx_write_queue.empty()` because we just filled `bufs` from it, which copies everything out.

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1090/commits/69804513b746746062e789fe6aba5237a00f098e

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1090/commits/69804513b746746062e789fe6aba5237a00f098e

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1090/commits/6e55bda1d590535ed38b2bd9366d2ce38c2a6118

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1090/commits/6e55bda1d590535ed38b2bd9366d2ce38c2a6118

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1090/commits/6e55bda1d590535ed38b2bd9366d2ce38c2a6118

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3928,15 +3962,18 @@ namespace eosio {
...
 
       fc_dlog(vote_logger, "bcast ${t} vote: block #${bn} ${id}.., ${v}, key ${k}..",
                 ("t", exclude_peer ? "received" : "our")("bn", block_header::num_from_id(msg->block_id))("id", msg->block_id.str().substr(8,16))
                 ("v", msg->strong ? "strong" : "weak")("k", msg->finalizer_key.to_string().substr(8,16)));
 
-      boost::asio::post( my_impl->thread_pool.get_executor(), [exclude_peer, msg{std::move(send_buffer)}]() mutable {
-         my_impl->dispatcher.bcast_vote_msg( exclude_peer, std::move(msg) );
+      boost::asio::post( my_impl->thread_pool.get_executor(), [exclude_peer, msg]() mutable {
+            buffer_factory buff_factory;
+            auto send_buffer = buff_factory.get_send_buffer( *msg );
```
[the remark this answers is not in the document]
greg7mdp: I'm not really seeing why we have the caching in buffer factory (and pass a buffer factory to the posted lambdas) rather than just pass the serialized buffer to the posted lambdas?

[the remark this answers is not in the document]
heifner: In this particular case the `vote_message_ptr` is passed to the lambda so it can be serialized in the net thread pool instead of on the calling thread. The `buffer_factory` caching is not being used for votes, just provides the convenient interface.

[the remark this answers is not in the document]
greg7mdp: I was looking at `dispatch_manager::bcast_block()` too. What is the benefit of using the buffer_factory there, instead of just calling `get_send_buffer` before the call to `for_each_block_connection` and capturing the send_buffer in the lambda instead of the buffer_factory?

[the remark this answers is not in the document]
heifner: For `bcast_block` the benefit is the node does not have to serialize the block if all peers have already received the block.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: System Stability
component: P2P
summary: Update net-plugin to prioritize blocks, votes, and messages to ensure timeline arrival of critical information.  
Note:end

PR #1096, opened 2025-01-08
resolve a noisy cmake deprecation warning when using latest cmake version

spoonincode:
latest cmake 3.31 has started to produce warnings such as
```
CMake Deprecation Warning at libraries/eos-vm/CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```
A variation of this is printed over 10 times for spring. Resolve this warning for projects we control. Needs also,
AntelopeIO/abieos#36
AntelopeIO/appbase#38
AntelopeIO/chainbase#55
AntelopeIO/eos-vm#29

### DIFF
--- a/CMakeModules/VersionUtils.cmake
+++ b/CMakeModules/VersionUtils.cmake
@@ -1,5 +1,3 @@
-cmake_minimum_required(VERSION 3.5)
-
 function(GENERATE_VERSION_METADATA)
    # Execute `git` to grab the corresponding data.
    execute_process(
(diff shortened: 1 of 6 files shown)


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Resolve cmake warnings on future version support. 
Note:end

spoonincode:
Thanks for the reviews; I still need another one on the linked AntelopeIO/abieos#36. It will help getting this merged before submodules continue to drift out which is a real pain.

spoonincode:
thanks

PR #1100, opened 2025-01-15
relax `checktime_hashing_fail` contract caching timing to eliminate spurious test failures

spoonincode:
On a fast Zen5 CPU, I see spurious failures on this line about 5% of the time (the action completes faster than 8ms resulting in the exception not firing). Reducing the limit to 6ms resolves the problem. But this is rather fragile and has been tweaked a number of times in the past. Since the purpose of this first action is to just get the contract cached, why not run a simple action that is guaranteed to pass to prime the cache?


linh2931 (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Eliminate spurious test failures by relaxing contract cache timing
Note:end

Issue #425, opened 2024-07-27
Test failure: performance_test_basic_read_only_trxs 


heifner:
https://github.com/AntelopeIO/spring/actions/runs/10119591629/job/27988450741

heifner:
One of the read-only trx failed:

```
[...]

PR #1102, opened 2025-01-15
[1.0.4] bump eos-vm submodule to fix spurious access violations in `performance_test_basic_read_only` test

spoonincode:
Brings in AntelopeIO/eos-vm#30 to resolve test failures seen in #425

TBD how to provide a test for this; it's tricky from a timing standpoint.


heifner (commented):
Is there anything we can do in the performance_test_basic_read_only test or some other test to make this more likely to fail (before the fix)?

spoonincode:
I added a test that should generally reliably fail before this fix. A (failing) demo run with the new two commits for the test applied on top of 1.0 is at https://github.com/AntelopeIO/spring/actions/runs/12801721083

tests/read_only_trx_overlap_timeout.py
```diff
@@ -0,0 +1,161 @@
...
+    deployTestContracts()
+
+    # start a background thread that constatly runs a ROtrx that is never expected to complete in max-transaction-time
+    thr = threading.Thread(target = longROtrxThread)
+    thr.start()
+
+    endTime = time.time() + testLengthSeconds
+    # and then run some other ROtrx that should complete successfully
+    while time.time() < endTime:
+        results = sendTransaction(tightloopAccountName, 'doit', {"count": 1000000}, opts='--read')  #1 million is a good number to always take <10ms
```
spoonincode, line 151: Though I am a little worried about the sensitivity across runs of this time requirement

spoonincode: and yup this value is no good in CI but great locally. hmmm I will have to think some on what to do. The test is more accurate the larger this number is (without "going over"). So maybe it needs to try and sniff out a good value before starting the test.

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Fix spurious access violations and bump eos-vm submodule.
Note:end

tests/read_only_trx_overlap_timeout.py
spoonincode: I've pushed up some changes to the test I was hoping would get it working in CI but no... still just way too sensitive to the timing. Not sure where to go from here on it.

spoonincode: I made at least one more change, cec196e, to try and improve the failures, but no luck. I think the test needs to be run with dedicated resources (really I think all the NP/LR tests should run with dedicated resources)

heifner (approved):

tests/read_only_trx_overlap_timeout.py
heifner: One counter-point to NP/PR tests running in a more dedicated environment is that the chaos-monkey of the current approach has shaken out bugs we would never see in a more stable environment.

greg7mdp (approved):

Issue #1103, opened 2025-01-15
fc::unpack() does not clear target for std::optional, std::shared_ptr, std::set, and std::list


linh2931:
`fc::unpack()` has several issues:

1. `std::optional`: the target is not reset if the source does not have a value. This will causes mismatch of source and target if the target already has a value, as that value is kept.
2. `std::shared_ptr`: the target is not reset if the source is null_ptr.
3. `std::set`: the target is not set to empty if the source is empty, and the target is not clear if the source is not empty ending up with elements from both source and target.
4. `std::list`: the target is not set to empty if the source is empty, and the target is not clear if the source is not empty ending up with elements from both source and target.

PR #1104, opened 2025-01-15
Fix upacking() issues for std::optional, std::shared_ptr, std::set, and std::list

linh2931:
This PR fixes a number of `unpacking()` issues:

1. `std::optional`: the target is not reset if the source does not have a value. This will causes mismatch of source and target if the target already has a value, as that value is kept.
2. `std::shared_ptr`: the target is not reset if the source is null_ptr.
3. `std::set`: the target is not set to empty if the source is empty, and the target is not clear if the source is not empty ending up with elements from both source and target.
4. `std::list`: the target is not set to empty if the source is empty, and the target is not clear if the source is not empty ending up with elements from both source and target.

The PR adds comprehensive tests for packing/unpacking `std::optional`, `std::shared_ptr`, `std::set`, and `std::list`.

Resolves https://github.com/AntelopeIO/spring/issues/1103

### DIFF
--- a/libraries/libfc/include/fc/io/raw.hpp
+++ b/libraries/libfc/include/fc/io/raw.hpp
@@ -207,7 +207,7 @@ namespace fc {
          auto tmp = std::make_shared<std::remove_const_t<T>>();
          fc::raw::unpack( s, *tmp );
          v = std::move(tmp);
-      }
+      } else { v.reset(); }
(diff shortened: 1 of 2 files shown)


heifner (approved):

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Fix `unpacking()` issues for several standard types.
Note:end

PR #1107, opened 2025-01-16
Fix interrupt transaction race condition

### LINKED ISSUE #1095: test failure: restart-scenarios-if-test-resync
https://github.com/AntelopeIO/spring/actions/runs/12656370912/job/35269365364

heifner:
The support for interrupting transaction on new best head, #1047, has a race condition such that an interrupt could interrupt a speculative transaction instead of a transaction in apply block. This would cause the transaction to fail and not be re-tried on the node. 

This PR refactors the `platform_timer` to keep track if it was trigger by an explicit interrupt or by the timer timing out. This allows for `transaction_context::checktime` to directly tell if a transaction was interrupted and raise an appropriate interrupt exception. If a transaction is interrupted it can then be retried similar to how a transaction is retried if it hits a block boundary.

Resolves #1095

### DIFF
--- a/libraries/chain/include/eosio/chain/platform_timer.hpp
+++ b/libraries/chain/include/eosio/chain/platform_timer.hpp
@@ -16,7 +16,6 @@ struct platform_timer {
    void start(fc::time_point tp);
    void stop();
    void interrupt_timer();
-   void _expire_now(); // called by internal timer
 
    /* Sets a callback for when timer expires. Be aware this could might fire from a signal handling context and/or
       on any particular thread. Only a single callback can be registered at once; trying to register more will
@@ -43,8 +42,9 @@ struct platform_timer {
    state_t timer_state() const { return _state; }
 
 private:
-   std::atomic<state_t> _state = state_t::stopped;
+   void expire_now();
 
+   std::atomic<state_t> _state = state_t::stopped;
 
    struct impl;
    constexpr static size_t fwd_size = 8;
--- a/libraries/chain/include/eosio/chain/platform_timer.hpp
+++ b/libraries/chain/include/eosio/chain/platform_timer.hpp
@@ -7,8 +7,6 @@
 
 #include <atomic>
 
-#include <signal.h>
-
 namespace eosio { namespace chain {
 
 struct platform_timer {
@@ -17,7 +15,8 @@ struct platform_timer {
 
    void start(fc::time_point tp);
    void stop();
-   void expire_now();
+   void interrupt_timer();
+   void _expire_now(); // called by internal timer
 
    /* Sets a callback for when timer expires. Be aware this could might fire from a signal handling context and/or
       on any particular thread. Only a single callback can be registered at once; trying to register more will
@@ -35,9 +34,18 @@ struct platform_timer {
       _expiration_callback_data = user;
    }
 
-   std::atomic_bool expired = true;
+   enum class state_t {
+      running,
+      timed_out,
+      interrupted,
+      stopped
+   };
+   state_t timer_state() const { return _state; }
 
 private:
+   std::atomic<state_t> _state = state_t::stopped;
+
+
    struct impl;
    constexpr static size_t fwd_size = 8;
    fc::fwd<impl,fwd_size> my;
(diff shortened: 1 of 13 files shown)


libraries/chain/include/eosio/chain/platform_timer.hpp
```diff
@@ -17,7 +15,8 @@ struct platform_timer {
 
    void start(fc::time_point tp);
    void stop();
-   void expire_now();
+   void interrupt_timer();
+   void _expire_now(); // called by internal timer
```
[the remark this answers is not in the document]
heifner: Not without some friends and with the multiple impls that becomes rather a pain.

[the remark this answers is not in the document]
heifner: Hmm, I'm a bit surprised that worked.

libraries/chain/platform_timer_accuracy.cpp
```diff
@@ -39,9 +39,10 @@ void compute_and_print_timer_accuracy(platform_timer& timer) {
       for(unsigned int i = 0; i < loops; ++i) {
          auto start = std::chrono::high_resolution_clock::now();
          timer.start(fc::time_point(fc::time_point::now().time_since_epoch() + fc::microseconds(interval)));
-         while(!timer.expired) {}
+         while(timer.timer_state() == platform_timer::state_t::running) {}
          auto end = std::chrono::high_resolution_clock::now();
          int timer_slop = std::chrono::duration_cast<std::chrono::microseconds>(end-start).count() - interval;
+         timer.stop();
```
greg7mdp, line 45: Not seeing why this added line is necessary?

libraries/chain/platform_timer_asio_fallback.cpp
```diff
@@ -75,18 +75,25 @@ void platform_timer::start(fc::time_point tp) {
...
       call_expiration_callback();
    }
 }
 
 void platform_timer::stop() {
-   if(expired)
+   if(_state == state_t::stopped)
       return;
 
    my->timer->cancel();
```
greg7mdp, line 95: Wouldn't it make sense to call `my->timer->cancel();` only if `(_state == state_t::running)`?

libraries/chain/platform_timer_asio_fallback.cpp
```diff
@@ -75,18 +75,25 @@ void platform_timer::start(fc::time_point tp) {
...
-   if (expired.compare_exchange_strong(expected, true)) {
+   state_t expected = state_t::running;
+   if (_state.compare_exchange_strong(expected, state_t::timed_out)) {
+      call_expiration_callback();
+   }
+}
+
+void platform_timer::interrupt_timer() {
+   state_t expected = state_t::running;
+   if (_state.compare_exchange_strong(expected, state_t::interrupted)) {
```
greg7mdp, line 86: Why don't these call `my->timer->cancel()`?

Isn't it possible that someone calls `interrupt_timer() ` (which updates the state but doesn't cancel the `async_wait`), and then calls `start(fc::time_point::maximum())`, and may be surprised when the previous `async_wait` expires the timer.

libraries/chain/platform_timer_accuracy.cpp
heifner: I added it because I originally added an assert that start() only called when timer is stopped. However, we don't actually honor that constraint currently elsewhere. Seems like it is better to honor that constraint normally.

ericpassmore:
Note:start
category: System Stability
component: Internal
summary: Fix a race condition where an interrupt could interrupt a speculative transaction.
Note:end

libraries/chain/platform_timer_asio_fallback.cpp
heifner: We would need to add a mutex around the timer to do that. I think the intention is that stop() should be called before calling start() again.

libraries/chain/platform_timer_accuracy.cpp
heifner: Decided it best to fix the invariant that `stop()` always called before `stop()`. Surprise-surprise the issue there is with deferred-transactions.

libraries/chain/platform_timer_asio_fallback.cpp
heifner: This now needs to set to stopped, so the assert on start() is not triggered.

greg7mdp (approved):

spoonincode:
interestingly, ARM prior to 8.3 generates calls to library code for these CAS. It's probably benign library code that doesn't change much and can be audited. Still, it may be a compelling reason to require 8.3+ if we ever have proper ARM support.

spoonincode (approved):
It would be nice to acquire/release the atomic, but that can be done later

PR #1115, opened 2025-01-23
P2P: Add p2p-listen-endpoint :trx & :blk support

### LINKED ISSUE #1101: P2P: listen endpoint option to specify no blocks
https://hackmd.io/-766AO7CTjCb99TOAt_jSQ

Add option to `p2p-listen-endpoint` `(:trx)` which tells all connections on that endpoint to not relay blocks.
This would be useful for public relay nodes so blocks are not sent to it.

Currently, blocks are sent to public relay nodes (all peers) after a block header has been validated and the block has not yet been received from the node.

A test I ran locally with 17 public P2P peers, echoed back to 5 on average for every block. It is highly doubtful those peers needed those blocks.

heifner:
Add ability to specify `p2p-listen-endpoint` connections to be transaction or block only. This allows, for example, a public P2P relay node to specify a public `--p2p-listen-endpoint 0.0.0.0:9876:trx` so that all connected peers will not send blocks to the relay node. The connected peers will only send transactions. This will reduce the inbound network traffic to the public relay node. Requires the connecting peer to have this PR incorporated. Older versions of nodeos will ignore the request to only send trx/blocks.

```
  --p2p-listen-endpoint arg (=0.0.0.0:9876:0)
                                        The actual host:port[:trx|:blk][:<rate-
                                        cap>] used to listen for incoming p2p
                                        connections. May be used multiple
                                        times. The optional rate cap will limit
                                        per connection block sync bandwidth to
                                        the specified rate. Total allowed
                                        bandwidth is the rate-cap multiplied by
                                        the connection count limit. A number
                                        alone will be interpreted as bytes per
                                        second. The number may be suffixed with
                                        units. Supported units are: 'B/s',
                                        'KB/s', 'MB/s, 'GB/s', 'TB/s', 'KiB/s',
                                        'MiB/s', 'GiB/s', 'TiB/s'. Transactions
                                        and blocks outside sync mode are not
                                        throttled. The optional 'trx' and 'blk'
                                        indicates to peers that only
                                        transactions 'trx' or blocks 'blk'
                                        should be sent. Examples:
                                           192.168.0.100:9876:1MiB/s
                                           node.eos.io:9876:trx:1512KB/s
                                           node.eos.io:9876:0.5GB/s
                                           [2001:db8:85a3:8d3:1319:8a2e:370:734
                                        8]:9876:250KB/s
```

Resolves #1101


ericpassmore:
Wow very cool

ericpassmore:
Note:start
category: Other
component: P2P
summary: Allow :trx and :blk support in p2p endpoints
Note:end

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,80 @@ namespace detail {
       return block_sync_rate_limit;
    }
 
-} // namespace detail
+   /// @return host, port, remainder
+   inline std::tuple<std::string, std::string, std::string> split_host_port_remainder(const std::string& peer_add) {
```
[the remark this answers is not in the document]
heifner: I personally would prefer to leave as is. I don't think the effort to improve this would be worth the effort. That being said, I could certainly do this. `string_view` I think is problematic. The boost interfaces are `std::string`. Also port, type, and sometimes host and remainder would be under the string size optimization which would make string_view not any better than string.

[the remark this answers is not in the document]
greg7mdp: fair enough!

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,80 @@ namespace detail {
...
-            limit = std::string(address, last_colon_location+1);
+                        "Invalid address specification ${a}; IPv6 addresses must be enclosed in square brackets.", ("a", peer_add));
          }
       }
+      string::size_type colon = p != string::npos ? peer_add.find(':', p) : string::npos;
+      if (colon == string::npos || colon == 0) {
+         return {};
+      }
+      string::size_type colon2 = peer_add.find(':', colon + 1);
+      string host = (p > 0) ? peer_add.substr( 0, p+1 ) : peer_add.substr( 0, colon );
```
greg7mdp, line 61: Wait, at this point, I think we are guaranteed that:
- `p > 0`
- `p + 1 == colon`

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,80 @@ namespace detail {
...
-         if( std::count(address.begin()+right_bracket_location, address.end(), ':') > 1 ) {
-            listen_addr = std::string(address, 0, last_colon_location);
-            limit = std::string(address, last_colon_location+1);
-         }
-      } else {
-         if( auto colon_count = std::count(address.begin(), address.end(), ':'); colon_count > 1 ) {
+      string::size_type p = peer_add[0] == '[' ? peer_add.find(']') : 0;
+      if (p == 0) {
+         if( auto colon_count = std::count(peer_add.begin(), peer_add.end(), ':'); colon_count >= 7 ) {
             EOS_ASSERT( colon_count <= 2, chain::plugin_config_exception,
```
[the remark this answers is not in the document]
heifner: Thanks. I'll fix that.
The existing tests expect `plugin_config_exception` which makes sense since this should only throw and report an error for configured options like `p2p-listen-endpoint`.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,80 @@ namespace detail {
...
-                        "Invalid address specification ${addr}; IPv6 addresses must be enclosed in square brackets.", ("addr", address));
-            listen_addr = std::string(address, 0, last_colon_location);
-            limit = std::string(address, last_colon_location+1);
+                        "Invalid address specification ${a}; IPv6 addresses must be enclosed in square brackets.", ("a", peer_add));
          }
       }
+      string::size_type colon = p != string::npos ? peer_add.find(':', p) : string::npos;
+      if (colon == string::npos || colon == 0) {
+         return {};
```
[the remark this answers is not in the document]
heifner: I don't follow. I reworked the logic, please check again.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: I reworked the logic, please check again.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,71 @@ namespace detail {
...
+                    "Invalid address specification ${a}; IPv6 addresses must be enclosed in square brackets.", ("a", peer_add));
+         return {};
+
+      }
+      string::size_type colon = peer_add.find(':', end_bracket+1);
+      if (colon == string::npos || colon == 0) {
+         return {};
+      }
+      string::size_type colon2 = peer_add.find(':', colon + 1);
+      string host = (end_bracket > 0) ? peer_add.substr( 0, end_bracket+1 ) : peer_add.substr( 0, colon );
```
greg7mdp, line 69: This would successfully parse a peer spec such as:
`"[2001:db8:85a3:8d3:1319:8a2e:370:7348]whatever junk I want here:9876:250KB/s"`

Which could be fixed by checking that, `end_bracket == 0 || end_bracket+1 == colon`.

Then this line could be simpler (the only case where `end_bracket > 0 && end_bracket+1 != colon` if when there is junk text between the end bracket and the colon):

`string host = peer_add.substr( 0, colon );`

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,71 @@ namespace detail {
...
-      } else {
-         if( auto colon_count = std::count(address.begin(), address.end(), ':'); colon_count > 1 ) {
-            EOS_ASSERT( colon_count <= 2, chain::plugin_config_exception,
-                        "Invalid address specification ${addr}; IPv6 addresses must be enclosed in square brackets.", ("addr", address));
-            listen_addr = std::string(address, 0, last_colon_location);
-            limit = std::string(address, last_colon_location+1);
-         }
+      constexpr bool should_throw = true;
+      auto [host, port, remainder] = detail::split_host_port_remainder(address, should_throw);
+      auto listen_addr = host + ":" + port;
```
greg7mdp, line 99: Is ":" a valid listen address?
For example, when `peer_add = "192.168.0.1"`, I think `split_host_port_remainder` would return 3 empty strings without throwing, which would cause `listen_addr ` to be ":".

My suggestion at line 63 would address this I think.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,71 @@ namespace detail {
...
+            EOS_ASSERT(!should_throw, chain::plugin_config_exception,
+                       "Invalid address specification ${a}, IPv6 no closing square bracket", ("a", peer_add) );
+            return {};
+         }
+      } else if (colon_count >= 7) {
+         EOS_ASSERT(!should_throw, chain::plugin_config_exception,
+                    "Invalid address specification ${a}; IPv6 addresses must be enclosed in square brackets.", ("a", peer_add));
+         return {};
+
+      }
```
greg7mdp, line 63:
```suggestion
      } else if (colon_count < 1 || colon_count > 3) {
￼         EOS_ASSERT(!should_throw, chain::plugin_config_exception,
￼                    "Invalid address specification ${a}; bad colon count", ("a", peer_add));
      }
```

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,71 @@ namespace detail {
...
+      auto [host, port, remainder] = detail::split_host_port_remainder(peer_add, should_throw);
+      if (host.empty()) return {};
+
+      string::size_type end = remainder.find_first_of( " :+=.,<>!$%^&(*)|-#@\t" ); // future proof by including most symbols without using regex
+      std::string type = remainder.substr(0, end);
+
+      return {std::move(host), std::move(port), std::move(type)};
+   }
+
+   /// @return listen address, type [trx|blk], and block sync rate limit (in bytes/sec) of address string
```
greg7mdp, line 94: Does it return the `type [trx|blk]`? I don't see it.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -40,29 +40,71 @@ namespace detail {
...
+      using std::string;
+      // host:port[:trx|:blk][:<rate>]   // rate is discarded
+      if (peer_add.empty()) return {};
+
+      constexpr bool should_throw = false;
+      auto [host, port, remainder] = detail::split_host_port_remainder(peer_add, should_throw);
+      if (host.empty()) return {};
+
+      string::size_type end = remainder.find_first_of( " :+=.,<>!$%^&(*)|-#@\t" ); // future proof by including most symbols without using regex
+      std::string type = remainder.substr(0, end);
```
greg7mdp, line 89: What is the benefit of doing it this way vs:

```
      std::string type;
      if (remainder.starts_with("blk") || remainder.starts_with("trx"))
         type = remainder.substr(0, 3);
```

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3715,15 +3721,21 @@ namespace eosio {
...
+         return;
       }
       case normal : {
          if (protocol_version >= proto_block_nack) {
             if (!msg.req_blocks.ids.empty()) {
                const block_id_type& id = msg.req_blocks.ids.back();
-               peer_dlog( this, "received request_message:normal #${bn}:${id}", ("bn", block_header::num_from_id(id))("id",id) );
+               peer_dlog( this, "${d} request_message:normal #${bn}:${id}",
+                          ("d", is_blocks_connection() ? "received" : "ignoring")("bn", block_header::num_from_id(id))("id",id) );
+               if (!is_blocks_connection())
```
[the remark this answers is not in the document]
heifner: If we add a votes connection or other connection type then there would be a distinction. That might be future proofing we don't need. But if anything I think we would remove the `only` versions.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1115/commits/6c2442b14f7724c44625cc8b8f862b1cada1449d

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1115/commits/6c2442b14f7724c44625cc8b8f862b1cada1449d

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1115/commits/6c2442b14f7724c44625cc8b8f862b1cada1449d

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1115/commits/6c2442b14f7724c44625cc8b8f862b1cada1449d

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1115/commits/6c2442b14f7724c44625cc8b8f862b1cada1449d

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: I didn't think of that, good point!

spoonincode (approved):

Issue #1119, opened 2025-01-23
don't stop an already stopped `platform_timer`


spoonincode:
This long standing issue can result in thousands of needless syscalls a second.

PR #1123, opened 2025-01-25
don't pause billing timer when fetching EOS VM OC's `code_descriptor`

spoonincode:
This pause & restart of the billing timer occurs for _every_ action executed by OC. It's a substantial amount of overhead (calling in to the kernel twice and doing whatever overhead needed to manage the timer) to be burdened with for every single action. I think this was added because in 1.1 `consume_compile_thread_queue()` is a wee bit heavier than prior releases (because it needs to take a lock a couple of times; for every action :disappointed: ). But since the number of outstanding compiles is limited to OC compile threads, the time spent in there is still very low. It wouldn't surprise me if this pause/resume is alone consuming more unaccounted for time than the nominal execution of `consume_compile_thread_queue()`.

Noticed during work of #1119


greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: VM-OC
summary: Reduce pause and restart of billing timer for EOS VM OC. 
Note:end

PR #1127, opened 2025-01-26
remove a log message when OC compilation completes for running contract

spoonincode:
When I start from snapshot I can sometimes see this log once or twice. imo this ends up being deceptive because it looks like OC only tiered up 1 or 2 contracts (at least based on the current wording of the message). We already have another ilog a few lines down that logs when an interruption actually occurs. So, I don't really think we need this one.

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -115,7 +115,6 @@ struct eosvmoc_tier {
       void async_compile_complete(boost::asio::io_context& ctx, const digest_type& code_id, fc::time_point queued_time) {
          if (executing_code_hash.load() == code_id) { // is action still executing?
             auto elapsed = fc::time_point::now() - queued_time;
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -115,7 +115,6 @@ struct eosvmoc_tier {
       void async_compile_complete(boost::asio::io_context& ctx, const digest_type& code_id, fc::time_point queued_time) {
          if (executing_code_hash.load() == code_id) { // is action still executing?
             auto elapsed = fc::time_point::now() - queued_time;
-            ilog("EOS VM OC tier up for ${id} compile complete ${t}ms", ("id", code_id)("t", elapsed.count()/1000));
```
heifner, line 118: Fine to remove it. Could have made it a `tlog`. It was nice when testing as I could easily see how long it took for the queued up compiles to finish.

heifner (approved):

ericpassmore:
Note:start
category: Logging
component: OC
summary: Remove redundant log message when OC compilation completes.
Note:end

PR #1128, opened 2025-01-27
allow the dev&test synchronous EOS VM OC mode without adjusting a compile time option

### LINKED ISSUE #1114: don't gate synchronous OC mode behind a compile-time option
While this option is only for development & testing use, gating it behind a compile time option means automated testing systems (like chicken dance) can't simply just pull a built binary from CI and use this option.

spoonincode:
Synchronous EOS VM OC mode is only for development & testing use, but gating it behind a compile time option means automated testing systems (like chicken dance) can't simply just pull a built binary from CI and use this mode. Now allow synchronous EOS VM OC via `wasm-runtime=eos-vm-oc-forced` (will also need to set `eos-vm-oc-enable=off`). This `eos-vm-oc-forced` choice isn't documented anywhere and it's unlikely any normal user would stumble on it. Closes #1114

### DIFF
--- a/libraries/chain/wasm_interface.cpp
+++ b/libraries/chain/wasm_interface.cpp
@@ -120,7 +120,7 @@ std::istream& operator>>(std::istream& in, wasm_interface::vm_type& runtime) {
       runtime = eosio::chain::wasm_interface::vm_type::eos_vm;
    else if (s == "eos-vm-jit")
       runtime = eosio::chain::wasm_interface::vm_type::eos_vm_jit;
-   else if (s == "eos-vm-oc")
+   else if (s == "forced-eos-vm-oc")
       runtime = eosio::chain::wasm_interface::vm_type::eos_vm_oc;
    else
       in.setstate(std::ios_base::failbit);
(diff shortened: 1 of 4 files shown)


libraries/chain/wasm_interface.cpp
```diff
@@ -120,7 +120,7 @@ std::istream& operator>>(std::istream& in, wasm_interface::vm_type& runtime) {
       runtime = eosio::chain::wasm_interface::vm_type::eos_vm;
    else if (s == "eos-vm-jit")
       runtime = eosio::chain::wasm_interface::vm_type::eos_vm_jit;
-   else if (s == "eos-vm-oc")
+   else if (s == "forced-eos-vm-oc")
```
heifner, line 123: Maybe `eos-vm-oc-forced` so they all start with `eos-vm`.

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Allow synchronous EOS VM OC intended for development and testing.
Note:end

PR #1129, opened 2025-01-27
enable `fsgsbase` instructions on Linux 5.9+ for EOS VM OC

spoonincode:
EOS VM OC uses the GS segment register for the base of its WASM memory. Prior to Linux 5.9 the only way to adjust the base address is via a syscall. For 5.9+, combined with modern CPUs (Intel Ivy Bridge+, 2012; AMD Bulldozer+, 2011), it's possible to instead use the `fsgsbase` instructions from userspace without needing to make a syscall. Switching to these instructions eliminates _at least_ one syscall per action, and oftentimes more for something like EVM. 

I ran a replay test over blocks 96328736-96627942 on an i7-12700K (Alder Lake, 2021). I saw a surprising ~3.5% improvement. This far exceeds my expectations (should have done this years ago). I was anticipating that only older CPUs subject to Meltdown would incur such heavy overhead on this seemingly simplistic syscall. What might explain this is that Alder Lake is subject to the swapgs spectre variant, but I have not confirmed that is the source of the slow down.

I had planed to dig out an 8th gen Core CPU (with Meltdown) to benchmark with but now I'm not sure I'll go through the trouble since the gains are so impressive on modern gen.

I have created,
https://github.com/AntelopeIO/spring/wiki/Environment-Variables-Controlling-CPU-Features
to track the various environment variables controlling CPU instruction usage.

### DIFF
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/gs_seg_helpers.c
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/gs_seg_helpers.c
@@ -3,8 +3,15 @@
 #include <asm/prctl.h>
 #include <sys/prctl.h>
 #include <sys/mman.h>
+#include <sys/auxv.h>
+#include <elf.h>
+#include <immintrin.h>
 
 int arch_prctl(int code, unsigned long* addr);
+
+#ifndef HWCAP2_FSGSBASE
+#define HWCAP2_FSGSBASE (1 << 1)
+#endif
 
 #define EOSVMOC_MEMORY_PTR_cb_ptr GS_PTR struct eos_vm_oc_control_block* const cb_ptr = ((GS_PTR struct eos_vm_oc_control_block* const)(EOS_VM_OC_CONTROL_BLOCK_OFFSET));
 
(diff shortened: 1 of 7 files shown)


libraries/chain/webassembly/runtimes/eos-vm-oc/gs_seg_helpers.c
```diff
@@ -64,3 +70,56 @@ void* eos_vm_oc_get_bounce_buffer_list() {
...
+              ↓
+      argc, argv[0], ..., argv[argc - 1], NULL, evniron0, environ1, ..., NULL
+   */
+   const int argc = *(int*)(_dl_argv - 1);
+   char** my_environ = _dl_argv + argc + 1;
+   while(*my_environ != NULL) {
+      const char disable_str[] = "SPRING_DISABLE_FSGSBASE";
+      if(strncmp(*my_environ++, disable_str, strlen(disable_str)) == 0)
+         return 0;
+   }
```
spoonincode, line 105: This is the same mess as I used in bls,
https://github.com/AntelopeIO/bls12-381/pull/4#discussion_r1279807568
Don't really know what to do here. Disabling optimizations via environment variable is nice. Seems hard to believe `_dl_argv` would ever deviate.

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: OC
summary: Utilize modern CPU instructions to speed up system and contract execution.
Note:end

PR #1130, opened 2025-01-27
don't start kernel level platform timer when applying blocks

spoonincode:
When passing `fc::time_point::max()` to `platform_timer::start()`, `platform_timer` recognizes that,
https://github.com/AntelopeIO/spring/blob/909603c103e81744c6898e18f48cbcd1a08e542f/libraries/chain/platform_timer_posix.cpp#L57-L62
and pretends the timer is running without actually starting the timer. Then when `stop()` is called it would still make a request to the kernel to stop the never started timer. There is nothing logically wrong with that, it's just a wasteful syscall that isn't needed that can end up happening thousands of times a second.

My memory was that historically when applying a block, `transaction_context::init()` would effectively `platform_timer::start(fc::time_point::max())`. That would result in the condition above and it's why I wrote #1119 worded the way it is. But I can't actually find evidence of this in history so I may be misremembering it.

Instead, today, `transaction_context::init()` passes a time 1 year in the future when applying a block. This actually makes the situation _worse_ than what I thought: now for every transaction while applying a block a timer is entirely needlessly started and stopped.

So this PR,
1) Makes `transaction_context::init()` do `platform_timer::start(fc::time_point::max())`; the timer somehow needs to indicate `running` state for things like `checktime()` to work so unless wanting to touch a lot more code we can't just entirely be hands off the timer. We could also do something like a `platform_timer::start_forever()` instead if not liking overloading the meaning of `start()`.
2) The `platform_timer` now remembers when it was started with `fc::time_point::max()` and when stopped skips the kernel call.

There are a handful of ways to go about accomplishing various nuances in what this PR does. Any strong opinions for a different approach can be explored.

### DIFF
--- a/libraries/chain/platform_timer_asio_fallback.cpp
+++ b/libraries/chain/platform_timer_asio_fallback.cpp
@@ -56,11 +56,14 @@ platform_timer::~platform_timer() {
 }
 
 void platform_timer::start(fc::time_point tp) {
+   assert(_state == state_t::stopped);
    if(tp == fc::time_point::maximum()) {
       _state = state_t::running;
+      timer_running_forever = true;
       return;
    }
    fc::microseconds x = tp.time_since_epoch() - fc::time_point::now().time_since_epoch();
+   timer_running_forever = false;
    if(x.count() <= 0)
       _state = state_t::timed_out;
    else {
@@ -89,11 +92,14 @@ void platform_timer::interrupt_timer() {
 }
 
 void platform_timer::stop() {
-   if(_state == state_t::stopped)
+   const state_t prior_state = _state;
+   if(prior_state == state_t::stopped)
+      return;
+   _state = state_t::stopped;
+   if(prior_state == state_t::timed_out || timer_running_forever)
       return;
 
    my->timer->cancel();
-   _state = state_t::stopped;
 }
 
 }}
--- a/libraries/chain/platform_timer_asio_fallback.cpp
+++ b/libraries/chain/platform_timer_asio_fallback.cpp
@@ -57,9 +57,9 @@ platform_timer::~platform_timer() {
 
 void platform_timer::start(fc::time_point tp) {
    assert(_state == state_t::stopped);
-   if(tp == fc::time_point::maximum()) {
+   timer_running_forever = tp == fc::time_point::maximum();
+   if(timer_running_forever) {
       _state = state_t::running;
-      timer_running_forever = true;
       return;
    }
    fc::microseconds x = tp.time_since_epoch() - fc::time_point::now().time_since_epoch();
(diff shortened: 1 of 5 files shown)


libraries/chain/platform_timer_kqueue.cpp
```diff
@@ -88,20 +88,23 @@ platform_timer::~platform_timer() {
...
+      _state = state_t::timed_out;
    else {
       struct kevent64_s aTimerEvent;
       EV_SET64(&aTimerEvent, my->timerid, EVFILT_TIMER, EV_ADD|EV_ENABLE|EV_ONESHOT, NOTE_USECONDS|NOTE_CRITICAL, x.count(), (uint64_t)this, 0, 0);
 
-      expired = false;
+      _state = state_t::running;
       if(kevent64(kqueue_fd, &aTimerEvent, 1, NULL, 0, KEVENT_FLAG_IMMEDIATE, NULL) != 0)
-         expired = true;
+         _state = state_t::timed_out;
```
spoonincode, line 107: It looks like this wasn't updated completely in #1107. We should really refactor the code a bit to move the common state logic out of the platform specific impls.

libraries/chain/platform_timer_posix.cpp
```diff
@@ -88,11 +90,14 @@ void platform_timer::interrupt_timer() {
 }
 
 void platform_timer::stop() {
-   if(_state == state_t::stopped)
+   const state_t prior_state = _state;
+   if(prior_state == state_t::stopped)
+      return;
+   _state = state_t::stopped;
+   if(prior_state == state_t::timed_out || timer_running_forever)
```
spoonincode, line 97: Another minor tweak here is to not stop the timer if it has already expired. Obviously that's a rather rare scenario, but might as well be thorough while tweaking all this.

heifner (approved):

libraries/chain/platform_timer_asio_fallback.cpp
```diff
@@ -56,11 +56,14 @@ platform_timer::~platform_timer() {
...
 
 void platform_timer::start(fc::time_point tp) {
+   assert(_state == state_t::stopped);
    if(tp == fc::time_point::maximum()) {
       _state = state_t::running;
+      timer_running_forever = true;
       return;
    }
    fc::microseconds x = tp.time_since_epoch() - fc::time_point::now().time_since_epoch();
+   timer_running_forever = false;
```
greg7mdp, line 66: No need to change, just a remark. Personally I prefer initializing members once if possible, as in:
```suggestion
   timer_running_forever = tp == fc::time_point::maximum();
   if(timer_running_forever ) {
      _state = state_t::running;
      return;
   }
   fc::microseconds x = tp.time_since_epoch() - fc::time_point::now().time_since_epoch();
```

greg7mdp (approved):

libraries/chain/platform_timer_asio_fallback.cpp
spoonincode: that does look good here

ericpassmore:
Note:start
category: Other
component: Internal
summary: More efficient system calls related to platform timer during application of blocks.
Note:end

PR #1135, opened 2025-01-28
Test: Create unique log files for nodeos

heifner:
When `nodeos` is quickly relaunched in the integration tests, it is possible that a `nodeos` log file is overwritten because it has the same launch time and file name. This happens, for example, in `nodeos_read_terminate_at_block_test.py`. Make sure a unique log file name is created for each run of `nodeos`.


spoonincode (approved):

greg7mdp (approved):

greg7mdp (commented):
Maybe add a comment such as `# make sure we don't overwrite an existing log file.`

greg7mdp (approved):

PR #1137, opened 2025-01-29
Make eos-vm-oc-whitelist option multitoken

heifner:
All other whitelist options are multi-token.  This PR changes `eos-vm-oc-whitelist` to also be multi-token.

Multi-token allows for multiple values to be specified for `--eos-vm-oc-whitelist`; e.g. `--eos-vm-oc-whitelist xsat other another`.

Note: if any `eos-vm-oc-whitelist` option is specified then `xsat` needs to be explicitly added. The default is overridden if any value is specified. For example, to add `other`, use `--eos-vm-oc-whitelist xsat other`.

The default value of `eos-vm-oc-whitelist` is `xsat`. If `xsat` is not desired, then an alternative must be specified. If there is a desire to not whitelist any accounts, then use `eosio.null` or any other account without a contract.

Also adds a logging of `eos-vm-oc-whiltelist` accounts on startup.


spoonincode (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: OC
summary: All the config option `--eos-vm-oc-whitelist` to accept multiple tokens.
Note:end

PR #1141, opened 2025-02-03
fix overlapped read only transaction timeouts on macOS

spoonincode:
Needs AntelopeIO/eos-vm#34. This remains broken on 1.0.x but since there will not be any more 1.0.x releases not fixing it there.


heifner (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Fix overlapped read only transaction timeouts on macOS.
Note:end

greg7mdp (approved):

Issue #1145, opened 2025-02-05
Spring 1.1 Misc Errors During Replay


ericpassmore:
Getting misc errors during replay test. Found in 
- run 20
   - job 140547858842528
   - start block 16150309
   - end block 17323276
- run 54
   - job 140547858857456
   - start block 46911265
   - end block 48116875

[nodeos140547858857456.log](https://github.com/user-attachments/files/18675035/nodeos140547858857456.log)
[nodeos140547858842528.log](https://github.com/user-attachments/files/18675033/nodeos140547858842528.log)

```
error 2025-02-05T14:21:39.934 nodeos    controller.cpp:3883           apply_block          ] e.to_detail_string(): 3100000 misc_exception: Miscellaneous exception
Setting a platform_timer callback when one already exists
    {}
    nodeos  platform_timer.hpp:36 set_expiration_callback
eosluckydice <= eosio::onerror pending console output: 
    {"console":"","account":"eosio","action":"onerror","receiver":"eosluckydice"}
    nodeos  apply_context.cpp:134 exec_one
 
warn  2025-02-05T14:21:39.934 nodeos    controller.cpp:4361           replay_irreversible_ ] 3100000 misc_exception: Miscellaneous exception
Setting a platform_timer callback when one already exists
    {}
    nodeos  platform_timer.hpp:36 set_expiration_callback
eosluckydice <= eosio::onerror pending console output: 
    {"console":"","account":"eosio","action":"onerror","receiver":"eosluckydice"}
    nodeos  apply_context.cpp:134 exec_one

    {}
    nodeos  controller.cpp:3891 apply_block

info  2025-02-05T14:21:39.934 nodeos    main.cpp:165                  operator()           ] nodeos version v1.1.0-rc1 v1.1.0-rc1-dcb9cc9bec4c175d38a03efedf2b50431ea35c71
```

ericpassmore:
Another Misc error 
- Run 78
   - Start Block: 68664955
   - End Block: 69871994

ericpassmore:
Last commit hash that worked
```
info  2025-02-03T22:02:19.525 nodeos    main.cpp:165                  operator()           ] nodeos version v1.1.0-dev v1.1.0-dev-4a5095aa3efd7518d7f669622a0b8884b5c3f3f6
[...]

PR #1146, opened 2025-02-05
Revert "fix overlapped read only transaction timeouts on macOS"

heifner:
Reverts AntelopeIO/spring#1141


heifner:
Issue #1145 still happening with this reverted.

PR #1148, opened 2025-02-06
Performance: Store packed block in signed_block

### LINKED ISSUE #1062: Performance: Store packed block in block_state
`packed_transaction` stores both the packed and unpacked transaction for quick access to both forms. The same could be done for `block_state` for the `block`.

One area this would help is: https://github.com/AntelopeIO/spring/pull/1061#discussion_r1880261439
Also would make sending out the packed block to `net_plugin` quicker.

heifner:
Cache the packed block data in `signed_block` so that it can be retrieved without having to `pack` it at time of serialization. 

Resolves #1062

### DIFF
--- a/libraries/libfc/include/fc/io/datastream.hpp
+++ b/libraries/libfc/include/fc/io/datastream.hpp
@@ -192,6 +192,8 @@ public:
    bool read( char* d, size_t s ) {
       if (ds.read(d, s)) {
          auto size = mirror.size();
+         if (mirror.capacity() < size + s)
+            mirror.reserve(std::bit_ceil(size + s));
          mirror.resize(size + s);
          memcpy(mirror.data() + size, d, s);
          return true;


libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -85,21 +85,26 @@ namespace eosio { namespace chain {
 
    using block_extension = block_extension_types::block_extension_t;
 
+   using signed_block_ptr = std::shared_ptr<const signed_block>;
+   using mutable_block_ptr = std::unique_ptr<signed_block>;
```
linh2931, line 89: Why not keep the original name `mutable_signed_block_ptr`? That's more clearer.

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -129,3 +142,20 @@ FC_REFLECT_DERIVED(eosio::chain::transaction_receipt, (eosio::chain::transaction
 FC_REFLECT(eosio::chain::additional_block_signatures_extension, (signatures));
 FC_REFLECT(eosio::chain::quorum_certificate_extension, (qc));
 FC_REFLECT_DERIVED(eosio::chain::signed_block, (eosio::chain::signed_block_header), (transactions)(block_extensions) )
+
+namespace fc::raw {
+   template <typename Stream>
+   void unpack(Stream& s, eosio::chain::signed_block& v) {
+      try {
+         if constexpr (requires { s.extract_mirror(); }) {
```
linh2931, line 150: Don't quite get this code. Why the `if constexpr` check? Why the `4096` below?

libraries/chain/controller.cpp
```diff
@@ -1590,7 +1578,7 @@ struct controller_impl {
 
                // blog.append could fail due to failures like running out of space.
                // Do it before commit so that in case it throws, DB can be rolled back.
-               blog.append( (*bitr)->block, (*bitr)->id(), irreversible_mode() ? f.get() : it++->get() );
+               blog.append( (*bitr)->block, (*bitr)->id(), (*bitr)->block->packed_signed_block() );
```
linh2931, line 1581: Now it is cleaner.

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -129,3 +142,20 @@ FC_REFLECT_DERIVED(eosio::chain::transaction_receipt, (eosio::chain::transaction
 FC_REFLECT(eosio::chain::additional_block_signatures_extension, (signatures));
 FC_REFLECT(eosio::chain::quorum_certificate_extension, (qc));
 FC_REFLECT_DERIVED(eosio::chain::signed_block, (eosio::chain::signed_block_header), (transactions)(block_extensions) )
+
```
linh2931, line 145: Why not put this code some where in fc?

tests/block_log.cpp
```diff
@@ -36,7 +36,7 @@ struct block_log_fixture {
       }
       else {
          eosio::chain::genesis_state gs;
-         log->reset(gs, std::make_shared<eosio::chain::signed_block>());
+         log->reset(gs, eosio::chain::signed_block::create_signed_block(eosio::chain::signed_block::create_mutable_block({})));
```
linh2931, line 39: Why this complicated?

libraries/chain/block_log.cpp
```diff
@@ -1094,9 +1095,13 @@ namespace eosio { namespace chain {
          }
 
          signed_block_ptr retry_read_block_by_num(uint32_t block_num) final {
-            auto ds = catalog.ro_stream_for_block(block_num);
-            if (ds)
-               return read_block(*ds, block_num);
+            uint64_t block_size = 0;
+
+            auto ds = catalog.ro_stream_and_size_for_block(block_num, block_size);
```
linh2931, line 1100: Not from your changes. I wish `ro_stream_and_size_for_block` returned `ds` and `size` explicitly.

libraries/chain/include/eosio/chain/block.hpp
heifner: Because in most cases it is not "signed" yet.

libraries/chain/include/eosio/chain/block.hpp
heifner: Because if `signed_block` is being unpacked with a `datastream_mirror` then we can use the `extract_mirror()` to populate `signed_block::packed_block`. If a different Stream type is being used, then wrap it in a `datastream_mirror` to capture the packed block data. The 4096 is used to reserve a size in the vector; open to removing the `+ 4096` if we think it might be wasteful.

tests/block_log.cpp
heifner: The idea is to make it difficult to misuse `signed_block`. We want `packed_block` data of `signed_block` to always be populated. Could create a method of `signed_block` that just does this, but since that would only be useful in tests I didn't.

libraries/chain/include/eosio/chain/block.hpp
heifner: Not sure I follow. `signed_block` is defined in `libraries/chain` which depends on `libraries/libfc`.
I did try to move the forward declare out of `fc`, but couldn't get it to work without the `fc` forward declare.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Cache the packed block data in `signed_block` so that it can be retrieved without having to `pack` it at time of serialization.
Note:end

libraries/chain/include/eosio/chain/block.hpp
linh2931: Can you add a comment? We have other cases where mutable and non-mutable use the same root like `mutale_db()` and `db()`.

libraries/chain/include/eosio/chain/block.hpp
linh2931: I meant `unpack(Stream& s, eosio::chain::signed_block& v)` is defined in `raw_fwd.hpp`, why not put the implementation there. I think about it more. It is also OK to remain here.

linh2931 (approved):

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -85,21 +85,28 @@ namespace eosio { namespace chain {
...
    public:
       signed_block() = default;
-      explicit signed_block( const signed_block_header& h ):signed_block_header(h){}
       signed_block( signed_block&& ) = default;
       signed_block& operator=(const signed_block&) = delete;
       signed_block& operator=(signed_block&&) = default;
-      signed_block clone() const { return *this; }
+      mutable_block_ptr clone() const { return std::unique_ptr<signed_block>(new signed_block(*this)); }
+      static mutable_block_ptr create_mutable_block(const signed_block_header& h) { return std::unique_ptr<signed_block>(new signed_block(h)); }
+      static signed_block_ptr  create_signed_block(mutable_block_ptr&& b) { b->pack(); return signed_block_ptr{std::move(b)}; }
```
greg7mdp, line 106: I don't see why these should be inside `struct signed_block`. even if `pack()`  has tobe made public.

It is also a little strange to me that `bytes packed_block;` is inside `signed_block`? Why not have a separate type `packed_signed_block`? Maybe it would require more changes?

libraries/libfc/include/fc/io/datastream.hpp
```diff
@@ -176,6 +176,38 @@ class datastream<Container, typename std::enable_if_t<(std::is_same_v<std::vecto
...
+public:
+   explicit datastream_mirror( DataStream& ds, size_t reserve = 0 ) : ds(ds) {
+      mirror.reserve(reserve);
+   }
+
+   void skip( size_t s ) { ds.skip(s); }
+   bool read( char* d, size_t s ) {
+      if (ds.read(d, s)) {
+         auto size = mirror.size();
+         mirror.resize(size + s);
```
greg7mdp, line 195: This is potentially very inefficient. I suggest `mirror.resize(std::bit_ceil(size + s));`

greg7mdp (commented):
I think I know why adding the:
`namespace eosio::chain { struct signed_block; }`

at the top of `block.hpp` didn't work. It is because some files like `block_state_legacy.cpp` include `block_state_legacy.hpp` first, and then `block_state_legacy.hpp` includes `block_header_state_legacy.hpp` *before* `block.hpp`.

So for files like this the `raw_fwd.hpp` is included before the forward delaration of `signed_block` is seen.

libraries/chain/include/eosio/chain/block.hpp
heifner: They are constructor like methods. Seems like they belong in `signed_block`, also allows `pack()` to be private without adding friends of the external functions.

We emit/signal `signed_block` out to plugins. `signed_block` is a well known and used type. I see this as an internal cached optimization to make `packed_signed_block()` efficient. Creating a different type means that we have to pass around a `packed_signed_block` everywhere we use `signed_block` today.

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -129,3 +144,20 @@ FC_REFLECT_DERIVED(eosio::chain::transaction_receipt, (eosio::chain::transaction
...
+
+namespace fc::raw {
+   template <typename Stream>
+   void unpack(Stream& s, eosio::chain::signed_block& v) {
+      try {
+         if constexpr (requires { s.extract_mirror(); }) {
+            fc::reflector<eosio::chain::signed_block>::visit( fc::raw::detail::unpack_object_visitor<Stream, eosio::chain::signed_block>( v, s ) );
+            v.packed_block = s.extract_mirror();
+         } else {
+            fc::datastream_mirror<Stream> ds(s, sizeof(eosio::chain::signed_block) + 4096);
```
greg7mdp, line 156: I don't think the `sizeof(eosio::chain::signed_block)` is very significant since most of the memory used is likely in the `deque<transaction_receipt>   transactions; ` whose `sizeof()` is probably around 16. I think I would just simplify to just 4096, but it probably doesn't matter much and is fine as is.

libraries/chain/block_log.cpp
```diff
@@ -661,10 +661,11 @@ namespace eosio { namespace chain {
 
          signed_block_ptr read_block_by_num(uint32_t block_num) final {
             try {
-               uint64_t pos = get_block_pos(block_num);
+               auto [ pos, size ] = get_block_position_and_size(block_num);
                if (pos != block_log::npos) {
                   block_file.seek(pos);
-                  return read_block(block_file, block_num);
+                  fc::datastream_mirror ds(block_file, size);
```
greg7mdp, line 667: Cool, here we dimension the mirror vector to exactly the right size.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3190,8 +3191,10 @@ namespace eosio {
             return true;
       }
 
-      auto ds = pending_message_buffer.create_datastream();
-      fc::raw::unpack( ds, which );
+      auto mb_ds = pending_message_buffer.create_datastream();
+      fc::raw::unpack( mb_ds, which );
+
+      fc::datastream_mirror ds(mb_ds, message_length);
```
greg7mdp, line 3197: cool, good dimensioning of the vector here as well. I see that my suggestion on the `resize()` was not as critical  as I originally thought.

gdb4713 (approved):

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -108,9 +115,17 @@ namespace eosio { namespace chain {
...
+
+      const bytes& packed_signed_block() const { assert(!packed_block.empty()); return packed_block; }
+
+   private:
+      friend struct block_state;
+      friend struct block_state_legacy;
+      template<typename Stream> friend void fc::raw::unpack(Stream& s, eosio::chain::signed_block& v);
+      void pack() { packed_block = fc::raw::pack( *this ); }
+
+      bytes packed_block; // packed this
```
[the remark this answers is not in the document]
heifner: That would be nice, but I'm not seeing how to do that without adding a mutex.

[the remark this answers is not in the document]
greg7mdp: What for? Doesn't `std::future` already include the functionality of a mutex + condition_variable?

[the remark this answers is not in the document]
heifner: Only one thread is allowed to access the `std::future`. Until that one thread has extracted the future data and stored it in the `packed_block` no other thread would be allowed to access the `packed_block` or even know if it is valid. 

`shared_future` is no better as each thread would need their own copy of the s`hared_future`. "Access to the same shared state from multiple threads is safe if each thread does it through its own copy of a shared_future object."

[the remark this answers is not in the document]
greg7mdp: More complicated than I thought. Thanks!

greg7mdp (approved):

gdb4713 (commented):
Ok

PR #1150, opened 2025-02-06
[1.1.0] Fix deferred trx processing

### LINKED ISSUE #1145: Spring 1.1 Misc Errors During Replay

heifner:
Currently with #1027 a long running trx is interrupted when the oc compile completes. This causes an issue with deferred trx processing in old blocks where deferred trxs exist. Disable interrupt of transaction when oc compile completes for implicit transactions. Implicit transactions include deferred `onerror` and `onblock`. Not interrupting `onblock` seems like an ok thing here as `onblock` is part of the main system contract.

Also stop() timer on oc interrupt so start() is always called on a stopped timer.
Also call `checktime()` after setup of cleanup in eos-vm-oc so that `set_expiration_callback(nullptr, nullptr);` is called on `checktime` exception.

chicken-dance mismatch still on this branch which is being tracked by https://github.com/AntelopeIO/spring/issues/1152

Resolves #1145

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2690,6 +2690,8 @@ struct controller_impl {
       } catch ( const boost::interprocess::bad_alloc& ) {
          throw;
       } catch( const fc::exception& e ) {
+         // apply_onerror for deferred trxs is implicit so interrupt oc not allowed
+         assert(e.code() != interrupt_oc_exception::code_value);
          handle_exception(e);
       } catch ( const std::exception& e ) {
          auto wrapper = fc::std_exception_wrapper::from_current_exception(e);
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2644,7 +2644,8 @@ struct controller_impl {
 
       transaction_checktime_timer trx_timer(timer);
       const packed_transaction trx( std::move( etrx ) );
-      transaction_context trx_context( self, trx, trx.id(), std::move(trx_timer), bb.action_receipt_digests().store_which(), start );
+      transaction_context trx_context( self, trx, trx.id(), std::move(trx_timer), bb.action_receipt_digests().store_which(),
+                                       start, transaction_metadata::trx_type::implicit );
 
       if (auto dm_logger = get_deep_mind_logger(trx_context.is_transient())) {
          dm_logger->on_onerror(etrx);
@@ -2817,7 +2818,8 @@ struct controller_impl {
       auto& bb = std::get<building_block>(pending->_block_stage);
 
       transaction_checktime_timer trx_timer( timer );
-      transaction_context trx_context( self, *trx->packed_trx(), gtrx.trx_id, std::move(trx_timer), bb.action_receipt_digests().store_which() );
+      transaction_context trx_context( self, *trx->packed_trx(), gtrx.trx_id, std::move(trx_timer), bb.action_receipt_digests().store_which(),
+                                       start, transaction_metadata::trx_type::scheduled );
       trx_context.leeway =  fc::microseconds(0); // avoid stealing cpu resource
       trx_context.block_deadline = block_deadline;
       trx_context.max_transaction_time_subjective = max_transaction_time;
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -165,9 +165,10 @@ struct eosvmoc_tier {
          }
 #endif
          // do not allow oc interrupt if no undo as the transaction needs to be undone to restart it.
-         // do not allow oc interrupt if implicit as deferred trx onerror execute outside the transaction retry in transaction_context.
+         // do not allow oc interrupt if implicit as deferred trx onerror execute as implicit outside the transaction
+         // retry in transaction_context. Also scheduled manage the undo stack explicitly.
          const bool allow_oc_interrupt = attempt_tierup && context.is_applying_block() &&
-                                         context.trx_context.has_undo() && !context.trx_context.is_implicit();
+                                         context.trx_context.has_undo() && !context.trx_context.is_implicit() && !context.trx_context.is_scheduled();
          auto ex = fc::make_scoped_exit([&]() {
             if (allow_oc_interrupt) {
                eos_vm_oc_compile_interrupt = false;
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -164,7 +164,10 @@ struct eosvmoc_tier {
             }
          }
 #endif
-         const bool allow_oc_interrupt = attempt_tierup && context.is_applying_block() && context.trx_context.has_undo();
+         // do not allow oc interrupt if no undo as the transaction needs to be undone to restart it.
+         // do not allow oc interrupt if implicit as deferred trx onerror execute outside the transaction retry in transaction_context.
+         const bool allow_oc_interrupt = attempt_tierup && context.is_applying_block() &&
+                                         context.trx_context.has_undo() && !context.trx_context.is_implicit();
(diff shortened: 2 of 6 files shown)


libraries/chain/controller.cpp
```diff
@@ -2644,7 +2644,8 @@ struct controller_impl {
 
       transaction_checktime_timer trx_timer(timer);
       const packed_transaction trx( std::move( etrx ) );
-      transaction_context trx_context( self, trx, trx.id(), std::move(trx_timer), bb.action_receipt_digests().store_which(), start );
+      transaction_context trx_context( self, trx, trx.id(), std::move(trx_timer), bb.action_receipt_digests().store_which(),
+                                       start, transaction_metadata::trx_type::implicit );
```
linh2931, line 2648: onblock is implicit. But I don't see it see anywhere.

heifner: https://github.com/AntelopeIO/spring/blob/f0d379a83309387d4e64e000823c928e20257396/libraries/chain/controller.cpp#L3031-L3031

It is set via `trx->get_trx_type()`.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -164,7 +164,11 @@ struct eosvmoc_tier {
             }
          }
 #endif
-         const bool allow_oc_interrupt = attempt_tierup && context.is_applying_block() && context.trx_context.has_undo();
+         // do not allow oc interrupt if no undo as the transaction needs to be undone to restart it.
+         // do not allow oc interrupt if implicit as deferred trx onerror execute as implicit outside the transaction
+         // retry in transaction_context. Also scheduled manage the undo stack explicitly.
+         const bool allow_oc_interrupt = attempt_tierup && context.is_applying_block() &&
+                                         context.trx_context.has_undo() && !context.trx_context.is_implicit() && !context.trx_context.is_scheduled();
```
linh2931, line 171: Do we need both `!context.trx_context.is_implicit()` and `!context.trx_context.is_scheduled()`? Is `!context.trx_context.is_scheduled()` sufficient?

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -164,7 +164,11 @@ struct eosvmoc_tier {
             }
          }
 #endif
-         const bool allow_oc_interrupt = attempt_tierup && context.is_applying_block() && context.trx_context.has_undo();
+         // do not allow oc interrupt if no undo as the transaction needs to be undone to restart it.
+         // do not allow oc interrupt if implicit as deferred trx onerror execute as implicit outside the transaction
```
linh2931, line 168: This wording `if implicit as deferred trx` is a little confusing. Implicit trx and deferred trx are different.

spoonincode:
Can you describe a call path that yielded the failure a little more? Because,
> Also call checktime() after setup of cleanup in eos-vm-oc so that set_expiration_callback(nullptr, nullptr); is called on checktime exception.

Couldn't this on its own have been a problem in prior releases?

heifner:
> Can you describe a call path that yielded the failure a little more? Because,
> 
> > Also call checktime() after setup of cleanup in eos-vm-oc so that set_expiration_callback(nullptr, nullptr); is called on checktime exception.
> 
> Couldn't this on its own have been a problem in prior releases?

This has been this way since 2.0.0. I noticed it during visual inspection of all calls to `set_expiration_callback`. Certainly seems like this could have been an issue at any point. Maybe there was something that prevent(ed) checktime from ever being triggered there.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
heifner: I reworked the comment. https://github.com/AntelopeIO/spring/pull/1150/commits/845576c52b8cabecf6b315dbc44988debd6b1924

linh2931 (approved):

spoonincode:
Sorry I should have written those as 2 separate questions more clearly. Can you describe the (or a) flow that was causing the problem?

heifner:
> Sorry I should have written those as 2 separate questions more clearly. Can you describe the (or a) flow that was causing the problem?

## onerror case

While applying a block a deferred transaction is processed. During the processing of the deferred transaction the action was interrupted via oc interrupt. This throws the `interrupt_oc_exception` transaction out of `apply` (line 191):

https://github.com/AntelopeIO/spring/blob/845576c52b8cabecf6b315dbc44988debd6b1924/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L185-L195

This `interrupt_oc_exception` is expected to be caught in `transaction_context`:
https://github.com/AntelopeIO/spring/blob/845576c52b8cabecf6b315dbc44988debd6b1924/libraries/chain/transaction_context.cpp#L371-L375

However, `onerror` calls `trx_context.execute_action` directly from the controller:
https://github.com/AntelopeIO/spring/blob/845576c52b8cabecf6b315dbc44988debd6b1924/libraries/chain/controller.cpp#L2673-L2673

Which means the `interrupt_oc_exception` would be caught here:
https://github.com/AntelopeIO/spring/blob/845576c52b8cabecf6b315dbc44988debd6b1924/libraries/chain/controller.cpp#L2693-L2693

This `interrupt_oc_exception` would cause the `onerror` to fail. But the failure is captured into a trace of the `onerror` (not thrown on out). This results in a `hard_fail` instead of a `soft_fail`. Also the trx_context is undone instead of squashed.

### deferred trx case

I'm not really sure why interrupting deferred trxs exec() would be a problem. I was thinking it was because it created its own undo_session. But it shouldn't when validating a block. I now don't see why interrupting a deferred trx would cause any issue. I does, because one of the runs had the not interrupt for implicit but do interrupt for scheduled and that chicken-dance failed with an invalid `soft_fail`.

libraries/chain/transaction_context.cpp
```diff
@@ -59,6 +59,7 @@ namespace eosio::chain {
       undo();
       *trace = transaction_trace{}; // reset trace
       initialize();
+      transaction_timer.stop();
```
heifner, line 62: This was not needed. The `undo()` call above calls `transaction_timer.stop()`. I can remove it, but having it here makes it more obvious that it is being called.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Fix deferred trx process to support replay-ability. 
Note:end

Issue #730, opened 2024-09-09
Verify block id as part of block header validation for savanna blocks


heifner:
For legacy blocks, the block id can't be verified until the `action_mroot` (merkle root of action receipts) is calculated which requires applying the transactions. For Savanna blocks the `action_mroot` is the `finality_mroot` (merkle root of the finality tree) which is known before applying the transactions. 

For savanna blocks, verify the block id as part of block header validation.

linh2931:
What about `transaction_mroot`? It is a part of block ID. It is unknown until transactions are applied.

heifner:
> What about `transaction_mroot`? It is a part of block ID. It is unknown until transactions are applied.

It is known before they are applied, see: https://github.com/AntelopeIO/spring/blob/main/libraries/chain/controller.cpp#L4127

linh2931:
Thanks. I was looking at the transaction_mroot for producing.

linh2931:
`finality_mroot` is known after `valid` structure built. To build `valid` structure, we need `action_mroot` (not the one in the header but the merkle root of action receipts) https://github.com/AntelopeIO/spring/blob/af51472f5157cd3181fde8dfc7233ab6c95c76c3/libraries/chain/block_state.cpp#L286
It is only available after transactions are applied. That's why 2 stages are currently used (step 7 in the Changes Required section): https://github.com/AntelopeIO/leap/issues/2080#:~:text=Similarly%2C%20when%20a,rejected%20as%20invalid.

Currently finality_tree root is checked by
https://github.com/AntelopeIO/spring/blob/af51472f5157cd3181fde8dfc7233ab6c95c76c3/libraries/chain/controller.cpp#L3791
block id is checked by
https://github.com/AntelopeIO/spring/blob/af51472f5157cd3181fde8dfc7233ab6c95c76c3/libraries/chain/controller.cpp#L3812

Not sure if we can move the checks any earlier.

I also think the following check is unnecessary as it uses the same value to check itself
https://github.com/AntelopeIO/spring/blob/af51472f5157cd3181fde8dfc7233ab6c95c76c3/libraries/chain/controller.cpp#L4152

linh2931:
But we can verify `action_mroot` (Finality Tree Root) as early as in `verify_basic_proper_block_invariants()`. I will do that.

PR #995, opened 2024-10-31
[1.0.3] Improve validation

linh2931:
Improve validation.

PR #1164, opened 2025-02-12
Validate finality mroot of Savanna blocks as a part of header validation

linh2931:
For Savanna blocks, add `finality mroot` validation to `verify_basic_proper_block_invariants()` as a part of header validation.

https://github.com/AntelopeIO/spring/issues/730 called for validating block ID as as part of block header validation. While investigating the issue, it is discovered that `new_qc_claim`s `block_num` and `finality mroot` could be validated during the header validation. `new_qc_claim`'s `block_num` validation was done by https://github.com/AntelopeIO/spring/pull/995.  Block ID validation is not possible at header validation stage as new `transaction_mroot` is not available before transactions are applied.

This PR adds `finality mroot` validation and a unit test for the validation.

Resolves https://github.com/AntelopeIO/spring/issues/730

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -4062,10 +4062,10 @@ struct controller_impl {
 
          // compute finality mroot using previous block state and new qc claim
          auto        computed_finality_mroot = prev.get_finality_mroot_claim(new_qc_claim);
-         const auto& supplied_action_mroot   = b->action_mroot;
-         EOS_ASSERT( computed_finality_mroot == supplied_action_mroot, block_validate_exception,
-                     "computed finality mroot (${computed}) does not match supplied action mroot ${supplied} by header extension. Block number: ${b}, block id: ${id}",
-                     ("computed", computed_finality_mroot)("supplied", supplied_action_mroot)("b", block_num)("id", id) );
+         const auto& supplied_finality_mroot  = b->action_mroot;
+         EOS_ASSERT( computed_finality_mroot == supplied_finality_mroot, block_validate_exception,
+                     "computed finality mroot (${computed}) does not match supplied finality mroot ${supplied} by header extension. Block number: ${b}, block id: ${id}",
+                     ("computed", computed_finality_mroot)("supplied", supplied_finality_mroot)("b", block_num)("id", id) );
       }
 
       return std::optional{qc_proof};
--- a/unittests/block_tests.cpp
+++ b/unittests/block_tests.cpp
@@ -441,7 +441,7 @@ BOOST_FIXTURE_TEST_CASE( invalid_action_mroot_test, tester )
                            [] (const fc::exception &e)->bool {
                               return e.code() == block_validate_exception::code_value &&
                                      e.to_detail_string().find("computed finality mroot") != std::string::npos &&
-                                     e.to_detail_string().find("does not match supplied action mroot") != std::string::npos;
+                                     e.to_detail_string().find("does not match supplied finality mroot") != std::string::npos;
                            });
 }
 


libraries/chain/controller.cpp
```diff
@@ -4066,6 +4066,15 @@ struct controller_impl {
                   "QC is_strong (${s1}) in block extension does not match is_strong_qc (${s2}) in header extension. Block number: ${b}",
                   ("s1", qc_proof.is_strong())("s2", new_qc_claim.is_strong_qc)("b", block_num) );
 
+      if (prev.valid) {
+         // compute finality mroot using previous block state and new qc claim
+         auto computed_finality_mroot = prev.get_finality_mroot_claim(new_qc_claim);
+         auto supplied_action_mroot = b->action_mroot;
```
heifner, line 4072: No need to make a copy.

libraries/chain/controller.cpp
```diff
@@ -4066,6 +4066,15 @@ struct controller_impl {
                   "QC is_strong (${s1}) in block extension does not match is_strong_qc (${s2}) in header extension. Block number: ${b}",
                   ("s1", qc_proof.is_strong())("s2", new_qc_claim.is_strong_qc)("b", block_num) );
 
+      if (prev.valid) {
```
heifner, line 4069: I may have missed something here, but this doesn't look thread safe. This can be filled in when validating the block which could be happening at the same time this is running on the net thread. Can you instead check `validated` and then `assert(prev.valid)`? Is it correct that `prev.valid` is set when `validated==true`?

libraries/chain/controller.cpp
```diff
@@ -4066,6 +4066,15 @@ struct controller_impl {
...
+         // compute finality mroot using previous block state and new qc claim
+         auto computed_finality_mroot = prev.get_finality_mroot_claim(new_qc_claim);
+         auto supplied_action_mroot = b->action_mroot;
+         EOS_ASSERT( computed_finality_mroot == supplied_action_mroot, block_validate_exception,
+                     "computed finality mroot (${computed}) does not match supplied action mroot ${supplied} by header extension. Block number: ${b}",
```
heifner, line 4074: Should include block id as well.

heifner (changes requested):

unittests/forked_tests_savanna.cpp
```diff
@@ -115,8 +115,10 @@ BOOST_FIXTURE_TEST_CASE(fork_with_bad_block_savanna, savanna_cluster::cluster_t)
          if (j <= i) {
             auto copy_b = std::make_shared<signed_block>(b->clone());
             if (j == i) {
-               // corrupt this block (forks[j].blocks[j] is corrupted)
-               copy_b->action_mroot._hash[0] ^= 0x1ULL;
+               // Corrupt this block (forks[j].blocks[j] is corrupted).
+               // Do not corrupt the block by modifying action_mroot, as action_mroot is checked
+               // by block header validation, _nodes[0].push_block(b) would fail.
+               copy_b->confirmed++;
```
linh2931, line 121: @greg7mdp Could you double check if this change is correct or not?

heifner: Seems like any corrupt will work, `confirmed` is as good as anything else.

libraries/chain/controller.cpp
linh2931: It was used twice; that's why I created a copy. But it is used in the same EOS_ASSERT; will remove it for simplicity.

heifner: You could use `const auto& supplied_action_mroot = b->action_mroot;`

libraries/chain/controller.cpp
linh2931: `valid` https://github.com/AntelopeIO/spring/blob/e43fd1e468276e1791471159e2e75503f13db6aa/libraries/chain/include/eosio/chain/block_state.hpp#L77 is for Finality Merkle Tree while `validated`https://github.com/AntelopeIO/spring/blob/e43fd1e468276e1791471159e2e75503f13db6aa/libraries/chain/include/eosio/chain/block_state.hpp#L81 is mostly for forkbd. They are independent. Since I am using previous block's `valid` structure, it should be already constructed. I don't see a thread safe problem.

heifner: During the transition to Savanna it is modified here:
https://github.com/AntelopeIO/spring/blob/53ad484eb68863d8aac86151e0c71c18de7fcf63/libraries/chain/controller.cpp#L1385-L1385
https://github.com/AntelopeIO/spring/blob/53ad484eb68863d8aac86151e0c71c18de7fcf63/libraries/chain/controller.cpp#L1406-L1406

Even outside Savanna transition it is modified here:
https://github.com/AntelopeIO/spring/blob/53ad484eb68863d8aac86151e0c71c18de7fcf63/libraries/chain/controller.cpp#L804-L804
This can happen while the net threads are accessing `prev.valid`.

I believe it is correct that `valid` will be filled in when `validated` is true. Since `validated` is atomic you could check that value here instead of `valid`.

libraries/chain/controller.cpp
linh2931: This https://github.com/AntelopeIO/spring/blob/5247431b84f2c6d712494e650d71d3169fe04667/libraries/chain/controller.cpp#L4200 checks block id.

unittests/forked_tests_savanna.cpp
linh2931: Now it occurs to me, should we check `confirmed` is 0 for Savanna blocks?

libraries/chain/controller.cpp
heifner: Sorry, I meant include the block id in the error message.

unittests/forked_tests_savanna.cpp
heifner: Yes. It is verified as part of block id, but wouldn't hurt to check it early.

greg7mdp: > @greg7mdp Could you double check if this change is correct or not?

Yes, seems fine to me as well.

libraries/chain/controller.cpp
linh2931: Thanks. That will avoid copying 32 bytes.

unittests/forked_tests_savanna.cpp
linh2931: Thanks.

heifner (approved):

unittests/forked_tests_savanna.cpp
linh2931: I ended up not adding the check of `confirmed == 0`. With the check turned on, it is hard to corrupt a block in `fork_with_bad_block_savanna`. As `confirmed` is not used in Savanna and checked later on by block ID, it is OK not being checked in the header validation.

libraries/chain/controller.cpp
```diff
@@ -4054,6 +4054,20 @@ struct controller_impl {
...
+      // `valid` structure can be modified while this function is running on net thread.
+      // Use is_valid() instead. It uses atomic `validated` and when it is true, `valid`
+      // has been constructed.
+      if (prev.is_valid()) {
+         assert(prev.valid);
+
+         // compute finality mroot using previous block state and new qc claim
+         auto        computed_finality_mroot = prev.get_finality_mroot_claim(new_qc_claim);
+         const auto& supplied_action_mroot   = b->action_mroot;
+         EOS_ASSERT( computed_finality_mroot == supplied_action_mroot, block_validate_exception,
```
greg7mdp, line 4066: We should probably use either `finality_mroot` or `action_mroot`, but not both.

greg7mdp (approved):

libraries/chain/controller.cpp
linh2931: Will change to use `finality_mroot` as it is a Savanna thing.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Validate finality mroot of Savanna blocks as a part of header validation.
Note:end

PR #1165, opened 2025-02-13
[1.1.0] disable boringssl debug symbols on release builds; restore reproducible builds in CI

spoonincode:
boringssl's upstream cmake file always adds `-ggdb` to the compile which damages our reproducibility due to a path difference. We might want to rectify the path difference if possible and/or masquerade the difference via compiler options, but disabling the debug symbol generation is far easier by negating the `-ggdb` with `-g0`

### DIFF
--- a/libraries/libfc/libraries/boringssl/CMakeLists.txt
+++ b/libraries/libfc/libraries/boringssl/CMakeLists.txt
@@ -3,6 +3,13 @@ target_compile_options(fipsmodule PRIVATE -Wno-error)
 target_compile_options(crypto PRIVATE -Wno-error)
 target_compile_options(decrepit PRIVATE -Wno-error)
 
+# boringssl's forced-on debug symbols interfere with reproducibility due to path differences; disable on Release builds
+if(CMAKE_BUILD_TYPE STREQUAL "Release")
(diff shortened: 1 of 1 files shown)


libraries/libfc/libraries/boringssl/CMakeLists.txt
```diff
@@ -3,6 +3,13 @@ target_compile_options(fipsmodule PRIVATE -Wno-error)
 target_compile_options(crypto PRIVATE -Wno-error)
 target_compile_options(decrepit PRIVATE -Wno-error)
 
+# boringssl's forced-on debug symbols interfere with reproducibility due to path differences; disable on Release builds
+if(CMAKE_BUILD_TYPE STREQUAL "Release")
```
spoonincode, line 7: Checking `CMAKE_BUILD_TYPE` like this isn't ideal at all and maybe even borderline anti-pattern. But I'm not really sure what other indicator to use of "no thanks I don't want debug symbols"

greg7mdp (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Disable boringssl debug symbols on release builds.
Note:end

Issue #1000, opened 2024-10-31
Capture Core Dump for tests failing due to timeout


bhazzard:
Core Dump for tests failing due to timeout will help debug why they are hanging.

Examples:
https://github.com/AntelopeIO/spring/issues/919

spoonincode:
There was some work performed on this in on a branch with head 30667fd

PR #1168, opened 2025-02-14
add common method for registering different test types

spoonincode:
This is a change I was _required_ to make for #1000, but with that effort being buried I thought I'd still cherry pick this change out for opinions on taking it stand alone.

This removes a lot of boilerplate when declaring tests (especially LR & NP tests) by removing, for example,
```cmake
add_test(NAME nodeos_sanity_test COMMAND tests/nodeos_run_test.py -v --sanity-test ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
set_property(TEST nodeos_sanity_test PROPERTY LABELS nonparallelizable_tests)
```
and replacing it with
```cmake
add_np_test(NAME nodeos_run_test COMMAND tests/nodeos_run_test.py -v ${UNSHARE})
```

There is `add_p_test()`, `add_np_test()`, `add_lr_test()`, and `add_wasmspec_test()` that all do what you'd expect. Any place that continues to just use `add_test()` (like in a submodule) will still just keep working as expected.

`add_p_test()` is probably the most questionable, as it adds little utility over just `add_test()` since the parallelizable tests don't need a label. Though it does remove the `WORKING_DIRECTORY ${CMAKE_BINARY_DIR}` boilerplate used in spots.

Certainly a big fear with a change like this is inadvertently disabling or altering a test somehow. To ensure no test left behind, one thing we can look at is the number of tests printed in the CI run compared to last run on `main`. This is easiest to compare for parallelizable tests since there is a number printed at the end. A `grep | wc` can be used on the log files of  NP & LR tests to ensure the count matches up.

Another more through check is to create json output from ctest before and after the change,
```sh
ctest --show-only=json-v1 > before.json
ctest --show-only=json-v1 > after.json
```
Unfortunately there is a bunch of changes in the json due to "backtrace" related stuff (a graph of tests), so filter that out,
```sh
cat before.json | jq 'del(.version)|del(.backtraceGraph)|del(.tests[].backtrace)' > before-filtered.json
cat after.json | jq 'del(.version)|del(.backtraceGraph)|del(.tests[].backtrace)' > after-filtered.json
```
Now can do a `diff -u1 before-filtered.json after-filtered.json`. Unfortunately again this still a bit large because of changes such as
```diff
@@ -46872,3 +46872,3 @@
           "name": "WORKING_DIRECTORY",
-          "value": "/spring/build/unittests/wasm-spec-tests/generated-tests"
+          "value": "/spring/build"
         }
```
At this point I just scrolled through the output and manually verified the only thing in the diff was `WORKING_DIRECTORY` changes (due to the common `WORKING_DIRECTORY "${CMAKE_BINARY_DIR}"` now present that wasn't the case with every test in the past). Clearly the working directory being switch from, for example, `/spring/build/unittests/wasm-spec-tests/generated-tests` to `/spring/build` hasn't affected the ability of the test to run.

One other minor annoyance with this change I noticed is that the indirection removes the ▶️ from CLion's editor next to the `add_test()` to easily run the test.


heifner (approved):
Looks like an improvement to me.

CMakeModules/AddTestHelpers.cmake
```diff
@@ -0,0 +1,39 @@
...
+   if(arg_COST)
+      set_tests_properties("${arg_NAME}" PROPERTIES COST ${arg_COST})
+   endif()
+   if(arg_TIMEOUT)
+      set_tests_properties("${arg_NAME}" PROPERTIES TIMEOUT ${arg_TIMEOUT})
+   endif()
+endfunction()
+
+function(add_p_test)
+   cmake_parse_arguments(PARSE_ARGV 0 arg "" "NAME;COST;TIMEOUT" "COMMAND")
```
linh2931, line 15: Is this `cmake_parse_arguments` duplicate as `setup_test_common` calls it too?

unittests/CMakeLists.txt
```diff
@@ -96,11 +96,11 @@ foreach(TEST_SUITE ${UNIT_TESTS}) # create an independent target for each test s
...
-        add_test(NAME ${TRIMMED_SUITE_NAME}_unit_test_${RUNTIME} COMMAND unit_test --run_test=${SUITE_NAME} --report_level=detailed --color_output -- --${RUNTIME})
+        add_p_test(NAME ${TRIMMED_SUITE_NAME}_unit_test_${RUNTIME} COMMAND unit_test --run_test=${SUITE_NAME} --report_level=detailed --color_output -- --${RUNTIME})
         # add a duplicate test to run without fsgsbase instructions for a limited number of OC wasm tests
         if(RUNTIME STREQUAL "eos-vm-oc" AND TRIMMED_SUITE_NAME MATCHES "^wasm_part")
-           add_test(NAME ${TRIMMED_SUITE_NAME}_unit_test_${RUNTIME}-nofsgs COMMAND unit_test --run_test=${SUITE_NAME} --report_level=detailed --color_output -- --${RUNTIME})
-           set_tests_properties(${TRIMMED_SUITE_NAME}_unit_test_${RUNTIME}-nofsgs PROPERTIES ENVIRONMENT "SPRING_DISABLE_FSGSBASE=1" COST 5000)
+           add_p_test(NAME ${TRIMMED_SUITE_NAME}_unit_test_${RUNTIME}-nofsgs COMMAND unit_test --run_test=${SUITE_NAME} --report_level=detailed --color_output -- --${RUNTIME} COST 5000)
+           set_tests_properties(${TRIMMED_SUITE_NAME}_unit_test_${RUNTIME}-nofsgs PROPERTIES ENVIRONMENT "SPRING_DISABLE_FSGSBASE=1")
```
linh2931, line 103: Why is `COST 5000` moved up?

CMakeModules/AddTestHelpers.cmake
spoonincode: good eye -- I have to admit I'm not 100% sure on the semantics of `cmake_parse_arguments` so I went with the conservative choice to just repeat all arguments. During development I did test that just `cmake_parse_arguments(PARSE_ARGV 0 arg "" "NAME" "")` worked but I wasn't really sure it worked right in all cases.

That said -- the item you placed the comment on, `add_p_test` does not seem to need to parse anything since unlike `add_wasmspec_test`/`add_np_test`/`add_lr_test` it does not need even `NAME`. So it looks fine to remove this line from `add_p_test`, I will try.

unittests/CMakeLists.txt
linh2931: Thanks for your detailed explanation.

PR #1178, opened 2025-02-18
Switch to using `eos-system-contracts` in libtester tests

### LINKED ISSUE #1177: Switch to using `eos-system-contracts` in libtester tests
Now that we are not updating the `reference-contracts` repo anymore, switch the libtester tests to using `eos-system-contracts`.

greg7mdp:
Resolves #1177.

Now that we are not updating the `reference-contracts` repo anymore, switch the libtester tests to using `eos-system-contracts`.

This reverts commit 3fdc35407f9eb3ee2b08e4982052629123a5ce6a.

### DIFF
--- a/.cicd/defaults.json
+++ b/.cicd/defaults.json
@@ -3,7 +3,7 @@
        "target":"4.1",
        "prerelease":true
     },
-    "referencecontracts":{
-       "ref":"main"
+    "eossystemcontracts":{
+       "ref":"release/3.2"
     }
  }
--- a/.cicd/defaults.json
+++ b/.cicd/defaults.json
@@ -4,6 +4,6 @@
        "prerelease":true
     },
     "eossystemcontracts":{
-       "ref":"release/3.2"
+       "ref":"main"
     }
  }
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,8 +18,8 @@ on:
         - default
         - true
         - false
-      override-reference-contracts:
-        description: 'Override reference contracts ref'
+      override-eos-system-contracts:
+        description: 'Override eos-system-contracts ref'
         type: string
 
 permissions:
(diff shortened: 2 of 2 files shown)


.cicd/defaults.json
```diff
@@ -3,7 +3,7 @@
        "target":"4.1",
        "prerelease":true
     },
-    "referencecontracts":{
+    "eossystemcontracts":{
        "ref":"main"
```
spoonincode, line 7: Any reason to not set it back to a release too?

greg7mdp: I feel that `main` is better, but maybe I'm not understanding the process well (my first time doing this)

For example I am working on some changes in `eos-system-contracts` , and for the tests to pass I'll update `defaults.json` to use my branch which has the changes. After my branch is merged to main, I think we'll have to switch `defaults.json` back to main for the tests to continue passsing.

spoonincode: I think the easiest way to imagine a problem with `main` is if spring's `release/1.2` is using `main`, and spring's `main` is using `main`, and then system contract's `main` changes in such a way that it requires spring's `main`. Now it's impossible for any PRs to `release/1.2` to pass since they'll still fetch contract's `main` that won't work on `release/1.2`.

So it's probably preferable to stick with something stable unless needing a branch for some reason, and at the very least when we prepare to branch a release branch we should try to pin it down.

Keep in mind too that reference contracts has no release branches, so that's a big reason it was left to languish as `main` even though that's not ideal.

All that said, eos-system-contracts is using `main` for spring and `main` for cdt so :woman_shrugging:

spoonincode (approved):

.cicd/defaults.json
spoonincode: No it does not require a hard fork. It could simply be a new interface that libtester provides that contracts tests begin to use.

PR #1181, opened 2025-02-19
[1.1.1] fix flat to pruned block log conversion

spoonincode:
Conversion of non-trivially long block logs from flat to pruned has been broken for quite some time. I suspect as early as leap 4.0 via AntelopeIO/leap#532. Unfortunately despite there being a lot of tests around pruned logs, none of them were "big" enough to stumble on the problem because they'd effectively only prune new entries -- not existing entries. A new test has been added that adds enough blocks to demonstrate the failure. And the fix is as simple as ensuring the blog index is opened before performing pruning.

Combination of simple fix + reported in wild: should go to stable.

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -1160,6 +1160,7 @@ namespace eosio { namespace chain {
             // convert from  non-pruned block log to pruned if necessary
             if (!preamble.is_currently_pruned()) {
                block_file.open(fc::cfile::update_rw_mode);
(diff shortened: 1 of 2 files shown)


heifner (approved):

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Fix to allow very large block logs to convert from non-pruned block log to pruned if necessary.
Note:end

PR #1108, opened 2025-01-17
P2P: Fix duplication connection logic

heifner:
The duplication logic for P2P connections was overly complicated. The `node_id` should be more than sufficient to distinguish between two connections.

Resolves #1092

PR #1188, opened 2025-02-20
[1.1.1 -> main] Revert "P2P: Fix duplication connection logic

heifner:
Reverts #1108 

The simplified de-dup logic of #1108 allows for both sides to determine the other is duplicate if the connections happen at the same time. The previous more complicated logic does allow duplicate connections in about half the cases, but prevents the race-condition of both sides deciding the other is a duplicate.

I wonder with the new block-nack feature and future limit of connections if we should even attempt to remove duplicate connections. In one sense, it seems like if a user wants to configure both sides to connect to each other that we should just allow it. It is obviously an important connection. Also all the reconnects and determination of duplicate and disconnect is in itself wasteful.

Note the previous version support here is not really needed, but just kept this as a straight revert.

Merges `release/1.1` into `main` including #1186

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3416,6 +3416,7 @@ namespace eosio {
       peer_fork_db_head_block_num = msg.fork_db_head_num;
       fc::unique_lock g_conn( conn_mtx );
       last_handshake_recv = msg;
+      auto c_time = last_handshake_sent.time;
       g_conn.unlock();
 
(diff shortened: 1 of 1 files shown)


spoonincode (approved):

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: P2P
summary: Revert simplified duplicate connection logic and revert to previous logic to prevent duplicate connections.
Note:end

Issue #1194, opened 2025-02-23
[regression] /v1/chain/get_info is not available on non-chain API endpoints in v1.1.0


matthewdarwin:
/v1/chain/get_info should always be available.

In v1.0.x, with the following configuration

```
http-server-address = http-category-address
http-category-address = chain_ro,:8888
http-category-address = chain_rw,:8888
http-category-address = db_size,localhost:8889
http-category-address = net_ro,localhost:8889
http-category-address = net_rw,localhost:8889
http-category-address = producer_ro,localhost:8889
http-category-address = producer_rw,localhost:8889
http-category-address = snapshot,localhost:8889
```

You could run  locahost:8889/v1/chain/get_info http request and it worked
In 1.1.0 it does not.

```
$ curl localhost:8889/v1/chain/get_info
{"code":404,"message":"Not Found","error":{"code":0,"name":"exception","what":"unspecified","details":[{"message":"Unknown Endpoint","file":"beast_http_session.hpp","line_number":186,"method":"handle_request"}]}}#   ```

PR #1197, opened 2025-02-24
[1.1.1 -> main] Fix get_info not available on non-chain API endpoints

linh2931:
`get_info` was mistakenly removed from `node` category (https://github.com/AntelopeIO/spring/pull/920) and caused it not available on non-chain API endpoints

This PR
* Moves it back to `node` category
* Fixes a bug of `plugin_http_api_test` that it does not exit with an exit value. That causes `plugin_http_api_test` to always appear successful even though it fails
* Add a test of `get_info` on non-chain API endpoint

Merges release/1.1 into main including https://github.com/AntelopeIO/spring/pull/1195

Resolves https://github.com/AntelopeIO/spring/issues/1194

### DIFF
--- a/plugins/chain_api_plugin/chain_api_plugin.cpp
+++ b/plugins/chain_api_plugin/chain_api_plugin.cpp
@@ -127,7 +127,7 @@ void chain_api_plugin::plugin_startup() {
 
    // Run get_info on http thread only
    _http_plugin.add_async_api({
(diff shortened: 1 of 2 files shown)


heifner (approved):

spoonincode (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Restore `get_info` to non-chain API endpoints.
Note:end

Issue #1198, opened 2025-02-24
trace_plugin_test fails to keep test log if the test fails


linh2931:
`trace_plugin_test.py` attempts to keep test log when test fails, but it fails to keep the log due to a bug. The log is useful for debugging test failures.

PR #1200, opened 2025-02-24
[1.1.1 -> main] Fix log file not kept by trace_plugin_test when the test fails

linh2931:
Fix the issue that `trace_plugin_test` fails to keep log file if the test fails.

Merges release/1.1 into main including https://github.com/AntelopeIO/spring/pull/1199

Resolves https://github.com/AntelopeIO/spring/issues/1198


spoonincode (approved):

ericpassmore:
Note:start
category: Tests
component: Other
summary: Retain trace log file after test failure.
Note:end

PR #1204, opened 2025-02-25
[1.1.2] P2P, Prometheus: fix threading issue

heifner:
`connection_start_time` accessed from connection strand and also from any net thread when Prometheus enabled.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -920,7 +920,7 @@ namespace eosio {
       std::string                      remote_endpoint_ip     GUARDED_BY(conn_mtx);
       boost::asio::ip::address_v6::bytes_type remote_endpoint_ip_array GUARDED_BY(conn_mtx);
 
-      std::chrono::nanoseconds         connection_start_time{0};
+      std::atomic<std::chrono::nanoseconds>  connection_start_time;
 
       // block nack support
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -920,7 +920,7 @@ namespace eosio {
       std::string                      remote_endpoint_ip     GUARDED_BY(conn_mtx);
       boost::asio::ip::address_v6::bytes_type remote_endpoint_ip_array GUARDED_BY(conn_mtx);
 
-      std::chrono::nanoseconds         connection_start_time{0};
+      std::atomic<std::chrono::nanoseconds>  connection_start_time;
```
linh2931, line 923: You can copy the one line PR description to here as comment too.

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
heifner: The comment would quickly be out of date. I want to also access this in connection status which is http threads.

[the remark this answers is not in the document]
heifner: It was found via inspection. I didn't see this fail. 
`start_session` is called on re-connect which could happen at the same time as the connection status running. The likely hood is extremely slim, but it is not technically correct.

ericpassmore:
Note:start
category: P2P
component: Internal
summary: Prometheus: fix threading issue.
Note:end

PR #1206, opened 2025-02-26
Fix poor hash function for `chain::name`

### LINKED ISSUE #1185: Fix poor hash function for `chain::name`
The [hash]( https://github.com/AntelopeIO/spring/blob/767b3489281bb66653ea8887e0ae8592c9e8e9d0/libraries/chain/include/eosio/chain/name.hpp#L179-L187) function provided for `chain::name` is poor. 

It does `std::hash<uint64_t>()(name.to_uint64_t())`. The stdlib implementation on my machine returns the `uint64_t` value unchanged.

The issue is that eosio names shorter than 13 characters will have all zeroes in least significant bits. So for example all ten character names will have the 14 least significant bits set to zero.

The worst case is for unordered containers which use a `power of two` array size to store values, and compute the index of the entry with a mask to avoid the more expensive modulo operation. For these we will have a 100% collision rate (as long as the array size is `< 2^14`).

Even for unordered containers which compute the bucket index using `hash_value % bucket_count()`, the collision rate would be very high.

greg7mdp:
Resolves #1185.

### DIFF
--- a/libraries/chain/include/eosio/chain/name.hpp
+++ b/libraries/chain/include/eosio/chain/name.hpp
@@ -179,9 +179,16 @@ namespace eosio::chain {
 namespace std {
    template<> struct hash<eosio::chain::name> : private hash<uint64_t> {
       typedef eosio::chain::name argument_type;
-      size_t operator()(const argument_type& name) const noexcept
-      {
-         return hash<uint64_t>::operator()(name.to_uint64_t());
+
+      size_t operator()(const argument_type& name) const noexcept {
+         assert(sizeof(size_t) == sizeof(uint64_t));
+         uint64_t v         = name.to_uint64_t();
+         auto     num_zeros = std::countr_zero(v);
+         if (num_zeros >= 4) {
+            uint64_t shift = 4 + ((num_zeros - 4) / 5) * 5;
+            return v >> shift;
+         }
+         return v;
       }
    };
 };


heifner (changes requested):

libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -179,9 +179,16 @@ namespace eosio::chain {
 namespace std {
    template<> struct hash<eosio::chain::name> : private hash<uint64_t> {
       typedef eosio::chain::name argument_type;
-      size_t operator()(const argument_type& name) const noexcept
-      {
-         return hash<uint64_t>::operator()(name.to_uint64_t());
+
+      size_t operator()(const argument_type& name) const noexcept {
+         assert(sizeof(size_t) == sizeof(uint64_t));
```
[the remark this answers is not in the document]
greg7mdp: Thanks!

heifner (approved):

libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -179,9 +179,16 @@ namespace eosio::chain {
...
+
+      size_t operator()(const argument_type& name) const noexcept {
+         static_assert(sizeof(size_t) == sizeof(uint64_t));
+         uint64_t v         = name.to_uint64_t();
+         auto     num_zeros = std::countr_zero(v);
+         if (num_zeros >= 4) {
+            uint64_t shift = 4 + ((num_zeros - 4) / 5) * 5;
+            return v >> shift;
+         }
+         return v;
```
spoonincode, line 191: Seems overly complex to me -- a byteswap solves the same problem and resolves to just a single x86 instruction. Function can just be the static assert and `return bswap_64(name.to_int64_t())`

greg7mdp: Great, thanks! https://github.com/AntelopeIO/spring/pull/1206/commits/a4c82895f541f130ab51194578ccb6571a3e0549

heifner (approved):

spoonincode (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Improve hash function for `chain::name`
Note:end

PR #1207, opened 2025-02-27
P2P: 1MB send/receive buffer

heifner:
Set a 1MB send/receive buffer for P2P similar to what we have for SHiP.


spoonincode (approved):

ericpassmore:
Note:start
category: Other
component: P2P
summary: Increase send/receive buffer to 1Mb.
Note:end

PR #1208, opened 2025-02-27
P2P: Add new peer-log-format options

heifner:
- Add new options `_peer` and `_nver` to `peer-log-format`.
  - `_peer` - The configured `p2p-peer-address` or reported address if incoming.
  - `_nver` - The P2P network protocol version of the connection. e.g. 9 is `proto_savanna`, 10 is `proto_block_nack`.
  - Changed the default to use `_peer` instead  of `_name` as `_peer` is easier to correlate with configured `p2p-peer-address`.
- Add log when not sending a block because connection is busy sending already.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2810,7 +2810,6 @@ namespace eosio {
                   boost::asio::post(cp->strand, [cp, send_buffer{std::move(send_buffer)}, bnum]() {
                      cp->latest_blk_time = std::chrono::steady_clock::now();
                      peer_dlog( cp, "bcast block_notice ${b}", ("b", bnum) );
-                     cp->enqueue_buffer( msg_type_t::block_notice_message, queued_buffer::queue_t::general, send_buffer, 0, no_reason );
                      cp->enqueue_buffer( msg_type_t::block_notice_message, std::nullopt, queued_buffer::queue_t::general, send_buffer, no_reason );
                   });
                   return;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1657,20 +1661,23 @@ namespace eosio {
...
+      do_queue_write(block_num);
    }
 
    // called from connection strand
-   void connection::do_queue_write() {
-      if( !buffer_queue.ready_to_send(connection_id) || closed() )
+   void connection::do_queue_write(block_num_type block_num) {
+      if( !buffer_queue.ready_to_send(connection_id) || closed() ) {
+         if (block_num > 0) {
+            peer_dlog(this, "not sending block yet ${n}", ("n", block_num) );
```
linh2931, line 1671: It would be more helpful if you can print out the exact reason: not ready_to send or closed.

Why only log when block_num > 0?

Not sure `c->do_queue_write(0);` below is a good choice. Maybe you can use optional if block_num is not always available.

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp

linh2931 (approved):

greg7mdp (approved):
I think it might be cleaner to have another `enqueue_buffer` version, which doesn't take a `std::optional<block_num_type> block_num`, but which calls the other one with `std::nullopt`. But fine as it is if you prefer.

Issue #1212, opened 2025-02-27
SC: Introduce `sync_call` Protocol Feature


linh2931:
The task includes definition and activation of `sync_call` protocol feature, and a basic test to show `sync_call` protocol feature can be indeed activated.

PR #1222, opened 2025-02-28
SC: SYNC_CALL protocol feature definition and registration

linh2931:
* Define and register SYNC_CALL protocol feature.
* Add a test to verify the definition is correct, dependency is met, and SYNC_CALL can be activated.
* Fix a fragility of former preactivate_all_but_disable_deferred_trx such that no code changes are required when future protocol features are added
* Rename `full_except_do_not_disable_deferred_trx` to `full_prior_to_disable_deferred_trx` and `preactivate_all_but_disable_deferred_trx` to `preactivate_all_prior_to_disable_deferred_trx` to better reflect what they are.

Resolves https://github.com/AntelopeIO/spring/issues/1212.


libraries/chain/protocol_feature_manager.cpp
```diff
@@ -363,6 +363,17 @@ host function call will trigger a transition to the Savanna consensus algorithm.
...
+         (  builtin_protocol_feature_t::sync_call, builtin_protocol_feature_spec{
+            "SYNC_CALL",
+            fc::variant("49f6fbc4fee045bc5c8b09a96b2b5c096afdc320ac29b54e0f3805c660b88483").as<digest_type>(),
+            // SHA256 hash of the raw message below within the comment delimiters (exclude newline after /*) (do not modify message below).
+/*
+Builtin protocol feature: SYNC_CALL
+
+Enables synchronous calls to functions in other and own contracts.
+*/
+            {}
```
heifner, line 375: I think this should depend on `builtin_protocol_feature_t::savanna`. If we ever mint a new genesis we want to be able to know that this can't be enabled without all the other protocol features.

heifner (changes requested):

libraries/chain/protocol_feature_manager.cpp
linh2931: Thanks. I was thinking about the dependency on SAVANNA but thought it could work standalone. But it is safer to have all previous protocol features enabled. Will add the dependency.

heifner (approved):

heifner (approved):

ericpassmore:
Note:start
category: Other
component: Contracts
summary: Define and register SYNC_CALL protocol feature.
Note:end

spoonincode (approved):

PR #1223, opened 2025-02-28
P2P: Avoid thread hop on incoming blocks

heifner:
Optimize processing of incoming blocks by removing an unneeded thread hop on nominal case.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -430,6 +430,9 @@ namespace eosio {
       /** @} */
 
       alignas(hardware_destructive_interference_sz)
+      fc::mutex                             accept_blk_mtx; // avoid extra work of validating same block at same time
+
+      alignas(hardware_destructive_interference_sz)
       fc::mutex                             expire_timer_mtx;
       boost::asio::steady_timer             expire_timer GUARDED_BY(expire_timer_mtx) {thread_pool.get_executor()};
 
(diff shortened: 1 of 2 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3947,103 +3950,89 @@ namespace eosio {
...
-               fc_dlog(logger, "unlinkable_block ${bn} : ${id}, previous ${pn} : ${pid}",
-                       ("bn", ptr->block_num())("id", id)("pn", block_header::num_from_id(ptr->previous))("pid", ptr->previous));
-            }
-            boost::asio::post(c->strand, [c, id, blk_num=ptr->block_num(), close_mode]() {
-               peer_dlog( c, "rejected block ${bn} ${id}", ("bn", blk_num)("id", id) );
-               my_impl->sync_master->rejected_block( c, blk_num, close_mode );
-            });
-            return;
+         // This mutex is not about thread-safety; the call to accept_block is thread-safe.
+         // This mutex is to avoid extra work of validating the same block header on multiple threads at the same time.
```
spoonincode, line 3976: Feels like there ought to be something even more lightweight than a mutex given how this comment describes the problem being solved. But I'm not immediately thinking of anything with certainty given unfamiliarity here.. Maybe an atomic that simply indicates the current block id being processed: if the atomic matches then you know this block is already being processed on another thread and can bail in any effort processing it on this thread.

Of course the problem is that we only have 16 byte lockfree atomics and block id is 32 bytes. But matching against the first or last 16 of 32 seems fine enough in this use case.

(not a suggestion for a change just thinking about it)

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3947,103 +3950,89 @@ namespace eosio {
...
+         // The mutex is used instead of posting to the dispatcher strand as the post to the dispatcher strand adds
+         // a thread switch which adds a delay in processing.
```
spoonincode, line 3978: That's going to be true for post() but quite possibly dispatch() solves that problem? Both the dispatcher strand and connection strand run on the same thread pool (same `io_context`). So I would have expected "`dispatcher_strand.dispatch()`" to immediately call the callback without a thread hop (unless the dispatcher strand is already executing something else at the moment).

heifner: Thanks. It looks like `dispatch()` does solve the problem.

plugins/net_plugin/net_plugin.cpp
spoonincode: I suppose a single atomic breaks down once you have 3 threads in here at once since if 2 of those threads are the same id then you would allow the unneeded accept_block. Doesn't break anything just nerfs the purpose. Of course if 3+ threads are often going to be contesting this mutex it's effectively a bunch of thread start/stops anyways which might nullify any gains you get from trying to avoid the duplicate work depending on how long the work takes.

Almost like you need a lock free unordered hash here.

Or  you could even use a small vector of atomics; e.g. 3 of them will ensure if 4 threads simultaneously are here everything is still fully guarded.

plugins/net_plugin/net_plugin.cpp
spoonincode: I think in many ways what you've done with the mutex is kinda sorta what the dispatch ultimately does for you. It LGTM as is if you like it this way. There is a small possibility the dispatch will be more efficient than the mutex when the mutex is contested because a contested mutex generally implies a thread stop/start, but this stop/start may not be the case for when dispatch posts because whatever thread the strand is currently running on could pick up the post without and start/stop of a thread since that thread is already running (all this would depend on asio's impl that I'm not sure about so this could be an optimistic take).

heifner: In my bit of testing it does seem like the `dispatch()` is better than the mutex for the nominal case. Plus it is simpler than adding a mutex.

plugins/net_plugin/net_plugin.cpp
greg7mdp: > I suppose a single atomic breaks down once you have 3 threads in here at once since if 2 of those threads are the same id then you would allow the unneeded accept_block. Doesn't break anything just nerfs the purpose. Of course if 3+ threads are often going to be contesting this mutex it's effectively a bunch of thread start/stops anyways which might nullify any gains you get from trying to avoid the duplicate work depending on how long the work takes.
> 
> Almost like you need a lock free unordered hash here.
> 
> Or you could even use a small vector of atomics; e.g. 3 of them will ensure if 4 threads simultaneously are here everything is still fully guarded.

Maybe a solution would be to use the double check locking (1 atomic + 1 mutex) which is designed to avoid locking the mutex in the most common case (example here is for the initialization of static variables, but you get the idea).
```c++
atomic<bool> x_init(false);

if (!x_init.load(memory_order_acquire)) {
   lock_guard<mutex> _(x_init_mtx);
   if (!x_init.load(std::memory_order_relaxed)) {
      // initialize x here
      ...
      x_init.store(true, memory_order_release);
   }
}
```

plugins/net_plugin/net_plugin.cpp
greg7mdp: Am I right that this PR has two unrelated changes:
1. The switch from `post()` to `dispatch()`,  which avoids a thread hop
2. All the rest of the PR, which implements a better way to avoid validating the same block twice (although I think it could still happen, since two threads can check in `create_block_handle` that `id` is not in `fork_db`, then proceed to validate `id` before adding it to `fork_db`)

ericpassmore:
Note:start
category: Other
component: P2P
summary: Optimize processing of incoming blocks by removing an unneeded thread hop on nominal case.
Note:end

plugins/net_plugin/net_plugin.cpp
heifner: The original dispatcher `post()` was to avoid two block header validations at the same time. This PR is an alternative, more efficient mechanism to achieve the same thing.

It is possible via the `create_block_handle` to have more than one thread validating the same block header. That is why the net_plugin serializes the calls to `create_block_handle` to avoid that. In practice, it is only called from net_plugin.

spoonincode (approved):

plugins/net_plugin/net_plugin.cpp
heifner: That would work but seems rather complicated compared to current implementation.

greg7mdp: It would also allow to validate blocks (including signatures) in parallel, although we'd still have to add them to `fork_db` in order. Not sure if that would increase syncing performance or not though.

greg7mdp (approved):

Issue #1169, opened 2025-02-14
EPIC: Gossip-based Automanagement of BP Network


bhazzard:
- BPs should keep each other informed of the connection details of other BPs
- BPs should be able to validate the provenance of a BPs connection details
- BP nodes should maintain connection to other BP nodes
   - When the BP schedule changes
   - When a BP changes their endpoint

Issues in other Repos:
- [x] https://github.com/eosnetworkfoundation/eos-system-contracts/issues/184

PR #1228, opened 2025-03-04
Retrieval of peers public keys from chainbase db.

### LINKED ISSUE #1183: Add support for accessing the on-chain peer public keys
Block producers can now store on-chain the public keys that want to use to sign the new `gossip_bp_peers` network message, using the new actions `regpeerkey` and `delpeerkey` of `eos-system-contracts`.

This will add code in spring to maintain a map container (`name` -> `public_key`) which will provide this information, and which can be efficiently accessed fron the multiple threads of net_plugin.

greg7mdp:
Resolves #1183.

It is needed for https://github.com/AntelopeIO/spring/issues/1169.

### DIFF
--- a/plugins/net_plugin/peer_keys_db.cpp
+++ b/plugins/net_plugin/peer_keys_db.cpp
@@ -1,0 +1,46 @@
+#include <eosio/net_plugin/peer_keys_db.hpp>
+#include <eosio/chain/contract_table_objects.hpp>
+
+namespace eosio::chain {
+
+peer_keys_db_t::peer_keys_db_t()
+   : _peer_key_map(std::make_shared<peer_key_map_t>()) {
+}
+
+void peer_keys_db_t::on_accepted_block(const controller& chain) {
+   try {
+      const auto& head = chain.head();
+
+      // let's update once every 20 blocks, so every 10 seconds
+      if (head.is_valid() && block_header::num_from_id(head.id()) % 20 == 0) {
+         auto new_map = std::make_shared<peer_key_map_t>();
+
+         const auto& d = chain.db();
+         const auto* t_id =
+            d.find<table_id_object, by_code_scope_table>(boost::make_tuple("eosio"_n, "eosio"_n, "peerkeys"_n));
+         
+         if (t_id != nullptr) {
+            const auto&        idx = d.get_index<chain::key_value_index, chain::by_scope_primary>();
+            decltype(t_id->id) next_tid(t_id->id._id + 1);
+            auto               lower = idx.lower_bound(boost::make_tuple(t_id->id));
+            auto               upper = idx.lower_bound(boost::make_tuple(next_tid));
+
+            for (auto itr = lower; itr != upper; ++itr) {
+               const key_value_object& obj = *itr;
+               fc::datastream<const char *> ds(obj.value.data(), obj.value.size());
+
+               peer_key_map_t::value_type row;
+               fc::raw::unpack(ds, const_cast<name&>(row.first));
+               fc::raw::unpack(ds, row.second);
+               new_map->insert(row);
+            }
+         }
+
+         auto old_map = get_peer_key_map();
+         if (*new_map != *old_map)
+            _peer_key_map = new_map;
+      }
+   } FC_LOG_AND_DROP(("Error when updating peer_keys_db"));
+}
+
+} // namespace eosio::chain
--- a/plugins/net_plugin/CMakeLists.txt
+++ b/plugins/net_plugin/CMakeLists.txt
@@ -1,6 +1,7 @@
 file(GLOB HEADERS "include/eosio/net_plugin/*.hpp" )
 add_library( net_plugin
              net_plugin.cpp
+             peer_keys_db.cpp
              ${HEADERS} )
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.0)
--- a/plugins/net_plugin/include/eosio/net_plugin/peer_keys_db.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/peer_keys_db.hpp
@@ -1,0 +1,27 @@
+#pragma once
+
+#include <eosio/chain/controller.hpp>
+
+namespace eosio::chain {
+
+/**
+ * This class caches the on-chain public keys that BP use to sign the `gossip_bp_peers`
+ * network message. These public keys are populated using the actions regpeerkey and
+ * delpeerkey of eos-system-contracts.
+ */
+
+class peer_keys_db_t {
+public:
+   using peer_key_map_t = std::map<name, public_key_type>;
+
+   peer_keys_db_t();
+
+   void on_accepted_block(const controller& chain);        // update the map every N blocks, a small delay is fine
+
+   std::shared_ptr<peer_key_map_t> get_peer_key_map() { return _peer_key_map; }
+
+private:
+   std::atomic<std::shared_ptr<peer_key_map_t>> _peer_key_map;
+};
+
+} // namespace eosio::chain
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2,6 +2,7 @@
 #include <eosio/net_plugin/protocol.hpp>
 #include <eosio/net_plugin/net_utils.hpp>
 #include <eosio/net_plugin/auto_bp_peering.hpp>
+#include <eosio/net_plugin/peer_keys_db.hpp>
 #include <eosio/chain/types.hpp>
 #include <eosio/chain/controller.hpp>
 #include <eosio/chain/exceptions.hpp>
@@ -390,6 +391,7 @@ namespace eosio {
       unique_ptr< sync_manager >       sync_master;
       dispatch_manager                 dispatcher {thread_pool.get_executor()};
       connections_manager              connections;
+      peer_keys_db_t                   peer_keys_db;
 
       /**
        * Thread safe, only updated in plugin initialize
@@ -4080,6 +4082,9 @@ namespace eosio {
          on_pending_schedule(*pending_producers);
       }
       on_active_schedule(chain_plug->chain().active_producers());
+
+      // update peer public keys from chainbase db every so often
+      peer_keys_db.on_accepted_block(chain_plug->chain());
    }
 
    // called from application thread


plugins/net_plugin/peer_keys_db.cpp
```diff
@@ -0,0 +1,46 @@
...
+   : _peer_key_map( boost::shared_ptr<peer_key_map_t>(new peer_key_map_t)) {
+}
+
+void peer_keys_db_t::on_accepted_block(const controller& chain) {
+   try {
+      const auto& head = chain.head();
+
+      // let's update once every 20 blocks, so every 10 seconds
+      if (head.is_valid() && block_header::num_from_id(head.id()) % 20 == 0) {
+         auto new_map = boost::shared_ptr<peer_key_map_t>(new peer_key_map_t);
```
heifner, line 16: `std::shared_ptr` and might as well use `make_shared`.

plugins/net_plugin/peer_keys_db.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+               peer_key_map_t::value_type row;
+               fc::raw::unpack(ds, const_cast<name&>(row.first));
+               fc::raw::unpack(ds, row.second);
+               new_map->insert(row);
+            }
+         }
+
+         auto old_map = get_peer_key_map();
+         if (*new_map != *old_map)
```
heifner, line 40: I think you can do `new_map.get() != old_map.get()` and avoid the complete map compare.

heifner (changes requested):

plugins/net_plugin/peer_keys_db.cpp
```diff
@@ -0,0 +1,46 @@
...
+         const auto* t_id =
+            d.find<table_id_object, by_code_scope_table>(boost::make_tuple("eosio"_n, "eosio"_n, "peerkeys"_n));
+         
+         if (t_id != nullptr) {
+            const auto&        idx = d.get_index<chain::key_value_index, chain::by_scope_primary>();
+            decltype(t_id->id) next_tid(t_id->id._id + 1);
+            auto               lower = idx.lower_bound(boost::make_tuple(t_id->id));
+            auto               upper = idx.lower_bound(boost::make_tuple(next_tid));
+
+            for (auto itr = lower; itr != upper; ++itr) {
```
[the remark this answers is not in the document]
heifner: There are 679 registered producers in EOS. Granted there are only 30 finalizers, but that can grow unbounded.

plugins/net_plugin/peer_keys_db.cpp
greg7mdp: I think I have to use `boost_shared_ptr` to initialize a `boost::atomic_shared_ptr`, but I'll use `boost::make_shared`.

plugins/net_plugin/peer_keys_db.cpp
greg7mdp: I do want to do a complete map compare to see if there are any new keys in the map.

plugins/net_plugin/peer_keys_db.cpp
```diff
@@ -0,0 +1,46 @@
...
+peer_keys_db_t::peer_keys_db_t()
+   : _peer_key_map( boost::shared_ptr<peer_key_map_t>(new peer_key_map_t)) {
+}
+
+void peer_keys_db_t::on_accepted_block(const controller& chain) {
+   try {
+      const auto& head = chain.head();
+
+      // let's update once every 20 blocks, so every 10 seconds
+      if (head.is_valid() && block_header::num_from_id(head.id()) % 20 == 0) {
```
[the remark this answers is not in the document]
greg7mdp: My understanding is that people can call the new actions at any time to populate (or update) the public key for their node, and we shouldn't wait for the next producer schedule version change to take it into account.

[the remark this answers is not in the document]
heifner: Then maybe a version on the peer key table?

plugins/net_plugin/peer_keys_db.cpp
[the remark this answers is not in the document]
greg7mdp: A few hundred `name`/`public keys` pairs is tiny, a few KB. I feel that adding extra code is not desirable it since it will not make any measurable difference, but it may not be the prevalent viewpoint at ENF :-)

spoonincode:
Apologies if this was covered in some design doc I missed: Was using a ROtrx considered instead of fishing in the tables manually?

plugins/net_plugin/peer_keys_db.cpp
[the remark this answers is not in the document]
spoonincode: But what if the table is maliciously made large?

greg7mdp:
> Apologies if this was covered in some design doc I missed: Was using a ROtrx considered instead of fishing in the tables manually?

I didn't consider it. Wouldn't really know how to do it either.

heifner:
> Apologies if this was covered in some design doc I missed: Was using a ROtrx considered instead of fishing in the tables manually?

Interesting idea. Currently read-only trxs are not enabled on all nodes, but that could be changed if we wanted to use them "internally" for something like this. Also I wonder how much overhead that would be compared to direct table access.

plugins/net_plugin/peer_keys_db.cpp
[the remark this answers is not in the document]
greg7mdp: I'll address this with the version as well.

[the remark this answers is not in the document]
spoonincode: I'm not following how a version helps prevent needing to iterate over 100000 rows?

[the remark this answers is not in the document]
spoonincode: An interesting property about using chainbase directly here is that we ought to be able to get it to tells us what rows have been removed/created/modified fairly easily since it already knows all that. Leaning on it prevents needing to iterate through all the rows. Seems like we'd still have to make a full copy of the map though, which also may be expensive with 100000 entries.

All that said not sure it's a wise idea to lean on any chainbase magic since it makes it harder to move away from it.

I do like @heifner's thought that maybe the primary key on this table ought to be the producer name so we can just pluck out the ones we need super quick.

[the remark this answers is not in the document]
heifner: My idea of a version was either a diff table with just a version in it that is incremented on every mod to the peer table. Or maybe some special producer_name and use the public_key as a version number.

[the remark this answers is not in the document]
spoonincode: Maybe another option is for the contract to be responsible for maintaining the table with only current producer keys, and it swaps them in/out as needed with the large set of all registered producer keys.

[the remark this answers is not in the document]
heifner: Maybe even just add it as an extension to the finalizer table.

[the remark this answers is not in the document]
greg7mdp: > Maybe even just add it as an extension to the finalizer table.

iiuc, and you mean adding it to the finalizer table in `eos-system-contracts`, I think this would be a quite difficult and risky change, esp since the finalizer table is provided to spring via a host function and also is used by SHiP and Deepmind.

However, what about adding a host function to pass the information straight to the controller?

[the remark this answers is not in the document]
greg7mdp: @heifner one question I have, should we consider both proposers and finalizers?
Also should we use the current producers and the ones on the next schedule.
Should we consider the current finalizers + all the ones pending?

Upon reflection, I think it is better to just efficiently reflect the table from chainbase, using the version scheme I described above.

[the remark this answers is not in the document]
heifner: I think I agree with that. The net_plugin can track the current proposers and finalizers easily. It already tracks the active and pending proposer schedule. The account has to pay for the on-chain ram, duplicating it in the node seems like would be safe. We could limit the copy time to say 5ms so it can't be weaponized.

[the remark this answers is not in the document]
heifner: However, why bother keeping a copy. Seems like chainbase access should only be slightly slower than C++ memory lookup. You can use hash lookup instead of chainbase lookup, but seems like not worth making a copy.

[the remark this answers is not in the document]
greg7mdp: > However, why bother keeping a copy. 

So it can be accessed without locks from any `net_plugin` thread.

ericpassmore:
Note:start
category: Other
component: P2P
summary: Retrieval of peers public keys from chainbase db.
Note:end

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,74 @@
...
+   uint64_t res = 0;
+   fc::raw::unpack(ds, res);
+   return std::optional<uint64_t>(res);
+}
+
+size_t peer_keys_db_t::update_peer_keys(const controller& chain) {
+   size_t num_updated = 0;
+   try {
+      const auto& db = chain.db();
+      auto cb_version = _get_version(db);
```
spoonincode, line 33: I am not really seeing any benefit to using chainbase's revision over just simply the applied block num? That's really what the contract is tracking -- the last block num the row was modified at. (`version` seems a misnomer in that aspect) Relying on chainbase revision == head block num might be an implementation detail.

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,74 @@
...
+         public_key_type row_key;
+
+         const auto& obj = *itr2;
+         fc::datastream<const char*> ds(obj.value.data(), obj.value.size());
+         // must match `struct peer_key;` in eosio.system.hpp
+         // -------------------------------------------------
+         fc::raw::unpack(ds, row_name);
+         fc::raw::unpack(ds, row_version);
+         fc::raw::unpack(ds, row_key);
+         assert(row_version > _version && row_version <= *cb_version);
```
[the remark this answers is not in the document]
greg7mdp: How could a row *not* be well formed though?

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+size_t peer_keys_db_t::update_peer_keys(const controller& chain) {
+   size_t num_updated = 0;
+   try {
+      const auto& db = chain.db();
+      auto cb_version = _get_version(db);
+      if (!cb_version || *cb_version == _version)
+         return num_updated;                    // nothing to do
+      assert(*cb_version > _version);
```
spoonincode, line 36: Won't this potentially assert on fork switches?

libraries/chain/peer_keys_db.cpp
greg7mdp: This tracks the version of the `peerkeys` table, and this version is incremented every time the `peerkeys` table if modified.
What I meant with `cb_version` is `the version of peerkeys in chainbase`, vs the version of `peerkeys` we currently have in the `unordered_flat_map`.

libraries/chain/peer_keys_db.cpp
greg7mdp: Hum, good point, let me think about it.

libraries/chain/peer_keys_db.cpp
spoonincode: But afaict you're just using chainbase's revision as a proxy for current block num. I don't believe that's a pattern used elsewhere -- normally you'd look at forkdb or the block passed in accepted_block.

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+      const auto* t_id =
+         db.find<table_id_object, by_code_scope_table>(boost::make_tuple("eosio"_n, "eosio"_n, "peerkeys"_n));
+      assert(t_id != nullptr);
+      const auto& secidx = db.get_index<index64_index, by_secondary>();
+
+      auto lower = secidx.lower_bound(std::make_tuple(t_id->id._id, _version + 1));
+      auto upper = secidx.upper_bound(std::make_tuple(t_id->id._id, *cb_version));
+
+      auto new_map = boost::make_shared<peer_key_map_t>(*get_peer_key_map());
```
spoonincode, line 46: Do we have confidence that given the size of this map can be maliciously controlled that performance will not be a problem at 100k, 500k, etc entries when doing the deep copy? It may be possible someone can make bloated 512KB PUB_WA in these entries as well, making this a very heavy copy (this may be something that the contract can defend against though by simply disallowing PUB_WA here). I'd like to see more data on the overhead and/or defense against abuse.

greg7mdp: Good point for the PUB_WA. I'll disallow it in the contract. 

[regpeerkey](https://github.com/eosnetworkfoundation/eos-system-contracts/blob/fea9b67c5da8d3ed5168212c8a19ba67999c15d7/contracts/eosio.system/include/eosio.system/eosio.system.hpp#L1678) registers a key for an existing account.

To register 500k peerkeys, you'd have to create 500k accounts. At a cost of 1GB of ram.

The cost for the table would only be 50MB. And because those updates would not occur in a single block it would not take an excessive time. It doesn't seem that that would be the most effective attack against spring.

I do believe that if someone created 500K accounts, there are likely many other issues in Spring that would degrade more that the user registering 500K peer keys and using 50MB of RAM.

There is a possible attack by creating an account with a name, registering a key, then deleting the account which I think would return the ram cost, and then doing the same thing over with a different name. I can add some protection against that if necesary.

libraries/chain/peer_keys_db.cpp
greg7mdp: > just using chainbase's revision as a proxy for current block num

Where do I do that? `peer_keys_db_t::_get_version()` returns the version stored in the `peerkeysver` table.

libraries/chain/peer_keys_db.cpp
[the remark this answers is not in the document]
greg7mdp: You means the serialized data we unpack? How it it possible that the system contract would not populate a correct row?

libraries/chain/peer_keys_db.cpp
greg7mdp: disallowing webauthn keys: https://github.com/eosnetworkfoundation/eos-system-contracts/pull/185/commits/fd86ed8c39ebff23b805461e18a03f6bb7019ab1

libraries/chain/peer_keys_db.cpp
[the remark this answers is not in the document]
greg7mdp: > a malformed table row can abort the whole process if asserts are enabled.

I feel that that's the whole point of enabling asserts. You *want* the process to abort if something that shouldn't be possible does happen.

libraries/chain/peer_keys_db.cpp
spoonincode: whoops you are right I misread the earlier code

libraries/chain/peer_keys_db.cpp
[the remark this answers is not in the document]
spoonincode: That's not how I remember us describing the use case for asserts. I remember asserts being present to describe invariants that cannot be true. In this case this isn't an invariant but rather input validation on what the contract is communicating to nodeos. asserts should be impossible from the viewpoint of nodeos but in this case a misbehaving contract can force an assert to be true.

libraries/chain/peer_keys_db.cpp
spoonincode: I wasn't _too_ worried about changes; worst case (assuming action executes 100µs) I guess you would end up with 40000 changed keys to iterate through. I was more worried about the unbounded size of the map being copied. Nobody who owns 1GB, 10GB, or whatever should be able to cause large delays inside of nodeos.

libraries/chain/peer_keys_db.cpp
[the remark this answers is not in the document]
greg7mdp: > invariants that cannot be true

According to how we understand the code to work IMO! And that should include contract code.
Really an assert says: If my code and the compiler are correct, this cannot possibly be true.

Different from user input validation that we use `EOS_ASSERT` for.

libraries/chain/peer_keys_db.cpp
greg7mdp: The map with 10,000 keys is copied in `0.059ms`. I don't see the `large delays inside of nodeos` happening.

libraries/chain/peer_keys_db.cpp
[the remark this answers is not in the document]
spoonincode: I don't consider the system contract and nodeos to be one entity, so communication between them would qualify as requiring input validation. Even the system contract isn't worthy enough to blindly trust to the point of allowing it to fail asserts.

spoonincode (changes requested):
requires further discussion:
https://github.com/AntelopeIO/spring/pull/1228#discussion_r1987907429

libraries/chain/peer_keys_db.cpp
greg7mdp: PS: I wish there was a `fc::public_key` type which didn't include the `webauthn::public_key` in the std::variant, it would reduce the size to around 34 bytes from 80.

spoonincode: Somewhere north of 100k things starting getting slow. I got 3.7ms for 500k and 7.3ms for 1M. And that's for pretty much the fastest box I have. At 10k I got 9µs which is in the same ballpark as what you got.

On my middling EPYC 7443P I got a whopping 23.2ms with 500k. It seems almost hard to believe.

greg7mdp: Thanks for checking! Yes, I also tried 100K and 1M and saw that it doesn't scale linearly, most likely degrading when the data copy (incl. both maps) doesn't fit in L2 or L3 caches anymore.

I have an idea for addressing these long copy times in the (unlikely) event that the map grows large. We can take advantage that it doesn't matter if we wait till the next block for the data to be updated, so when the map is large we can post the copy to an thread and add the new entries on the next `on_accepted_block` where the copy is ready.

libraries/chain/peer_keys_db.cpp
greg7mdp: Should be addressed by https://github.com/AntelopeIO/spring/pull/1228/commits/2bcfc00b690c2ab997e24c3815486139db0655d8 and 
https://github.com/eosnetworkfoundation/eos-system-contracts/pull/185/commits/90773cb692cbb5f07c83db3924f9ef62ad9ca3fe

which now update the reflected maps only up to lib.

libraries/chain/CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ add_library( eosio_chain
              symbol.cpp
              whitelisted_intrinsics.cpp
              thread_utils.cpp
+             peer_keys_db.cpp
```
heifner, line 130: Originally this was in net_plugin. I think that is a better place for it.

libraries/chain/peer_keys_db.cpp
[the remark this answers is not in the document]
greg7mdp: I have EOS_ASSERTs now so we should be fine.

libraries/chain/peer_keys_db.cpp
greg7mdp: Implemented the flip-flop solution https://github.com/AntelopeIO/spring/pull/1228/commits/f7384d574be6b307f52d6a3d9a347444db54c523

libraries/chain/peer_keys_db.cpp
heifner: Doesn't matter what signal you attach to, the state of the node is as of HEAD unless you are running in IRREVERSIBLE mode.

libraries/chain/CMakeLists.txt
greg7mdp: I moved it in controller because I wanted to be able to test it in unit tests which I think is useful.

heifner: net_plugin has unit tests.

libraries/chain/peer_keys_db.cpp
heifner: Sorry, I misunderstood your comment. You are only pulling rows with block numbers up to lib from the table.

libraries/chain/peer_keys_db.cpp
heifner: I'm not sure how the separate thread solution would work. Couldn't you just put a time limit of say 5ms per accept block? If it is delayed in getting the data no real harm especially if the data is an attack.

libraries/chain/CMakeLists.txt
greg7mdp: The unit test is in `eos-system-contracts/tests` and accesses the controller via the tester. I don't think the cmake system is plumbed for accessing net_plugin, but I could be wrong.

greg7mdp: Also I think it is probably smart to keep direct chainbase access in `libraries/chain` and `chain_plugin`.

libraries/chain/peer_keys_db.cpp
greg7mdp: > Couldn't you just put a time limit of say 5ms per accept block? 

I implemented the flip-flop solution, so a time limit is not needed anymore. I think the current version is safe.

libraries/chain/CMakeLists.txt
heifner: That is a fair argument. I think we should move the chainbase access out of chain_plugin, just never was a priority.

greg7mdp: > I didn't realize you were using it in `eos-system-contracts/tests`. Seems like those tests should be moved to Spring as they test Spring functionality.

Well, I need the contracts from `eos-system-contract` to modify chainbase memory when registering keys using the `regpeerkey` action, so the test has to be in  `eos-system-contracts/tests`. But in the same test I verify the `reading from chainbase memory` part which is in `libraries/chain`

greg7mdp:
Silly me, I realize there is an issue with my flip-flop scheme... moving the PR back to draft.

libraries/chain/peer_keys_db.cpp
spoonincode: > I'm not sure how the separate thread solution would work

I was thinking the copy and merge would occur on a separate thread.

greg7mdp:
Code is ready for review again. I ended up using a `mutex`, but considering the updates to the hash_map are quite fast and infrequent (per block containing new peer keys), I feel it shouldn't be an issue.
I could use a `shared_mutex` to only read lock for `operator[]()`, but I doubt it would be an improvement.

libraries/chain/include/eosio/chain/peer_keys_db.hpp
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * This class caches the on-chain public keys that BP use to sign the `gossip_bp_peers`
+ * network message. These public keys are populated using the actions regpeerkey and
+ * delpeerkey of eos-system-contracts.
+ */
+class peer_keys_db_t {
+public:
+   using peer_key_map_t = boost::unordered_flat_map<name, public_key_type, std::hash<name>>;
+
+   peer_keys_db_t();
```
heifner, line 18: This should be disabled by default. Most nodes have no use for this and there is no need to update it on those nodes. I was thinking that could be added later, but probably better to get that in before merging this.

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,79 @@
...
+   std::lock_guard g(_m);
+   size_t num_updated = 0;
+   try {
+      if (lib_number <= _block_num)
+         return num_updated;                      // nothing to do
+
+      const auto& db = chain.db();
+      const auto  table_ref = boost::make_tuple("eosio"_n, "eosio"_n, "peerkeys"_n);
+      const auto* t_id      = db.find<table_id_object, by_code_scope_table>(table_ref);
+      EOS_ASSERT(t_id != nullptr, plugin_exception, "cannot retrieve `peerkeys` table");
```
heifner, line 32: `plugin_exception` should be something else. Either a new one or maybe just `misc_exception`. Since this is caught and dropped below, misc is probably fine. Maybe better would be to `elog` and return instead of throw and catch below.

heifner: Same goes for the `plugin_exception`s below. There the `misc_exception` is likely easier.

heifner (changes requested):

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,80 @@
...
+      const auto lower = secidx.lower_bound(std::make_tuple(t_id->id._id, static_cast<uint64_t>(_block_num + 1)));
+      const auto upper = secidx.upper_bound(std::make_tuple(t_id->id._id, static_cast<uint64_t>(lib_number)));
+
+      if (upper == lower) {
+         // no new keys registered
+         _block_num = lib_number;
+         return num_updated;
+      }
+
+      std::lock_guard g(_m); // we only need to protect access to _peer_key_map
```
heifner, line 44: I thought this had to be a `fc::lock_guard`.

greg7mdp: https://github.com/AntelopeIO/spring/pull/1228/commits/985ec9b02226288692ec98e9decaf4eb22f3d941

libraries/chain/peer_keys_db.cpp
greg7mdp: > Maybe better would be to `elog` and return instead of throw and catch below.

How would it be better? this is not something that is expected to happen ever.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -427,6 +427,9 @@ namespace eosio::chain {
 
          chain_id_type get_chain_id()const;
 
+         void set_peer_keys_retrieval_active(bool active);          // thread safe
```
heifner, line 430: This is not thread safe.

heifner: There is no need for it to be thread safe. Just remove the comment.

libraries/chain/CMakeLists.txt
heifner: I think I would rather test the eos-system-contracts in Spring than test Spring in eos-system-contracts. Not sure @spoonincode thoughts on this. I can approve as is, but seems cleaner to move all the tests to Spring.

spoonincode: > I think I would rather test the eos-system-contracts in Spring than test Spring in eos-system-contracts

The short answer is I agree with that.

I had a longer answer how maybe the ideal way to test this feature is in spring but _not_ via eos-system-contracts, because my understanding from prior discussions is that reference-contracts/eos-system-contracts being part of spring's tests has been considered somewhat[^1] an anti-pattern, especially outside of integration level tests. But since this will ultimately involve net plugin tests too perhaps it will eventually be part of an integration test.

[^1]:Or maybe not even 'somewhat' depending who you ask

greg7mdp: It is hard to make sure that the code writing into chainbase in `eos-system-contracts` is in sync with the code reading the data in spring, if we are not using both spring code and  `eos-system-contracts`.

For example when I change the code in  `eos-system-contracts` to store `block_num` as `uint32_t` instead of `uint64_t`, the test failed because the code was not in sync with spring.  I wouldn't want to lose that connection by duplicating the contract code in spring.

heifner (approved):

libraries/chain/CMakeLists.txt
spoonincode: Thinking on it more, it's clearer that the right place to test is in spring for two reasons -- 
1) I take it eventually there will be an integration test for the feature as a whole (i.e. with net stuff tested) too and that can only be done in spring anyways. But even if not making such a test, 
2) contract tests should not depend on random internals of libchain because it prevents us from creating a clean libtester api. In other words, contract unit tests directly interacting beyond the tester apis is not something we want to encourage -- we want to reduce the leakiness not add more. As an extreme example of why it's a problem, consider if we want to rewrite the system contracts tests in Vert[^1]. It's literally impossible to have this test with Vert.

I see @heifner has approved so I won't block the PR on this issue either. I'll rereview shortly.

[^1]: This might not even be an 'extreme' example since it has been floated more than once!

libraries/chain/CMakeLists.txt
```diff
@@ -127,11 +127,16 @@ add_library( eosio_chain
...
              platform_timer_accuracy.cpp
              ${PLATFORM_TIMER_IMPL}
              ${HEADERS}
              )
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.0)
+  target_compile_options(eosio_chain PUBLIC -Wthread-safety)
+endif()
```
spoonincode, line 138: In the future we should consider making this common somewhere and not copy paste it all over.

Also, since this is `PUBLIC` it means everything that links to `libchain` also gets it enabled. This is true today for everything that links to `net_plugin` -- for example nodeos' `main.cpp` currently have `-Wthread-safety` enabled. Is that a problem? If it's not a problem, all the more reason to consider just enabling it globally.

libraries/chain/include/eosio/chain/peer_keys_db.hpp
```diff
@@ -0,0 +1,40 @@
...
+   // safe to be called from any thread
+   std::optional<public_key_type> get_peer_key(name n) const;
+
+   // safe to be called from any thread
+   size_t size() const;
+
+private:
+   std::optional<uint64_t> _get_version(const chainbase::database& db);
+
+   bool               _active;               // if not active (the default), no update occurs
```
spoonincode, line 34: Looks like this needs a `= false`

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,80 @@
...
+            // -------------------------------------------------
+            fc::raw::unpack(ds, row_name);
+            EOS_ASSERT(row_name.good(), misc_exception, "deserialized invalid name from `peerkeys`");
+
+            fc::raw::unpack(ds, row_block_num);
+            EOS_ASSERT(row_block_num > static_cast<uint64_t>(_block_num), misc_exception,
+                       "deserialized invalid version from `peerkeys`");
+
+            fc::raw::unpack(ds, row_key);
+            EOS_ASSERT(row_key.valid(), misc_exception, "deserialized invalid public key from `peerkeys`");
```
spoonincode, line 67: I am suspicious this opens up the possibility for a denial of service. It looks like you should just ignore entries with a bad key instead of failing to update all entries.

spoonincode (changes requested):
lgtm other then https://github.com/AntelopeIO/spring/pull/1228/files#r1996432630 and https://github.com/AntelopeIO/spring/pull/1228/files#r1996452813

libraries/chain/peer_keys_db.cpp
greg7mdp: I believe this is what we are doing with the `try/catch` block within the loop.

spoonincode: you are right I got cross eyed with the double-uped try blocks

libraries/chain/include/eosio/chain/peer_keys_db.hpp
greg7mdp: Thanks https://github.com/AntelopeIO/spring/pull/1228/commits/c60b6e9611835f0747caf4d65920f7a795a38632

libraries/chain/CMakeLists.txt
greg7mdp: Would you like me to add it to the top `CMakeLists.txt`?

spoonincode (approved):

libraries/chain/CMakeLists.txt
spoonincode: I wasn't really sure what the side effect was for enabling `-Wthread-safety` on code that isn't prepared for it. We're already doing that a little because of net_plugin and now we're going to be doing it a lot because of libchain. If that's fine maybe we blanket enable it at top anyways, if that's bad then maybe we should change these to `PRIVATE` instead. Doesn't need to be done now.

greg7mdp: I was just trying it out and it builds fine. Checking it in to see if there is any test issue. I can undo that commit if desired.

greg7mdp: > I wasn't really sure what the side effect was for enabling `-Wthread-safety` on code that isn't prepared for it.

I believe that it should be fine (a no-op)

libraries/chain/peer_keys_db.cpp
```diff
@@ -0,0 +1,85 @@
...
+            // must match `struct peer_key;` in eosio.system.hpp
+            // -------------------------------------------------
+            fc::raw::unpack(ds, row_name);
+            EOS_ASSERT(row_name.good(), misc_exception, "deserialized invalid name from `peerkeys`");
+
+            fc::raw::unpack(ds, row_block_num);
+            EOS_ASSERT(row_block_num > static_cast<uint64_t>(_block_num), misc_exception,
+                       "deserialized invalid version from `peerkeys`");
+
+            fc::raw::unpack(ds, row_version); // no need to check, all versions must include the `key` optional
```
spoonincode, line 67: Similar to  https://github.com/eosnetworkfoundation/eos-system-contracts/pull/185/files#r2001804923 -- comment does not match my understanding of prior discussion. It seems like if version != 0 this should skip the row.

Issue #1213, opened 2025-02-27
SC: Boilerplate code for host functions call, get_call_data, and set_call_return_value


linh2931:
This task implements boilerplate code for the new host functions `call`, `get_call_data`, and `set_call_return_value`. It defines, registers, and puts them into host function whitelist. Actual implementation of the host functions is tracked by separate issues.

A unit test using `wasm` should be added to show the new host functions are available (without unresolved symbols when deploying) after `sync_call` protocol feature is activated.

PR #1238, opened 2025-03-05
SC: Implement boilerplate code for new host functions call, get_call_data,  and set_call_return_value

linh2931:
* Implements boilerplate code for the new host functions `call`, `get_call_data`, and `set_call_return_value`. In particular, defines, registers, and adds them to host function whitelist. Actual implementation of the host functions is to be done in separate PRs.
* Adds tests to show that after sync_call protocol feature is activated,  a contract using the new host functions can be deployed without "unresolved symbols" error and those new host functions can be called.

Resolves https://github.com/AntelopeIO/spring/issues/1213


libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -476,6 +476,11 @@ REGISTER_CF_HOST_FUNCTION(action_data_size);
 REGISTER_CF_HOST_FUNCTION(current_receiver);
 REGISTER_HOST_FUNCTION(set_action_return_value);
 
+// sync call api
+REGISTER_HOST_FUNCTION(call);
+REGISTER_CF_HOST_FUNCTION(get_call_data);
+REGISTER_HOST_FUNCTION(set_call_return_value);
```
spoonincode, line 482: Does it make sense for `get_call_data` to be allowed in CFA? If it isn't possible to `call` in a CFA can `get_call_data` ever do anything useful in a CFA?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -684,6 +684,57 @@ namespace webassembly {
...
+          *                  // point or the signature does not match the format,
+          *                  // or the called function_name is not recognized by the receiver.
+          *                  // Otherwise, the call aborts.
+          *                  uint64_t flags          = 0;
+          *               };
+          *
+          *               struct data_t {
+          *                  data_header      header;
+          *                  span<const char> arguments;
+          *               };
```
spoonincode, line 716: Is this the appropriate location for the `data` documentation? The protocol is agnostic to what `data` is, so nodeos doesn't care... until output is shown in any API calls I guess. So I'm not sure..

linh2931: I was thinking about adding something like this is a recommended data format or updated CDT expects to use this. You are correct, this is not the right place to specify it. Spring does not care `data`. Will remove it.

libraries/chain/webassembly/runtimes/eos-vm.cpp
heifner: A CFA is designed so they can be removed. They should have no side-effects. They can only assert. If they are on the blockchain then they didn't assert, so you can just not run them. Having a `set_call_return_value` would mean that the trace would have to be different. Also would need to somehow store that for creating the action digest.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic_mapping.hpp
```diff
@@ -187,6 +187,9 @@ inline constexpr auto get_intrinsic_table() {
       "env.tapos_block_prefix",
       "env.tapos_block_num",
       "env.get_action",
+      "env.call",
+      "env.get_call_data",
+      "env.set_call_return_value",
```
heifner, line 192: Don't these have to be added at the end?

heifner (changes requested):

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic_mapping.hpp
linh2931: Thanks! Yes, they can only be added at the bottom. Wondering why none of OC tests in CICD fails.

spoonincode: None of the tests have an existing code cache (populated with a version of nodeos prior to this change).

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: Thanks for the explanation. I just thought a CFA cannot access (read/write) the state. 

For comparison, existing `read_action_data` can be used in CFA while `set_action_return_value` cannot.

Maybe I should change `get_call_data` to `read_call_data` to be consistent with `read_action_data`?

@arhag How do you think about sync calls in CFAs?

arhag: I don't think we should allow sync calls in context-free actions.

linh2931: Thanks. I will keep all of them not to be called in context-free actions as is. Will add a comment there.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic_mapping.hpp
```diff
@@ -282,6 +279,9 @@ inline constexpr auto get_intrinsic_table() {
       "env.bls_fp_mul",
       "env.bls_fp_exp",
       "env.set_finalizers",
+      "env.call",
+      "env.get_call_data",
+      "env.set_call_return_value",
       "eosvmoc_internal.check_memcpy_params"
```
heifner, line 285: Seems like they should come at the very end.

linh2931: Sorry, I should have been more careful. From `eosvmoc_internal` I just thought that was a last item guard.

heifner (approved):

spoonincode (approved):

ericpassmore:
Note:start
category: Other
component: Contracts
summary: Initial checkin with new host functions supporting sync calls.
Note:end

PR #1246, opened 2025-03-07
avoid memset & memcpy on non-trivially copyable `fc::variant` to eliminate a bunch of warnings

spoonincode:
clang20 is firing off a megaton of warnings such as
```
warning: first argument in call to 'memset' is a pointer to non-trivially copyable type 'fc::variant'
```
and
```
warning: first argument in call to 'memcpy' is a pointer to non-trivially copyable type 'fc::variant'
```

Not sure if anyone has strong opinions on proper way to resolve the warning (besides a larger refactor). I think we could keep the memcpy/memset and explicitly cast the pointers but that seems more hacky than using a std::array like done here.

### DIFF
--- a/libraries/libfc/include/fc/variant.hpp
+++ b/libraries/libfc/include/fc/variant.hpp
@@ -8,8 +8,6 @@
 #include <unordered_set>
 #include <vector>
 
-#include <string.h> // memset
-
 #include <fc/string.hpp>
 #include <fc/time.hpp>
 #include <fc/container/deque_fwd.hpp>
@@ -349,7 +347,6 @@ namespace fc
         template<typename T>
         explicit variant( const std::optional<T>& v )
         {
-           memset( this, 0, sizeof(*this) );
            if( v.has_value() ) *this = variant(*v);
         }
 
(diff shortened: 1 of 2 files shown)


libraries/libfc/include/fc/variant.hpp
```diff
@@ -362,8 +359,8 @@ namespace fc
         void    clear();
       private:
         void    init();
-        double  _data;                ///< Alligned according to double requirements
-        char    _type[sizeof(void*)]; ///< pad to void* size
+        //enough room to store pointers, doubles, uint64s. doubled to allow 1 extra byte to store type at the end
+        alignas(double) std::array<char, std::max(sizeof(uintmax_t ), sizeof(double)) * 2> _data = {};
```
spoonincode, line 363: afaict we just need a +1 but I doubled it so it fits nicely in 16 byte register (and 16 bytes is what it was previously too)

linh2931 (approved):

heifner (approved):

ericpassmore:
Note:start
category: Chores
component: Internal
summary: Cleanup memcpy and memset to avoid warnings on non-trivially copyable type.
Note:end

Issue #1043, opened 2024-11-22
Make OC allow for nested Executor execution


spoonincode:
When WASM finishes executing OC needs to restore the value of context pointer instead of setting it to 0.

Possibly some minor changes to check time.

spoonincode:
There will also need to be a change to allow access to the new WASM export

linh2931:
Need to do `checktime` for OC too:
https://github.com/AntelopeIO/spring/pull/1251#discussion_r2001813904

linh2931:
https://github.com/AntelopeIO/spring/pull/1266#pullrequestreview-2737485085

linh2931:
Need to add the plumbing for the new `sync_call` WASM entry point.

Issue #1218, opened 2025-02-27
SC: Enforce sync call constraints


linh2931:
* Do not allow  inline actions or require_recipients in sync calls.
* Do not allow action related host functions in sync calls.

This should be done after https://github.com/AntelopeIO/spring/issues/1261

PR #1251, opened 2025-03-12
SC:  Implement the core functionalities of sync calls

linh2931:
This PR implements the core functionalities of sync calls, in particular the  `call` host function. The Unit tests show basic call flows are working:
* Sync calls whose receivers are the same as the senders
* Sync calls whose receivers are different from the senders
* Nested sync calls
* Recursive sync calls
* Sequential sync calls

Notes for reviewing:
* Please roughly follow the order of `libraries/chain/webassembly/sync_call.cpp`, `libraries/chain/apply_context.cpp` and `libraries/chain/webassembly/runtimes/eos-vm.cpp` to review
* `eosvm-oc` not working yet. https://github.com/AntelopeIO/spring/issues/1043  (sync call unit tests skip eosvm-oc)
* Please double check if `transaction_checktime_timer` related changes are correct
* Tests will be updated using return values and console output after sync call return value and call trace are implemented (next immediate tasks)
* Reset of apply_context iterator caches is tracked by a separate issue for keeping this PR focused on main call flow https://github.com/AntelopeIO/spring/issues/1252

Resolves https://github.com/AntelopeIO/spring/issues/1214

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -60,13 +60,11 @@ apply_context::apply_context(controller& con, transaction_context& trx_ctx, uint
    context_free = trace.context_free;
 }
 
-apply_context::apply_context(controller& con, transaction_context& trx_ctx)
+apply_context::apply_context(controller& con, transaction_context& trx_ctx, name sender, name receiver, std::span<const char> data)
 :control(con)
 ,db(con.mutable_db())
 ,trx_context(trx_ctx)
-,recurse_depth(0)
-,first_receiver_action_ordinal(0)
-,action_ordinal(0)
+,sync_call_ctx({sender, receiver, data})
 ,idx64(*this)
 ,idx128(*this)
 ,idx256(*this)
@@ -277,10 +275,7 @@ void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<c
 
          try {
             // use a new apply_context for a new sync call
-            apply_context a_ctx(control, trx_context);
-            a_ctx.sync_call_ctx = sync_call_context{.sender   = get_sync_call_sender(),
-                                                    .receiver = receiver,
-                                                    .data     = std::move(data)};
+            apply_context a_ctx(control, trx_context, get_sync_call_sender(), receiver, std::move(data));
 
             control.get_wasm_interface().do_sync_call(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, a_ctx);
          } catch( const wasm_exit&) {}
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -287,10 +287,8 @@ void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<c
    } catch (const boost::interprocess::bad_alloc&) {
       throw;
    } catch(const fc::exception& e) {
-      ilog("fc::exceptio");
       handle_exception(e);
    } catch (const std::exception& e) {
-      ilog("std::exception");
       auto wrapper = fc::std_exception_wrapper::from_current_exception(e);
       handle_exception(wrapper);
    }
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -248,6 +248,8 @@ void apply_context::exec()
 
 void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<const char> data)
 {
+   assert(sync_call_ctx.has_value() ^ (act != nullptr)); // can be only one of action and sync call
+
    dlog("receiver: ${r}, flags: ${f}, data size: ${s}",
         ("r", receiver)("f", flags)("s", data.size()));
 
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -293,6 +293,12 @@ void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<c
       handle_exception(wrapper);
    }
 }
+
+action_name apply_context::get_sync_call_sender() const {
+   // Current context's receiver is the sender of next sync call
+   return sync_call_ctx ? sync_call_ctx->receiver : get_receiver();
+}
+
 bool apply_context::is_account( const account_name& account )const {
    return nullptr != db.find<account_object,by_name>( account );
 }
@@ -1147,11 +1153,6 @@ action_name apply_context::get_sender() const {
       const action_trace& creator_trace = trx_context.get_action_trace( trace.creator_action_ordinal );
       return creator_trace.receiver;
    }
-   return action_name();
-}
-
-action_name apply_context::get_sync_call_sender() const {
-   // TBD implement with sync call trace implementation
    return action_name();
 }
 
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -292,6 +292,8 @@ void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<c
       auto wrapper = fc::std_exception_wrapper::from_current_exception(e);
       handle_exception(wrapper);
    }
+
+   trx_context.checktime(); // protect against the case where during the removal of the callback, the timer expires.
 }
 
 action_name apply_context::get_sync_call_sender() const {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -60,6 +60,21 @@ apply_context::apply_context(controller& con, transaction_context& trx_ctx, uint
    context_free = trace.context_free;
 }
 
+apply_context::apply_context(controller& con, transaction_context& trx_ctx)
+:control(con)
+,db(con.mutable_db())
+,trx_context(trx_ctx)
+,recurse_depth(0)
+,first_receiver_action_ordinal(0)
+,action_ordinal(0)
+,idx64(*this)
+,idx128(*this)
+,idx256(*this)
+,idx_double(*this)
+,idx_long_double(*this)
+{
+}
+
 void apply_context::exec_one()
 {
    auto start = fc::time_point::now();
@@ -233,6 +248,56 @@ void apply_context::exec()
 
 } /// exec()
 
+void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<const char> data)
+{
+   dlog("receiver: ${r}, flags: ${f}, data size: ${s}",
+        ("r", receiver)("f", flags)("s", data.size()));
+
+   auto* code = control.db().find<account_object, by_name>(receiver);
+   EOS_ASSERT(code != nullptr, action_validate_exception,
+              "sync call's receiver account ${r} does not exist", ("r", receiver));
+
+   EOS_ASSERT((flags & 0xFFFFFFFFFFFFFFFE) == 0, action_validate_exception,
+              "sync call's flags ${f} can only set bit 0", ("f", flags));
+
+   auto handle_exception = [&](const auto& e)
+   {
+      // TBD need to store the exception in the sync call trace
+      throw;
+   };
+
+   try {
+      try {
+         const account_metadata_object* receiver_account = &db.get<account_metadata_object, by_name>( receiver);
+         if (receiver_account->code_hash == digest_type()) {
+            // TBD store the info in sync call trace
+            ilog("receiver_account->code_hash empty");
+            return;
+         }
+
+         try {
+            // use a new apply_context for a new sync call
+            apply_context a_ctx(control, trx_context);
+            a_ctx.sync_call_info = sync_call{.sender   = get_sync_call_sender(),
+                                             .receiver = receiver,
+                                             .data     = std::move(data)};
+
+            control.get_wasm_interface().do_sync_call(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, a_ctx);
+         } catch( const wasm_exit&) {}
+      } FC_RETHROW_EXCEPTIONS(warn, "sync call exception on ${receiver}", ("receiver", receiver))
+   } catch (const std::bad_alloc&) {
+      throw;
+   } catch (const boost::interprocess::bad_alloc&) {
+      throw;
+   } catch(const fc::exception& e) {
+      ilog("fc::exceptio");
+      handle_exception(e);
+   } catch (const std::exception& e) {
+      ilog("std::exception");
+      auto wrapper = fc::std_exception_wrapper::from_current_exception(e);
+      handle_exception(wrapper);
+   }
+}
 bool apply_context::is_account( const account_name& account )const {
    return nullptr != db.find<account_object,by_name>( account );
 }
@@ -1087,6 +1152,11 @@ action_name apply_context::get_sender() const {
       const action_trace& creator_trace = trx_context.get_action_trace( trace.creator_action_ordinal );
       return creator_trace.receiver;
    }
+   return action_name();
+}
+
+action_name apply_context::get_sync_call_sender() const {
+   // TBD implement with sync call trace implementation
    return action_name();
 }
 
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -278,9 +278,9 @@ void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<c
          try {
             // use a new apply_context for a new sync call
             apply_context a_ctx(control, trx_context);
-            a_ctx.sync_call_info = sync_call{.sender   = get_sync_call_sender(),
-                                             .receiver = receiver,
-                                             .data     = std::move(data)};
+            a_ctx.sync_call_ctx = sync_call_context{.sender   = get_sync_call_sender(),
+                                                    .receiver = receiver,
+                                                    .data     = std::move(data)};
 
             control.get_wasm_interface().do_sync_call(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, a_ctx);
          } catch( const wasm_exit&) {}
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -296,8 +296,10 @@ void apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<c
    trx_context.checktime(); // protect against the case where during the removal of the callback, the timer expires.
 }
 
+// Returns the sender of any sync call initiated by this apply_context or sync_call_ctx
 action_name apply_context::get_sync_call_sender() const {
-   // Current context's receiver is the sender of next sync call
+   // The sync call is initiated by this apply_context or its sync_call_ctx.
+   // That's why the context's receiver is the sender of the sync call.
    return sync_call_ctx ? sync_call_ctx->receiver : get_receiver();
 }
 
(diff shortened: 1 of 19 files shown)


libraries/chain/include/eosio/chain/platform_timer.hpp
```diff
@@ -26,17 +26,21 @@ struct platform_timer {
       callback still execute if the timer expires and stop() is called nearly simultaneously.
       However, set_expiration_callback() is synchronized with the callback.
    */
-   void set_expiration_callback(void(*func)(void*), void* user) {
+   void set_expiration_callback(void(*func)(void*), void* user, bool appending = false) {
```
linh2931, line 29: If the approach is deemed OK, I will replace this default parameter with regular parameter, and define an enum for the boolean, like `{muilt_callbacks_allowed, muilt_callbacks_disallowed}`.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -129,6 +130,33 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
+                iface, "env", "sync_call",
+                scall->sender.to_uint64_t(),
+                scall->receiver.to_uint64_t(),
+                static_cast<uint64_t>(scall->data.size())); // size_t can be uint32_t or uint64_t depending on architecture. Safely cast to uint64_t to always match the type expected by sync_call entry point.
+         };
+
+         execute(context, bkend, exec_ctx, wasm_alloc, fn);
+      }
+
       void apply(apply_context& context) override {
```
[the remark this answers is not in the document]
linh2931: Done by https://github.com/AntelopeIO/spring/pull/1251/commits/fc53065e42ac1de6b7c09c6e0cc94f755a36e377

libraries/chain/wasm_interface.cpp
```diff
@@ -91,6 +91,11 @@ namespace eosio { namespace chain {
       my->apply( code_hash, vm_type, vm_version, context );
    }
 
+   void wasm_interface::do_sync_call( const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version, apply_context& context ) {
+      ilog("wasm_interface::do_sync_call");
```
spoonincode, line 95: looks like a debug message left in

libraries/chain/include/eosio/chain/platform_timer.hpp
```diff
@@ -60,18 +64,18 @@ struct platform_timer {
...
          }
+
          _callback_variables_busy.store(false, std::memory_order_release);
       }
    }
 
    std::atomic_bool _callback_variables_busy = false;
-   void(*_expiration_callback)(void*) = nullptr;
-   void* _expiration_callback_data;
+   std::vector<std::pair<void(*)(void*), void*>> _expiration_callbacks;
```
spoonincode, line 78: I've been thinking about the changes in this file some (or more generally just the semantics of timeout in the code). I am not 100% sure I completely grok the world yet to know for sure all the ramifications.

I think the code as is probably has a race. Notice how the callbacks are not called if the "spin lock" is currently held (it behaves like a "trylock"). This "lock" is held in set_expiration_callback(). So consider when a sync call completes the timed_run completes and then goes to remove its callback and during the removal of the callback the timer expires. This expiration will be ignored from EOS VM's perspective (all code sections remain executable). So then if we reenter the caller's JIT code that JIT code can no longer be timed out and can run forever.

I think for this particular example it can be resolved by ensuring we do a manual `checktime()` call after the callee completes but before we return control to the caller.

libraries/chain/include/eosio/chain/platform_timer.hpp
```diff
@@ -26,17 +26,21 @@ struct platform_timer {
...
+      EOS_ASSERT(!(!appending && func && !_expiration_callbacks.empty()), misc_exception, "Setting a platform_timer callback when one already exists");
 
-      _expiration_callback = func;
-      _expiration_callback_data = user;
+      if (!func && !_expiration_callbacks.empty()) {
+         _expiration_callbacks.pop_back();
+         return;
+      }
+
+      _expiration_callbacks.push_back({func, user});
```
spoonincode, line 43: It does seem kind of unfortunate we need to maintain a stack of callbacks when it seems like there will only effectively be 2 unique ones registered. We could remove the stack with something like a `swap_expiration_callback()` and place the onus on restoring the value on whoever did the swap (much like we have a requirement whoever registers a callback today clears it). Some other approaches I can think of maybe go against  JITs current design (for example instead of wiring up the callback to a timed_run we somehow wire it up directly to the backend allocator). I guess we can keep this as is for now but it'd be nice to find something better.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -129,37 +130,67 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
          _runtime(runtime),
          _instantiated_module(std::move(mod)) {}
 
+      void do_sync_call(apply_context& context) override {
+         backend_t                                bkend;
+         typename eos_vm_runtime<Impl>::context_t exec_ctx;
+         vm::wasm_allocator                       wasm_alloc;
```
spoonincode, line 136: It would be preferable to not create and destroy this for every call; it's "expensive" from the standpoint of allocating and freeing pages each time.

Can leave like this for now but,

Have you checked that this is not leaking memory? `wasm_allocator` requires an explicit `free()` call to free its resources  and I don't believe it's called anywhere.

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -621,6 +628,8 @@ class apply_context {
       bool                          privileged   = false;
       bool                          context_free = false;
 
+      std::optional<sync_call_context> sync_call_ctx{};  // only one of act and sync_call_ctx can be present
```
greg7mdp, line 631: Instead of having `apply_context` having members that can be there or not, I think it would be cleaner to have something like (also that would remove the need for a separate class `sync_call_context`):

```
struct context_base {
   ...  // all common members
};

struct apply_context : public context_base {
   uint32_t recurse_depth; ///< how deep inline actions can recurse
   uint32_t first_receiver_action_ordinal = 0;
   uint32_t action_ordinal                = 0;
};

struct call_context : public context_base {
   account_name          sender;
   account_name          receiver;
   std::span<const char> data; // includes function name, arguments, and other information
};
```

libraries/chain/webassembly/runtimes/eos-vm.cpp
greg7mdp: > `wasm_allocator` requires an explicit `free()` call to free its resources 

Odd that it `mmap`s in its constructor and doesn't have a destructor to `munmap`.

spoonincode: Yeah I do suspect free() should just be moved to be the dtor. Especially since it looks like free() isn't being called anywhere. (I haven't investigated it much though)

libraries/chain/wasm_interface.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1251/commits/dcfaa50c2e6b429be652c0db47e8dae108ab9dcb

unittests/sync_call_tests.cpp
```diff
@@ -0,0 +1,504 @@
...
+   (data (i32.const 0) "sync_call in same contract called")
+)
+)=====";
+
+// Verify sync call to a function in the same account works
+BOOST_AUTO_TEST_CASE(same_account) { try {
+   validating_tester t;
+
+   if( t.get_config().wasm_runtime == wasm_interface::vm_type::eos_vm_oc ) {
+      // skip eos_vm_oc for now.
```
heifner, line 52: why?

heifner (changes requested):

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -129,37 +130,67 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
+
+         const std::optional<sync_call_context> sync_call_ctx  = context.get_sync_call_ctx();
```
heifner, line 138: Seems expensive to make this call of the sync call data.

libraries/chain/apply_context.cpp
```diff
@@ -233,6 +246,59 @@ void apply_context::exec()
...
+      handle_exception(e);
+   } catch (const std::exception& e) {
+      auto wrapper = fc::std_exception_wrapper::from_current_exception(e);
+      handle_exception(wrapper);
+   }
+}
+
+action_name apply_context::get_sync_call_sender() const {
+   // Current context's receiver is the sender of next sync call
+   return sync_call_ctx ? sync_call_ctx->receiver : get_receiver();
```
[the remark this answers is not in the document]
spoonincode: ah okay good missed that

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -129,48 +130,90 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
          } catch(eosio::vm::timeout_exception&) {
             context.trx_context.checktime();
          } catch(eosio::vm::wasm_memory_exception& e) {
             FC_THROW_EXCEPTION(wasm_execution_error, "access violation: ${d}", ("d", e.detail()));
          } catch(eosio::vm::exception& e) {
             FC_THROW_EXCEPTION(wasm_execution_error, "eos-vm system failure: ${d}", ("d", e.detail()));
          }
+
+         if (multi_expr_callbacks_allowed) {
+            context.trx_context.checktime(); // protect against the case where during the removal of the callback, the timer expires.
```
spoonincode, line 203: I haven't thought it through entirely but my hunch is this check needs to be higher up (like maybe at the end of `execute_sync_call()` or such). Think about a case where the callee is being run by OC but the caller by JIT. We would need to add this checktime in both OC & JIT; or preferably just one common place higher up.

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: Thanks. I changed to use a pre-created pool of allocators to avoid expensive allocation for each sync call.

https://github.com/AntelopeIO/spring/pull/1251/commits/48c3e592215a318b7b950b505722a54e53bfbd4d

unittests/sync_call_tests.cpp
linh2931: OC changes have not been done.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -152,6 +152,7 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
          };
 
          execute(context, bkend, exec_ctx, wasm_alloc, fn, true);
+         context.control.return_sync_call_wasm_allocator(); // return the wasm_allocator obtainded by get_sync_wasm_allocator back to the pool
```
spoonincode, line 155: Does this need to be `make_scoped` so it still runs upon exception?

spoonincode: alternatively maybe `sync_call_wasm_alloc_index` can just be reset to 0 upon on a new apply context

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: I moved it to execute_sync_call for now and added a comment to the OC issue. We will find a common place when doing OC change.

https://github.com/AntelopeIO/spring/pull/1251/commits/fc220492545a08e21482a7aebf9fcd6fddf993a3

libraries/chain/controller.cpp
```diff
@@ -1301,6 +1303,9 @@ struct controller_impl {
          if( shutdown ) shutdown();
       } );
 
+      sync_call_wasm_alloc.resize(16); // Will change to use max_sync_call_depth of global property object in future version
```
spoonincode, line 1306: Seems like this would need to be resized on each read only thread too?

spoonincode: It might make more sense to do more of that work on #1257  -- it would have to resize upon the parameters changed etc

libraries/chain/controller.cpp
```diff
@@ -1013,6 +1013,8 @@ struct controller_impl {
    thread_local static platform_timer timer; // a copy for main thread and each read-only thread
 #if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
    thread_local static vm::wasm_allocator wasm_alloc; // a copy for main thread and each read-only thread
+   thread_local static std::vector<vm::wasm_allocator> sync_call_wasm_alloc; // used for sync call. expensive to create one for each call
+   uint32_t            sync_call_wasm_alloc_index;
```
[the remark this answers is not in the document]
linh2931: Thanks for your good eyes! I intended it to be thread_local but did not do that. Will correct it.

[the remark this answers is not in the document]
linh2931: It was a thread_local below in https://github.com/AntelopeIO/spring/blob/fc220492545a08e21482a7aebf9fcd6fddf993a3/libraries/chain/controller.cpp#L5073

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: We cannot just reset to 0 upon on a new apply context, as the apply context can make multiple sequential sync calls. I changed to use scoped exit mechanism.

https://github.com/AntelopeIO/spring/pull/1251/commits/06bdd12b595735e7a885e3422925037f5d58eaa2

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: Fixed in https://github.com/AntelopeIO/spring/pull/1251/commits/06bdd12b595735e7a885e3422925037f5d58eaa2

libraries/chain/controller.cpp
linh2931: Yes, on #1257, I will do the resize changes using the new parameters.

linh2931: I added the initialization of sync_call_wasm_alloc on read threads at start up https://github.com/AntelopeIO/spring/pull/1251/commits/06bdd12b595735e7a885e3422925037f5d58eaa2. Will update it in #1257 with actual sync_call_depth.

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: I added a test to make 500 sync calls in a loop to verify wasm allocators are properly released after each call (we are not running out of them).

https://github.com/AntelopeIO/spring/pull/1251/commits/b1cbe35ce8114475972aebf0b994af7b1da39db6

libraries/chain/include/eosio/chain/apply_context.hpp
linh2931: Thanks. The Implementation section of the design document talked about this too. This PR focuses on demonstrating the feasibility of sync calls. I will create a task to track it.

linh2931: https://github.com/AntelopeIO/spring/issues/1261

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -597,8 +601,11 @@ class apply_context {
       bool is_privileged()const { return privileged; }
       action_name get_receiver()const { return receiver; }
       const action& get_action()const { return *act; }
+      const action* get_action_ptr()const { return act; }
```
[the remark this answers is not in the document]
linh2931: All the validations which disallow certain functions under sync calls are planned to be done by https://github.com/AntelopeIO/spring/issues/1218. Probably I should create a separate issues for functions in apply_context.

libraries/chain/apply_context.cpp
```diff
@@ -233,6 +246,59 @@ void apply_context::exec()
...
+   auto handle_exception = [&](const auto& e)
+   {
+      // TBD need to store the exception in the sync call trace
+      throw;
+   };
+
+   try {
+      try {
+         const account_metadata_object* receiver_account = &db.get<account_metadata_object, by_name>( receiver);
+         if (receiver_account->code_hash == digest_type()) {
```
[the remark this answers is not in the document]
linh2931: Thanks.

https://github.com/AntelopeIO/spring/pull/1251/commits/81a602015aed6bb10287c0e4f22ce78fc8ffe810

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
linh2931: > Should this be an assert? Is it valid to call outside of a sync call? Also I don't understand what the comment is trying to convey.

A sync call be be initiated by a sync call (using sync_call_ctx) or an action (using apply_context). It is valid for this function to be called outside of a sync call.

I clarified the comments and move the function to the private section.

https://github.com/AntelopeIO/spring/pull/1251/commits/c2f95c0bc73bf0721f59ffe026886497c2fbb342

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1251/commits/d031917abdc6f9ec8152f5edcf43581f36e47bca

spoonincode:
It seems like a sync call can not be privileged. I can't tell if that's intentional or not (it's not mentioned in design doc), and I don't know what to immediately think about it (i.e. does it prevent useful use cases of sync calls?)

linh2931:
> It seems like a sync call can not be privileged. I can't tell if that's intentional or not (it's not mentioned in design doc), and I don't know what to immediately think about it (i.e. does it prevent useful use cases of sync calls?)

@arhag, any comments?

libraries/chain/controller.cpp
linh2931: I created a task for resizing the vector as it requires some experiment. https://github.com/AntelopeIO/spring/issues/1269

spoonincode: Probably also dependent on if we allow privileged sync calls, since a privileged sync call might, for example, reduce the sync call depth beyond the current depth

heifner (approved):

spoonincode:
@linh2931 which PR do you want to plumb in privileged support?

linh2931:
> @linh2931 which PR do you want to plumb in privileged support?

I plan to do it in the restructured sync_call_context class and need to consider changing `max_sync_call_depth` while in a sync call (probably not allow). I just created an issue to track the task
https://github.com/AntelopeIO/spring/issues/1279

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -621,6 +627,11 @@ class apply_context {
       bool                          privileged   = false;
       bool                          context_free = false;
 
+      std::optional<sync_call_context> sync_call_ctx{};  // only one of act and sync_call_ctx can be present
+
+      // Returns the sender of any sync call initiated by this apply_context or its sync_call_ctx
+      action_name get_sync_call_sender() const;
```
spoonincode, line 633: It is strange that this is `action_name` for a name that isn't an action. But I see it's the same for `get_sender()`

spoonincode (approved):

libraries/chain/include/eosio/chain/apply_context.hpp
linh2931: In https://github.com/AntelopeIO/spring/pull/1273, `get_sync_call_sender()` is replaced with the virtual function `get_sender()`.

ericpassmore:
Note:start
category: Other
component: SyncCalls
summary: Initial commit of core functionality for sync calls.
Note:end

Issue #1256, opened 2025-03-13
SC: Resource limits


linh2931:
This documents preliminary resource limits for sync calls agreed at today's (03/13/2025) standup meeting.

1. Max depth of nested sync calls: 16
2. Max number of bytes of sync call data (input) and return value (output) each: 512 KB

PR #1257, opened 2025-03-14
SC: Introduce chain_config_v2 for max_sync_call_depth and max_sync_call_data_size

linh2931:
This PR introduces a new version of chain_config `chain_config_v2` for sync call parameters:  `max_sync_call_depth` and `max_sync_call_data_size`. They are on chain parameters and require BP agreement for changes.

A new version of snapshot (`snapshot_v9`) is required to accommodate  `chain_config_v2`.

`abieos` needs to be updated https://github.com/AntelopeIO/abieos/issues/39

Tests have been updated to verify the new parameters can be read and set.

Resolves https://github.com/AntelopeIO/spring/issues/1256


libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -322,6 +387,41 @@ inline DataStream &operator<<(DataStream &s, const eosio::chain::data_entry<eosi
...
+   using namespace eosio::chain;
+
+   //initial requirements were to skip packing field if it is not activated.
+   //this approach allows to spam this function with big buffer so changing this behavior
+   //moreover:
+   //The contract has no way to know that the value was skipped and is likely to behave incorrectly.
+   //When the protocol feature is not activated, the old version of nodeos that doesn't know about
+   //the entry MUST behave the same as the new version of nodeos that does.
+   //Skipping known but unactivated entries violates this.
+   EOS_ASSERT(entry.is_allowed(), unsupported_feature, "config id ${id} is no allowed", ("id", entry.id));
```
spoonincode, line 408: "not allowed"

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -419,6 +519,34 @@ inline DataStream &operator>>(DataStream &s, eosio::chain::data_entry<eosio::cha
...
+ * that is used as part of unpacking data_range<chain_config_v2, ...>
+ * @param s datastream
+ * @param entry contains config reference and particular id
+ * @throws unsupported_feature if protocol feature for particular id is not activated
+ */
+template <typename DataStream>
+inline DataStream &operator>>(DataStream &s, eosio::chain::data_entry<eosio::chain::chain_config_v2, eosio::chain::config_entry_validator> &entry){
+   using namespace eosio::chain;
+
+   EOS_ASSERT(entry.is_allowed(), unsupported_feature, "config id ${id} is no allowed", ("id", entry.id));
```
spoonincode, line 533: "not allowed"

libraries/chain/chain_config.cpp
```diff
@@ -45,6 +45,14 @@ void chain_config_v1::validate() const {
                "max action return value size should be less than MAX_SIZE_OF_BYTE_ARRAYS" );
 }
 
+void chain_config_v2::validate() const {
+   chain_config_v1::validate();
+   EOS_ASSERT( 1 <= max_sync_call_depth, action_validate_exception,
+               "max sync call depth should be at least 1" );
+   EOS_ASSERT( max_sync_call_data_size <= MAX_SIZE_OF_BYTE_ARRAYS, action_validate_exception,
+               "max ssync call data size should be less than MAX_SIZE_OF_BYTE_ARRAYS" );
```
spoonincode, line 53: "max sync"

spoonincode: also it might be nicer to output the value of `MAX_SIZE_OF_BYTE_ARRAYS` here instead of just "MAX_SIZE_OF_BYTE_ARRAYS"

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -184,6 +184,67 @@ struct chain_config_v1 : chain_config_v0 {
...
+
+   inline const Base& base() const {
+      return static_cast<const Base&>(*this);
+   }
+
+   void validate() const;
+
+   template<typename Stream>
+   friend Stream& operator << ( Stream& out, const chain_config_v2& c ) {
+      return c.log(out) << "\n";
```
spoonincode, line 214: Should this have something else? v1 is quite different,
https://github.com/AntelopeIO/spring/blob/ea867bfee0890c49d4ca262f9927ee6dadf05958/libraries/chain/include/eosio/chain/chain_config.hpp#L182-L184

linh2931:
> It looks like ship's ABI needs to have `chain_config_v2` added. I think we can wait on touching abieos for now though, as long as we don't forget later.
> 
> I think snapshot version should be increased to v9 too due to `chain_config_v2` too?

Thanks! Yes, snapshot version needs to be increased to v9. I will change ship and snapshot for now and create an issue for abieos so we won't forget.

libraries/chain/chain_config.cpp
spoonincode: I now see this is just copied how v1 also prints it as  `MAX_SIZE_OF_BYTE_ARRAYS`. Can keep it like this for consistency (or change both I guess).

spoonincode:
I noticed this will probably need to be changed to use chain_config_v1
https://github.com/AntelopeIO/spring/blob/b9293222813cb37bffaec225819148e64d2765f7/libraries/chain/include/eosio/chain/global_property_object.hpp#L59-L73

I'm guessing we're also going to need to create a `snapshot_global_property_object_v8` since the current serialized GPO has a chain_config_v1

linh2931:
> > It looks like ship's ABI needs to have `chain_config_v2` added. I think we can wait on touching abieos for now though, as long as we don't forget later.
> > I think snapshot version should be increased to v9 too due to `chain_config_v2` too?
> 
> Thanks! Yes, snapshot version needs to be increased to v9. I will change ship and snapshot for now and create an issue for abieos so we won't forget.

V9 snapshot: https://github.com/AntelopeIO/spring/pull/1257/commits/35cee9e668b8eb72914f6c0df4cefe36d3723939

linh2931:
> I noticed this will probably need to be changed to use chain_config_v1
> 
> https://github.com/AntelopeIO/spring/blob/b9293222813cb37bffaec225819148e64d2765f7/libraries/chain/include/eosio/chain/global_property_object.hpp#L59-L73
> 
> I'm guessing we're also going to need to create a `snapshot_global_property_object_v8` since the current serialized GPO has a chain_config_v1

Thanks. Done in the snapshot commit https://github.com/AntelopeIO/spring/commit/aa28ececcf56c9d4d6429f15ce729a30b5296d7a.

It seems the version numbers for Legacy snapshot_global_property_object grows sequentially. I named it `snapshot_global_property_object_v6`

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1257/commits/c77065ab4cd07db17cdd94598170cc1cfc8caea5
I also corrected "no allowed" in all other places in the same file.

libraries/chain/chain_config.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1257/commits/318d440e9a8279df2ca491af899d031fca5d1da5
I updated v1's too.

linh2931:
> > > It looks like ship's ABI needs to have `chain_config_v2` added. I think we can wait on touching abieos for now though, as long as we don't forget later.
> > > I think snapshot version should be increased to v9 too due to `chain_config_v2` too?
> > 
> > 
> > Thanks! Yes, snapshot version needs to be increased to v9. I will change ship and snapshot for now and create an issue for abieos so we won't forget.
> 
> V9 snapshot: [35cee9e](https://github.com/AntelopeIO/spring/commit/35cee9e668b8eb72914f6c0df4cefe36d3723939)

SHiP ABI update: https://github.com/AntelopeIO/spring/pull/1257/commits/2ffd8980a1566054bded1d55d44d3d388b03d3bf

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: Do I miss something of this comment? `Stream& log(Stream& out) const` is for `v1`'s `log` function, while `<<` is for v2's `<<` operator.

spoonincode: Looks good I got mixed up

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -184,6 +184,60 @@ struct chain_config_v1 : chain_config_v0 {
...
+   }
+
+   friend inline bool operator == ( const chain_config_v2& lhs, const chain_config_v2& rhs ) {
+      //add v1 parameters comarison here
```
heifner, line 218: This comment is misleading, probably can just remove it.

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -184,6 +184,60 @@ struct chain_config_v1 : chain_config_v0 {
...
+      return ( std::tie(lhs.max_sync_call_depth, lhs.max_sync_call_data_size) ==
+               std::tie(lhs.max_sync_call_depth, rhs.max_sync_call_data_size) ) &&
+             lhs.base() == rhs.base();
+   }
+
+   friend inline bool operator != ( const chain_config_v2& lhs, const chain_config_v2& rhs ) {
+      return !(lhs == rhs);
+   }
```
heifner, line 226: `operator!=` should not be needed with c++20.

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -184,6 +184,60 @@ struct chain_config_v1 : chain_config_v0 {
...
+
+   inline chain_config_v2& operator= (const chain_config_v0& b) {
+      chain_config_v0::operator= (b);
+      return *this;
+   }
```
heifner, line 231: Why do you need an assignment operator from v0 that leaves v1 and v2 data alone? If this is needed, seems like an explicit `copy()` method would be safer.

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -419,6 +512,34 @@ inline DataStream &operator>>(DataStream &s, eosio::chain::data_entry<eosio::cha
...
+      fc::raw::unpack(s, entry.config.max_sync_call_data_size);
+      break;
+      default:
+      eosio::chain::data_entry<chain_config_v1, config_entry_validator> base_entry(entry);
+      fc::raw::unpack(s, base_entry);
+   }
+
+   return s;
+}
+
```
heifner, line 542: Is this a duplicate? Looks the same as the one above.

libraries/chain/include/eosio/chain/global_property_object.hpp
```diff
@@ -66,11 +66,25 @@ namespace eosio::chain {
...
+         // minimum_version and maximum_version refer to chain_snapshot_header
+         // version. snapshot_global_property_object_v6 applies to version 7 & 8
+         static constexpr uint32_t minimum_version = 7;
+         static constexpr uint32_t maximum_version = 8;
+         static_assert(chain_snapshot_header::minimum_compatible_version <= maximum_version,
+                       "snapshot_global_property_object_v6 is no longer needed");
+
+         std::optional<block_num_type>       proposed_schedule_block_num;
+         producer_authority_schedule         proposed_schedule;
+         chain_config_v1                     configuration;
```
heifner, line 84: `chain_config_v2`. How did tests pass? Need a test with non-default values.

heifner: Nice, `kv_configuration` is gone.

heifner (changes requested):

libraries/chain/include/eosio/chain/global_property_object.hpp
spoonincode: I think `chain_config_v1` is right here since this for v7 & v8 snapshots.

libraries/chain/include/eosio/chain/chain_config.hpp
greg7mdp: > `operator!=` should not be needed with c++20.

No need to define `operator ==()` when `= default;` is the same thing.

spoonincode:
btw I realize now that AntelopeIO/leap#1770 introduces an inconsistency. The premise is that we shouldn't return `wasm_config` unless the protocol feature is activated. ok sounds good. But we still always return `max_action_return_value_size` too -- even if _that_ protocol feature is not enabled. This PR adds two more items returned even when their protocol feature is not activated.

I'm not sure what the right answer is.. we could potentially scrub the `get_consensus_parameters` reply for any config item related to an unactivated feature. That seems kinda hacky but would at least be consistent with AntelopeIO/leap#1770 and the confusion it solves.

libraries/chain/include/eosio/chain/global_property_object.hpp
linh2931: Yes, this is for V7 and V8. That's why `chain_config_v1`.

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: This is required by https://github.com/AntelopeIO/spring/blob/2ffd8980a1566054bded1d55d44d3d388b03d3bf/libraries/chain/controller.cpp#L2568

spoonincode:
Also need to update this part of ship,
https://github.com/AntelopeIO/spring/blob/2ffd8980a1566054bded1d55d44d3d388b03d3bf/libraries/state_history/include/eosio/state_history/serialization.hpp#L332-L333

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: It's not a duplicate. It is for v2. The one above is for v1.

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1257/commits/93f0295e3ab8e056ab42270d2a28620bcee562e4

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: Thanks. Removed. 

https://github.com/AntelopeIO/spring/pull/1257/commits/93f0295e3ab8e056ab42270d2a28620bcee562e4

spoonincode: Don't we need the `= default` still?  I think this will skip equality check of v2 members as is

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: Actually there are 3 places using it.

Done by https://github.com/AntelopeIO/spring/pull/1257/commits/1890e98892e07c274bb2d434ca2a3c15f6c87ca5

spoonincode: What makes it safer? At least the old way didn't have to repeat every item manually (which seems _less_ safe)

libraries/chain/webassembly/privileged.cpp
```diff
@@ -230,7 +230,7 @@ namespace eosio { namespace chain { namespace webassembly {
       cfg.validate();
       context.db.modify( context.control.get_global_properties(),
          [&]( auto& gprops ) {
-              gprops.configuration = cfg;
+              gprops.configuration.copy_from_v0(cfg);
```
spoonincode, line 233: > Why do you need an assignment operator from v0 that leaves v1 and v2 data alone? 

This might be why. Does this change technically cause a consensus violation? The old code would leave `max_action_return_value_size` unchanged but this code will reset it to `config::default_max_action_return_value_size`. (I don't see any reason one can't call `set_blockchain_parameters_packed()` after activating `ACTION_RETURN_VALUE`)

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: I thought about that too. The new version explicitly sets v1 and v2. But that could be done without manually setting the long list of v0 fields. Should I revert to the original version, only adding the v1 and v2 portion?

libraries/chain/webassembly/privileged.cpp
linh2931: Good catch! Will revert.

libraries/chain/include/eosio/chain/chain_config.hpp
spoonincode: If we don't like the unusual `operator=` aspect of this, we could still have a `copy()` that called `chain_config_v0::operator= (other)` I think? That way you don't have the unusual `operator=` and don't need to repeat each v0 item (personally I'd suggest we def avoid that). This is all mostly a style thing though; the original unusual `operator=` doesn't bother me all that much (esp now that we understand why it's there), so need to hear more @heifner thoughts on it.

libraries/chain/webassembly/privileged.cpp
spoonincode: I don't think you need to entirely revert but rather `copy_from_v0()` shouldn't stomp on v1 & v2 items.

spoonincode: > I don't think you need to entirely revert

(pending the style question in the other thread)

libraries/chain/include/eosio/chain/chain_config.hpp
linh2931: I like this compromise.

linh2931: https://github.com/AntelopeIO/spring/pull/1257/commits/7a19b34dc68cfe0c28f4e73ea5754385f9420ab1

linh2931:
> Also need to update this part of ship,
> 
> https://github.com/AntelopeIO/spring/blob/2ffd8980a1566054bded1d55d44d3d388b03d3bf/libraries/state_history/include/eosio/state_history/serialization.hpp#L332-L333

Thanks. Done.

https://github.com/AntelopeIO/spring/pull/1257/commits/806fd71cf8975ce90aa351b7703a96aa060d5656

libraries/chain/include/eosio/chain/chain_config.hpp
heifner: I like the compromise as well.

spoonincode:
> > Also need to update this part of ship,
> > https://github.com/AntelopeIO/spring/blob/2ffd8980a1566054bded1d55d44d3d388b03d3bf/libraries/state_history/include/eosio/state_history/serialization.hpp#L332-L333
> 
> Thanks. Done.
> 
> [806fd71](https://github.com/AntelopeIO/spring/commit/806fd71cf8975ce90aa351b7703a96aa060d5656)

This part needs to be changed to 2,
https://github.com/AntelopeIO/spring/blob/806fd71cf8975ce90aa351b7703a96aa060d5656/libraries/state_history/include/eosio/state_history/serialization.hpp#L334
(it's manually encoding the variant index).

linh2931:
> ebd7f51

Thank you so much for this! I did not notice that. https://github.com/AntelopeIO/spring/pull/1257/commits/ebd7f51004aa15c527ef10a57ce1a663de421a9a

I have also added chain_config_v2 to abieos sync_call branch. Unit test test_deltas_global_property_history needs that.

libraries/chain/include/eosio/chain/chain_config.hpp
heifner: I mean lines 385-412 above

linh2931: lines 385 - 412 are for `operator<<` (packing), here are for `operator>>` (unpacking).

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -184,6 +184,55 @@ struct chain_config_v1 : chain_config_v0 {
...
+   }
+
+   // no need to define !=
+   bool operator == (const chain_config_v2&) const = default;
+
+   inline void copy_from_v0(const chain_config_v0& b) {
+      // copy v0 fields
+      chain_config_v0::operator= (b);
+
+      // leave v1 and v2 fields alone as changing them might break cosensus
```
spoonincode, line 224: "consensus"

linh2931: Fixed b https://github.com/AntelopeIO/spring/pull/1257/commits/30d934cac99a93d151ece9603657a45df7d3093c

linh2931:
> btw I realize now that [AntelopeIO/leap#1770](https://github.com/AntelopeIO/leap/pull/1770) introduces an inconsistency. The premise is that we shouldn't return `wasm_config` unless the protocol feature is activated. ok sounds good. But we still always return `max_action_return_value_size` too -- even if _that_ protocol feature is not enabled. This PR adds two more items returned even when their protocol feature is not activated.
> 
> I'm not sure what the right answer is.. we could potentially scrub the `get_consensus_parameters` reply for any config item related to an unactivated feature. That seems kinda hacky but would at least be consistent with [AntelopeIO/leap#1770](https://github.com/AntelopeIO/leap/pull/1770) and the confusion it solves.

I created https://github.com/AntelopeIO/spring/issues/1289 to track this.

unittests/snapshot_tests.cpp
```diff
@@ -388,10 +388,10 @@ void compatible_versions_test()
    std::filesystem::copy(source_log_dir / "blocks.index", config.blocks_dir / "blocks.index");
```
heifner, line 388: The `abieos` submodule update should be via a PR to `abieos`. I don't see one.

heifner (approved):
Make sure to update to merged abieos commit.

spoonincode (approved):

unittests/snapshot_tests.cpp
linh2931: As the changes were small and I'd to have a quick test, I went ahead making the abieos changes in the `sync_call` branch. I will make sure all the future changes will be done from a branch created out of the `sync_call` branch.

PR #1265, opened 2025-03-21
P2P: Add gossip of BP peers

### LINKED ISSUE #1245: Add P2P gossip_bp_peer message
Add new gossip_bp_peer message for sending bp peer connection information.

## P2P: Auto-management of Private BP connections

The current `p2p-auto-bp-peer` feature is not useful for Savanna consensus as it only maintains connections to immediate BPs in the schedule. With Savanna consensus, connection to all finalizers (currently BPs) is required. See #1227, fixed by #1244.

* Update `p2p-auto-bp-peer` to maintain connections to the top 21 BPs.
    * Continues to tracks proposer schedule to know which peer keys are currently active.
    * If at least one configured `p2p-auto-bp-peer` then the feature is activated. The `p2p-auto-bp-peer` list is used to bootstrap the process. `p2p-peer-address` entries will also be included if they are bp peers; established by the `gossip_bp_peer` handshake. 
* Add gossip of connections to verifiable BP peers.
      
* New P2P gossip message
   * ```
     struct gossip_bp_peers {
        struct bp_peer {
           eosio::name producer_name;
           std::string server_address;
           // sig over [producer_name, server_address]
           signature sig;
        };
        std::vector<bp_peer> peers;
     };
   * Configured `p2p-signature-provider` will verify received `gossip_bp_peers` and send out `gossip_bp_peers` if any new peers received.
   * `gossip_bp_peers` message only sent to verified gossip bp peers.
   * Any `bp_peer` entry where all peers can't be connected to and verified as a bp peer is dropped from tracked state. This is how no longer valid `bp_peer` entries are removed.
   * The `server_address` allows for one BP to have a number of peers and backup peers. A limit of 4 `bp_peer` entries is allowed per producer_name. The `server_address` acts to distinguish between them.
* `p2p-peer-address` peers will also be used for bootstrapping as long as one `p2p-auto-bp-peer` is configured and will always be connected; subject to the `p2p-peer-limit`. The `p2p-peer-limit` only applies to `p2p-peer-address`, not to gossip bp peers.
    * Connection flow:
        * If configured with `p2p-signature-provider` and `p2p-auto-bp-peer` then send a signed empty `gossip_bp_peer`. 
            * If peer receives a `gossip_bp_peers` with new information it will send the new merged gossip_bp_peers to all of its previously verified bp gossip peers.
* Gossip peers will remain connected when the associated proposer/finalizer is in the top 21.
* When a `p2p-auto-bp-peer` rotates into the the top 21 then it will connect to its `p2p-auto-bp-peer`s and send an empty `gossip_bp_peers` to have current `gossip_bp_peers` sent back to it.
* When a node rotates out of the top 21 then peers will disconnect from it unless they explicitly connected via `p2p-peer-address`. Also the bp peer that rotated out will itself disconnect from any bp peers and wait to be rotated back in before participating in the bp peer gossip.
* The on-chain peer list can be queried every block. If a version can be quickly queried this is likely the easiest and quickest method. Only bp peer nodes need to make this state query. See #1228.

heifner:
Adds new `gossip_bp_peers_message` message to P2P protocol for gossip of available "BP" peers. Normally these peers would be canary relay nodes to the block producer and block finalizer nodeos.

- Gossip connection information is only shared with nodes that are connected to another "BP" node that has a peer key registered on-chain.

- Block producers should register a new peer key on chain. See https://github.com/eosnetworkfoundation/eos-system-contracts/pull/185
- Block producers then run the canary relay node(s) with the private key of the peer key and their block producer account. 
- For example, block producer `defproducera`.
  -  Register a public peer key on change via `regpeerkey` with values: `defproducera` `EOS6y94JK7Dsg3iJCrnvdbyLPZJvaAhfBSU44eU8jkPd4VHX9CJzp`
  - Add the following to their canary relay node.
```
       p2p-producer-peer = defproducera
       signature-provider = EOS6y94JK7Dsg3iJCrnvdbyLPZJvaAhfBSU44eU8jkPd4VHX9CJzp=KEY:5J6Z64Jqexample
       p2p-auto-bp-peer = p2p.peer.com:9876
```

The `p2p-producer-peer` matches the `producer-name` of the BP. The key provided to `signature-provider` is the public/private key of the on-chain key specified with `regpeerkey`. The `p2p-auto-bp-peer` is a known BP peer. This is not technically needed if another BP peer has your endpoint listed in their `p2p-auto-bp-peer` list. 

`p2p-auto-bp-peer` can be specified for as many known bp peers. These will be used along with gossip peers as producers rotate into the producer schedule. They are not gossiped out to other peers. Only this nodes connection info along with other received gossip BP info is gossiped to other peers.

Resolves #1245


greg7mdp:
I'm not getting why you changed `update_peer_keys` to return a `flat_set`?

heifner:
> I'm not getting why you changed `update_peer_keys` to return a `flat_set`?

To know which ones are new or updated.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -68,46 +105,110 @@ class bp_connection_manager {
...
+   }
+
+   // Called at startup and when peer key changes
+   // empty modified_keys means to update all
+   void update_bp_producer_peers(const chain::controller& cc, const flat_set<account_name>& modified_keys, const std::string& server_address)
+   {
+      if (config.my_bp_peer_accounts.empty())
+         return;
+      fc::lock_guard gm(mtx);
+      fc::lock_guard g(gossip_bps.mtx);
```
greg7mdp, line 158: When locking multiple mutexes, I believe we should use [scoped_lock](https://en.cppreference.com/w/cpp/thread/scoped_lock) to avoid deadlocks. Unfortunately I don't believe we have a `fc` version of it.

heifner: Correct, there is not an fc version. Could add it, but didn't think it was worth it. The important thing is to always do it in the same order.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -68,46 +105,110 @@ class bp_connection_manager {
...
+            peer.sig = self()->sign_compact(*pk, peer.digest());
+            if (auto i = prod_idx.find(boost::make_tuple(e, boost::cref(server_address))); i != prod_idx.end()) {
+               gossip_bps.index.modify(i, [&peer](auto& v) {
+                  v = peer;
+               });
+            } else {
+               gossip_bps.index.emplace(peer);
+            }
+         }
+         first = false;
```
greg7mdp, line 183: I'm having a hard time figuring out what `first` is used for?
I wonder if `first = false;` should not be moved one line up. And maybe lines 168-170 executed only if `modified_keys.empty() `? 
Or maybe I completely misunderstands it?

heifner: `// normally only one bp peer account except in testing scenarios or test chains`

`first` is the first configured bp peer account (in the rare cases there are more than one).

It only needs to re-sign the `signed_empty` if the first bp peer account has been modified. If the second or later is modified (but not the first) then no reason to re-sign.

heifner: I do like that better. Good point about the "first". It doesn't matter which one is used, just need only one.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -140,20 +360,31 @@ class bp_connection_manager {
...
+               auto& prod_idx = gossip_bps.index.get<by_producer>();
+               for (const auto& account : pending_connections) {
+                  if (auto i = config.bp_peer_addresses.find(account); i != config.bp_peer_addresses.end()) {
+                     fc_dlog(self()->get_logger(), "connect to manual bp peer ${p}", ("p", i->second));
+                     self()->connections.resolve_and_connect(i->second, self()->get_first_p2p_address() );
+                  }
+                  auto r = prod_idx.equal_range(account);
+                  for (auto i = r.first; i != r.second; ++i) {
+                     fc_dlog(self()->get_logger(), "connect to gossip bp peer ${p}", ("p", i->server_address));
+                     self()->connections.resolve_and_connect(i->server_address, self()->get_first_p2p_address() );
```
greg7mdp, line 377: should we skip this connection attempt (and log message) if we already connected to the same `i->server_address` on line 372?

heifner: `resolve_and_connect` does skip if already connected. It doesn't log it. Could add a log statement, might be useful.

heifner: Oh, ok. Those will now be followed up with what happens.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
              established_client_connection(new_connection) && num_established_clients() > self()->connections.get_max_client_count();
    }
 
+   // thread safe
+   // removes invalid entries from msg
+   bool validate_gossip_bp_peers_message( gossip_bp_peers_message& msg ) {
+      if (msg.peers.empty())
+         return false;
+      if (msg.peers.size() != 1 || !msg.peers[0].server_address.empty()) { // initial case, no server_addresses to validate
```
greg7mdp, line 237: I was a little misled because I would expect that the comment *after* the `{` describes the `then` state. May I suggest something like:
```suggestion
      bool initial_msg = msg.peers.size() == 1 && msg.peers[0].server_address.empty());
      if (!initial_msg) { 
```

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+            const auto& [host, port, type] = net_utils::split_host_port_type(addr);
+            return !host.empty() && !port.empty();
+         };
+         std::map<eosio::name, size_t> producers;
+         const gossip_bp_peers_message::bp_peer* prev = nullptr;
+         for (const auto& peer : msg.peers) {
+            if (peer.producer_name.empty())
+               return false;
+            if (!valid_address(peer.server_address))
+               return false;
```
greg7mdp, line 248: seems that one invalid address will cause us to ignore the whole message. Not what the top comment says:
`// removes invalid entries from msg`.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -68,46 +105,108 @@ class bp_connection_manager {
...
+            EOS_ASSERT(pk, chain::plugin_config_exception, "No on-chain peer key found for ${n}", ("n", e));
+            fc_dlog(self()->get_logger(), "Signing with producer_name ${p} key ${k}", ("p", e)("k", *pk));
+            if (e == *config.my_bp_peer_accounts.begin()) { // use the first one of the set, doesn't matter which one is used, just need one
+               gossip_bp_peers_message::bp_peer signed_empty{.producer_name = e}; // .server_address not set for initial message
+               signed_empty.sig = self()->sign_compact(*pk, signed_empty.digest());
+               initial_gossip_msg_factory.set_initial_send_buffer(signed_empty);
+            }
+            auto& prod_idx = gossip_bps.index.get<by_producer>();
+            gossip_bp_peers_message::bp_peer peer{.producer_name = e, .server_address = server_address};
+            peer.sig = self()->sign_compact(*pk, peer.digest());
```
greg7mdp, line 173: Are we not signing with our own producer key here? If so why do we need to check `modified_keys.contains(e)`? Wouldn't we have restarted our node if we changed our registered key on-chain?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+            if (peer.producer_name.empty())
+               return false;
+            if (!valid_address(peer.server_address))
+               return false;
+            if (prev != nullptr) {
+               if (prev->producer_name == peer.producer_name) {
+                  if (prev->server_address == peer.server_address)
+                     return false; // duplicate entries not allowed
+               } else if (prev->producer_name > peer.producer_name) {
+                  return false; // required to be sorted
```
greg7mdp, line 254: What is the benefit of requiring the entries to be sorted?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+         const gossip_bp_peers_message::bp_peer* prev = nullptr;
+         for (const auto& peer : msg.peers) {
+            if (peer.producer_name.empty())
+               return false;
+            if (!valid_address(peer.server_address))
+               return false;
+            if (prev != nullptr) {
+               if (prev->producer_name == peer.producer_name) {
+                  if (prev->server_address == peer.server_address)
+                     return false; // duplicate entries not allowed
```
greg7mdp, line 252: what is the benefit of doing this check?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+         }
+      }
+
+      controller& cc = self()->chain_plug->chain();
+
+      fc::lock_guard g(gossip_bps.mtx);
+      auto& sig_idx = gossip_bps.index.get<by_sig>();
+      for (auto i = msg.peers.begin(); i != msg.peers.end();) {
+         const auto& peer = *i;
+         try {
```
greg7mdp, line 269: Can the `try/catch` block be moved into the if below (line 270)?

Also instead of all these `erase(); continue;`, I think it would be somewhat cleaner to have a lambda:
```
auto is_peer_msg_valid = [&](const bp_peer& peer) -> bool { ...};
```

so the loop would become:

```
for (auto i = msg.peers.begin(); i != msg.peers.end();) {
   if (!is_peer_msg_valid(*i) 
      i = msg.peers.erase(i);
   else
      ++i;
}
```

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+   }
+
+   // thread-safe
+   bool update_gossip_bps(const gossip_bp_peers_message& msg) {
+      // providing us with full set
+      fc::unique_lock g(gossip_bps.mtx);
+      auto& idx = gossip_bps.index.get<by_producer>();
+      bool diff = false;
+      for (const auto& peer : msg.peers) {
+         if (auto i = idx.find(boost::make_tuple(peer.producer_name, boost::cref(peer.server_address))); i != idx.end()) {
```
greg7mdp, line 305: What if the peer changes the server address it uses? Old addresses will remain in `gossip_bps`, right?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+            if (*i != peer) {
+               gossip_bps.index.modify(i, [&peer](auto& m) {
```
greg7mdp, line 307: Maybe add a comment that this is to update the signature.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+               diff = true;
+            }
+         } else {
+            gossip_bps.index.insert(peer);
+            diff = true;
+         }
+      }
+      g.unlock();
+      if (diff) {
+         connect_to_active();
```
greg7mdp, line 319: Why not have the `connect_to_active()` be done in the caller? That will also avoid the `g.unlock()`.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+   void connect_to_active() {
+      // do not hold mutexes when calling resolve_and_connect which acquires connections mutex since other threads
+      // can be holding connections mutex when trying to acquire these mutexes
+      flat_set<std::string> addresses;
+      {
+         fc::lock_guard gm(mtx);
+         active_bps = active_bp_accounts(active_schedule);
+
+         fc::lock_guard g(gossip_bps.mtx);
+         auto& prod_idx = gossip_bps.index.get<by_producer>();
```
greg7mdp, line 334: can be `const`.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+      if (diff) {
+         connect_to_active();
+      }
+      return diff;
+   }
+
+   // thread-safe
+   void connect_to_active() {
+      // do not hold mutexes when calling resolve_and_connect which acquires connections mutex since other threads
+      // can be holding connections mutex when trying to acquire these mutexes
```
greg7mdp, line 327: We do hold a mutex in `on_pending_schedule` though? Is it not an issue there?
Also, can we avoid this code duplication?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -181,8 +422,18 @@ class bp_connection_manager {
                              std::back_inserter(peers_to_drop));
          fc_dlog(self()->get_logger(), "peers to drop: ${p}", ("p", to_string(peers_to_drop)));
 
+         fc::lock_guard g(gossip_bps.mtx);
+         auto& prod_idx = gossip_bps.index.get<by_producer>();
          for (const auto& account : peers_to_drop) {
-            self()->connections.disconnect(config.bp_peer_addresses[account]);
+            if (auto i = config.bp_peer_addresses.find(account); i != config.bp_peer_addresses.end()) {
+               fc_dlog(self()->get_logger(), "disconnect to manual bp peer ${p}", ("p", i->second));
+               self()->connections.disconnect(i->second);
```
greg7mdp, line 430: `disconnect` takes a lock of `connections_mtx`... maybe it would also be better to call the `disconnect()`s after releasing the `gossip_bps.mtx` mutex?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: It makes the duplicate check simpler.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: Duplicates can bloat the message. Do not allow them to prevent someone spamming a node with a large number of duplicates.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
greg7mdp: Should we check that `sign_compact` does find the private key associated to `*pk`, and that the signature stored in `peer.sig` is not empty?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: Yes. If they change it most likely the previous one is no longer valid in which case the connection will fail. If it remains valid then everyone will connect to both. We should recommend that if they change their server address that they also register a new key on-chain.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +229,127 @@ class bp_connection_manager {
...
+   // removes invalid entries from msg
+   bool validate_gossip_bp_peers_message( gossip_bp_peers_message& msg ) {
+      if (msg.peers.empty())
+         return false;
+      if (msg.peers.size() != 1 || !msg.peers[0].server_address.empty()) { // initial case, no server_addresses to validate
+         auto valid_address = [](const std::string& addr) -> bool {
+            const auto& [host, port, type] = net_utils::split_host_port_type(addr);
+            return !host.empty() && !port.empty();
+         };
+         std::map<eosio::name, size_t> producers;
```
greg7mdp, line 242: We validate that producers are sorted by name. Why do we need a `std::map` at all here?

Something like the following would work I think:

```
      if (prev->producer_name == peer.producer_name) {
         ++num_entries;
         if (num_entries >  4)
            return false; // only 4 entries per producer allowed
      } else
         num_entries = 0;
}

File: plugins/net_plugin/net_plugin.cpp [hunk 17]
```diff
@@ -915,7 +843,8 @@ namespace eosio {
       std::atomic<uint16_t>   protocol_version = 0;
       uint16_t                net_version = net_version_max;
       std::atomic<uint16_t>   consecutive_immediate_connection_close = 0;
-      std::atomic<bool>       is_bp_connection = false;
+      std::atomic<bool>       is_gossip_bp_connection = false;
+      std::atomic<bool>       is_configured_bp_connection = false;
```
greg7mdp, line 847: Could these be replaced by an `enum class connection_type { non_bp, bp_config, bp_gossip };`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3941,6 +3761,73 @@ namespace eosio {
...
+      is_gossip_bp_connection = true;
+
+      assert(!msg.peers.empty()); // checked by validate_gossip_bp_peers_message()
+      if (msg.peers.size() == 1 && msg.peers[0].server_address.empty()) {
+         // initial message case, send back our entire collection
+         send_gossip_bp_peers_message();
+      } else {
+         bool diff = my_impl->update_gossip_bps(msg);
+         if (diff) { // update, let all our peers know about it
+            send_gossip_bp_peers_message_to_bp_peers();
```
greg7mdp, line 3795: I think we should wait a few seconds before sending to all peers, so that we allow those peers to get the initial message from the source, preventing them from rebroadcasting the message again.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3941,6 +3761,73 @@ namespace eosio {
...
+      const send_buffer_type& sb = my_impl->get_gossip_bp_send_buffer(factory);
+      enqueue_buffer(msg_type_t::gossip_bp_peers_message, {}, queued_buffer::queue_t::general, sb, no_reason);
+   }
+
+   // called from connection strand, thread safe
+   void connection::send_gossip_bp_peers_message_to_bp_peers() {
+      assert(my_impl->bp_gossip_enabled());
+      my_impl->connections.for_each_connection([this](const connection_ptr& c) {
+         gossip_buffer_factory factory;
+         if (this != c.get() && c->is_gossip_bp_connection && c->socket_is_open()) {
```
greg7mdp, line 3822: Se we don't want to update the `is_configured_bp_connection` peers?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: Yes, that is an issue. Thanks.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: Yep.

greg7mdp:
My understanding is that we didn't want to directly populate the server IP addresses on-chain to avoid disclosing these publicly.
Is it an issue that they appear in clear in the gossip messages?

heifner:
> My understanding is that we didn't want to directly populate the server IP addresses on-chain to avoid disclosing these publicly. Is it an issue that they appear in clear in the gossip messages?

They will only appear in the clear in gossip messages sent to peers that have been verified to be a bp gossip peer.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4636,6 +4542,14 @@ namespace eosio {
 
          cc.aggregated_vote().connect( broadcast_vote );
          cc.voted_block().connect( broadcast_vote );
+
+         if (bp_gossip_enabled()) {
+            cc.set_peer_keys_retrieval_active(true);
+            // update peer public keys from chainbase db
+            cc.update_peer_keys(cc.head().irreversible_blocknum());
+            // pass in empty set so validation that peer key exists is enforced
```
greg7mdp, line 4551: that passing of empty set is not necessary or useful afaict.
Since this is the first time we call `cc.update_peer_keys()`, if the key can be found using `cc.get_peer_key(e)`, then it will also be present in the returned `flat_set`.

heifner: If the node configures a `p2p-producer-peer` that is not registered on-chain then we want an error on startup.

greg7mdp: Ah yes, I just realized that thanks!

plugins/net_plugin/net_plugin.cpp
heifner: No, manually configured bp connections are not shared with others.

plugins/net_plugin/net_plugin.cpp
greg7mdp: Suppose:
- node A broadcasts a new peer to B, C, D, E.
- B receives it first and broadcasts to A, C, D, E
- C, D, E receive B's message (before A's) and each broadcast to 4 nodes.

If B had waited, C, D, and E would have received A's message first and would not have rebroadcasted (3rd step above)

greg7mdp: So a peer key update will cause a volley of (N-1)**2 gossip messages, if N is the number of finalizers. Could be an issue if N was to become significantly higher than 21.

heifner: This scheme does not work for large number of finalizers. A new approach will be needed if/when the number of finalizers is expanded. Similarly vote aggregation will have to be handled differently. Presumably whatever scheme used for vote aggregation can also be used to connections.

greg7mdp: What do you think of that:

> One way to avoid this squaring of the messages would be if the gossip messages also included information to addresses which are notified, so the recipients would not send duplicate messages to these.

heifner: You don't know peer actually received the message, only that it was sent. The processing of these messages seems extremely quick so not sure that is anything to worry about.

greg7mdp:
I wonder if we should send the vote messages only on the peer gossip connections?
Since every BP will be commected to most other BPs via `bp-gossip` connections, it should be enough to send the vote messages on these connections only I think.

ericpassmore:
Note:start
category: Other
component: P2P
summary: Add gossip of BP peers.
Note:end

plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
```diff
@@ -104,5 +105,5 @@ namespace eosio {
 }
 
 FC_REFLECT( eosio::connection_status, (peer)(remote_ip)(remote_port)(connecting)(syncing)
-                                      (is_bp_peer)(is_socket_open)(is_blocks_only)(is_transactions_only)
+                                      (is_bp_peer)(is_bp_gossip_peer)(is_socket_open)(is_blocks_only)(is_transactions_only)
```
heifner, line 108: Add bp-peer-name and rank.

libraries/chain/controller.cpp
```diff
@@ -3384,12 +3389,16 @@ struct controller_impl {
       return onblock_trace;
    } /// start_block
 
-   void update_peer_keys(fc::time_point deadline) {
+   void update_peer_keys() {
+      // if syncing or replaying old blocks don't bother updating peer keys
```
spoonincode, line 3393: good call I didn't think of this when reviewing other PR

libraries/chain/include/eosio/chain/peer_keys_db.hpp
```diff
@@ -16,28 +16,37 @@ class peer_keys_db_t {
    using peer_key_map_t = boost::unordered_flat_map<name, peer_info_t, std::hash<name>>;
    using new_peers_t    = flat_set<name>;
```
greg7mdp, line 17: Can we make it a `boost::unordered_flat_set`?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -68,46 +108,106 @@ class bp_connection_manager {
...
-   void mark_bp_connection(Connection* conn) const {
-      /// mark an connection as a bp connection if it connects to an address in the bp peer list, so that the connection
-      /// won't be subject to the limit of max_client_count.
+   void mark_configured_bp_connection(Connection* conn) const {
+      /// mark an connection as a configured bp connection if it connects to an address in the bp peer list,
+      /// so that the connection won't be subject to the limit of max_client_count.
       auto space_pos = conn->log_p2p_address.find(' ');
       // log_p2p_address always has a trailing hex like `localhost:9877 - bc3f55b`
       std::string addr = conn->log_p2p_address.substr(0, space_pos);
       if (config.bp_peer_accounts.count(addr)) {
```
greg7mdp, line 193: I'm not sure howw reliable that is, wouldn't this fail if `set_configured_bp_peers` was called with strings having a space after the coma, or at the end of the string after the ip address.
Why not store a std::pair of `host, port` as returned from `net_utils::split_host_port_type()` instead of a string?

libraries/chain/controller.cpp
```diff
@@ -1303,7 +1303,7 @@ struct controller_impl {
             transaction_metadata::trx_type::read_only);
 
          // allow a max of 20ms for getpeerkeys
-         auto trace = push_transaction(metadata, deadline, fc::milliseconds(20), 0, false, 0);
+         auto trace = push_transaction(metadata, fc::time_point::maximum(), fc::milliseconds(20), 0, false, 0);
```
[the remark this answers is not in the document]
heifner: This is now only called for applied blocks, not produced blocks. We have no deadline for applying blocks.

libraries/chain/include/eosio/chain/peer_keys_db.hpp
heifner: Actually no. We use `std::set_union` which doesn't work for `unordered_*`.

spoonincode:
I don't think it was changed in this PR but looking through the code I see it creates and verifies canonical signatures. Maybe consider turning that off (passing false) now from the start of this feature

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +230,141 @@ class bp_connection_manager {
...
+            prev = &peer;
+         }
+      }
+
+      const controller& cc = self()->chain_plug->chain();
+      auto is_peer_key_valid = [&](const gossip_bp_peers_message::bp_peer& peer) -> bool {
+         try {
+            std::optional<peer_info_t> peer_info = cc.get_peer_info(peer.producer_name);
+            if (peer_info && peer_info->key) {
+               public_key_type pk(peer.sig, peer.digest());
```
spoonincode, line 275: I think we want to immediately kick out SIG_WA. We did make the contract disallow PUB_WA.

libraries/chain/include/eosio/chain/peer_keys_db.hpp
greg7mdp: Where do we use `new_peers_t ` for `std::set_union`?

heifner: I did go ahead and rename to `peer_name_set_t` and use throughout.

https://github.com/AntelopeIO/spring/blob/596135dc238ffb98c07dc31fab353011b536500e/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp#L422-L422

heifner: https://github.com/AntelopeIO/spring/pull/1265/commits/596135dc238ffb98c07dc31fab353011b536500e

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1265/commits/f8166430a538187d3fd8085b0e7730b0efc6679f

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1265/commits/3b8145f94d14b9ea06e3b7696ed1b84e910b7ffe

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +232,152 @@ class bp_connection_manager {
...
+               gossip_bps.index.erase(idx.find(peer.producer_name));
+            }
+            gossip_bps.index.insert(peer);
+            diff = true;
+         }
+      }
+      return diff;
+   }
+
+   flat_set<std::string> find_gossip_bp_addresses(const peer_name_set_t& accounts, const char* desc) const {
```
greg7mdp, line 347: Is there a reason to use a `flat_set` as a return value, as opposed to a `std::vector`?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +232,152 @@ class bp_connection_manager {
...
+            } else { // unknown key
+               // ok, might have just been deleted or dropped out of top ranking
+               fc_dlog(self()->get_logger(), "Failed to find peer key ${p}", ("p", peer.producer_name));
+               return false;
+            }
+         } catch (fc::exception& e) {
+            fc_dlog(self()->get_logger(), "Exception recovering peer key ${p}, error: ${e}", ("p", peer.producer_name)("e", e.to_detail_string()));
+            fatal_error = true;
+            return false; // invalid key
+         }
```
greg7mdp, line 299: Can something throw an exception in this lambda?
If not we can remove this `fatal_error` stuff.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +232,152 @@ class bp_connection_manager {
...
+      // can be holding connections mutex when trying to acquire these mutexes
+      flat_set<std::string> addresses;
+      {
+         fc::lock_guard gm(mtx);
+         active_bps = active_bp_accounts(active_schedule);
+         addresses = find_gossip_bp_addresses(active_bps, "connect");
+      }
+
+      for (const auto& add : addresses) {
+         self()->connections.resolve_and_connect(add, self()->get_first_p2p_address());
```
greg7mdp, line 377: This will likely fill the debug log with:

`Already connected to ${p}`

mabe pass a flag to `resolve_and_connect` to avoid logging if already connected.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: Yes. `public_key_type` constructor can throw on invalid format for example.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1265/commits/f778d714a3202f9647b37188c35ac8ea13af90db

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -130,6 +232,152 @@ class bp_connection_manager {
...
+            fc_dlog(self()->get_logger(), "Exception recovering peer key ${p}, error: ${e}", ("p", peer.producer_name)("e", e.to_detail_string()));
+            fatal_error = true;
+            return false; // invalid key
+         }
+         return true;
+      };
+
+      fc::lock_guard g(gossip_bps.mtx);
+      auto& sig_idx = gossip_bps.index.get<by_sig>();
+      for (auto i = msg.peers.begin(); i != msg.peers.end() && !fatal_error;) {
```
greg7mdp, line 305: I don't think we need to exit the loop early if `fatal_error` is true (it is after all an optimization for a very unlikely case).

And if we don't, we can use `std::erase` instead of that for loop which makes the code clearer I feel.

greg7mdp: nit: Maybe `invalid_message` would be more descriptive than `fatal_error`.

heifner: I think it is important to exit early on a bad `public_key`. Seems like those could be crafted to be expensive to evaluate. As soon as one is determined to be invalid, don't bother trying to verify signature on any others.

heifner: https://github.com/AntelopeIO/spring/pull/1265/commits/75c63e39f526c5ffe5d2de44e95fe98f3c521905

greg7mdp: > Seems like those could be crafted to be expensive to evaluate.
I don't think it is a very likely attack vector from one of the top-50 producers, but OK.

greg7mdp (approved):

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: It could be anyone that is connected to the node.

spoonincode: That aspect is what I've been thinking on some and why suggested disallowing SIG_WA -- those have the most opportunity to create pain on the very first signature in the message.

I have been trying to think of what else could be problematic here. I don't see a limit for the size of `gossip_bp_peers_message` which means you could have as many peers as you can stuff in 8MB (or until hitting max array elements). So someone can flood `gossip_bp_peers_message`s with ~110k peers each taking ~40µs to recover key leading to over 4 seconds of work here?

heifner: Only `max_bp_peers_per_producer` (4) per producer is allowed. The number allowed here will be limited to 4x(max-allowed-by-contract) which is currently 200 (4x50).

spoonincode: That limit is enforced prior to checking the signatures though. So as an attacker why can't I just submit a message with names `a`, `b`, `c`, etc all with valid signatures.

heifner: `a`, `b`, `c`, have to be in the top 50 to eval the signature.
https://github.com/AntelopeIO/spring/blob/75c63e39f526c5ffe5d2de44e95fe98f3c521905/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp#L281-L282

greg7mdp: > It could be anyone that is connected to the node.

I thought the peer gossip was restricted to the top-50 producers. Maybe we should ignore `peer_gossip` messages if they don't come from one of the top-50 producers.

heifner: I take that back. We do send a full bp peer message to the peers sent to us. We only validate the initial one, after that we trust the info that gossip peer sends up.

heifner: If the message is not from one of the top-50 we know about then we will not consider that a valid message and will disconnect.

greg7mdp: So this means that the attack with invalid signatures is not very likely, right. Not advocating for a change, but just that these messages are among a trusted group (all trusted to see the IP of all block producers, so if there was a rogue element it could probably do more damage by ddos'ing producers as they are scheduled to produce than by forging public keys that take time to validate).

heifner: Well anyone could send a crafted `gossip_bp_peers_message` we have to validate to know it is from a valid producer. 

This conversation did make me realize that it was a bit too strict. Just because a peer is no longer in the top producers does not mean we should `fatal` disconnect from them.

greg7mdp: > Well anyone could send a crafted `gossip_bp_peers_message` we have to validate to know it is from a valid producer.

Hum, I wonder if the whole message (the `std::vector<bp_peer>` + sender's name) should not be signed by the peer sending it. Then the first thing we could do is validate that the signature matches for one  connection IP, so we authenticate that it comes from one of the current `top-50` producers.

heifner: We could certainly do that. Not sure it is worth the extra processing. Open to what others think.

spoonincode (approved):

Issue #1219, opened 2025-02-27
SC: Validate `sync_call` entry point signature


linh2931:
`sync_call` entry point is optional. But if it is present, validate its signature at contract deployment.

linh2931:
Use `sync_call_supported` in `code_object` to validate sync calls.

PR #1266, opened 2025-03-21
SC: Validate sync call entry point

linh2931:
Implement this feature earlier to help prevent errors in WASM test contracts.

* Before a contract is inserted into `wasm_instantiation_cache`, check if `sync_call` entry point is  present in the contract; if present, validate the signature. Set `receiver_supports_sync_call` flag of the cache entry accordingly.
* When a sync call is made, check `receiver_supports_sync_call` of the receiver. If `false`, return `-1` and let the caller decide what to do about the failure.
* Tests are added for those validation.

Resolves  https://github.com/AntelopeIO/spring/issues/1219

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -259,8 +259,8 @@ int32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::spa
    const auto max_sync_call_data_size = control.get_global_properties().configuration.max_sync_call_data_size;
    EOS_ASSERT(data.size() <= max_sync_call_data_size, sync_call_call_data_exception,
               "sync call call data size must be less or equal to ${s} bytes", ("s", max_sync_call_data_size));
-   EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::last), sync_call_validate_exception,  // all but last `std::bit_width(sync_call_flags::last)` bits must be 0s
-              "only ${last} least significant bits of sync call's flags (${flags}) can be set", ("last", std::bit_width(static_cast<uint64_t>(sync_call_flags::last)))("flags", flags));
+   EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::all_allowed_bits), sync_call_validate_exception,  // all but `std::bit_width(all_allowed_bits)` LSBs must be 0s
+              "only ${bits} least significant bits of sync call's flags (${flags}) can be set", ("bits", std::bit_width(static_cast<uint64_t>(sync_call_flags::all_allowed_bits)))("flags", flags));
 
    auto handle_exception = [&](const auto& e)
    {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -247,7 +247,7 @@ void apply_context::exec()
 
 } /// exec()
 
-int32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<const char> data) {
+int64_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<const char> data) {
    assert(sync_call_ctx.has_value() ^ (act != nullptr)); // can be only one of action and sync call
 
    dlog("receiver: ${r}, flags: ${f}, data size: ${s}",
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -247,7 +247,7 @@ void apply_context::exec()
 
 } /// exec()
 
-uint32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<const char> data) {
+int32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::span<const char> data) {
    assert(sync_call_ctx.has_value() ^ (act != nullptr)); // can be only one of action and sync call
 
    dlog("receiver: ${r}, flags: ${f}, data size: ${s}",
@@ -283,7 +283,11 @@ uint32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::sp
          try {
             // use a new apply_context for a new sync call
             apply_context a_ctx(control, trx_context, get_sync_call_sender(), receiver, flags, std::move(data));
-            control.get_wasm_interface().do_sync_call(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, a_ctx);
+
+            auto rc = control.get_wasm_interface().do_sync_call(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, a_ctx);
+            if (rc == sync_call_return_code::receiver_not_support_sync_call) {  // Currently -1 means there is no valid sync call entry point
+               return -1;
+            }
 
             // store return value
             if (a_ctx.sync_call_ctx.has_value()) {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -259,8 +259,8 @@ uint32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::sp
    const auto max_sync_call_data_size = control.get_global_properties().configuration.max_sync_call_data_size;
    EOS_ASSERT(data.size() <= max_sync_call_data_size, sync_call_call_data_exception,
               "sync call call data size must be less or equal to ${s} bytes", ("s", max_sync_call_data_size));
-   EOS_ASSERT((flags & ~0b11) == 0, sync_call_validate_exception,  // all but last 2 bits are 0s
-              "only two least significant bits of sync call's flags (${flags}) can be set", ("flags", flags)); // bit 0 for read-only, bit 1 for no-op if no sync call entry point exists
+   EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::last), sync_call_validate_exception,  // all but last `sync_call_flags::last` bits must be 0s
+              "only ${last} least significant bits of sync call's flags (${flags}) can be set", ("last", static_cast<uint64_t>(sync_call_flags::last))("flags", flags));
 
    auto handle_exception = [&](const auto& e)
    {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -259,8 +259,8 @@ int32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::spa
    const auto max_sync_call_data_size = control.get_global_properties().configuration.max_sync_call_data_size;
    EOS_ASSERT(data.size() <= max_sync_call_data_size, sync_call_call_data_exception,
               "sync call call data size must be less or equal to ${s} bytes", ("s", max_sync_call_data_size));
-   EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::last), sync_call_validate_exception,  // all but last `sync_call_flags::last` bits must be 0s
-              "only ${last} least significant bits of sync call's flags (${flags}) can be set", ("last", static_cast<uint64_t>(sync_call_flags::last))("flags", flags));
+   EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::last), sync_call_validate_exception,  // all but last `std::bit_width(sync_call_flags::last)` bits must be 0s
+              "only ${last} least significant bits of sync call's flags (${flags}) can be set", ("last", std::bit_width(static_cast<uint64_t>(sync_call_flags::last)))("flags", flags));
 
    auto handle_exception = [&](const auto& e)
    {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -60,11 +60,12 @@ apply_context::apply_context(controller& con, transaction_context& trx_ctx, uint
    context_free = trace.context_free;
 }
 
-apply_context::apply_context(controller& con, transaction_context& trx_ctx, name sender, name receiver, std::span<const char> data)
+// For sync calls
+apply_context::apply_context(controller& con, transaction_context& trx_ctx, name sender, name receiver, uint64_t flags, std::span<const char> data)
 :control(con)
 ,db(con.mutable_db())
 ,trx_context(trx_ctx)
-,sync_call_ctx({sender, receiver, data})
+,sync_call_ctx(sync_call_context(sender, receiver, flags, data))
 ,idx64(*this)
 ,idx128(*this)
 ,idx256(*this)
@@ -258,8 +259,8 @@ uint32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::sp
    const auto max_sync_call_data_size = control.get_global_properties().configuration.max_sync_call_data_size;
    EOS_ASSERT(data.size() <= max_sync_call_data_size, sync_call_call_data_exception,
               "sync call call data size must be less or equal to ${s} bytes", ("s", max_sync_call_data_size));
-   EOS_ASSERT((flags & 0xFFFFFFFFFFFFFFFE) == 0, sync_call_validate_exception,
-              "sync call's flags ${f} can only set bit 0", ("f", flags));
+   EOS_ASSERT((flags & ~0b11) == 0, sync_call_validate_exception,  // all but last 2 bits are 0s
+              "only two least significant bits of sync call's flags (${flags}) can be set", ("flags", flags)); // bit 0 for read-only, bit 1 for no-op if no sync call entry point exists
 
    auto handle_exception = [&](const auto& e)
    {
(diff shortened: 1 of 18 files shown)


libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -301,7 +322,17 @@ std::unique_ptr<wasm_instantiated_module_interface> eos_vm_runtime<Impl>::instan
...
       eos_vm_host_functions_t::resolve(bkend->get_module());
+
+      // check sync_call entry point
+      sync_call_supported = false;
+      uint32_t sync_call_idx = bkend->get_module().get_exported_function("sync_call");
+      if (sync_call_idx < std::numeric_limits<uint32_t>::max()) {
+         // signature has already been checked when the contract was deployed
+         sync_call_supported = true;
+      }
+
```
linh2931, line 335: Instead of caching `sync_call_supported`, I think I should move this code to `do_sync_call`. it is cleaner and much fewer files are touched.

linh2931: Much cleaner. https://github.com/AntelopeIO/spring/pull/1266/commits/e94850e022228c7211133b0eb9cbed7e92f7c1a3
https://github.com/AntelopeIO/spring/pull/1266/commits/ebce5e2913b96b209cab732992fc31125c132b08

libraries/chain/sync_call_context.cpp
```diff
@@ -0,0 +1,21 @@
+#include <eosio/chain/sync_call_context.hpp>
+
+namespace eosio::chain {
+
+sync_call_context::sync_call_context(account_name sender, account_name receiver, uint64_t flags, std::span<const char>data)
+   : sender(sender)
+   , receiver(receiver)
+   , flags(flags)
+   , data(std::move(data))
```
heifner, line 9: I think move of a span might be a bit misleading. span doesn't own the data. Probably better to remove the `std::move`.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -105,6 +105,17 @@ void validate( const bytes& code, const wasm_config& cfg, const whitelisted_intr
...
       EOS_ASSERT((apply_type == vm::host_function{{vm::i64, vm::i64, vm::i64}, {}}), wasm_serialization_error, "apply has wrong type");
+
+      // Check sync-call entry point after the protocol feature is activated.
+      // It is optional for a contract to provide sync calls.
+      // To prevent problems, if sync call entry point is present, its signature must be correct
+      if (control.is_builtin_activated(builtin_protocol_feature_t::sync_call)) {
+         uint32_t sync_call_idx = bkend.get_module().get_exported_function("sync_call");
+         if (sync_call_idx < std::numeric_limits<uint32_t>::max()) {
+            const vm::func_type& sync_call_type = bkend.get_module().get_function_type(sync_call_idx);
+            EOS_ASSERT((sync_call_type == vm::host_function{{vm::i64, vm::i64, vm::i32}, {}}), wasm_serialization_error, "sync call entry point has wrong type");
```
heifner, line 116: This is checking signature right? Maybe "sync call entry point has wrong function signature"

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -131,17 +142,29 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
 
       void do_sync_call(apply_context& context) override {
+         const std::optional<sync_call_context>& sync_call_ctx  = context.get_sync_call_ctx();
+         assert(sync_call_ctx.has_value());
+         assert(!context.get_action_ptr());
+
+         uint32_t sync_call_idx = _instantiated_module->get_module().get_exported_function("sync_call");
+         if (sync_call_idx == std::numeric_limits<uint32_t>::max()) {
+            if (sync_call_ctx->no_op_if_receiver_not_support_sync_call()) {
+               dlog("receiver does not have sync call entry point or the entry point is invalid, and the no-op flag is set. just returns");
```
heifner, line 152: Probably do not want to leave in this debug log.

heifner (changes requested):

libraries/chain/sync_call_context.cpp
linh2931: Thanks. https://github.com/AntelopeIO/spring/pull/1266/commits/bbe2ffab05239a2f3b4228e8dfc88f8c5b0c88e0

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1266/commits/d84de41da5ca3bb78fd17d35b95f6d608885c1e7

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1266/commits/d76df03a9d578285f79a4cc893dd4a615fdf911c

heifner (approved):

libraries/chain/sync_call_context.cpp
```diff
@@ -0,0 +1,21 @@
...
+   , data(data)
+{
+}
+
+bool sync_call_context::is_read_only()const {
+   return flags & 0b1; // LSB. bit index 0
+}
+
+bool sync_call_context::no_op_if_receiver_not_support_sync_call()const {
+   return flags & 0b10;  // second bit from LSB. bit index 1
```
spoonincode, line 18: Is there some pattern we can use to get rid of the naked bit constants here and
https://github.com/AntelopeIO/spring/blob/cdb1bebefacb07b4b7bed9d9d9547c4d519928e0/libraries/chain/apply_context.cpp#L262

Simplest being something like
```c++
enum sync_call_flags {
   read_only = 1<<0;
   noop.. = 1<<1;
   last = noop..;
}
```
then flags are valid if `flags <= last`.

Lots of other potential approaches too.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -131,17 +142,28 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
+         uint32_t sync_call_idx = _instantiated_module->get_module().get_exported_function("sync_call");
+         if (sync_call_idx == std::numeric_limits<uint32_t>::max()) {
+            if (sync_call_ctx->no_op_if_receiver_not_support_sync_call()) {
+               return;
+            } else {
+               EOS_ASSERT(false, sync_call_not_supported_by_receiver_exception, "receiver does not support sync calls");
+            }
+         }
+         // If the contract contains sync_call entry point, its signature had already been
+         // validated when the contract was deployed.
```
spoonincode, line 158: I don't think this comment is entirely correct -- its signature will be validated provided that the contract was deployed after the activation of `SYNC_CALL`. If someone deploys a contract that exports `sync_call` with an unexpected signature prior to activation of `SYNC_CALL` we will have never checked its signature ahead of time.

In this case here it seems okayish (not UB or such) because EOS VM does validation on each `call()` that the parameters and parameter count match up with the export. But it does mean if code is deployed with bad signature prior to `SYNC_CALL` and the call is made with `no_op_if_receiver_not_support_sync_call` you'll _not_ get a noop but rather some other eosvm assert. In practice we would never expect this situation, but it is quirky behavior enshrined in the protocol we would need to remember and replicate in OC too. I feel like we should tighten this up so the rule is simpler to describe.

i.e. the rule should be something like "when a sync call is made to a contract that does not export an expected `sync_call`, it aborts with `sync_call_not_supported_by_receiver_exception` unless the sync call flag `no_op_if_receiver_not_support_sync_call` is set which makes it a noop"

I thought last time this topic was discussed we were going to cache away a bool somewhere that indicates if `sync_call` is properly exported. Ideally we do that in a common location so we are not repeating this check in each wasm runtime. Unfortunately I'm not sure exactly where that should go -- I initially thought `wasm_cache_entry` was the right place but seems not since OC can skip it.

Personally I don't think we should bother adding additional checks to `setcode` because even if we add a check there we still have have to be prepared for a mismatched `sync_call` anyways -- the check doesn't really help simplify anything in nodeos.[^1] Not sure how strongly I feel about it in this case.

[^1]: There is an argument that the check in `setcode` makes it more upfront to a user they got something wrong. There is also an argument the other way that by making the validation more strict we potentially  prevent valid code that is deployed today from being redeployed -- though of course that's very academic for this scenario.

libraries/chain/sync_call_context.cpp
linh2931: Thanks!

https://github.com/AntelopeIO/spring/pull/1266/commits/287e1c585382223ef97edf44ea5d829be8c13c42

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: I like your simplified rule -- not checking the signature.

But we will miss one feature "someone wants a no-op if the signature is bad". But I think it is very rare and we don't need to support it, at least for now.

libraries/chain/sync_call_context.cpp
heifner: If we are getting rid of the naked bit constants, I would recommend you do it like this:
https://github.com/AntelopeIO/spring/blob/e186a90a6215405eee3c85a24e9e4358560c0eee/libraries/chain/include/eosio/chain/account_object.hpp#L67-L72

libraries/chain/webassembly/runtimes/eos-vm.cpp
spoonincode: "expected `sync_call`" was meant to imply proper signature.

So the rule should be something like
> When a sync call is made to a contract that does not export `sync_call` with a proper signature, it aborts with `sync_call_not_supported_by_receiver_exception` unless the sync call flag `no_op_if_receiver_not_support_sync_call` is set which makes it a noop

The code does _not_ follow this simple rule right now. The behavior is more complex -- behavior is dependent on when the code was deployed.

libraries/chain/sync_call_context.cpp
spoonincode: No preference from me; using `has_field()` seems, if I'm understanding right, to offer some minor additional paranoia that is good.

My main disappointment is that there is no automatic way to get the last item, thus requiring us to manually specify it. My hunch is it's possible to construct something fancy with constexpr to resolve that but not something I'm advocating we do now.

linh2931: I will change to use `has_field()` so it is more consistent with existing code style.

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: Per team discussion, if the receiver does not have valid entry point, return -1 and let the caller decide what to do about the failure.

https://github.com/AntelopeIO/spring/pull/1266/commits/4aead0a461fe61c95b6ac3dc6fed4b30f46b84d4

libraries/chain/apply_context.cpp
```diff
@@ -258,8 +259,8 @@ uint32_t apply_context::execute_sync_call(name receiver, uint64_t flags, std::sp
    const auto max_sync_call_data_size = control.get_global_properties().configuration.max_sync_call_data_size;
    EOS_ASSERT(data.size() <= max_sync_call_data_size, sync_call_call_data_exception,
               "sync call call data size must be less or equal to ${s} bytes", ("s", max_sync_call_data_size));
-   EOS_ASSERT((flags & 0xFFFFFFFFFFFFFFFE) == 0, sync_call_validate_exception,
-              "sync call's flags ${f} can only set bit 0", ("f", flags));
+   EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::last), sync_call_validate_exception,  // all but last `sync_call_flags::last` bits must be 0s
+              "only ${last} least significant bits of sync call's flags (${flags}) can be set", ("last", static_cast<uint64_t>(sync_call_flags::last))("flags", flags));
```
spoonincode, line 263: Using `sync_call_flags::last` here doesn't seem right. If last is 1<<3 then this message will say "only 8 least significant bits of sync call's flags (${flags}) can be set". That value should be 4 instead of 8. I think you want the `std::bit_width(sync_call_flags::last)`

libraries/chain/sync_call_context.cpp
linh2931: Looked at this further. We need to verify bits other than the first bit can have a value. The `has_field` does not provide a way to do that. Looks like the existing implementation is a simple one.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -694,10 +694,10 @@ namespace webassembly {
           *                the sync call. LSB bit indicates read-only. All other bits
           *                are reserved to be 0.
           * @param data - the data of the sync call, which may include function name, arguments and other information.
-          * @return the number of bytes of the return value of the call (to be used by next get_call_return_value) .
+          * @return the number of bytes of the return value of the call (to be used by next get_call_return_value) if succeeded, return -1 if sync_call entry point does not exist or has invalid signature .
           *
          */
-         uint32_t call(name receiver, uint64_t flags, span<const char> data);
+         int32_t call(name receiver, uint64_t flags, span<const char> data);
```
spoonincode, line 700: I know this is academic but it's the cost of `CONFIGURABLE_WASM_LIMITS2`'s infinite knobs: `int32_t` isn't enough for greater than 2GB unless the contract treats this as unsigned. Unless there is a compelling reason to keep it 32-bit I'd suggest moving to `int64_t` so that the semantics are clearer for these academic edge cases.

libraries/chain/apply_context.cpp
linh2931: Thanks for noticing this!

https://github.com/AntelopeIO/spring/pull/1266/commits/64eb9285134a2134ba9cf4efe569a55c48185fb0

libraries/chain/include/eosio/chain/webassembly/interface.hpp
linh2931: The value returned by `call()` (the number of sync call return value) is used as the size component for constructing `memory` argument in `get_call_return_value(span<char> memory)`. In WASM, the memory span is represented as  (param i32 i32). Is it safe to cast down from i64 to i32 when the user uses it?

libraries/chain/include/eosio/chain/webassembly/runtime_interface.hpp
```diff
@@ -13,15 +13,15 @@ class apply_context;
...
+      virtual sync_call_return_code do_sync_call(apply_context& context) = 0;
 
       virtual ~wasm_instantiated_module_interface();
 };
 
 class wasm_runtime_interface {
    public:
       virtual std::unique_ptr<wasm_instantiated_module_interface> instantiate_module(const char* code_bytes, size_t code_size,
-                                                                                     const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version) = 0;
+                                                                                     const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version, bool& sync_call_supporte) = 0;
```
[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/1266/commits/268deaa871d2e8d3cfd915a1d0c1b25b3558b2fa

libraries/chain/apply_context.cpp
greg7mdp: I think the check `flags <= static_cast<uint64_t>(sync_call_flags::last) is wrong. For example if two bits out of two were set, `flags` would be `3` and `static_cast<uint64_t>(sync_call_flags::last)` would be 2.

linh2931: Thanks. I renamed `last` and added comments to prevent future confusions. Also added flags validation tests.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
linh2931: Another point: the return value size is limited by `max_sync_call_data_size`, which is likely less than `int32_t` size.

linh2931: @arhag Any comment about this?

arhag: I don't really have a position on this one.

linh2931: To prevent any potential edge cases, I changed to return type to `int64_t`.

Thanks.

https://github.com/AntelopeIO/spring/pull/1266/commits/f6605d3abd44969205395e1d175f1ccbe552f945

spoonincode (approved):
We are going to need to do some non trivial refactoring when wiring up OC but we can keep this as is for now

linh2931:
> We are going to need to do some non trivial refactoring when wiring up OC but we can keep this as is for now

I copied the link to this comment to the OC issue https://github.com/AntelopeIO/spring/issues/1043

Issue #1247, opened 2025-03-10
Test failure: plugin_http_api_test


spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/13681617299/job/38255916119

I am having a little bit of trouble discerning if this is #992 resurfaced or a new problem

greg7mdp:
https://github.com/AntelopeIO/spring/actions/runs/13796200464/job/38589020620

spoonincode:
https://github.com/AntelopeIO/spring/actions/runs/13925820804/job/38970941917

heifner:
https://github.com/AntelopeIO/spring/actions/runs/13999157091/job/39202003597

linh2931:
https://github.com/AntelopeIO/spring/actions/runs/14205458471/job/39802684595#step:4:479

heifner:
A current one that should have the extra output of #1270.
https://github.com/AntelopeIO/spring/actions/runs/14339780836/job/40196701196

PR #1270, opened 2025-03-24
Fix incorrect names (`preActivate` instead of `activate`) and add debugging info

greg7mdp:
This is work on issue #1247.
I was not able to figure out why it seems that sometimes, we are missing one activated protocol feature.
This PR adds some debugging code to print out the missing feature when the issue occurs.

> note: Issue may have been present for a long time, but was hidden as a failure of the test to return an exit value caused it to always succeed (fixed in https://github.com/AntelopeIO/spring/pull/1195).

### DIFF
--- a/tests/TestHarness/Cluster.py
+++ b/tests/TestHarness/Cluster.py
@@ -1111,7 +1111,7 @@ class Cluster(object):
             return None
 
         if pfSetupPolicy == PFSetupPolicy.FULL:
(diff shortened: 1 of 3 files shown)


heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
category: Tests
component: Internal
summary: Logs protocol activations during tests.
Note:end

Issue #1261, opened 2025-03-19
SC: Refactor apply_context and sync_call_context classes


linh2931:
apply_context and sync_call_context can be made out from a common base class. Care should be taken in modifying the host function interface class, which currently uses apply_context as the gateway to the host.

PR #1273, opened 2025-03-25
SC: Introduce host_context class as the interface to the host and as the base class for apply_context and sync_call_context

linh2931:
Introduces `host_context` as the interface to the host, and makes `apply_context` and `sync_call_context` as the subclasses of `host_context`.  `host_context` implements the common functionalities, which `apply_context` and `sync_call_context` implement functionalities particular to action execution and sync call execution respectively.

This makes code cleaner, more natural to add new functionalities, simpler to identify which host functions are applicable to actions or sync calls.

Resolves https://github.com/AntelopeIO/spring/issues/1261

Please note, enforcement of host function constraints is to be implemented by https://github.com/AntelopeIO/spring/issues/1218 next.

### DIFF
--- a/libraries/chain/include/eosio/chain/apply_context.hpp
+++ b/libraries/chain/include/eosio/chain/apply_context.hpp
@@ -87,7 +87,6 @@ class apply_context : public host_context {
 
       bool is_context_free()const override { return context_free; }
       bool is_privileged()const override { return privileged; }
-      action_name get_receiver()const override { return receiver; }
       const action& get_action()const override { return *act; }
       const action* get_action_ptr()const override { return act; }
 
--- a/libraries/chain/include/eosio/chain/apply_context.hpp
+++ b/libraries/chain/include/eosio/chain/apply_context.hpp
@@ -503,6 +503,7 @@ class apply_context {
       void exec_one();
       void exec();
       uint32_t execute_sync_call(name receiver, uint64_t flags, std::span<const char> data);
+      uint32_t get_call_return_value(std::span<char> memory) const;
       uint32_t get_call_data(std::span<char> memory) const;
       void set_call_return_value(std::span<const char> return_value);
       void execute_inline( action&& a );
--- a/libraries/chain/include/eosio/chain/apply_context.hpp
+++ b/libraries/chain/include/eosio/chain/apply_context.hpp
@@ -17,11 +17,11 @@ class apply_context : public host_context {
    public:
 
       void exec_one();
-      void exec() override;
+      void exec();
       void execute_inline( action&& a ) override;
       void execute_context_free_inline( action&& a ) override;
       void schedule_deferred_transaction( const uint128_t& sender_id, account_name payer, transaction&& trx, bool replace_existing ) override;
-      bool cancel_deferred_transaction( const uint128_t& sender_id, account_name sender ) override;
+      bool cancel_deferred_transaction( const uint128_t& sender_id, account_name sender );
       bool cancel_deferred_transaction( const uint128_t& sender_id ) override { return cancel_deferred_transaction(sender_id, receiver); }
 
       // Not callable from apply_context (actions)
@@ -88,7 +88,7 @@ class apply_context : public host_context {
       bool is_context_free()const override { return context_free; }
       bool is_privileged()const override { return privileged; }
       const action& get_action()const override { return *act; }
-      const action* get_action_ptr()const override { return act; }
+      const action* get_action_ptr()const { return act; }
 
       action_name get_sender() const override;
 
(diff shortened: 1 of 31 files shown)


greg7mdp:
To avoid these massive changes, maybe you could have renamed the files first:
apply_context.cpp -> host_context.cpp
apply_context.hpp -> host_context.hpp
...

And then do the changes since most of the `host_context` files would be unchanged.

linh2931:
> To avoid these massive changes, maybe you could have renamed the files first:
> 
> ```
> apply_context.cpp -> host_context.cpp
> apply_context.hpp -> host_context.hpp
> ...
> ```
> 
> And then do the changes since most of the `host_context.*` files would be unchanged.

This was essentially the way I did, but ended up making big changes once I started to consolidate the two subclasses.

linh2931:
> (not for this PR) makes me wonder if there should be a separate `readonly_context`

Would the `readonly_context` use a separate entry point?

spoonincode:
> > (not for this PR) makes me wonder if there should be a separate `readonly_context`
> 
> Would the `readonly_context` use a separate entry point?

Thinking on it more I think this is what I'm trying to say https://github.com/AntelopeIO/spring/pull/1277#issuecomment-2755459907

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -0,0 +1,601 @@
...
+
+   /// Fields:
+public:
+
+   controller&                      control;
+   chainbase::database&             db;  ///< database where state is stored
+   transaction_context&             trx_context; ///< transaction context in which the action is running
+   account_name                     receiver; ///< the code that is currently running
+   std::vector<char>                action_return_value;
+   std::optional<std::vector<char>> last_sync_call_return_value{}; /// return value of last sync call initiated by the current code (host context)
```
spoonincode, line 587: I should have thought about this in a prior PR, but afaict there doesn't seem to be any need to make this an optional?

linh2931: It does not need to be optional as we only return the size of the return value. 0 can be for both 0-length return value or return value not existing. Will make the change after I resolve the large number of files of merge conflicts.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -197,7 +198,7 @@ struct eosvmoc_tier {
          }
       }
 
-      sync_call_return_code do_sync_call( const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version, apply_context& context ) {
+      sync_call_return_code do_sync_call( const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version, sync_call_context& context ) {
          return get_instantiated_module(code_hash, vm_type, vm_version, context)->do_sync_call(context);
```
heifner, line 202: Am I missing something? Do sync calls not eosvmoc tierup?

spoonincode: Ultimately they will but nothing OC is plumbed through yet. Some stuff (outside OC) that we're ignoring right now is going to need to be refactored for that.

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -599,54 +79,25 @@ class apply_context {
...
       const action* get_action_ptr()const { return act; }
-      const std::optional<sync_call_context>& get_sync_call_ctx()const { return sync_call_ctx; }
-      std::optional<sync_call_context>& get_mutable_sync_call_ctx() { return sync_call_ctx; }
 
-      action_name get_sender() const;
+      action_name get_sender() const override;
 
       bool is_applying_block() const { return trx_context.explicit_billed_cpu_time; }
       bool is_eos_vm_oc_whitelisted() const;
       bool should_use_eos_vm_oc()const;
```
heifner, line 90: These will need to be moved up to `host_context`. I assume you will do that when you do the eosvmoc work.

linh2931: Yes, they will be moved up when we do the OC work (after EOSVM JIT and Interpreter).

libraries/chain/include/eosio/chain/host_context.hpp
linh2931: Done. https://github.com/AntelopeIO/spring/pull/1273/commits/6a079409221c7290f6c4dee8f92e9fff4c7cb0f5

linh2931:
https://github.com/AntelopeIO/spring/issues/1324 will be worked after this PR is merged.

spoonincode (approved):

PR #1277, opened 2025-03-26
SC: Enforce host function constraints in sync calls

linh2931:
The following host functions
1. `require_auth`,
2.  `require_auth2`,
3.  `require_recipient`,
4.  `read_action_data`,
5.  `action_data_size`,
6.  `set_action_return_value`,
7.  `get_context_free_data`,
8.  `send_inline`,
9.  `send_context_free_inline`,
10.  `send_deferred`,
11.  `cancel_deferred`
 
are not applicable for sync calls. Abort  the sync calls if they are used.

For each of the host functions, a test is added to verify the sync call is aborted.

Resolves https://github.com/AntelopeIO/spring/issues/1218

### DIFF
--- a/libraries/chain/sync_call_context.cpp
+++ b/libraries/chain/sync_call_context.cpp
@@ -53,15 +53,17 @@ action_name sync_call_context::get_sender() const {
    return receiver;
 }
 
-// EOS_ASSERTs and tests will be added for the following methods in next PR
 void sync_call_context::require_authorization(const account_name& account) {
+   EOS_ASSERT(false, sync_call_validate_exception, "require_auth is only allowed in actions");
 }
 bool sync_call_context::has_authorization(const account_name& account) const {
    return false;
 }
 void sync_call_context::require_authorization(const account_name& account, const permission_name& permission) {
+   EOS_ASSERT(false, sync_call_validate_exception, "require_auth2 is only allowed in actions");
 }
 void sync_call_context::require_recipient(account_name account) {
+   EOS_ASSERT(false, sync_call_validate_exception, "require_recipient is only allowed in actions");
 }
 bool sync_call_context::has_recipient(account_name account)const {
    return false;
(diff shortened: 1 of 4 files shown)


spoonincode:
Not new to this PR though it does it more, it is unfortunate that we have preconditions for whether a host call is valid in a particular "context" scattered about multiple locations,
1) sometimes in the host function wrapper (context free & privileged)
2) sometimes in the host function implementations themselves
3) sometimes in the apply_context/host_context/sync_call_context (could count this as more than 1 location)

I think all these checks should be in the host function wrappers, which already have a very flexible system for adding new preconditions. It was probably an oversight we didn't make use of those when adding ROtrx. What I'm not sure about is how/where to tackle this. We could potentially go ahead and refactor `main`'s RO preconditions to use the host function wrappers and then merge that through sync_call branches and follow suit.

linh2931:
> Not new to this PR though it does it more, it is unfortunate that we have preconditions for whether a host call is valid in a particular "context" scattered about multiple locations,
> 
> 1. sometimes in the host function wrapper (context free & privileged)
> 2. sometimes in the host function implementations themselves
> 3. sometimes in the apply_context/host_context/sync_call_context (could count this as more than 1 location)
> 
> I think all these checks should be in the host function wrappers, which already have a very flexible system for adding new preconditions. It was probably an oversight we didn't make use of those when adding ROtrx. What I'm not sure about is how/where to tackle this. We could potentially go ahead and refactor `main`'s RO preconditions to use the host function wrappers and then merge that through sync_call branches and follow suit.

Thanks for this suggestion! I will change to use host function wrappers for sync calls. Can we do ROtrx refactor later?

linh2931:
Close this to start a fresh new PR using host function wrappers.

spoonincode:
I was worried that if we didn't do the RO refactor first it'd either cause some merge conflict grief or we'd just need to do it here in the sync call branch (which means waiting on it until 2.0 instead of now but I suppose that's no real big deal). Actually even #1273 probably already will cause conflicts.

linh2931:
> Quote reply

Thanks. I will do this as one of the last tasks.

I plan to get he existing PRs in and do a merge from `main` to resolve any early conflicts. `apply_context.*` can be a pain to merge.

PR #1283, opened 2025-03-27
Minor cleanup in `buffer_factory.hpp`

greg7mdp:
- Use `send_buffer_type` consistently. 
- Avoid some `shared_ptr` copies.

### DIFF
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -205,12 +205,12 @@ public:
       return conn->bp_connection == Connection::bp_connection_type::non_bp && conn->socket_is_open() && conn->incoming_and_handshake_received();
    }
 
-   send_buffer_type get_gossip_bp_initial_send_buffer() {
+   const send_buffer_type& get_gossip_bp_initial_send_buffer() {
       fc::lock_guard g(mtx);
       return initial_gossip_msg_factory.get_initial_send_buffer();
    }
 
-   send_buffer_type get_gossip_bp_send_buffer(gossip_buffer_factory& factory) {
+   const send_buffer_type& get_gossip_bp_send_buffer(gossip_buffer_factory& factory) {
       return factory.get_send_buffer(gossip_bps);
    }
 
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -205,7 +205,7 @@ public:
       return conn->bp_connection == Connection::bp_connection_type::non_bp && conn->socket_is_open() && conn->incoming_and_handshake_received();
    }
 
-   const send_buffer_type& get_gossip_bp_initial_send_buffer() {
+   send_buffer_type get_gossip_bp_initial_send_buffer() {
       fc::lock_guard g(mtx);
       return initial_gossip_msg_factory.get_initial_send_buffer();
    }
--- a/plugins/net_plugin/include/eosio/net_plugin/buffer_factory.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/buffer_factory.hpp
@@ -10,7 +10,7 @@
 
 namespace eosio {
 
-   using send_buffer_type = std::shared_ptr<std::vector<char>>;
+   using send_buffer_type = std::unique_ptr<std::vector<char>>;
 
    struct buffer_factory {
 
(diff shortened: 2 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3154,7 +3154,7 @@ namespace eosio {
       peer_dlog(this, "Sending nack ${n}", ("n", block_header::num_from_id(block_id)));
 
       buffer_factory buff_factory;
-      auto send_buffer = buff_factory.get_send_buffer( block_nack_message{block_id} );
+      auto& send_buffer = buff_factory.get_send_buffer( block_nack_message{block_id} );
```
heifner, line 3157: `const auto&`

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -205,12 +205,12 @@ class bp_connection_manager {
       return conn->bp_connection == Connection::bp_connection_type::non_bp && conn->socket_is_open() && conn->incoming_and_handshake_received();
```
heifner, line 205: Need to revert bn256 submodule change.

heifner:
> * Use `unique_ptr` in buffer factory.

Update PR description.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
greg7mdp: Yes, I checked this in by mistake.
But actually, could you catchup with main in your branch? That would also resolve it.

heifner: Updated my branch

plugins/net_plugin/net_plugin.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1283/commits/3f12cb21350395d4829501551f37f0453c00f4d6

greg7mdp:
@heifner should we merge this into your branch?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -205,12 +205,12 @@ class bp_connection_manager {
...
    }
 
-   send_buffer_type get_gossip_bp_initial_send_buffer() {
+   const send_buffer_type& get_gossip_bp_initial_send_buffer() {
```
heifner, line 208: For thread-safety this needs to remain `send_buffer_type`.

greg7mdp: For my enlightenment, please explain why?

heifner: `bp_connection_manager`

thread A:
calls `get_gossip_bp_initial_send_buffer()`
acquires mutex mtx
assigns reference `send_buffer` to caller
releases mutex

thread B:
calls `update_bp_producer_peers()`
acquires mutex mtx
calls `initial_gossip_msg_factory.set_initial_send_buffer(signed_empty);`
starts to update `send_buffer`

thread A:
starts to use `send_buffer`

greg7mdp: Thanks. Missed that.
Still not 100% correct to read a value in one thread (`send_buffer`) which can be modified in another thread. 
maybe `send_buffer` should be an `atomic<shared_ptr>` in `gossip_buffer_initial_factory`

greg7mdp: > Still not 100% correct to read a value in one thread (`send_buffer`) which can be modified in another thread. maybe `send_buffer` should be an `atomic<shared_ptr>` in `gossip_buffer_initial_factory`

Oh wait I see we hold the mtx, so we are fine indeed.

heifner: That shouldn't be possible. A copy of the shared_ptr is made before releasing the mutex. The original shared_ptr is updated with new data, but that doesn't affect the copy.

`std::shared_ptr` - "All member functions (including copy constructor and copy assignment) can be called by multiple threads on different shared_ptr objects without additional synchronization even if these objects are copies and share ownership of the same object."

plugins/net_plugin/include/eosio/net_plugin/buffer_factory.hpp
```diff
@@ -185,14 +185,14 @@ namespace eosio {
       }
 
       /// requires set_initial_send_buffer to be called first
-      const send_buffer_type& get_initial_send_buffer() {
+      send_buffer_type get_initial_send_buffer() {
```
heifner, line 188: For consistency, I think also change: `get_gossip_bp_initial_send_buffer`

heifner (approved):

PR #1284, opened 2025-03-27
[RFC] hide `max_action_return_value_size` in `get_consensus_parameters` when feature not activated

spoonincode:
For a little background see AntelopeIO/leap#1660 and AntelopeIO/leap#1770. We decided to hide the `wasm_configuration` in `get_consensus_parameters` when `CONFIGURABLE_WASM_LIMITS2` is not activated. This helps resolve some confusion.

But I realized in https://github.com/AntelopeIO/spring/pull/1257#issuecomment-2730533968 that this introduces an inconsistency: we still show `max_action_return_value_size` even if `ACTION_RETURN_VALUE` is not activated.

So this change will effectively hide `max_action_return_value_size` until `ACTION_RETURN_VALUE` is activated. This turned out really quite simple to do so makes it tempting to resolve the inconsistency this way.

If OK, will explore what a test would look like

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -2678,7 +2678,11 @@ read_only::get_consensus_parameters_results
 read_only::get_consensus_parameters(const get_consensus_parameters_params&, const fc::time_point& ) const {
    get_consensus_parameters_results results;
 
-   results.chain_config = db.get_global_properties().configuration;
+   if (db.is_builtin_activated(builtin_protocol_feature_t::action_return_value))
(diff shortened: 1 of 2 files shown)


spoonincode:
> If OK, will explore what a test would look like

ah okay there is a test already; so if OK I will explore fixing and improving the test

linh2931 (approved):
This looks great and simple! I can follow the same pattern for #1257 after you merge into the `main`.

ericpassmore:
Note:start
category: Other
component: Internal
summary: Only show action return information when `ACTION_RETURN` is activated.
Note:end

Issue #1268, opened 2025-03-21
SC: Enforce `max_sync_call_depth`


linh2931:
This should be done after https://github.com/AntelopeIO/spring/issues/1261

PR #1285, opened 2025-03-27
SC: enforce `max_sync_call_depth` for nested and recursive sync calls

linh2931:
Enforce on chain configuration parameter `max_sync_call_depth` for nested and recursive sync calls. Once the depth is reached, the call is aborted.

Notable things about the two new two tests for the PR:

1. One test has one function making sync calls to itself recursively, another has two functions on different accounts calling each other recursively. Both tests show sync calls work correctly when max depth is not reached, and calls are aborted when max depth is reached.

2. The test WASM contracts utilize `sender` and `receiver` arguments from `sync_call` entry point, which demonstrates those two arguments are set correctly.

Resolves https://github.com/AntelopeIO/spring/issues/1268

### DIFF
--- a/libraries/chain/host_context.cpp
+++ b/libraries/chain/host_context.cpp
@@ -38,10 +38,7 @@ host_context::~host_context() = default;
 
 // called from apply_context or sync_call_context
 uint32_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std::span<const char> data) {
-   // If the call is initiated from an apply_context (an action), it is the first
-   // level of a sync call by the action, set the depth to 1.
-   // Otherwise the call is from another sync call, increment the depth by 1.
-   const uint32_t depth = dynamic_cast<apply_context*>(this) ? 1 : sync_call_depth + 1;
+   const uint32_t depth = sync_call_depth + 1;
    const auto max_depth = control.get_global_properties().configuration.max_sync_call_depth;
    EOS_ASSERT(depth <= max_depth, sync_call_depth_exception,
               "reached sync call max call depth ${max_depth}", ("max_depth", max_depth));
--- a/libraries/chain/host_context.cpp
+++ b/libraries/chain/host_context.cpp
@@ -6,11 +6,11 @@
 
 namespace eosio::chain {
 
-host_context::host_context(controller& con, transaction_context& trx_ctx, uint32_t depth)
+// used to create an apply context
+host_context::host_context(controller& con, transaction_context& trx_ctx)
    : control(con)
    , db(con.mutable_db())
    , trx_context(trx_ctx)
-   , recurse_depth(depth)
    , idx64(*this)
    , idx128(*this)
    , idx256(*this)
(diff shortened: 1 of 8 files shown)


libraries/chain/host_context.cpp
```diff
@@ -62,8 +71,13 @@ uint32_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std
          }
 
          try {
+            // If the call is initiated from an apply_context (an action), it is the first
+            // level of a sync call by the action, set septh to 1.
+            // Otherwise the call is from another sync call, increment the call depth by 1.
+            const uint32_t depth = dynamic_cast<apply_context*>(this) ? 1 : recurse_depth + 1;
```
heifner, line 77: This seems rather ick to me. Can you not pass in the recurse_depth instead.

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -582,6 +583,7 @@ class host_context {
    controller&                      control;
    chainbase::database&             db;  ///< database where state is stored
    transaction_context&             trx_context; ///< transaction context in which the action is running
+   uint32_t                         recurse_depth; ///< how deep inline actions can recurse
```
heifner, line 586: I thought sync calls can't make inline action calls. Why is this here?

heifner (changes requested):

libraries/chain/host_context.cpp
linh2931: Thanks for this good catch! I was trying to consolidate the depth field but forgot about recurse_depth could be already set by inline actions.

Fixed by https://github.com/AntelopeIO/spring/pull/1285/commits/7d16b88cf172aa95a78344e0d3c17f9aea463426

libraries/chain/include/eosio/chain/host_context.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1285/commits/7d16b88cf172aa95a78344e0d3c17f9aea463426

libraries/chain/host_context.cpp
```diff
@@ -34,12 +38,22 @@ host_context::~host_context() = default;
 
 // called from apply_context or sync_call_context
 uint32_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std::span<const char> data) {
-   auto* code = control.db().find<account_object, by_name>(call_receiver);
+   // If the call is initiated from an apply_context (an action), it is the first
+   // level of a sync call by the action, set the depth to 1.
+   // Otherwise the call is from another sync call, increment the depth by 1.
+   const uint32_t depth = dynamic_cast<apply_context*>(this) ? 1 : sync_call_depth + 1;
```
heifner, line 44: I think we should refactor to remove this `dynamic_cast`. Can't you just pass in the depth instead?

heifner (changes requested):

libraries/chain/host_context.cpp
linh2931: Yes, it can be removed because `sync_call_depth` default initial value is `0`. Now the code is cleaner. Thanks.

https://github.com/AntelopeIO/spring/pull/1285/commits/c37b00d14b30a0a79fbefc9d96e5a5b8d2feff11

heifner (approved):

spoonincode (approved):

PR #1296, opened 2025-03-31
SC: Enforce host function preconditions for sync calls

linh2931:
The following host functions are not allowed to be called from sync calls:
1. `require_auth()`,
2.  `require_auth2()`,
3. `has_auth()`,
4.  `require_recipient()`,
5.  `read_action_data()`,
6.  `action_data_size()`,
7.  `set_action_return_value()`,
8.  `get_context_free_data()`,
9.  `send_inline()`,
10.  `send_context_free_inline()`,
11.  `send_deferred()`,
12.  `cancel_deferred()`

The following host functions are not allowed to be called from actions:
1. `get_call_data()`,
2. `set_call_return_value()`

This PR enforces those preconditions via host function definition wrappers, which is cleaner than the checks are scattered in the function implementation. Thanks @spoonincode for the suggestion!

For each of the host functions, a test is added to verify the precondition is checked.

Resolves https://github.com/AntelopeIO/spring/issues/1218

### DIFF
--- a/libraries/chain/include/eosio/chain/host_context.hpp
+++ b/libraries/chain/include/eosio/chain/host_context.hpp
@@ -500,7 +500,7 @@ public:
     * @throws missing_auth_exception If no sufficient permission was found
     */
    virtual void require_authorization(const account_name& account) { EOS_ASSERT(false, unaccessible_api, "default implemention should never be used"); } // This function should never be called in sync calls due to require_auth host wrapper preconditions. This EOS_ASSERT double makes sure that.
-   virtual bool has_authorization(const account_name& account) const = 0;
+   virtual bool has_authorization(const account_name& account) const { assert(false); __builtin_unreachable(); }
    virtual void require_authorization(const account_name& account, const permission_name& permission) { EOS_ASSERT(false, unaccessible_api, "default implemention should never be used"); }
 
    /**
--- a/libraries/chain/include/eosio/chain/host_context.hpp
+++ b/libraries/chain/include/eosio/chain/host_context.hpp
@@ -484,8 +484,8 @@ public:
 
 public:
    /// Constructor and destructor:
-   host_context(controller& con, transaction_context& trx_ctx, uint32_t depth);
-   host_context(controller& con, transaction_context& trx_ctx, uint32_t depth,  account_name receiver);
+   host_context(controller& con, transaction_context& trx_ctx);
+   host_context(controller& con, transaction_context& trx_ctx, account_name receiver, uint32_t sync_call_depth );
    virtual ~host_context();
 
    /// Authorization methods:
@@ -583,10 +583,11 @@ public:
    controller&                      control;
    chainbase::database&             db;  ///< database where state is stored
    transaction_context&             trx_context; ///< transaction context in which the action is running
-   uint32_t                         recurse_depth; ///< how deep inline actions can recurse
    account_name                     receiver; ///< the code that is currently running
    std::vector<char>                action_return_value;
-   std::optional<std::vector<char>> last_sync_call_return_value{}; /// return value of last sync call initiated by the current code (host context)
+
+   std::optional<std::vector<char>> last_sync_call_return_value{}; // return value of last sync call initiated by the current code (host context)
+   uint32_t                         sync_call_depth = 0; // depth for sync call
 
    generic_index<index64_object>                                  idx64;
    generic_index<index128_object>                                 idx128;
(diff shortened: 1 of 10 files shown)


libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -513,10 +513,10 @@ REGISTER_HOST_FUNCTION(get_permission_last_used);
 REGISTER_HOST_FUNCTION(get_account_creation_time);
 
 // authorization api
-REGISTER_HOST_FUNCTION(require_auth);
-REGISTER_HOST_FUNCTION(require_auth2);
+REGISTER_HOST_FUNCTION(require_auth, action_check);
+REGISTER_HOST_FUNCTION(require_auth2, action_check);
 REGISTER_HOST_FUNCTION(has_auth);
```
heifner, line 518: If `require_auth` not allowed, why is `has_auth`? Isn't `require_auth` equivalent of `check(has_auth())`?

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -599,6 +601,9 @@ class host_context {
    // act pointer may be invalidated on call to trx_context.schedule_action
    iterator_cache<key_value_object>    keyval_cache;
    std::string                         _pending_console_output;
+
+   bool                                _is_action    = true;  // used to avoid using dynamic_cast to identify type of host_context
+   bool                                _is_sync_call = false; // used to avoid using dynamic_cast to identify type of host_context
```
heifner, line 606: These can be `const`.

heifner (changes requested):

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -554,12 +554,12 @@ class host_context {
 public:
    vector<account_name> get_active_producers() const;
 
-   virtual int get_action( uint32_t type, uint32_t index, char* buffer, size_t buffer_size )const = 0;
-   virtual int get_context_free_data( uint32_t index, char* buffer, size_t buffer_size )const = 0;
+   virtual int get_action( uint32_t type, uint32_t index, char* buffer, size_t buffer_size ) const { EOS_ASSERT(false, unaccessible_api, "default implemention should never be used"); }
+   virtual int get_context_free_data( uint32_t index, char* buffer, size_t buffer_size )const { EOS_ASSERT(false, unaccessible_api, "default implemention should never be used"); }
```
spoonincode, line 558: well, I guess for these couldn't just do an `assert()`, but would also need a `__builtin_unreachable()` to silence the warning of not having a return value

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -599,6 +601,9 @@ class host_context {
    // act pointer may be invalidated on call to trx_context.schedule_action
    iterator_cache<key_value_object>    keyval_cache;
    std::string                         _pending_console_output;
+
+   const bool                          _is_action    = true;  // used to avoid using dynamic_cast to identify type of host_context
+   const bool                          _is_sync_call = false; // used to avoid using dynamic_cast to identify type of host_context
```
spoonincode, line 606: An alternative to these would be for `is_action()` and `is_sync_call()` to be virtual and return false on host_context, and then `apply_context` and `sync_call_context` can override the appropriate one to return true. I'm indifferent on which approach to take, just mentioning it.

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: It is decided to not allow `has_auth()` for consistencies.

https://github.com/AntelopeIO/spring/pull/1296/commits/675c6b25288ad1c1cbcb8345d77aa1cb6ab9cc07

Thanks.

libraries/chain/include/eosio/chain/host_context.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1296/commits/05b812af54d98d0f7cb5e537adc106f1412fc6d9

libraries/chain/include/eosio/chain/host_context.hpp
linh2931: They were removed by https://github.com/AntelopeIO/spring/pull/1296/commits/0b3559cc857cfd90dc735c4b76be489b81410a92

libraries/chain/include/eosio/chain/host_context.hpp
linh2931: I like your idea. It removes 2 member fields and makes the constructor a little simpler.

https://github.com/AntelopeIO/spring/pull/1296/commits/0b3559cc857cfd90dc735c4b76be489b81410a92

spoonincode (approved):

heifner (approved):

PR #1308, opened 2025-04-01
Eliminate duplicate P2P connections

### LINKED ISSUE #1255: P2P: Fix duplicate connection logic
See #1243 for additional info.

The current scheme still leaves duplicates in some race conditions. A simpler approach was recently attempted. See https://github.com/AntelopeIO/spring/pull/1108. However, this approach had a race condition so that it would sometimes disconnect both connections; reverted in https://github.com/AntelopeIO/spring/pull/1186. The issue is that the connection and handshake messages can be sent from each node at the same time. Both nodes can consider the other a duplicate at the same time and both disconnect.

We believe the race condition can be solved using the following technique.

- On disconnect because of duplicate connection, do not attempt an immediate retry of the connections.
- When connecting to a peer, verify there is not already a open connection to that peer node_id.
- Maintain the logic of closing a duplicate if node's `node_id` less than peer `node_id`.

heifner:
Fix for an issue where `net_plugin` would allow some duplicate connections to remain open.

Simplified to just checking both incoming and outgoing connections for duplicate.

Resolves #1255

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -924,7 +924,7 @@ namespace eosio {
       boost::asio::steady_timer        sync_response_expected_timer GUARDED_BY(sync_response_expected_timer_mtx);
 
       alignas(hardware_destructive_interference_sz)
-      std::atomic<go_away_reason>      no_retry{no_reason};
+      std::atomic<go_away_reason>      no_retry{go_away_reason::no_reason};
 
       alignas(hardware_destructive_interference_sz)
       mutable fc::mutex                conn_mtx; //< mtx for last_handshake_recv .. remote_endpoint_ip
@@ -1545,8 +1545,8 @@ namespace eosio {
       auto [on_fork, unknown_block] = block_on_fork(msg_head_id);
       if( unknown_block ) {
          peer_ilog( this, "Peer asked for unknown block ${mn}, sending: benign_other go away", ("mn", msg_head_num) );
-         no_retry = benign_other;
-         enqueue( go_away_message( benign_other ) );
+         no_retry = go_away_reason::benign_other;
+         enqueue( go_away_message{ go_away_reason::benign_other } );
       } else {
          // if peer on fork, start at their last fork_db_root_num, otherwise we can start at msg_head+1
          if (on_fork)
@@ -1619,7 +1619,7 @@ namespace eosio {
    void connection::check_heartbeat( std::chrono::steady_clock::time_point current_time ) {
       if( latest_msg_time > std::chrono::steady_clock::time_point::min() ) {
          if( current_time > latest_msg_time + hb_timeout ) {
-            no_retry = benign_other;
+            no_retry = go_away_reason::benign_other;
             if( !peer_address().empty() ) {
                peer_wlog(this, "heartbeat timed out for peer address");
                close(true);
@@ -1821,8 +1821,8 @@ namespace eosio {
          peer_requested.reset(); // unable to provide requested blocks
          block_sync_send_start = 0ns;
          block_sync_frame_bytes_sent = 0;
-         no_retry = benign_other;
-         enqueue( go_away_message( benign_other ) );
+         no_retry = go_away_reason::benign_other;
+         enqueue( go_away_message{ go_away_reason::benign_other } );
       }
       return true;
    }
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3492,57 +3491,37 @@ namespace eosio {
                set_connection_type( msg.p2p_address);
             else
                peer_dlog(this, "Invalid handshake p2p_address ${p}", ("p", msg.p2p_address));
+         } else {
+            // check if peer requests no trx or no blocks
```
linh2931, line 3495: This comment is a little confusing.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3492,57 +3491,37 @@ namespace eosio {
...
+               return false;
+            fc::unique_lock g_check_conn( check->conn_mtx );
+            fc_dlog( logger, "dup check: connected ${c}, ${l} =? ${r}",
+                     ("c", check->connected())("l", check->last_handshake_recv.node_id)("r", msg.node_id) );
+            if(check->connected() && check->last_handshake_recv.node_id == msg.node_id) {
+               if (my_impl->node_id < msg.node_id) {
+                  fc_dlog( logger, "not duplicate, my_impl->node_id '${lhs}' < msg.node_id '${rhs}'",
+                           ("lhs", my_impl->node_id)("rhs", msg.node_id) );
+                  // only the connection from lower node_id to higher node_id will be considered as a duplicate,
+                  // so there is no chance for both connections to be closed
```
linh2931, line 3511: The is simpler than the complicated prior version.

linh2931 (approved):

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3492,57 +3491,37 @@ namespace eosio {
...
+               return true;
             }
-         } else {
-            peer_dlog(this, "skipping duplicate check, addr == ${pa}, id = ${ni}", ("pa", peer_address())("ni", msg.node_id));
-
-            // check if peer requests no trx or no blocks
-            set_peer_connection_type(msg.p2p_address);
+            return false;
+         };
+         if (my_impl->connections.any_of_connections(std::move(is_duplicate))) {
```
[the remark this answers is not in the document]
heifner: No reason, I removed it.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3492,57 +3491,37 @@ namespace eosio {
...
-               if(check.get() == this)
+         peer_dlog( this, "checking for duplicate" );
+         auto is_duplicate = [&](const connection_ptr& check) {
+            if(check.get() == this)
+               return false;
+            fc::unique_lock g_check_conn( check->conn_mtx );
+            fc_dlog( logger, "dup check: connected ${c}, ${l} =? ${r}",
+                     ("c", check->connected())("l", check->last_handshake_recv.node_id)("r", msg.node_id) );
+            if(check->connected() && check->last_handshake_recv.node_id == msg.node_id) {
+               if (my_impl->node_id < msg.node_id) {
```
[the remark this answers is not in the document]
heifner: In that scenario the node does keep both connections. The peer will then drop one of the two connections leaving you with one connection.

greg7mdp:
I wonder if it would not be preferable to do the duplicate connection check when `(msg.generation >= 1) ` (instead of `==`)?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3492,57 +3491,37 @@ namespace eosio {
...
-         } else {
-            peer_dlog(this, "skipping duplicate check, addr == ${pa}, id = ${ni}", ("pa", peer_address())("ni", msg.node_id));
-
-            // check if peer requests no trx or no blocks
-            set_peer_connection_type(msg.p2p_address);
+            return false;
+         };
+         if (my_impl->connections.any_of_connections(is_duplicate)) {
+            peer_dlog( this, "sending go_away duplicate, msg.p2p_address: ${add}", ("add", msg.p2p_address) );
+            go_away_message gam(duplicate);
```
greg7mdp, line 3520: I don't think the constructor of `go_away_message` serves any useful purpose, and makes the code here at least more awkward. Without it we could do:
`enqueue(go_away_message{duplicate, conn_node_id});`

greg7mdp: bonus point if making `go_away_reason` an `enum class` :-)

greg7mdp (approved):

ericpassmore:
Note:start
category: Other
component: P2P
summary: Fix for an issue where `net_plugin` would allow some duplicate connections to remain open. 
Note:end

Issue #1279, opened 2025-03-26
SC: privileged support


linh2931:
https://github.com/AntelopeIO/spring/pull/1251#issuecomment-2743935409

This is to be done after the refactoring apply_context and sync_call_context https://github.com/AntelopeIO/spring/issues/1261

It also needs to be considered the impact of changing `max_sync_call_depth` in a sync call (should not be allowed) to the wasm allocator vector size https://github.com/AntelopeIO/spring/issues/1269

Issue #1269, opened 2025-03-21
SC: Update `sync_call_wasm_alloc` upon `max_sync_call_depth` changes


linh2931:
`sync_call_wasm_alloc` is a thread_local vector storing `max_sync_call_depth` of wasm allocators per thread (main and read-only threads). When `max_sync_call_depth` changes, the vector must be resized.

One way is to emit a signal (new max_sync_call_depth) to producer_plugin, the plugin posts a task to the thread pool, and the pool runs the task on each thread in the pool. But it is not clear how to make sure the pool runs the task on each thread.

@spoonincode suggested use something like boost lockfree queue to have a central pool of wasm allocators.

@greg7mdp suggested make  max_sync_call_depth  an atomic, and check its value on each thread when accessing sync_call_wasm_alloc , and resize the vector if the value of max_sync_call_depth  has changed.

Need to do some experiment.

PR #1311, opened 2025-04-02
SC: privilege support

linh2931:
Allow privileged host functions to be called if the receiver has the privileged permission.

Note `set_parameters_packed` is not allowed in sync calls as change of `max_sync_call_depth` will interfere with active sync calls. Updating wasm allocator queues upon changed  `max_sync_call_depth` will be done by https://github.com/AntelopeIO/spring/issues/1269

Resolve https://github.com/AntelopeIO/spring/issues/1279

### DIFF
--- a/libraries/chain/host_context.cpp
+++ b/libraries/chain/host_context.cpp
@@ -75,6 +75,11 @@ uint32_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std
          try {
             // use a new sync_call_context for next sync call
             sync_call_context call_ctx(control, trx_context, get_sync_call_sender(), call_receiver, depth, flags, data);
+
+            if (!control.skip_trx_checks()) {
+               call_ctx.privileged = receiver_account->is_privileged();
+            }
+
             control.get_wasm_interface().do_sync_call(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
 
             // store return value
--- a/libraries/chain/webassembly/runtimes/eos-vm.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm.cpp
@@ -420,7 +420,7 @@ REGISTER_LEGACY_HOST_FUNCTION(preactivate_feature, privileged_check);
 REGISTER_HOST_FUNCTION(set_resource_limits, privileged_check);
 REGISTER_LEGACY_HOST_FUNCTION(get_resource_limits, privileged_check);
 REGISTER_HOST_FUNCTION(get_parameters_packed, privileged_check);
-REGISTER_HOST_FUNCTION(set_parameters_packed, privileged_check);
+REGISTER_HOST_FUNCTION(set_parameters_packed, privileged_check, action_check); // Not allowed in sync calls. Change of max_sync_call_depth will interfere with active sync calls.
 REGISTER_HOST_FUNCTION(get_wasm_parameters_packed, privileged_check);
 REGISTER_HOST_FUNCTION(set_wasm_parameters_packed, privileged_check);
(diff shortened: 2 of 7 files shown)


libraries/chain/host_context.cpp
```diff
@@ -75,6 +75,11 @@ uint32_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std
          try {
             // use a new sync_call_context for next sync call
             sync_call_context call_ctx(control, trx_context, get_sync_call_sender(), call_receiver, depth, flags, data);
+
+            if (!control.skip_trx_checks()) {
+               call_ctx.privileged = receiver_account->is_privileged();
+            }
```
heifner, line 81: Why not just always set `call_ctx.privileged` ?

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -420,7 +420,7 @@ REGISTER_LEGACY_HOST_FUNCTION(preactivate_feature, privileged_check);
 REGISTER_HOST_FUNCTION(set_resource_limits, privileged_check);
 REGISTER_LEGACY_HOST_FUNCTION(get_resource_limits, privileged_check);
 REGISTER_HOST_FUNCTION(get_parameters_packed, privileged_check);
-REGISTER_HOST_FUNCTION(set_parameters_packed, privileged_check);
+REGISTER_HOST_FUNCTION(set_parameters_packed, privileged_check, action_check); // Not allowed in sync calls. Change of max_sync_call_depth will interfere with active sync calls.
```
heifner, line 423: Does this prevent any useful sub-divide of the system contract?

heifner (changes requested):

libraries/chain/host_context.cpp
linh2931: `exec_one()` of apply_context has this check for action execution. I did the same for consistency. Probably we can always set `call_ctx.privileged`.

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: I hated to break the uniformity of privileged host function preconditions too. The worse is this restriction is due to implementation of sync call wasm allocator queue resizing.

With the restriction, implementation would be simpler. We can resize right at when `set_parameters_packed` is executed.

Without the restriction, we know `set_parameters_packed` cannot be called from read-only threads and in read-window. On the main thread, we will need to check if `max_sync_call_depth` has changed. If changed, we resize the wasm allocator queue.

@spoonincode, @arhag What do you think about this?

spoonincode: Personally I lean toward allowing it. The complication with resizing can be avoided by simply never shrinking the allocator vector/queue during operation. (yeah this leaves some resources unused until restart but seems no big deal)

arhag: I think we should allow it.

linh2931: Thanks. I will revert to allow it.

linh2931: https://github.com/AntelopeIO/spring/pull/1311/commits/ed1f95544b7c71aa19b41d2e111ebda5dca108f6

libraries/chain/host_context.cpp
linh2931: I think it about more, and have changed to always set it at sync_call_context constructor. Thanks.

https://github.com/AntelopeIO/spring/pull/1311/commits/2500086af7d096a0f9004c52bea9918679b36e28

heifner (approved):

spoonincode (approved):

Issue #185, opened 2024-05-21
Remove eosio.system_tests unittests


heifner:
eosio.system_tests were added back in: https://github.com/AntelopeIO/leap/pull/2054

See the commit comment: 1057d50428401d21aca70cd33404fcecebcf3bb0

> Port over some reference contract tests that provide coverage not found in existing unittests. Most of these are likely redundant, but change_inflation, votepay_share_invariant, and votepay_transition2 triggered a producer_double-confirming error no other tests found.

Need to add a test that provides coverage on producer double-confirming error outside of the eosio.system contract tests.

PR #1316, opened 2025-04-03
Update to peer_keys_db_t to match changes in eos-system-contracts pr #185

greg7mdp:
Updates for changes in https://github.com/eosnetworkfoundation/eos-system-contracts/pull/185.


gdb4713 (approved):

libraries/chain/controller.cpp
```diff
@@ -3339,9 +3382,18 @@ struct controller_impl {
...
       guard_pending.cancel();
+
       return onblock_trace;
    } /// start_block
 
+   void run_readonly_transactions() {
+      // update peer public keys from chainbase db using a readonly trx
+      auto block_num = chain_head.block_num();
+      if (block_num % 120 == 0) {
+         peer_keys_db.update_peer_keys(get_top_producer_keys()); // update once/minute
```
heifner, line 3393: You need a try-catch and drop here for if the system contract does not support the read-only trx.

greg7mdp: It seems that `push_transaction` silently fails and returns an empty return value when the action is not present.

libraries/chain/controller.cpp
```diff
@@ -1274,6 +1274,59 @@ struct controller_impl {
...
+         };
+
+         auto metadata = transaction_metadata::create_no_recover_keys(
+            std::make_shared<packed_transaction>(get_getpeerkeys_transaction()),
+            transaction_metadata::trx_type::read_only);
+
+         const auto& gpo = db.get<global_property_object>();
+
+         auto trace = push_transaction(metadata, fc::time_point::maximum(), fc::microseconds::maximum(),
+                                       gpo.configuration.min_transaction_cpu_usage, true, 0);
```
heifner, line 1309: No need to use `gpo.configuration.min_transaction_cpu_usage` for read-only, use `0` for `billed_cpu_time_us`. `explicit_billed_cpu_time` should be `false` not `true`. I think `max_transaction_time` can probably be left as maximum although we might consider using a smaller value. For `block_deadline`, if you are going to call this from `producer_plugin` then I think you should go ahead and use the actual `block_deadline`.

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/9c783fd7036de6e1efc44b64fd6d95df9e9a7a9a

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/9c783fd7036de6e1efc44b64fd6d95df9e9a7a9a

gdb4713 (approved):

libraries/chain/controller.cpp
```diff
@@ -3339,9 +3380,20 @@ struct controller_impl {
       }
 
       guard_pending.cancel();
+
       return onblock_trace;
    } /// start_block
 
+   void run_readonly_transactions(fc::time_point deadline) {
```
heifner, line 3387: I think a better name would be something like: `update_peer_keys()`

libraries/chain/include/eosio/chain/peer_keys_db.hpp
```diff
@@ -13,34 +13,34 @@ namespace eosio::chain {
...
 
    void set_active(bool b) { _active = b; }
 
    // must be called from main thread
-   size_t update_peer_keys(const controller& chain, uint32_t lib_number);
-
-   // safe to be called from any thread
-   std::optional<public_key_type> get_peer_key(name n) const;
+   // return the new peers either:
+   //  - added to the top-50
```
heifner, line 25: All the comments about "top-50" should be updated to just say top according to what contract returns.

libraries/chain/controller.cpp
```diff
@@ -1274,6 +1274,57 @@ struct controller_impl {
...
+            trx.set_reference_block(chain_head.id());
+            set_trx_expiration(trx);
+            return trx;
+         };
+
+         auto metadata = transaction_metadata::create_no_recover_keys(
+            std::make_shared<packed_transaction>(get_getpeerkeys_transaction()),
+            transaction_metadata::trx_type::read_only);
+
+         auto trace = push_transaction(metadata, deadline, fc::microseconds::maximum(), 0, false, 0);
```
heifner, line 1306: I think we should limit this read-only trx runtime to something reasonable. How about, 10ms.

unittests/contracts/eosio.system/peer_keys.cpp
```diff
@@ -0,0 +1,63 @@
...
+   peer_keys_table  peer_keys_table(get_self(), get_self().value);
+   producers_table  producers(get_self(), get_self().value);
+   constexpr size_t max_return = 50;
+
+   getpeerkeys_res_t resp;
+   resp.reserve(max_return);
+
+   auto idx = producers.get_index<"prototalvote"_n>();
+
+   for( auto it = idx.cbegin(); it != idx.cend() && resp.size() < max_return; ++it ) {
```
heifner, line 53: Consider adding a comment that this is simplified implementation than EOS system contract.

unittests/getpeerkeys_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+   for (size_t i=0; i<prod_names.size(); ++i) {
+      auto n = prod_names[i];
+      if (i % 2 == 0)
+         BOOST_REQUIRE_EQUAL(success(), regpeerkey(n, get_public_key(n)));
+   }
+
+   auto peerkeys = control->get_top_producer_keys(fc::time_point::maximum()); // call readonly action from controller
+   BOOST_REQUIRE_EQUAL(peerkeys.size(), 21);
+   //BOOST_REQUIRE(peerkeys == control->get_top_producer_keys()); // call readonly action from controller
```
heifner, line 38: remove

heifner (changes requested):

libraries/chain/controller.cpp
greg7mdp: I don't mind changing the name, but the reason I named it like that was because I though we may use other read-only transactions in the future to access chainbase data, so I thought better to use a non-specific name.

heifner: My vote would be to give it a more descriptive name.

libraries/chain/controller.cpp
greg7mdp: I did `20ms` because I am a little bit concerned that it does a number of db accesses and 10ms might be short in our overwhelmed ci/cd, but if you think it is too much please let me know and I'll change it.

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/9ca911a165eb9bafeb2a1309ffc9fabf6ad70e15 hopefully not too much!

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/ca9c2f193a449e760e44c5c0f9f7cf1bcde1eb65

unittests/contracts/eosio.system/peer_keys.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/406d92869e38336bdcd1cd02277bd6c63ea78dc8

unittests/getpeerkeys_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/e80a6e5837c7c29a5fc088901f7b23a442455b1f

libraries/chain/controller.cpp
heifner: I don't see https://github.com/AntelopeIO/spring/commit/9ca911a165eb9bafeb2a1309ffc9fabf6ad70e15 being any better than `run_readonly_transactiions`. I think it should describe more what is accomplishing as opposed to how it is accomplishing it.

greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/59e93360c2677ef1b15bc0949fd14d8191aed2ee

heifner (approved):

libraries/chain/peer_keys_db.cpp
```diff
@@ -5,85 +5,46 @@ namespace eosio::chain {
...
-         return num_updated;
+   fc::lock_guard g(_m);
+   new_peers_t res;
+
+   // update ranking of those removed from top selected producers
+   // -----------------------------------------------------------
+   for (auto& pi : _peer_info_map) {
+      if (!current.contains(pi.first)) {
+         pi.second.rank = std::numeric_limits<uint32_t>::max();
+         res.insert(pi.first);
```
heifner, line 34: I think you should remove the non-top ones from `_peer_info_map`. This can return ones that are no longer in the list as modified, but I don't think we need to keep them around.

greg7mdp: https://github.com/AntelopeIO/spring/pull/1316/commits/de28545b8fe0a231beba73a7fa2e9e2fc8315568

Issue #1170, opened 2025-02-17
Avoid block processing delays due to speculative trx execution


heifner:
Consider interrupting a speculative trx when a block is received. Currently a speculative trx can delay block processing for up to 150ms (current EOS Mainnet max trx time). Alternatively, do not start a speculative trx on a producer if it is less than max-transaction-time from start of first block of a round.

See https://github.com/AntelopeIO/spring/issues/1151#issuecomment-2660193692

bhazzard:
Could give minor benefits to reducing local forks and applying blocks faster.

Low priority, only pick up opportunistically.

PR #1322, opened 2025-04-04
[1.1.3] Reduce max read-only-read-window-time-us

### LINKED ISSUE #1318: 1.1 Limit read-only-read-window-time-us
#1288 Fixed an issue with read-only trxs being retried when `read-only-read-window-time-us` is set to larger than `max_block_cpu_usage`. The fix allows for read-only trxs via `read-only-read-window-time-us` to be configured to any large value. For example, 1 hr. Currently, read-only trxs are not interruptible by ctrl-c. Ability to interrupt read-only trxs by ctrl-c is being added via https://github.com/AntelopeIO/spring/issues/1170#issuecomment-2773007129 to Spring 1.2. 

For Spring 1.1.x, limit `read-only-read-window-time-us` to 200ms which corresponds to the current value of `max_block_cpu_usage` on EOS Mainnet. Once #1170 is added to 1.2, then `read-only-read-window-time-us` will then be allowed to be set to any desirable value, including larger than block time.

heifner:
For Spring 1.1 reduce the maximum allowed `read-only-read-window-time-us` to 200ms. This is done because Spring 1.1 does not support interrupting a read-only trx via ctrl-c. Therefore, only allow short run read-only trxs. Note the plan is the relax this restriction in Spring 1.2, see #1170. 

Resolves #1318

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1479,6 +1479,10 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
                  plugin_config_exception,
                  "read-only-read-window-time-us (${read}) must be at least greater than  ${min} us",
                  ("read", _ro_read_window_time_us)("min", _ro_read_window_minimum_time_us));
+      if (_ro_read_window_time_us > fc::milliseconds(200)) {
+         ilog("Reducing read-only-read-window-time-us to 200ms");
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

linh2931 (approved):

gdb4713 (approved):

ericpassmore:
Note:start
category: Other
component: Plugins
summary: Cap the maximum allowed `read-only-read-window-time-us` to 200ms.
Note:end

PR #1328, opened 2025-04-07
Interrupt read-only trxs via ctrl-c

### LINKED ISSUE #1323: Interrupt read-only trx on ctrl-c
Split off from #1170 as it can be worked separately and represents different functionality.

Add interrupt of read-only trxs on ctrl-c/SIGINT/SIGTERM/SIGPIPE. Do not interrupt read-only trxs when a block is received.

heifner:
Interrupt read-only transactions via ctrl-c/SIGINT/SIGTERM/SIGPIPE.
Read-only transactions are not interrupted when a new block is received. This means a node will continue to honor `read-only-read-window-time-us` even when set higher than block time. However, a node can be shutdown immediately via `ctrl-c`.

Resolves #1323


ericpassmore:
Note:start
category: Other
component: Internal
summary: Add interrupt of read-only trxs on ctrl-c/SIGINT/SIGTERM/SIGPIPE. Do not interrupt read-only trxs when a block is received.
Note:end

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -3051,6 +3056,23 @@ void producer_plugin::received_block(uint32_t block_num, chain::fork_db_add_t fo
...
 
+// thread-safe, called when ctrl-c/SIGINT/SIGTERM/SIGPIPE is received
+void producer_plugin::interrupt() {
+   fc_ilog(_log, "interrupt");
+   my->interrupt_read_only();
+   my->chain_plug->chain().interrupt_apply_block_transaction();
+}
+
+void producer_plugin_impl::interrupt_read_only() {
+   if (!_ro_timers.empty() && _ro_max_trx_time_us > fc::milliseconds(250)) {
```
linh2931, line 3067: Why not allow interrupt for any value of `_ro_max_trx_time_us`, if the user wants to do a ctrl-C?

tests/interrupt_read_only_trx_test.py
```diff
@@ -0,0 +1,179 @@
...
+    }
+    return apiNode.pushTransaction(trx, opts, silentErrors=True)
+
+def sendReadOnlyForeverPayloadless():
+    try:
+        sendTransaction('payloadless', action='doitforever', data={}, auth=[], opts='--read')
+    except Exception as e:
+        Print("Ignore Exception in sendReadOnlyForeverPayloadless: ", repr(e))
+
+def timeoutTest():
```
linh2931, line 145: Can this test be placed into `read_only_trx_test.py` so test setup code is not repeated?

heifner: I considered it, but the setup is different and seemed like it would overly complicate `read_only_trx_test.py`.

plugins/producer_plugin/producer_plugin.cpp
heifner: My thought was that if a read-only trx already has a limit of <= 250ms then it will finish in at most 250ms which is not very long so no reason to kill it, just let it finish.

linh2931 (approved):

linh2931 (approved):

heifner:
Not 100% sure this failure is due to the changes here, but seems likely: https://github.com/AntelopeIO/spring/actions/runs/14380363399/job/40323298195
That test failed because the node didn't shutdown. Looks like it was stuck somewhere in producer_plugin shutdown.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -3051,6 +3056,25 @@ void producer_plugin::received_block(uint32_t block_num, chain::fork_db_add_t fo
...
+   app().executor().stop(); // shutdown any blocking read_only_execution_task
+   my->interrupt_read_only();
+   my->chain_plug->chain().interrupt_apply_block_transaction();
+}
+
+void producer_plugin_impl::interrupt_read_only() {
+   // if read-only trx is going to finish in less than 250ms then might as well let it finish
+   if (!_ro_timers.empty() && _ro_max_trx_time_us > fc::milliseconds(250)) {
+      fc_ilog(_log, "interrupting read-only trxs");
+      for (auto& t : _ro_timers) {
```
greg7mdp, line 3071: nit - I think `auto` is fine for a pointer.

greg7mdp (approved):

Issue #1216, opened 2025-02-27
SC: Add `sync_call` traces to call traces


linh2931:
* Update `console` related host functions to make sure console output to be written into the sync call trace
* Store exception, if any, into the sync call trace.

Issue #1220, opened 2025-02-27
SC: Support the new `call` section in ABI


linh2931:

heifner:
https://github.com/AntelopeIO/spring/pull/1333#discussion_r2047517972

PR #1333, opened 2025-04-07
SC: Implement sync call trace

linh2931:
Impements sync call trace. Each individual sync call has a `sync_call_trace` struct containing all information about the sync call. An action trace has a vector of `sync_call_trace` for all sync calls happened within the call. The relationship between sync calls are identified by `ordinal` and `sender_ordinal`.

`console` field and condole debug prints will be implemented in a separate PR.

SHiP support is tracked by https://github.com/AntelopeIO/spring/issues/1334

Resolves https://github.com/AntelopeIO/spring/issues/1216

A sample sync call traces for
```
 [
  {
    "action_ordinal": 1,
    "creator_action_ordinal": 0,
    "closest_unnotified_ancestor_action_ordinal": 0,
    "receipt": {
      "receiver": "caller",
      "act_digest": "4169266343f52db133bded246813966fb493bf53a4c16ecea999ece1a6ca7f9f",
      "global_sequence": 48,
      "recv_sequence": 1,
      "auth_sequence": [
        [
          "caller",
          3
        ]
      ],
      "code_sequence": 1,
      "abi_sequence": 1
    },
    "receiver": "caller",
    "act": {
      "account": "caller",
      "name": "doit",
      "authorization": [
        {
          "actor": "caller",
          "permission": "active"
        }
      ],
      "data": ""
    },
    "context_free": false,
    "elapsed": 618,
    "console": "before calling callee1before calling callee2",
    "trx_id": "9189869e2c41413775edbaf1721d036dce527ab576a4e883a874d47f9f39bd2b",
    "block_num": 11,
    "block_time": "2020-01-01T00:00:05.000",
    "account_ram_deltas": [],
    "return_value": "",
    "call_traces": [
      {
        "call_ordinal": 1,
        "sender_ordinal": 0,
        "receiver": "callee1",
        "read_only": false,
        "data": "",
        "elapsed": 299,
        "console": "I am callee1",
        "console_markers": [
          12
        ],
        "return_value": "4920616d2063616c6c656531"
      },
      {
        "call_ordinal": 2,
        "sender_ordinal": 1,
        "receiver": "callee11",
        "read_only": false,
        "data": "4920616d2063616c",
        "elapsed": 145,
        "console": "I am callee11",
        "console_markers": [],
        "return_value": "4920616d2063616c6c65653131"
      },
      {
        "call_ordinal": 3,
        "sender_ordinal": 0,
        "receiver": "callee2",
        "read_only": false,
        "data": "",
        "elapsed": 115,
        "console": "I am callee2",
        "console_markers": [],
        "return_value": "4920616d2063616c6c656532"
      }
    ],
    "console_markers": [
      22,
      44
    ]
  }
]
```

A sample for exception propagation:

```
[
  {
    "action_ordinal": 1,
    "creator_action_ordinal": 0,
    "closest_unnotified_ancestor_action_ordinal": 0,
    "receiver": "caller",
    "act": {
      "account": "caller",
      "name": "doit",
      "authorization": [
        {
          "actor": "caller",
          "permission": "active"
        }
      ],
      "data": ""
    },
    "context_free": false,
    "elapsed": 677,
    "console": "",
    "trx_id": "9189869e2c41413775edbaf1721d036dce527ab576a4e883a874d47f9f39bd2b",
    "block_num": 11,
    "block_time": "2020-01-01T00:00:05.000",
    "account_ram_deltas": [],
    "except": {
      "code": 3050003,
      "name": "eosio_assert_message_exception",
      "message": "eosio_assert_message assertion failure",
      "stack": [
        {
          "context": {
            "level": "error",
            "file": "cf_system.cpp",
            "line": 14,
            "method": "eosio_assert",
            "hostname": "",
            "thread_name": "unit_test",
            "timestamp": "2025-04-14T21:00:04.716"
          },
          "format": "assertion failure with message: ${s}",
          "data": {
            "s": "eosio_assert by callee11"
          }
        },
        {
          "context": {
            "level": "warn",
            "file": "host_context.cpp",
            "line": 121,
            "method": "execute_sync_call",
            "hostname": "",
            "thread_name": "unit_test",
            "timestamp": "2025-04-14T21:00:04.716"
          },
          "format": "sync call exception ${receiver} <= ${sender} console output: ${console}",
          "data": {
            "receiver": "callee11",
            "sender": "callee1",
            "console": ""
          }
        },
        {
          "context": {
            "level": "warn",
            "file": "host_context.cpp",
            "line": 121,
            "method": "execute_sync_call",
            "hostname": "",
            "thread_name": "unit_test",
            "timestamp": "2025-04-14T21:00:04.716"
          },
          "format": "sync call exception ${receiver} <= ${sender} console output: ${console}",
          "data": {
            "receiver": "callee1",
            "sender": "caller",
            "console": ""
          }
        },
        {
          "context": {
            "level": "warn",
            "file": "apply_context.cpp",
            "line": 124,
            "method": "exec_one",
            "hostname": "",
            "thread_name": "unit_test",
            "timestamp": "2025-04-14T21:00:04.716"
          },
          "format": "${receiver} <= ${account}::${action} console output: ${console}",
          "data": {
            "console": "",
            "account": "caller",
            "action": "doit",
            "receiver": "caller"
          }
        }
      ]
    },
    "error_code": "10000000000000000000",
    "return_value": "",
    "call_traces": [
      {
        "call_ordinal": 1,
        "sender_ordinal": 0,
        "receiver": "callee1",
        "read_only": false,
        "data": "",
        "elapsed": 478,
        "console": "",
        "console_markers": [
          0
        ],
        "except": {
          "code": 3050003,
          "name": "eosio_assert_message_exception",
          "message": "eosio_assert_message assertion failure",
          "stack": [
            {
              "context": {
                "level": "error",
                "file": "cf_system.cpp",
                "line": 14,
                "method": "eosio_assert",
                "hostname": "",
                "thread_name": "unit_test",
                "timestamp": "2025-04-14T21:00:04.716"
              },
              "format": "assertion failure with message: ${s}",
              "data": {
                "s": "eosio_assert by callee11"
              }
            },
            {
              "context": {
                "level": "warn",
                "file": "host_context.cpp",
                "line": 121,
                "method": "execute_sync_call",
                "hostname": "",
                "thread_name": "unit_test",
                "timestamp": "2025-04-14T21:00:04.716"
              },
              "format": "sync call exception ${receiver} <= ${sender} console output: ${console}",
              "data": {
                "receiver": "callee11",
                "sender": "callee1",
                "console": ""
              }
            },
            {
              "context": {
                "level": "warn",
                "file": "host_context.cpp",
                "line": 121,
                "method": "execute_sync_call",
                "hostname": "",
                "thread_name": "unit_test",
                "timestamp": "2025-04-14T21:00:04.716"
              },
              "format": "sync call exception ${receiver} <= ${sender} console output: ${console}",
              "data": {
                "receiver": "callee1",
                "sender": "caller",
                "console": ""
              }
            }
          ]
        },
        "error_code": "10000000000000000000",
        "return_value": ""
      },
      {
        "call_ordinal": 2,
        "sender_ordinal": 1,
        "receiver": "callee11",
        "read_only": false,
        "data": "0000000000000000",
        "elapsed": 305,
        "console": "",
        "console_markers": [],
        "except": {
          "code": 3050003,
          "name": "eosio_assert_message_exception",
          "message": "eosio_assert_message assertion failure",
          "stack": [
            {
              "context": {
                "level": "error",
                "file": "cf_system.cpp",
                "line": 14,
                "method": "eosio_assert",
                "hostname": "",
                "thread_name": "unit_test",
                "timestamp": "2025-04-14T21:00:04.716"
              },
              "format": "assertion failure with message: ${s}",
              "data": {
                "s": "eosio_assert by callee11"
              }
            },
            {
              "context": {
                "level": "warn",
                "file": "host_context.cpp",
                "line": 121,
                "method": "execute_sync_call",
                "hostname": "",
                "thread_name": "unit_test",
                "timestamp": "2025-04-14T21:00:04.716"
              },
              "format": "sync call exception ${receiver} <= ${sender} console output: ${console}",
              "data": {
                "receiver": "callee11",
                "sender": "callee1",
                "console": ""
              }
            }
          ]
        },
        "error_code": "10000000000000000000",
        "return_value": ""
      }
    ],
    "console_markers": [
      0
    ]
  }
]
```


libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -45,6 +69,7 @@ namespace eosio::chain {
       std::optional<fc::exception>    except;
       std::optional<uint64_t>         error_code;
       std::vector<char>               return_value;
+      std::optional<std::vector<sync_call_trace>>  call_traces;
```
spoonincode, line 72: Why an optional vector? I'm guessing you don't want call_traces to appear in JSONified trace if an action didn't use sync calls at all?

linh2931: Yes, that is what I want. Should we show `call_traces` with an empty `""` instead if an action did not make any sync calls? @arhag

arhag: Wouldn't an empty vector be sufficient to unambiguously indicate the action didn't use sync calls at all?

I don't see the harm of always including the `call_traces` in the JSON output of the action trace with it being `[]` if no sync calls were made.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,30 @@ namespace eosio::chain {
...
+         , sender(sender)
+         , receiver(receiver)
+         , flags(flags)
+         , data(data.begin(), data.end())
+      {
+      }
+
+      uint32_t                      ordinal = 1;
+      const uint32_t                sender_ordinal = 0;
+      const account_name            sender;
```
spoonincode, line 34: Is sender redundant information since you have `sender_ordinal`?

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,30 @@ namespace eosio::chain {
...
+      const account_name            receiver;
+      const uint64_t                flags = 0;
+      const std::vector<char>       data;
+      fc::microseconds              elapsed;
+      string                        console;
+      std::optional<fc::exception>  except;
+      std::optional<uint64_t>       error_code;
+      int64_t                       return_value_size_or_error_id = 0; // if >=0: return value size, if -1: receiver not supporting sync calls
```
spoonincode, line 42: The return value size is redundant information since it's already implicitly part of the return_value. I am not sure how to best handle the validity of this and other items in the trace. It almost seems like really we have a variant made up of
* contract was runnable: `elapsed`, `console`, `except`,`error_code`, `return_value`
* contract was not runnable `error_id`
But using a variant here may be quite the departure from precedent.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,30 @@ namespace eosio::chain {
...
+         , flags(flags)
+         , data(data.begin(), data.end())
+      {
+      }
+
+      uint32_t                      ordinal = 1;
+      const uint32_t                sender_ordinal = 0;
+      const account_name            sender;
+      const account_name            receiver;
+      const uint64_t                flags = 0;
```
spoonincode, line 36: Does it make any sense to "unwrap" the flags in the user visible trace? For example instead of a consumer just seeing `flags: 1` or `flags: 2` and needing to know what the magic numbers are, they'd see for example
```
   "read_only": false,
   "some_other_flag": true
```

spoonincode (commented):
Adding a few comments on structure of user visible trace that has been nagging me -- I'm not really sure what the correct solution is for some of these. Especially whether there is precedence to lean in to for inconsistency or not.

libraries/chain/include/eosio/chain/trace.hpp
linh2931: I put `sender` here for simple reference. But it can be found from `sender_ordinal`. Will remove it.

libraries/chain/include/eosio/chain/trace.hpp
linh2931: I was struggling about this too. I would replace `return_value_size_or_error_id` with an optional `error_id`. If a contract is runnable, this field is not presented. If a contract is not runnable, this field shows up with reasons., while `elapsed`, `console`, `except`, `error_code`, and `return_value` are empty.

libraries/chain/include/eosio/chain/trace.hpp
linh2931: Good idea. I will unwrap it.

linh2931: One consideration is, if we use new bits in the future, we will need to add them to the call trace. But that should be OK as it is a protocol change.

libraries/chain/include/eosio/chain/trace.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1333/commits/3038697fe5791145081852aa7a6103a2c83c8be0

libraries/chain/include/eosio/chain/trace.hpp
spoonincode: That is true... I suppose it would be most disruptive for SHIP since it would require a whole new sync_call_trace_vx whenever a new flag is added.

linh2931:
> Adding a few comments on structure of user visible trace that has been nagging me -- I'm not really sure what the correct solution is for some of these. Especially whether there is precedence to lean in to for inconsistency or not.

I will add some comments about the new struct and its fields to help users.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,72 @@ namespace eosio::chain {
...
+      {
+      }
+
+      // receiver's ordinal. A sequence number starting with 1, unique within
+      // an action. It can be used to reference `call_trace` struct in `call_traces`
+      // vector in `action_trace` struct.
+      uint32_t                      ordinal = 1;
+
+      // sender's ordinal. If the caller is the action, sender_ordinal is 0.
+      const uint32_t                sender_ordinal = 0;
```
spoonincode, line 44: I should have noticed this sooner -- I think we'd want to make these varints (`fc::unsigned_int`). Ordinals in `action_trace` are also varints.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,72 @@ namespace eosio::chain {
    struct transaction_trace;
    using transaction_trace_ptr = std::shared_ptr<transaction_trace>;
 
+   // Identify the starting position of a sync call in the call's console log.
+   // This is used for pretty printing console logs.
+   struct console_marker {
+      uint32_t call_ordinal     = 0;
+      uint32_t console_position = 0;
+   };
```
spoonincode, line 27: Probably varint for these too

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,72 @@ namespace eosio::chain {
    struct transaction_trace;
    using transaction_trace_ptr = std::shared_ptr<transaction_trace>;
 
+   // Identify the starting position of a sync call in the call's console log.
+   // This is used for pretty printing console logs.
+   struct console_marker {
+      uint32_t call_ordinal     = 0;
```
spoonincode, line 25: When discussed on call I thought we came to the conclusion we only needed the position? i.e. type could be  `vector<unsigned_int> console_markers` and the size of `console_markers` is the number of sync calls performed I think[^1]. Can we still do it that way or was there a challenge with it?

[^1]: I seem to remember us saying number of sync calls minus 1 but that isn't making sense to me immediately :thinking:

libraries/chain/include/eosio/chain/trace.hpp
linh2931: Thanks. Will change.

libraries/chain/include/eosio/chain/trace.hpp
linh2931: I thought using the struct was more clear. But will change to one dimensional vector to save space, as this is to be used by SHiP. `n-1` should be OK as we know the start and end.

linh2931: Hm, when I started to change the code, it seemed to me we need `n` instead of `n-1`. Did I miss something?

If we have `n` sync calls, we should have `n` points to indicate where in the `console` we start the calls.

If we make only 1 sync call, I think we still need 1marker, not 0.

arhag: I think you are correct @linh2931.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -19,6 +19,63 @@ namespace eosio::chain {
...
+      // the console log produced by the call
+      string                        console;
+
+      // for each call directly made by the current sync call, identify the starting
+      // position in `console`.
+      // For example, suppose current sync call:
+      //    prints 10 chars; calls sync1; prints 50 chars; calls sycn2,
+      // console_markers would look like { 10, 60 }
+      // This is used for pretty printing console logs to show hierachy of all logs.
+      std::vector<fc::unsigned_int>  console_markers;
```
heifner, line 61: Should this be optional and not used if `contracts-console=false`? Or if not optional maybe it should be an empty vector if `contracts-console=false`.

linh2931: I think we should always provide it, as it can be used by `cleos` and other downstream tools to recreate call hierarchy for console logs.

heifner: If `contracts-console=false` there is no console data. `console == ""`.

linh2931: This https://github.com/AntelopeIO/spring/blob/067730e131378cdb6f7c77a38cb179353a8b4f4e/libraries/chain/apply_context.cpp#L206 always executes, while `contracts-console` is only for immediate console printing:
https://github.com/AntelopeIO/spring/blob/067730e131378cdb6f7c77a38cb179353a8b4f4e/libraries/chain/apply_context.cpp#L191-L193

heifner: That is going to be empty if `contracts-console=false`
https://github.com/AntelopeIO/spring/blob/7c352cedc2b63e9bb7f3926e442cf718aab428d5/libraries/chain/webassembly/console.cpp#L9-L9

linh2931: I think we keep it as an empty vector to match an empty console string if `contracts-console=false`?

libraries/chain/host_context.cpp
```diff
@@ -39,54 +39,83 @@ host_context::~host_context() = default;
...
-              "reached sync call max call depth ${max_depth}", ("max_depth", max_depth));
+   auto start = fc::time_point::now();
 
-   const auto max_data_size = control.get_global_properties().configuration.max_sync_call_data_size;
-   EOS_ASSERT(data.size() <= max_data_size, sync_call_call_data_exception,
-              "sync call call data size must be less or equal to ${max_data_size} bytes", ("max_data_size", max_data_size));
+   // As early as possible, create the call trace of this new sync call in the parent's
+   // (sender's) trace to record entire trace of the sync call, including any exceptions
+   auto& trace = get_root_action_trace();
+   const bool read_only = flags & static_cast<uint64_t>(sync_call_flags::read_only);
```
spoonincode, line 47: In an earlier PR we discussed using `has_field()` not sure how much it matters

libraries/chain/host_context.cpp
```diff
@@ -39,54 +39,83 @@ host_context::~host_context() = default;
...
+      finalize_call_trace(call_trace, start);
       throw;
    };
 
+   auto handle_call_failure = [&]()
+   {
+      auto& call_trace = get_call_trace(ordinal);
+      call_trace.error_id = -1;
+      finalize_call_trace(call_trace, start);
+      trx_context.checktime();
```
spoonincode, line 67: Probably not strictly necessary -- since no new WASM was run in this case, the expiration handlers weren't touched and we didn't run a risk of missing an expiration. Fine to keep as is since we may need to take more of a look on how expiration works in the future

libraries/chain/host_context.cpp
```diff
@@ -39,54 +39,83 @@ host_context::~host_context() = default;
...
+
+         EOS_ASSERT(flags <= static_cast<uint64_t>(sync_call_flags::all_allowed_bits), sync_call_validate_exception,  // all but `std::bit_width(all_allowed_bits)` LSBs must be 0s
+                    "only ${bits} least significant bits of sync call's flags (${flags}) can be set", ("bits", std::bit_width(static_cast<uint64_t>(sync_call_flags::all_allowed_bits)))("flags", flags));
+
          const account_metadata_object* receiver_account = &db.get<account_metadata_object, by_name>( call_receiver);
          if (receiver_account->code_hash.empty()) {
-            // TBD store the info in sync call trace
-            ilog("receiver_account->code_hash empty");
-            return return_value_size;
+            return handle_call_failure();
```
spoonincode, line 94: In this case the console marker will not be created. Seems like creation of console marker should be moved way up earlier in the function

libraries/chain/host_context.cpp
```diff
@@ -39,54 +39,83 @@ host_context::~host_context() = default;
...
+
             if (rc == sync_call_return_code::receiver_not_support_sync_call) {  //  Currently -1 means there is no valid sync call entry point
-               return -1;
+               return handle_call_failure();
             }
 
             // store return value
             last_sync_call_return_value = std::move(call_ctx.return_value);
             return_value_size = last_sync_call_return_value.size();
          } catch( const wasm_exit&) {}
```
spoonincode, line 117: Not new to this PR but I realize here that if a contract sets return value then `eosio_exit`s it looks like the return value is lost. Worse then that, highly likely there is a JIT vs OC difference here. Can make new issue for this but we need to track it some how if not fixed here.

libraries/chain/include/eosio/chain/sync_call_context.hpp
```diff
@@ -14,16 +14,22 @@ enum class sync_call_flags {
 
 class sync_call_context : public host_context {
 public:
-   sync_call_context(controller& con, transaction_context& trx_ctx, account_name sender, account_name receiver, bool privileged, uint32_t sync_call_depth, uint64_t flags, std::span<const char> data);
+   sync_call_context(controller& con, transaction_context& trx_ctx, uint32_t oridnal, action_trace& root_action_trace, account_name sender, account_name receiver, bool privileged, uint32_t sync_call_depth, uint64_t flags, std::span<const char> data);
```
spoonincode, line 17: spelling on ordinal

spoonincode (approved):

spoonincode:
I think these 3 should be addressed but otherwise LGTM
https://github.com/AntelopeIO/spring/pull/1333#discussion_r2042653748
https://github.com/AntelopeIO/spring/pull/1333#discussion_r2042685147
https://github.com/AntelopeIO/spring/pull/1333#discussion_r2042707796

spoonincode:
Also probably want to update the PR description with the new trace format. But there is plenty other documentation that will need to be updated one day too

libraries/chain/apply_context.cpp
```diff
@@ -121,7 +121,7 @@ void apply_context::exec_one()
                }
             }
          }
-      } FC_RETHROW_EXCEPTIONS( warn, "${receiver} <= ${account}::${action} pending console output: ${console}", ("console", _pending_console_output)("account", act->account)("action", act->name)("receiver", receiver) )
+      } FC_RETHROW_EXCEPTIONS( warn, "${receiver} <= ${account}::${action} console output: ${console}", ("console", trx_context.get_action_trace( action_ordinal ).console)("account", act->account)("action", act->name)("receiver", receiver) )
```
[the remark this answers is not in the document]
linh2931: Thanks! I renamed `get_root_action_trace()` (I did not like the name when I initially named it) with `get_current_action_trace()`, and replaced direct use of `trx_context.get_action_trace( action_ordinal )` with `get_current_action_trace()`. The code looks nicer.

https://github.com/AntelopeIO/spring/pull/1333/commits/26fbb7b27b9b2f5e2ae63b19837b4f3b9664cf97

libraries/chain/host_context.cpp
linh2931: Because the last element `all_allowed_bits` in the enum `sync_call_flags` is not a flag (or field in a sense). I did not use the `has_field` mechanism. Will revisit this later.

libraries/chain/host_context.cpp
linh2931: I put it here thinking if we could not make a sync call for any reasons before creating the sync call context, we should not put the sync call position in the markers, since no call was made.

spoonincode: No call was made but you added a trace for the call still. That means it becomes impossible to use the markers properly because the number of markers will not equal the number of sync call traces

libraries/chain/host_context.cpp
linh2931: https://github.com/AntelopeIO/spring/issues/1372

libraries/chain/host_context.cpp
linh2931: Yes, you are right. I will put them together.

linh2931: https://github.com/AntelopeIO/spring/pull/1333/commits/b361003952ccbcdd944c4e49bd31d7067542ba97

libraries/chain/include/eosio/chain/sync_call_context.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1333/commits/9318fe4ec60d14daafba381e84c3640c69976a81

linh2931:
> Also probably want to update the PR description with the new trace format. But there is plenty other documentation that will need to be updated one day too

I updated the PR description with the new trace format. It looks quite nice!

Yes, we will write good documentation when the feature is done.

linh2931:
> I think these 3 should be addressed but otherwise LGTM [#1333 (comment)](https://github.com/AntelopeIO/spring/pull/1333#discussion_r2042653748) [#1333 (comment)](https://github.com/AntelopeIO/spring/pull/1333#discussion_r2042685147) [#1333 (comment)](https://github.com/AntelopeIO/spring/pull/1333#discussion_r2042707796)

They have been addressed. Thanks!

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -576,6 +576,10 @@ namespace impl {
                }
             }
          } catch(...) {}
+
+         mvo("call_traces", act_trace.call_traces);
```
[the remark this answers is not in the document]
spoonincode: That requires sync calls to be plumbed in to ABI which they aren't yet. (I mean the overload technically doesn't but it won't do much useful without ABI support)

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -576,6 +576,10 @@ namespace impl {
...
+         mvo("console_markers", act_trace.console_markers);
```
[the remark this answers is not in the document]
spoonincode: We talked about this in multiple daily standups and afaict it matches what was agreed upon? If the JSON (_not what cleos pretty prints_) needs to be easier to manually grok we should discuss it again and come up with a new approach.

libraries/chain/include/eosio/chain/abi_serializer.hpp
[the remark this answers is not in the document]
heifner: Thanks. Added a link to this in https://github.com/AntelopeIO/spring/issues/1220. It can be worked there.

libraries/chain/include/eosio/chain/abi_serializer.hpp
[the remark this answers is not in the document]
heifner: That is true, we did discuss this. Seeing it in practice makes me think it is less than ideal. But the alternative of parsing the JSON console string is not great either. I withdraw my comment.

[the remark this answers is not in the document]
spoonincode: The only immediate thought I have is to change the console logs to be a `vector<variant<string, varint>>`. So the effective output is something like
```
"console": ["console output before a sync call", 1, "some stuff between two sync calls", 2, "after doing a couple sync calls"]
```
The numbers being the sync call ordinals.

Now of course if you output `vector<variant<string, varint>>` through standard variant mechanism it won't look like that. However, _we're already "manually" crafting the trace's JSON for ABI stuff_! So it isn't a stretch to go ahead and collapse the variant to the appropriate JSON types so it does look that nice as above. This does make the outputted JSON impossible to ingest through ABI again though, but I can't see any immediate reason that's a problem?

Of course the traces in ship would need to remain a proper `variant<string, varint>` which does introduce an inconsistency, at least technically. But ship never does ABI deserialization either so it's already not going to be exactly what the trace was from the node.

idk if it's worth it -- but I do believe it would improve readability some.

We also would still have the challenge of what to do about current `console` in action trace (the current approach is nice in that it doesn't touch it at all)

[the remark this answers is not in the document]
heifner: I'm also not sure it is worth it. But that could also be: 
```
"console": ["console output before a sync call", "some stuff between two sync calls", "after doing a couple sync calls", "", "after another sync call with no logging in between"]
```
Then it would be representable in ABI.

libraries/chain/apply_context.cpp
```diff
@@ -193,12 +193,21 @@ void apply_context::finalize_trace( action_trace& trace, const fc::time_point& s
...
 
+void apply_context::console_append(std::string_view val) {
+   action_trace& trace = get_current_action_trace();
+   trace.console += val;
+}
+
+void apply_context::store_console_marker() {
+   action_trace& trace = get_current_action_trace();
+   // Mark the starting point of upcoming sync call's console log
+   // when constructing coonsole log hierarchy in pretty printing
```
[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/1333/commits/439064519374d09f99c7946a88b792759b248724

libraries/chain/host_context.cpp
```diff
@@ -98,7 +98,15 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
             // use a new sync_call_context for next sync call
             sync_call_context call_ctx(control, trx_context, ordinal, get_root_action_trace(), get_sync_call_sender(), call_receiver, receiver_account->is_privileged(), depth, flags, data);
 
+            if (control.contracts_console()) {
```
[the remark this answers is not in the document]
linh2931: Doing here requires the `if` check in only one place. But this seems implementation details, it probably belongs to `store_console_marker()`. Will make the change.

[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/1333/commits/af07be839817c0aa08f6c0f7a77c10fa3902bc76

libraries/chain/include/eosio/chain/abi_serializer.hpp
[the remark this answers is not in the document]
arhag: I think it is fine as is.

I would rather not change the existing `console` field to not break any consumers that currently expect it as a string.

heifner (approved):

PR #1341, opened 2025-04-09
[1.1.4] Test fix: Wait an extra block for protocol activation

### LINKED ISSUE #1247: Test failure: plugin_http_api_test

heifner:
Protocol activation happens in the next block start_block, wait an extra block so that activation can happen. The test was failing because even though the transaction with the activate was found, the protocol feature had not actually been activated yet. Activation happens in the next start_block. Wait another block so that the protocol feature is activated.

Resolves #1247

### DIFF
--- a/tests/TestHarness/transactions.py
+++ b/tests/TestHarness/transactions.py
@@ -395,6 +395,8 @@ class Transactions(NodeosQueries):
                     Utils.Print("ERROR: Failed to preactive digest {}".format(digest))
                     return None
             self.waitForTransactionInBlock(trans['transaction_id'])
+            # protocol features are activated in the next start_block, wait one more block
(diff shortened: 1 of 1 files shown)


heifner:
> I take it this doesn't need to be more intelligent for the multi-producer case?

The test that keeps failing is single producer. Many of our tests do not wait on finality and are subject to fork issues. At some point maybe we should attempt a universal fix for that by using the new retry of send_transation2. For now, I think we just fix them as they come up.

linh2931 (approved):
Good find!

greg7mdp (approved):

Issue #1254, opened 2025-03-13
SC: Investigate removing thread_locals in eos_vm_runtime


linh2931:
`eos_vm_runtime` uses thread_locals for backend and execution context. It appears they can be replaced as on stack variables, as is being done for sync calls.

Thread local `wasm_allocator` in  controller should be revisited too.

Need more testing.

PR #1346, opened 2025-04-09
SC: A new approach to managing Wasm allocators in controller

linh2931:
Dynamically creating wasm allocators is expensive; we don't want to create an allocator for each level of a call. Currently a thread_local queue with a fixed size (`max_sync_call_depth`) is pre-created on each thread.  But `max_sync_call_depth` can be changed by BPs, which requires dynamically resizing the queue on each thread. Solutions seem to be complicated if keeping using the thread locals.

This PR implements a new approach to managing Wasm allocators,  using a pool with an internal Boost lockfree stack.

The advantages are
 1.  Lockfree for acquiring and releasing allocators.
 2. When `max_sync_call_depth` changes, the pool changes right away. No need to check if `max_sync_call_depth` has changed every time.
3. Avoid the use of thread_locals.
4. Simpler to reason.

Thanks to @spoonincode for suggesting looking into Boost lockfree data structures.
 
Note: It seems the wasm allocator implementation prior to sync call can be removed. But that's better to be done in https://github.com/AntelopeIO/spring/issues/1254

Resolves https://github.com/AntelopeIO/spring/issues/1269

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1014,10 +1014,8 @@ struct controller_impl {
    peer_keys_db_t                  peer_keys_db;
 
    thread_local static platform_timer timer; // a copy for main thread and each read-only thread
-#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
    thread_local static vm::wasm_allocator wasm_alloc; // a copy for main thread and each read-only thread
    wasm_alloc_pool wasm_allocator_pool;
-#endif
    wasm_interface  wasmif;
    app_window_type app_window = app_window_type::write;
 
@@ -5087,9 +5085,7 @@ struct controller_impl {
 }; /// controller_impl
 
 thread_local platform_timer controller_impl::timer;
-#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
 thread_local eosio::vm::wasm_allocator controller_impl::wasm_alloc;
-#endif
 
 const resource_limits_manager&   controller::get_resource_limits_manager()const
 {
@@ -6001,7 +5997,6 @@ uint32_t controller::earliest_available_block_num() const{
    return my->earliest_available_block_num();
 }
 
-#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
 vm::wasm_allocator& controller::get_wasm_allocator() {
    return my->wasm_alloc;
 }
@@ -6021,7 +6016,6 @@ void controller::set_wasm_alloc_pool_num_threads(uint32_t num_threads) {
 void controller::set_wasm_alloc_pool_max_call_depth(uint32_t depth) {
    my->wasm_allocator_pool.set_max_call_depth(depth);
 }
-#endif
 
 #ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
 bool controller::is_eos_vm_oc_enabled() const {
@@ -6358,10 +6352,8 @@ void controller_impl::on_activation<builtin_protocol_feature_t::sync_call>() {
       add_intrinsic_to_whitelist( ps.whitelisted_intrinsics, "set_call_return_value" );
    } );
 
-#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
    auto max_call_depth = db.get<global_property_object>().configuration.max_sync_call_depth;
    wasm_allocator_pool.set_max_call_depth(max_call_depth);
-#endif
 }
 
 /// End of protocol feature activation handlers
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -6001,11 +6001,11 @@ vm::wasm_allocator& controller::get_wasm_allocator() {
    return my->wasm_alloc;
 }
 
-vm::wasm_allocator* controller::acquire_sync_call_wasm_allocator() {
+std::shared_ptr<vm::wasm_allocator> controller::acquire_sync_call_wasm_allocator() {
    return my->wasm_allocator_pool.acquire();
 }
 
-void controller::release_sync_call_wasm_allocator(vm::wasm_allocator* alloc) {
+void controller::release_sync_call_wasm_allocator(std::shared_ptr<vm::wasm_allocator> alloc) {
    my->wasm_allocator_pool.release(alloc);
 }
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -26,6 +26,7 @@
 #include <eosio/chain/snapshot_detail.hpp>
 #include <eosio/chain/thread_utils.hpp>
 #include <eosio/chain/platform_timer.hpp>
+#include <eosio/chain/wasm_alloc_pool.hpp>
 #include <eosio/chain/block_header_state_utils.hpp>
 #include <eosio/chain/deep_mind.hpp>
 #include <eosio/chain/finalizer.hpp>
@@ -1015,10 +1016,9 @@ struct controller_impl {
    thread_local static platform_timer timer; // a copy for main thread and each read-only thread
 #if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
    thread_local static vm::wasm_allocator wasm_alloc; // a copy for main thread and each read-only thread
-   thread_local static std::vector<vm::wasm_allocator> sync_call_wasm_alloc; // used for sync call. expensive to create one for each call
-   thread_local static uint32_t sync_call_wasm_alloc_index;
+   wasm_alloc_pool wasm_allocator_pool;
 #endif
-   wasm_interface wasmif;
+   wasm_interface  wasmif;
    app_window_type app_window = app_window_type::write;
 
    typedef pair<scope_name,action_name>                   handler_key;
@@ -1304,11 +1304,6 @@ struct controller_impl {
          elog( "Exception in vote thread pool, exiting: ${e}", ("e", e.to_detail_string()) );
          if( shutdown ) shutdown();
       } );
-
-#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
-      sync_call_wasm_alloc.resize(16); // Will change to use max_sync_call_depth of global property object in future version
-      sync_call_wasm_alloc_index = 0;
-#endif
 
       set_activation_handler<builtin_protocol_feature_t::preactivate_feature>();
       set_activation_handler<builtin_protocol_feature_t::replace_deferred>();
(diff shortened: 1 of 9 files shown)


libraries/chain/wasm_alloc_pool.cpp
```diff
@@ -0,0 +1,92 @@
...
+      temp.push_back(alloc);
+   }
+
+   // update with new values
+   num_threads         = new_num_thread;
+   max_call_depth      = new_depth;
+   auto new_pool_size  = num_threads * max_call_depth;
+
+   // recreate the internal stack with the new capacity
+   stack = std::make_unique<boost::lockfree::stack<vm::wasm_allocator*>>(new_pool_size);
```
heifner, line 77: Why recreate the stack? You are using a variable-sized stack, seems like you can just add to it.

Also as is, `resize` can't be called concurrently with `acquire()` or `release()`. That seems fine, as those should only be called when the main-thread is idle. If you use the same stack and make `num_threads` and `max_call_depth` atomic you could make it safe to be called concurrently. Since we don't have that use-case, seems fine to just add a comment about the restriction.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1631,6 +1631,12 @@ void producer_plugin_impl::plugin_startup() {
       }
 
       if (_ro_thread_pool_size > 0) {
+#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
+         // inform controller's wasm_alloc_pool of number of updated total threads
+         // (number of main plus read only threads)
+         chain.set_wasm_alloc_pool_num_threads(1 + _ro_thread_pool_size);
```
heifner, line 1637: It is a no-op to call this, seems like it would be clearer to always call it.

libraries/chain/controller.cpp
```diff
@@ -6359,6 +6357,11 @@ void controller_impl::on_activation<builtin_protocol_feature_t::sync_call>() {
       add_intrinsic_to_whitelist( ps.whitelisted_intrinsics, "get_call_data" );
       add_intrinsic_to_whitelist( ps.whitelisted_intrinsics, "set_call_return_value" );
    } );
+
+#if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
```
heifner, line 6361: Personally I think it would be cleaner to move all the `#if` to `wasm_allocator_pool` internals. Then none would be needed sprinkled through-out. `wasm_allocactor_pool` would always exist, the methods would just be no-ops with the `#if` only in `wasm_alloc_pool.cpp`.

linh2931: Thanks. Will do that.

spoonincode: I didn't mention it on a prior PR, but it seems like any ifdef on `EOSIO_EOS_VM_RUNTIME_ENABLED` is unnecessary since currently we effectively consider EOS VM a fallback that is always available. Sure, technically it may not always be the case in the future, but if it ever comes to that there will probably be significant refactoring anyways.

libraries/chain/wasm_alloc_pool.cpp
spoonincode: By doing `make_unique` you're effectively adding a potential lock anyways since the system allocator may lock. afaict there is no effective difference in lockiness by doing the `make_unique` over letting the queue allocate more memory as needed since that would use the system allocator too. So i'm not really seeing a benefit in recreating it.

heifner: It helps with lockfree because it avoids `new`. But your use of `new` negates that. I think you should not add `boost::lockfree::fixed_sized<true>` and just let it dynamically grow without the re-create.

linh2931: The Boost manual says

> boost::lockfree::fixed_sized<>, defaults to boost::lockfree::fixed_sized<false>
> Can be used to completely disable dynamic memory allocations during push in order to ensure lockfree behavior.
> If the data structure is configured as fixed-sized, the internal nodes are stored inside an array and they are addressed by array indexing. This limits the possible size of the stack to the number of elements that can be addressed by the index type (usually 2**16-2), but on platforms that lack double-width compare-and-exchange instructions, this is the best way to achieve lock-freedom.

libraries/chain/controller.cpp
linh2931: Is `EOSIO_EOS_VM_JIT_RUNTIME_ENABLED` still necessary?

libraries/chain/wasm_alloc_pool.cpp
spoonincode: afaik As long as you `reserve()` the appropriate amount of space, and if you use `bounded_push()` the push will always be lock free. I think even `push()` will still be lockfree as long as you're still not exceeding what was reserved.

libraries/chain/controller.cpp
spoonincode: In this case no since it's `||`, but in general yes because we don't have EOS VM JIT on, for example, ARM

libraries/chain/wasm_alloc_pool.cpp
linh2931: Thank you both!

I removed the recreation. `resize()` is much simpler.
https://github.com/AntelopeIO/spring/pull/1346/commits/5dffde4ff4a811e7988510217f3d2ee43653e472

libraries/chain/include/eosio/chain/wasm_alloc_pool.hpp
```diff
@@ -0,0 +1,52 @@
...
+   // called on main thread from sync_call protocol feature activation or set_packed_parameters
+   void set_max_call_depth(uint32_t depth);
+
+private:
+   void resize(uint32_t new_num_thread, uint32_t new_depth);
+
+   uint32_t num_threads    = 1; // `1` for the main thread
+   uint32_t max_call_depth = 1; // prior to sync call protocol feature activated
+
+   std::unique_ptr<boost::lockfree::stack<vm::wasm_allocator*>> stack;
```
heifner, line 49: This doesn't have to be a unique_ptr anymore.

heifner: You have?
```
private:
  boost::lockfree::stack<vm::wasm_allocator*> stack{config::default_max_sync_call_depth}
```

linh2931: Thanks! I was using `(config::default_max_sync_call_depth)`. With your `{...}`, it worked.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -268,6 +269,12 @@ namespace eosio { namespace chain { namespace webassembly {
          [&]( auto& gprops ) {
               gprops.configuration = config_range.config;
       });
+
+      // only update wasm_allocator pool size when new max_sync_call_depth
+      // is greater than existing one
+      if (config_range.config.max_sync_call_depth > old_max_sync_call_depth) {
```
spoonincode, line 275: Since the `wasm_alloc_pool` already ignores shrinkage, it may be worthwhile considering avoiding this check here and thus the need to track `old_max_sync_call_depth`, to just keep everything simple. Your choice; just seems unnecessary to check in two places.

libraries/chain/include/eosio/chain/wasm_alloc_pool.hpp
```diff
@@ -0,0 +1,53 @@
...
+   // called on main thread from producer_plugin startup after number of read-only threads is determined
+   void set_num_threads(uint32_t num_threads);
+
+   // called on main thread from sync_call protocol feature activation or set_packed_parameters
+   void set_max_call_depth(uint32_t depth);
+
+private:
+   void resize(uint32_t new_num_thread, uint32_t new_depth);
+
+   uint32_t num_threads    = 1; // `1` for the main thread
```
spoonincode, line 47: I think it's really unfortunate we can't make this const and populated in the ctor. But the problem seems to be that chain_plugin's initialize (which is where controller gets created) occurs before producer_plugin's initialize (where read only thread count is determined). The only way to sidestep this ordering dilemma is to have chain_plugin dip its fingers in to config options of another (producer) plugin -- something it can technically do but is bad form and we should not do that. Oh well maybe one day we can something about this producer+chain plugin situation.

spoonincode: bummer :disappointed: I could only remember one instance off hand but didn't look

libraries/chain/webassembly/privileged.cpp
linh2931: I wanted to do the check as early as possible. But the simpler, the better.

https://github.com/AntelopeIO/spring/pull/1346/commits/e3007096723155eeb5bc265d7d20dc4ed4ff3e7c

libraries/chain/include/eosio/chain/wasm_alloc_pool.hpp
linh2931: If we make chain_plugin to look at producer_plugin's number of read-only threads config option, we also need to do all the validation performed by producer_plugin. We cannot simply use the value in the command line.

libraries/chain/wasm_alloc_pool.cpp
```diff
@@ -8,31 +8,24 @@ wasm_alloc_pool::wasm_alloc_pool()
...
 {
    // create 1 wasm allocator for the main thread
-   stack.push(new vm::wasm_allocator);
-}
-
-wasm_alloc_pool::~wasm_alloc_pool() {
-   vm::wasm_allocator* alloc;
-   while (stack.pop(alloc)) {
-      delete alloc;
-   }
```
heifner: I don't think the move to `shared_ptr` is better. This adds overhead simply to remove this destructor. Seems like `unique_ptr` would be a better fit.

linh2931: `unique_ptr` is not allowed. I got 

```
.../libraries/boost/libs/lockfree/include/boost/lockfree/stack.hpp: In instantiation of ‘class boost::lockfree::stack<std::unique_ptr<eosio::vm::wasm_allocator> >’:
.../libraries/chain/include/eosio/chain/wasm_alloc_pool.hpp:49:106:   required from here
.../libraries/boost/libs/lockfree/include/boost/lockfree/stack.hpp:73:58: error: static assertion failed: boost::is_copy_constructible<T>::value
   73 |     BOOST_STATIC_ASSERT(boost::is_copy_constructible<T>::value);
```

heifner: Then the naked pointer approach is likely better. Not sure we want to pay for the `shared_ptr`. Also the use of `shared_ptr` implies these are shared, but you explicitly do not share them.

spoonincode: > I can think of an impl that works for `aquire()` but not one that works with `release()` because `push()` seems to always make a copy?

oh heh, that doesn't matter, since now that I look at the error there is a static assert that says just that

plugins/producer_plugin/producer_plugin.cpp
heifner: Looks like you missed this. Not a big deal.

linh2931: It was removed by https://github.com/AntelopeIO/spring/pull/1346/commits/267018ee044dbe52953c000fdad89723ecb567f5 (at the bottom of that commit)

heifner: I meant that I think it would be cleaner to call even if `_ro_thread_pool_size == 0`

linh2931: Got it. Will do it in https://github.com/AntelopeIO/spring/pull/1378. Thanks.

libraries/chain/wasm_alloc_pool.cpp
linh2931: Done by https://github.com/AntelopeIO/spring/pull/1378/commits/cc7f17bc0c691d7635c6da566e46fb5caea9b5b7

PR #1357, opened 2025-04-11
no need to use `std::packaged_task`; asio supports returning a `std::future` directly

spoonincode:


libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -318,10 +319,8 @@ namespace eosio { namespace chain {
 
    // async on io_context and return future
    template<typename F>
-   auto post_async_task( boost::asio::io_context& ioc, F&& f ) {
-      auto task = std::make_shared<std::packaged_task<decltype( f() )()>>( std::forward<F>( f ) );
-      boost::asio::post( ioc, [task]() { (*task)(); } );
-      return task->get_future();
+   std::future<std::invoke_result_t<std::decay_t<F>>> post_async_task( boost::asio::io_context& ioc, F&& f ) {
```
greg7mdp, line 322: Did you specify the return type of `post_async_task` for clarity (as opposed to `auto`)?

greg7mdp (approved):

libraries/chain/include/eosio/chain/thread_utils.hpp
spoonincode: I did. I don't mind changing back if desired.

greg7mdp: It is a little verbose, but at least it makes the return value crystal clear. I say keep as is.

You could also use: 

`return boost::asio::post( ioc, std::packaged_task(std::forward<F>(f)) );`

which seems a bit clearer to me but again either way is fine.

spoonincode: `use_future` is more idiomatic. It's well documented and used in many of asio's examples. Whereas being able to pass a `packaged_task` is only mentioned deep in the reference docs and never used in examples.

PR #1361, opened 2025-04-11
[1.1.4] Correctly identify snapshot as forked out

### LINKED ISSUE #1355: Pending snapshot incorrectly identified as irreversible
Test failure: disaster_recovery

`release/1.1` branch

https://github.com/AntelopeIO/spring/actions/runs/14408005992/job/40409842148

node_00 before shutdown
```
info  2025-04-11T17:01:48.019 nodeos    controller.cpp:4523           operator()           ] successfully switched fork to new head 000000885c00cdbd3aebc163f40cec1a9e35663b036e5fa801a69274627d4ba1, removed {136:00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b}, applied {136:000000885c00cdbd3aebc163f40cec1a9e35663b036e5fa801a69274627d4ba1}
```

node_00 snapshot block didn't become irreversible. Returned block id of non-irreversible block 136.
```
info  2025-04-11T17:01:48.965 nodeos    pending_snapshot.hpp:54       finalize             ] Snapshot created at block 136 available at /__w/spring/spring/build/TestLogs/disaster_recovery9/node_00/snapshots/snapshot-00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b.bin
debug 2025-04-11T17:01:48.981 http-1    beast_http_session.hpp:481    send_response        ] Response: localhost:8888 HTTP/1.1 201 Created  Connection: close  Server: nodeos/v1.1.3  Content-Type: application/json  Content-Length: 328    {"head_block_id":"00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b","head_block_num":136,"head_block_time":"2025-04-11T17:01:47.500","version":8,"snapshot_name":"/__w/spring/spring/build/TestLogs/disaster_recovery9/node_00/snapshots/snapshot-00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b.bin"}
```

node_00
```
info  2025-04-11T17:02:01.045 nodeos    controller.cpp:1923           startup              ] Starting initialization from snapshot and block log 1-146, this may take a significant amount of time
info  2025-04-11T17:02:01.056 nodeos    controller.cpp:1935           startup              ] Snapshot loaded, head: 136 : 00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b
info  2025-04-11T17:02:01.057 nodeos    controller.cpp:1697           replay_block_log     ] existing block log, attempting to replay from 137 to 146 blocks
warn  2025-04-11T17:02:01.057 nodeos    controller.cpp:4371           replay_irreversible_ ] 3030001 unlinkable_block_exception: Unlinkable block
previous mismatch 000000885c00cdbd3aebc163f40cec1a9e35663b036e5fa801a69274627d4ba1 != 00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b
    {"p":"000000885c00cdbd3aebc163f40cec1a9e35663b036e5fa801a69274627d4ba1","id":"00000088d88fa3fcd93a0bf548545b6db55e2e1a785ea46827b0fbb1f7c4213b"}
    nodeos  block_header_state.cpp:494 next
```

heifner:
Snapshot was verifying that a block was irreversible by `chain.fetch_block_by_id(block_id)`. However this was checked on the `irreversible_block` signal which is signaled before forks are pruned from the fork database. Therefore, it was possible to find a non-irreversible block for a block_id that was non-irreversible but that had a block_num <= LIB.

Not sure it is worth the complexity to support lib_num < block_num, but kept the existing functionality. Looks like this really only handles the rare cases of starting from a snapshot and creating a snapshot on HEAD. Or switching from IRREVERSIBLE to HEAD mode with a snapshot scheduled on an irreversible block that becomes irreversible. Even in those cases this could still potentially fail if running without a block log.

The snapshot scheduler previously would not generate a snapshot for a forked out block. For example, if a snapshot was scheduled for block 7, but 7 was forked out, then it would not generate a snapshot on the canonical 7 block. With this PR, the scheduler will generate two snapshots for block 7 and throw the first one away.

Resolves #1355

### DIFF
--- a/libraries/chain/include/eosio/chain/pending_snapshot.hpp
+++ b/libraries/chain/include/eosio/chain/pending_snapshot.hpp
@@ -14,8 +14,8 @@ class pending_snapshot {
 public:
    using next_t = eosio::chain::next_function<T>;
 
-   pending_snapshot(const chain::block_id_type& block_id, const next_t& next, std::string pending_path, std::string final_path)
-       : block_id(block_id), next(next), pending_path(std::move(pending_path)), final_path(std::move(final_path)) {}
+   pending_snapshot(const chain::block_id_type& block_id, block_timestamp_type timestamp, const next_t& next, std::string pending_path, std::string final_path)
+       : block_id(block_id), timestamp(timestamp), next(next), pending_path(std::move(pending_path)), final_path(std::move(final_path)) {}
 
    uint32_t get_height() const {
       return chain::block_header::num_from_id(block_id);
@@ -33,36 +33,44 @@ public:
       return snapshots_dir / fc::format_string(".incomplete-snapshot-${id}.bin", fc::mutable_variant_object()("id", block_id));
    }
 
+   // call only with lib_id that is irreversible
    T finalize(const block_id_type& lib_id, const chain::controller& chain) const {
       auto lib_num = chain::block_header::num_from_id(lib_id);
-      auto block_num = chain::block_header::num_from_id(block_id);
+      auto block_num = get_height();
 
       assert(lib_num >= block_num);
 
-      // finalize called before forkdb is pruned of non-irreversible blocks, so this can find a non-irreversible block
-      auto block_ptr = chain.fetch_block_by_id(block_id);
-      auto in_chain = !!block_ptr && (lib_num > block_num || lib_id == block_id);
-      std::error_code ec;
+      bool valid = lib_id == block_id;
+      if (!valid) {
+         // Could attempt to look up the block_id, but not finding it doesn't necessarily mean it is not
+         // irreversible. Might be running without a block log or might have been loaded via a snapshot.
+         // Also, finalize called before forkdb is pruned of non-irreversible blocks, finding it doesn't necessarily
+         // mean it is irreversible.
+         if (lib_num > block_num) { // assume it is irreversible since unable to 100% determine
+            valid = true;
+         }
+      }
 
-      if(!in_chain) {
+      std::error_code ec;
+      if(!valid) {
          fs::remove(fs::path(pending_path), ec);
          ilog("Snapshot created at block id ${id} invalidated because block was forked out", ("id", block_id));
          EOS_THROW(chain::snapshot_finalization_exception,
-                   "Snapshotted block was forked out of the chain.  ID: ${block_id}",
-                   ("block_id", block_id));
+                   "Snapshotted block was forked out of the chain.  ID: ${id}", ("id", block_id));
       }
 
       fs::rename(fs::path(pending_path), fs::path(final_path), ec);
       EOS_ASSERT(!ec, chain::snapshot_finalization_exception,
                  "Unable to finalize valid snapshot of block number ${bn}: [code: ${ec}] ${message}",
-                 ("bn", get_height())("ec", ec.value())("message", ec.message()));
+                 ("bn", block_num)("ec", ec.value())("message", ec.message()));
 
-      ilog("Snapshot created at block ${bn} available at ${fn}", ("bn", block_ptr->block_num())("fn", final_path));
+      ilog("Snapshot created at block ${bn} available at ${fn}", ("bn", block_num)("fn", final_path));
 
-      return {block_id, block_ptr->block_num(), block_ptr->timestamp, chain::chain_snapshot_header::current_version, final_path};
+      return {block_id, block_num, timestamp, chain::chain_snapshot_header::current_version, final_path};
    }
 
    chain::block_id_type block_id;
+   block_timestamp_type timestamp;
    next_t next;
    std::string pending_path;
    std::string final_path;
--- a/libraries/chain/include/eosio/chain/pending_snapshot.hpp
+++ b/libraries/chain/include/eosio/chain/pending_snapshot.hpp
@@ -37,8 +37,7 @@ public:
       auto lib_num = chain::block_header::num_from_id(lib_id);
       auto block_num = chain::block_header::num_from_id(block_id);
 
-      EOS_ASSERT(lib_num >= block_num, chain::snapshot_finalization_exception,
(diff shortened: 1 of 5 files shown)


greg7mdp (approved):

libraries/chain/include/eosio/chain/pending_snapshot.hpp
```diff
@@ -33,9 +33,16 @@ class pending_snapshot {
...
+      auto lib_num = chain::block_header::num_from_id(lib_id);
+      auto block_num = chain::block_header::num_from_id(block_id);
+
+      EOS_ASSERT(lib_num >= block_num, chain::snapshot_finalization_exception,
+                 "finalize called for non-irreversible block ${bn}:${bid}", ("bn", block_num)("bid", block_id));
+
+      // finalize called before forkdb is pruned of non-irreversible blocks, so this can find a non-irreversible block
       auto block_ptr = chain.fetch_block_by_id(block_id);
-      auto in_chain = (bool) block_ptr;
+      auto in_chain = !!block_ptr && (lib_num > block_num || lib_id == block_id);
```
spoonincode, line 45: I'm not immediately understanding the rationale for this check and the assert above. What conditions mean that a simple `bool in_chain = lib_id == block_id` isn't sufficient? I guess tihs is what your second paragraph is getting at.

heifner: My thoughts exactly. The more I think about this, I think I should simplify this instead of leaving this "dead" code.

spoonincode: I wonder if there might be some corner case where where you start from a snapshot in HEAD mode and request a snapshot before receiving any new blocks. Maybe you'll never get on on_reverisble for that initial block since it was already irreversible?

heifner: Hmm, that does seem possible. I think we could just EOS_ASSERT on that with a message that LIB not signaled. You should have a snapshot for that HEAD since that is what you started with to begin with.  Might be possible to miss it as well by switching from IRREVERSIBLE to HEAD and the producer plugin on active when the blocks are applied. Again an EOS_ASSERT to the caller that LIB was missed seems like an ok fall-back for those cases.

spoonincode: In that case maybe instead of the `EOS_ASSERT` here, the logic that iterates through the pending snapshots could be modified,
https://github.com/AntelopeIO/spring/blob/17b468887e2db1301bf3892ba1caa72e21f2ebc1/libraries/chain/snapshot_scheduler.cpp#L52
It's the `<=` that causes this weirdness.

So in this loop we can then check if it's `<` (only will occur for these startup quirks afaict so far) then we can skip calling `finialize()` altogether (maybe still log an error :woman_shrugging:), and then if it's `==` the logic becomes a simple `bool in_chain = lib_id == block_id` I think?

Just an idea.

spoonincode: oh we can't do that since it's finalize that removes the pending snapshot. womp womp nevermind

heifner: For the cases outlined, I guess best to just leave it as is.

greg7mdp (approved):

libraries/chain/include/eosio/chain/pending_snapshot.hpp
```diff
@@ -33,9 +33,16 @@ class pending_snapshot {
       return snapshots_dir / fc::format_string(".incomplete-snapshot-${id}.bin", fc::mutable_variant_object()("id", block_id));
    }
 
-   T finalize(const chain::controller& chain) const {
+   T finalize(const block_id_type& lib_id, const chain::controller& chain) const {
+      auto lib_num = chain::block_header::num_from_id(lib_id);
+      auto block_num = chain::block_header::num_from_id(block_id);
+
+      EOS_ASSERT(lib_num >= block_num, chain::snapshot_finalization_exception,
```
spoonincode, line 40: How does this ever occur -- why not assert()?

heifner: It could be an assert() I believe. I'll update.

libraries/chain/include/eosio/chain/pending_snapshot.hpp
```diff
@@ -33,9 +33,15 @@ class pending_snapshot {
...
 
-   T finalize(const chain::controller& chain) const {
+   T finalize(const block_id_type& lib_id, const chain::controller& chain) const {
+      auto lib_num = chain::block_header::num_from_id(lib_id);
+      auto block_num = chain::block_header::num_from_id(block_id);
+
+      assert(lib_num >= block_num);
+
+      // finalize called before forkdb is pruned of non-irreversible blocks, so this can find a non-irreversible block
       auto block_ptr = chain.fetch_block_by_id(block_id);
```
spoonincode, line 43: imo it'd be nice to remove `block_ptr` since it and the new comment above are now red herrings on how this works -- the check for `block_ptr` for `in_chain` doesn't really have anything to do with _whether it's in chain_, it's just there to guard against bad usage of it later on. A more useful comment would be why `lib_num > block_num` is sufficient for being considered in chain. (what we discussed here about snapshots immediately at start)

But it's not easy to remove due to `block_ptr->timestamp` below.

heifner: I changed it to store timestamp so it doesn't need to be retrieved. Also improved the comments.

spoonincode: lgtm :shipit:

greg7mdp (approved):

PR #1374, opened 2025-04-15
Add `vaulta_genesis` file.

### LINKED ISSUE #1339: Add Mainnet genesis.json to repo
Would make it easy for users to find the correct one.

Currently that is served I think by https://github.com/CryptoLions/EOS-MainNet

greg7mdp:
Resolves #1339.

File copied from https://github.com/CryptoLions/EOS-MainNet.


genesis/vaulta_chain_id.txt
```diff
@@ -0,0 +1,3 @@
+{
+ "chain_id": "aca376f206b8fc25a6ed44dbdc66547c36c6c33e3a119ffbeaef943642f0e906"
+}
```
spoonincode, line 3: Since these files have json in them it probably makes sense for the files to be .json instead of .txt?

greg7mdp: Sure, I'll update them. 

Personally I think it would be nicer to have 3 subdirectories: 
- `vaulta_mainnet`
-  `jungle_testnet`
- `kylin_testnet`

What do you guys think?

heifner (approved):
Seems like we may want to link to genesis/README.md from /README.md. But that can come later.

genesis/vaulta_chain_id.txt
greg7mdp: > Since these files have json in them it probably makes sense for the files to be .json instead of .txt?

https://github.com/AntelopeIO/spring/pull/1374/commits/277b08cb13e438a85f703506bb995b8da9188f1a

spoonincode (approved):

PR #1378, opened 2025-04-15
SC: Implement EOSVM-OC support for sync calls

linh2931:
This PR implements `EOSVM-OC` support for sync calls. It unifies `apply()` and `do_sync_call()` into a single `execute()`, and in turn unifying `EOSVM/EOSVM-JIT` and `EOSOC` processing, making code simpler. Considerable refactoring is performed as a result.

In OC implementation itself, most changes are around `sync_call` entry point handling.

Resolves https://github.com/AntelopeIO/spring/issues/1043.


libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
```diff
@@ -51,7 +55,11 @@ class eosvmoc_instantiated_module : public wasm_instantiated_module_interface {
 };
 
 eosvmoc_runtime::eosvmoc_runtime(const std::filesystem::path data_dir, const eosvmoc::config& eosvmoc_config, const chainbase::database& db)
-   : cc(data_dir, eosvmoc_config, db), exec(cc), mem(wasm_constraints::maximum_linear_memory/wasm_constraints::wasm_page_size) {
+   : cc(data_dir, eosvmoc_config, db) {
+   for (auto i = 0; i < 14; ++i) {
```
linh2931, line 59: This will be changed to use lockfree stack pool.

We cannot do default 16, which results in mmap failure.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
```diff
@@ -262,7 +263,17 @@ void executor::execute(const code_descriptor& code, memory& mem, apply_context&
...
+
+            if (context.is_action()) {
+               void(*apply_func)(uint64_t, uint64_t, uint64_t) = (void(*)(uint64_t, uint64_t, uint64_t))(cb->running_code_base + code.apply_offset);
+               apply_func(context.get_receiver().to_uint64_t(), context.get_action().account.to_uint64_t(), context.get_action().name.to_uint64_t());
+            } else if (code.call_offset) {
+               ilog("here");
+               int64_t(*call_func)(uint64_t, uint64_t, uint32_t) = (int64_t(*)(uint64_t, uint64_t, uint32_t))(cb->running_code_base + *code.call_offset);
+               call_func(context.get_sender().to_uint64_t(), context.get_receiver().to_uint64_t(), static_cast<uint32_t>(static_cast<eosio::chain::sync_call_context&>(context).data.size()));
+            } else {
+               status = eosio::chain::execution_status::receiver_not_support_sync_call;
```
linh2931, line 275: How to check `sync_call` entry point signature?

spoonincode: imo we ought to not need to. The validity of the sync_call export doesn't need to be tracked and checked by each individual wasm runtime in a bespoke way -- we ought to be able to that in a single common location.

Unfortunately it seems we cannot use the instantiation cache as is, since OC can bypass that.

We could use the code index but perhaps that is considered improper. (that said, the code index already tracks non-protocol aspects of code like a reference count!)

I don't have other ideas at the moment.

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
```diff
@@ -51,7 +62,11 @@ class eosvmoc_instantiated_module : public wasm_instantiated_module_interface {
 };
 
 eosvmoc_runtime::eosvmoc_runtime(const std::filesystem::path data_dir, const eosvmoc::config& eosvmoc_config, const chainbase::database& db)
-   : cc(data_dir, eosvmoc_config, db), exec(cc), mem(wasm_constraints::maximum_linear_memory/wasm_constraints::wasm_page_size) {
+   : cc(data_dir, eosvmoc_config, db) {
+   for (auto i = 0; i < 4; ++i) {
+      exec.emplace_back(std::make_unique<eosvmoc::executor>(cc));
```
spoonincode, line 67: Ideally every executor in a thread shares the same resources. Not really sure best approach -- an executor could expose a `clone()` call to return a new executor sharing the same resources, or maybe there needs to be more refactoring such as a class that holds executor's resources that is just passed to a non-member `execute()` call (having trouble coming up with names for these things off hand).

I don't think we need to do this executor improvement immediately though

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
```diff
@@ -62,12 +77,31 @@ std::unique_ptr<wasm_instantiated_module_interface> eosvmoc_runtime::instantiate
...
+      exec_thread_local.emplace_back(std::make_unique<eosvmoc::executor>(cc));
+      mem_thread_local.emplace_back(std::make_unique<eosvmoc::memory>(eosvmoc::memory::sliced_pages_for_ro_thread));
+   }
 }
 
-thread_local std::unique_ptr<eosvmoc::executor> eosvmoc_runtime::exec_thread_local{};
-thread_local std::unique_ptr<eosvmoc::memory> eosvmoc_runtime::mem_thread_local{};
+thread_local std::vector<std::unique_ptr<eosvmoc::executor>> eosvmoc_runtime::exec_thread_local{};
+thread_local std::vector<std::unique_ptr<eosvmoc::memory>> eosvmoc_runtime::mem_thread_local{};
+thread_local uint32_t eosvmoc_runtime::ro_thread_index = 0;
```
spoonincode, line 105: I think we need the final stack approach before considering ready for review

unittests/sync_call_tests.cpp
```diff
@@ -1295,16 +1183,23 @@ static const char direct_recursive_wast[] = R"=====(
...
+      BOOST_REQUIRE_NO_THROW(t.push_action("caller"_n, "callwithinpt"_n, "caller"_n, mvo()("input", 3)));
+      return;
+   }
+
+   if( t.get_config().wasm_runtime == wasm_interface::vm_type::eos_vm_oc ) {
+      // Do a small recursive call for now
+      BOOST_REQUIRE_NO_THROW(t.push_action("caller"_n, "callwithinpt"_n, "caller"_n, mvo()("input", 3)));
+      return;
+   }
+
```
spoonincode, line 1202: Need to get the memory situation handled before ready for full review.

Currently the main thread memory is created with full 528 slices, and the RO thread memories are created with 10 slices.

For the simplest approach to get it floating, I'd suggest maybe just making every sync call use memory with a single slice.

Ideally we do something more complex where shallow sync calls get more slices than deeper sync calls (under the premise that shallow calls are more frequent). But we'd need to work out the math on this first.

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
linh2931: Fixed by latest refactoring.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
linh2931: Created an issue https://github.com/AntelopeIO/spring/issues/893

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
linh2931: Done

unittests/sync_call_tests.cpp
linh2931: Done with a single slice for now. Created an issue https://github.com/AntelopeIO/spring/issues/1393.

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
linh2931: Created an improvement issue for this https://github.com/AntelopeIO/spring/issues/1394

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
```diff
@@ -27,21 +27,31 @@ class eosvmoc_instantiated_module : public wasm_instantiated_module_interface {
...
+            auto mem  = _eosvmoc_runtime.acquire_call_mem();
+            auto cleanup = fc::make_scoped_exit([&](){
+               _eosvmoc_runtime.release_call_exec(exec);
+               _eosvmoc_runtime.release_call_mem(mem);
+            });
+            status = exec->execute(*cd, *mem, context);
+         } else if ( is_main_thread() ) {  // action on main thread
+            status = _eosvmoc_runtime.exec.execute(*cd, _eosvmoc_runtime.mem, context);
+         }
+         else {  // action on read only thread
```
heifner, line 50: Use consistent if/else indent/layout.

heifner (changes requested):

libraries/chain/include/eosio/chain/webassembly/runtime_interface.hpp
```diff
@@ -8,15 +8,16 @@ namespace IR {
...
+enum class execution_status : int64_t;  // full definition is in wasm_interface.hpp
 
 class wasm_instantiated_module_interface {
    public:
-      virtual void apply(apply_context& context) = 0;
+      //virtual void apply(apply_context& context) = 0;
 
-      virtual sync_call_return_code do_sync_call(sync_call_context& context) = 0;
+      //virtual execution_status do_sync_call(sync_call_context& context) = 0;
+      virtual execution_status execute(host_context& context) = 0;
```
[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/1378/commits/d3bdb5e74979b9458a6999f2dbc2a192aa5038c7

libraries/chain/webassembly/runtimes/eos-vm-oc.cpp
linh2931: Thanks. https://github.com/AntelopeIO/spring/pull/1378/commits/d60ed85b80f9e2a0dd230019c76f45f9eb8bc2d3

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/eos-vm-oc.hpp
```diff
@@ -31,6 +31,7 @@ struct code_descriptor {
    size_t code_begin;
    eosvmoc_optional_offset_or_import_t start;
    unsigned apply_offset;
+   std::optional<unsigned> call_offset;
```
spoonincode, line 34: I think this might cause a problem for existing code caches that may or may not be properly handled on start up. Changing the `code_descriptor` layout probably requires indicating something higher up (like at the header) that the format has been changed and the cache must be destroyed and recreated at start up.

spoonincode: No I don't believe that will help. The serialized format of `code_descriptor` has changed, and it isn't versioned. I think you need to change
https://github.com/AntelopeIO/spring/blob/8496995f58641fb19cff38966b92d0df10314813/libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp#L21
So that the cache file is completely removed and recreated

linh2931: Thanks! I missed that.

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
```diff
@@ -18,7 +18,7 @@ namespace eosio { namespace chain { namespace eosvmoc {
 static constexpr size_t header_offset = 512u;
 static constexpr size_t header_size = 512u;
 static constexpr size_t total_header_size = header_offset + header_size;
-static constexpr uint64_t header_id = 0x32434f4d56534f45ULL; //"EOSVMOC2" little endian
+static constexpr uint64_t header_id = 0x33434f4d56534f45ULL; //"EOSVMOC2" little endian
```
spoonincode, line 21: change the comment too

linh2931: Sorry about that.

spoonincode (approved):

heifner (approved):

PR #1384, opened 2025-04-16
Update to boost 1.88

greg7mdp:


greg7mdp:
Got many warnings with clang-20: `warning: first argument in call to 'memmove' is a pointer to non-trivially copyable type` ->
created issue in boost container: https://github.com/boostorg/container/issues/300

CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 cmake_minimum_required( VERSION 3.16 )
 
+if(POLICY CMP0144)
+  cmake_policy(SET CMP0144 NEW)   # avoid cmake warning, accept both lowercase and uppercase `<PackageName>_ROOT`
+endif()
```
spoonincode, line 5: What is responsible for that warning? Something we're doing or something in boost? I wonder if doing `cmake_minimum_required( VERSION 3.16...4.0 )` is better (though that opts us in to a _lot_ more new policies, perhaps we'd not want to opt in to any policies for a version we don't test in CI, but that's no real problem since we can move the cmake in reproducible build to anywhere we'd like)

greg7mdp: No, noth8ing to do with boost, it is because we have `find_package( cdt REQUIRED )` in a couple of `CMakeLists.txt`, and because `cdt` is in lower case. I get the warning with `cmake 3.29.6`

greg7mdp: > I wonder if doing `cmake_minimum_required( VERSION 3.16...4.0 )` is better

I think we'd have to go higher than that because `CMP0144` was introduced in cmake `3.27`.

https://cmake.org/cmake/help/latest/policy/CMP0144.html

greg7mdp: Also, can we get rid of that `EOSIO_CDT` stuff?
https://github.com/AntelopeIO/spring/blob/e08e1f2a197acc88987ca17332c44916735cd883/unittests/CMakeLists.txt#L5-L18

libraries/libfc/include/fc/io/random_access_file.hpp
```diff
@@ -394,7 +407,7 @@ class random_access_file {
       std::streamsize read(char* s, std::streamsize n) {
          ssize_t total_red = 0;
          while(n - total_red) {
-            ssize_t red = ctx->read_from(boost::asio::buffer(s+total_red, n-total_red), pos);
+            ssize_t red = ctx->read_from(mutable_buffers_1(s+total_red, n-total_red), pos);
```
spoonincode, line 410: No need to change this line, the problem can be solved above in 1 line,
```diff
--- a/libraries/libfc/include/fc/io/random_access_file.hpp
+++ b/libraries/libfc/include/fc/io/random_access_file.hpp
@@ -95,7 +95,7 @@ struct random_access_file_context {
    ssize_t read_from(const MutableBufferSequence& mbs, ssize_t offs) {
       struct iovec iov[IOV_MAX];
       int i = 0;
-      for(const boost::asio::mutable_buffer& b : mbs) {
+      for(const auto& b : boost::beast::buffers_range(mbs)) {
          iov[i].iov_base = b.data();
          iov[i].iov_len = b.size();
          if(++i == IOV_MAX)
```

CMakeLists.txt
spoonincode: > > I wonder if doing `cmake_minimum_required( VERSION 3.16...4.0 )` is better
> 
> I think we'd have to go higher than that because `CMP0144` was introduced in cmake `3.27`.
> 
> https://cmake.org/cmake/help/latest/policy/CMP0144.html

When you do `<min>...<max>` it enables all policies up to `max`. So on 3.27+ it would enable CMP0144 by default (and a bunch of others).

libraries/libfc/include/fc/io/random_access_file.hpp
greg7mdp: Thanks! Even better!

greg7mdp: https://github.com/AntelopeIO/spring/pull/1384/commits/f12b0a19ebd92589e30447dfddd9c53fb94c0f20

CMakeLists.txt
greg7mdp: > When you do `<min>...<max>` it enables all policies up to `max`. So on 3.27+ it would enable CMP0144 by default (and a bunch of others).

Sounds good to me. I'll do that!

spoonincode:
LGTM

unittests/misc_tests.cpp
```diff
@@ -1156,7 +1156,7 @@ BOOST_AUTO_TEST_CASE(stable_priority_queue_test) {
 
      appbase::execution_priority_queue pri_queue;
      auto io_serv = std::make_shared<boost::asio::io_context>();
-     auto work_ptr = std::make_unique<boost::asio::io_context::work>(*io_serv);
+     auto work_ptr = boost::asio::make_work_guard(*io_serv);
```
heifner, line 1159: Nit: not a `_ptr` type anymore.

greg7mdp: https://github.com/AntelopeIO/spring/pull/1384/commits/fa833316518b9de43c5a257803bd67be788b2a35

spoonincode:
Need to remove Ubuntu 20 stuff from documentation. At least these places
https://github.com/AntelopeIO/spring/blob/29f5fd2c38e0800f44c53e193843d777ef620c4b/README.md?plain=1#L17
https://github.com/AntelopeIO/spring/blob/29f5fd2c38e0800f44c53e193843d777ef620c4b/README.md?plain=1#L139-L142
https://github.com/AntelopeIO/spring/blob/29f5fd2c38e0800f44c53e193843d777ef620c4b/README.md?plain=1#L149-L150
https://github.com/AntelopeIO/spring/blob/29f5fd2c38e0800f44c53e193843d777ef620c4b/docs/00_install/index.md?plain=1#L14
might be others I just did a quick scan of places I knew about

spoonincode (approved):

heifner (approved):

PR #1385, opened 2025-04-16
Log if not producing because of stale production

heifner:
Add a log of `Not producing block because stale production not enabled` which was the only cases that didn't already log when a node would produce but doesn't for some reason. Noticed this on a recent test failure when looking at the logs.

Also sets the flag that prevents the `producer_plugin` from going into an infinite wait for a block. Since you are a configured producer you should keep trying to produce.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -2059,8 +2059,10 @@ producer_plugin_impl::determine_pending_block_mode(const fc::time_point& now,
    // If the next block production opportunity is in the present or future, we're synced.
    if (!_production_enabled) {
       _pending_block_mode = pending_block_mode::speculating;
-      if (_producers.find(scheduled_producer.producer_name) != _producers.end())
+      if (_producers.find(scheduled_producer.producer_name) != _producers.end()) {
          fc_elog(_log, "Not producing block because stale production not enabled, block ${t}", ("t", block_time));
+         not_producing_when_time = true;
+      }
    } else if (_producers.find(scheduled_producer.producer_name) == _producers.end()) {
       _pending_block_mode = pending_block_mode::speculating;
    } else if (num_relevant_signatures == 0) {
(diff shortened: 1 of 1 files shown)


spoonincode:
curiously when I tried this I got
```
error 2025-04-16T21:10:57.938 nodeos    producer_plugin.cpp:2063      determine_pending_bl ] Not producing block because stale production not enabled, block 2025-04-16T21:10:58.000
error 2025-04-16T21:10:58.438 nodeos    producer_plugin.cpp:2063      determine_pending_bl ] Not producing block because stale production not enabled, block 2025-04-16T21:10:58.500
```
and then output ceased. What causes only 2 logs to be printed instead of continuing at each slot?

heifner:
> and then output ceased. What causes only 2 logs to be printed instead of continuing at each slot?

Either no longer the producer for that slot or a recent block received.

spoonincode:
Just running `nodeos -peosio --delete-all` gives me only 2 logs

heifner:
> Just running `nodeos -peosio --delete-all` gives me only 2 logs

https://github.com/AntelopeIO/spring/blob/main/plugins/producer_plugin/producer_plugin.cpp#L2116-2125

This allows two messages before it stops because it is waiting for a block.

Thanks for pointing this out. I believe the correct behavior here is to set `not_producing_when_time = true` in this case. Then you repeatably get the error message for each block.

spoonincode (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2059,6 +2059,10 @@ producer_plugin_impl::determine_pending_block_mode(const fc::time_point& now,
    // If the next block production opportunity is in the present or future, we're synced.
    if (!_production_enabled) {
       _pending_block_mode = pending_block_mode::speculating;
+      if (_producers.find(scheduled_producer.producer_name) != _producers.end()) {
+         fc_elog(_log, "Not producing block because stale production not enabled, block ${t}", ("t", block_time));
```
linh2931, line 2063: Do you want to log the producer name?

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: I thought about it, but none of the other `Not producing block because` log statements log the producer. I didn't want to change any of the existing ones and just kept this consistent. Under normal operation the BP should know their one and only one producer name.

PR #1396, opened 2025-04-17
remove C++20 ifdefs in merkle.hpp

spoonincode:
Cleanup in file I was touching anyways for future experiment; seems non-contentious enough to take regardless of future experiment's uncertainty.


heifner (approved):

greg7mdp:
iirc, I think these `ifdefs` were needed because another repo was not on c++20. cdt?

spoonincode:
hm, maybe some contract's native unit tests (could be system contracts or EVM or such) somewhere is still being built with c++17? I don't get that though... it seems like c++20 would be pervasive through chain's header files -- all of which native contract unit testing might see

greg7mdp (approved):

Issue #1124, opened 2025-01-26
on every block at least 2 threads are created and destroyed via merkle async


spoonincode:
This `std::async()`,
https://github.com/AntelopeIO/spring/blob/909603c103e81744c6898e18f48cbcd1a08e542f/libraries/chain/include/eosio/chain/merkle.hpp#L35-L48
leads to the full creation and destruction of at least 2 threads for every block. Would be nice to figure out how to use a thread pool here.

greg7mdp:
This merkle computation is already significantly faster than the one in 5.0, even when not using `std::async`. And in my tests, the `async` version is faster still, even with the overhead of starting a thread in `std::async`.
I feel we should not add code complexity unless we can show it makes a significant difference.

bhazzard:
Not a high priority, to be explored opportunistically, considering impacts to code clarity and performance on replay / sync use cases.

PR #1398, opened 2025-04-17
[half-baked] use chain thread pool for merkles to avoid creation and destruction of threads every block

spoonincode:
Some experimentation on #1124 to use chain's thread pool instead of `std::async` which creates and destroys threads. The performance improvement from using a thread pool in a small isolated test does appear significant, though even if it was minor I'd still be interested because the creation and destruction of threads like this may have other tricky side effects in a large application. Plus it impairs externally setting thread affinity (e.g. forcing the main thread to stay on fastest core for fastest contract execution).

Approximately[^1] what I see is a 50-60% performance improvement.

[^1]: Unfortunately the box I use for super accurate performance testing -- where all cores are fixed to same speed and low power mode is distabled -- is tied up in other things at the moment. So I get some variance in this test based on what cores happen to get the threads etc

<details>
<summary>Performance Test Code</summary>

```cpp
BOOST_DATA_TEST_CASE(merkle_perf_test, boost::unit_test::data::make({8u, 42u, 660u, 2222u})){
   const std::vector<digest_type> digests = create_test_digests(sample);
   incremental_merkle_tree tree;
   for (size_t j=0; j<digests.size(); ++j)
      tree.append(digests[j]);

#ifdef NEWCODE
   eosio::chain::named_thread_pool<struct merkleit> thread_pool;
   thread_pool.start(4, [](const fc::exception&){}, [](size_t){});
#endif

   unsigned long long t0 = __builtin_readcyclecounter();
   for(unsigned i=0; i < 50000; ++i) {
      BOOST_CHECK_EQUAL(calculate_merkle(std::span(digests.begin(), digests.end())
#ifdef NEWCODE
         ,thread_pool.get_executor()
#endif
      ), tree.get_root());
   }
   unsigned long long t1 = __builtin_readcyclecounter();
   printf("%4u>>%llu\n", sample, t1-t0);
}
```
</details>

Problems (besides some aspects that feel a little ugly to me):

* chain thread pool has default of 2 threads while merkle stuff needs/wants 4. If we are content with launching 4 threads for merkle stuff seems no problem just changing chain thread default to 4. But more concerning,
* The chain thread pool works on other tasks. So it's possible in the larger application the merkle code will see increased latency in completion since it will be competing with those other tasks. asio `post()` doesn't guarantee ordering so it's possible in theory for merkle work to be starved for some pathological work load.
The solution here could be we just maintain a separate merkle thread pool. Again -- if we were content with launching 4 threads currently it seems no problem to just keep 4 threads hanging around all the time.
* I wasn't entirely sure how to plumb through the thread pool in to `finality_root` & `block_header_state` just yet. At the moment, `calculate_merkle()` will default to using the `system_executor` when not given an executor to use. This will use a process wide asio managed thread pool.which is not desirable (at least, it doesn't fit our current threading pattern). Though, funny enough, the performance is good strictly with using `system_executor` since that thread pool is persistent unlike the `aync()`s

I didn't run tests. I might have broke something.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -748,12 +748,12 @@ struct building_block {
                auto [transaction_mroot, action_mroot] = std::visit(
                   overloaded{[&](digests_t& trx_receipts) {
                                 // calculate_merkle takes 3.2ms for 50,000 digests (legacy version took 11.1ms)
-                                return std::make_pair(calculate_merkle(trx_receipts),
-                                                      calculate_merkle(*action_receipts.digests_s));
+                                return std::make_pair(calculate_merkle(trx_receipts, ioc),
+                                                      calculate_merkle(*action_receipts.digests_s, ioc));
                              },
                              [&](const checksum256_type& trx_checksum) {
                                 return std::make_pair(trx_checksum,
-                                                      calculate_merkle(*action_receipts.digests_s));
+                                                      calculate_merkle(*action_receipts.digests_s, ioc));
                              }},
                   trx_mroot_or_receipt_digests());
 
@@ -3904,7 +3904,7 @@ struct controller_impl {
                ab.apply_legacy<void>([&](assembled_block::assembled_block_legacy& abl) {
                   assert(abl.action_receipt_digests_savanna);
                   const auto& digests = *abl.action_receipt_digests_savanna;
-                  bsp->action_mroot_savanna = calculate_merkle(digests);
+                  bsp->action_mroot_savanna = calculate_merkle(digests, thread_pool.get_executor());
                });
             }
             auto& ab = std::get<assembled_block>(pending->_block_stage);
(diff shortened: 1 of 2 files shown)


spoonincode:
I think it comes down to whether we find it acceptable to access a global throughout the code for this use case.

If that's fine, then I _think_ we could could do something as simple as
```cpp
struct merkle_pool {
   merkle_pool() {
      thread_pool.start(4, /*...*/);
   }
   named_thread_pool<struct merkle> thread_pool;
};

asio::io_executor::executor_type get_merkle_thread_pool() {
   static merkle_pool the_merkle_pool;
   return the_merkle_pool.get_executor().get_executor();
}
```
and then just replace the `std::async()` with
```c++
post_async_task(get_merkle_thread_pool(), [](){/*...*/});
```

Now we have a named & properly sized (unlike `system_executor`) persisted thread pool. But it's global.

spoonincode:
replaced with #1412

PR #1403, opened 2025-04-18
Optimize fork database open/read

heifner:
Optimize opening/reading of the fork database by correctly setting the `pending_savanna_lib_id`. Also don't bother looking up qc claimed block if it can't result in a higher pending savanna LIB.

### DIFF
--- a/libraries/chain/fork_database.cpp
+++ b/libraries/chain/fork_database.cpp
@@ -132,9 +132,11 @@ namespace eosio::chain {
    template<class BSP>
    void fork_database_impl<BSP>::open_impl( const char* desc, const std::filesystem::path& fork_db_file, fc::cfile_datastream& ds, validator_t& validator ) {
       bsp_t _root = std::make_shared<bs_t>();
-      fc::raw::unpack( ds, pending_savanna_lib_id );
+      block_id_type savanna_lib_id;
+      fc::raw::unpack( ds, savanna_lib_id );
       fc::raw::unpack( ds, *_root );
-      reset_root_impl( _root );
+      reset_root_impl( _root ); // resets pending_savanna_lib_id
+      set_pending_savanna_lib_id_impl( savanna_lib_id );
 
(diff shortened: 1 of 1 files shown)


libraries/chain/fork_database.cpp
```diff
@@ -249,15 +251,20 @@ namespace eosio::chain {
...
-            // will not be valid or forked out. This is safe because the block is
-            // just acting as a carrier of this info. It doesn't matter if the block
-            // is actually valid as it simply is used as a network message for this data.
-            if (auto claimed = search_on_branch_impl(n->previous(), qc_claim.block_num, include_root_t::no)) {
-               auto& latest_qc_claim__block_ref = claimed->core.get_block_reference(claimed->core.latest_qc_claim().block_num);
-               set_pending_savanna_lib_id_impl(latest_qc_claim__block_ref.block_id);
+            // it is not possible to claim a future block, skip if pending is already a higher height
+            block_num_type current_lib = block_header::num_from_id(pending_savanna_lib_id);
+            block_num_type block_num = n->block_num();
+            if (block_num > current_lib) {
```
heifner, line 257: Thinking about this more, not sure it is worth the `if` check, since this should be extremely rare outside reading in the fork database. Might be better to revert this change. Thoughts?

heifner: Isn't `qc_claim.block_num` going to be <= `n->block_num()` ?

greg7mdp: > So `block_num > current_lib` is less likely to be true. Right?

Indeed, what I wrote above. So  we'll be *more* likely to skip the work within the `if`.

heifner: What can I say? It is Friday. lol, Thanks!

greg7mdp (approved):

linh2931 (approved):

Issue #1217, opened 2025-02-27
SC: Enforce read_only


linh2931:
When `read_only` flag is set, abort the call if state is modified.

PR #1404, opened 2025-04-18
SC: Support read only sync call 

linh2931:
* Refactor to use host function precondition check to disallow state modifying host functions in read-only transactions, instead of scattering the checks in various host function implementation.
* Add `is_read_only()` to host_context class to support both read-only transactions and read-only sync calls.
* Make sure read-only request is honored throughout a transaction and all children sync calls.
* Add comprehensive tests.

Resoles https://github.com/AntelopeIO/spring/issues/1217


libraries/chain/sync_call_context.cpp
```diff
@@ -37,7 +38,8 @@ void sync_call_context::set_call_return_value(std::span<const char> rv) {
 }
 
 bool sync_call_context::is_read_only()const {
-   return flags & static_cast<uint64_t>(sync_call_flags::read_only);
+   // If caller is read-only, callee must be read only too
+   return (is_caller_read_only || has_field(flags, sync_call_flags::read_only));
```
heifner, line 42: Why is `is_caller_read_only` not enough?

heifner (changes requested):

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -426,20 +426,20 @@ REGISTER_LEGACY_CF_ONLY_HOST_FUNCTION(get_context_free_data)
 // privileged api
 REGISTER_HOST_FUNCTION(is_feature_active, privileged_check);
 REGISTER_HOST_FUNCTION(activate_feature, privileged_check);
```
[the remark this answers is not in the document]
linh2931: `activate_feature` is an empty function.

https://github.com/AntelopeIO/spring/blob/870e4de66c4717b4a62b3ed42d2dc52c6177c946/libraries/chain/webassembly/privileged.cpp#L20-L22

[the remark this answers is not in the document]
heifner: I realize that. But for consistency seems like it should be included.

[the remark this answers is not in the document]
linh2931: Thanks. I will add that. Initially I thought about it but it did not have the usual `EOS_ASSERT(!context.trx_context.is_read_only(),...` in the existing code. That's why I did not include it.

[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/1404/commits/314924efba65b7bdec0296b29dd966d1c65fa26a

heifner (approved):

libraries/chain/sync_call_context.cpp
spoonincode: ah sorry maybe it's just Friday but I'm still not understanding this reasoning. It makes sense that a read only action can only execute a read only sync call, and that a read only sync call can only execute a read only sync call. But it seems like that would be enforced at the time of dispatching the sync call, not here?

linh2931: Now all the enforcement is done by the host function wrappers.

This is for the cases that an action is read only, far away from it in the sync call sequence, say `call_n`. `call_n`'s flag is not set; this does not mean `call_n` will modify the state. We should let `call_n` run and use the host function wrapper to make sure the state is not modified during `call_n`'s execution.

linh2931: Oh, not even for far away calls. We should allow a call no matter what the flags are. Abort it only when it actually attempt to modify states using the host function mechanism.

spoonincode: That delayed validation means that a readonly sync call can `call(flags=0)` and the call can complete successfully provided the call doesn't actually use any banned host functions. Is that the behavior you want? It would mean, for example, when looking at the stack of call traces, a `read_only: true` called a `read_only: false`

linh2931: Yes, that's what I want. The `read_only` field in call traces is the `ancestor`'s flag || current call's flag. So it will be `read_only: true` in all call_traces starting from the first one has `read_only` set.

A third party contract's `sync call function A` can call another `sync call B` without specifying read-only flag. `A` should be allowed to be called by any other contracts.

linh2931: Also even if the flags says `read-only`, we still do the checks during the function execution. We can just allow a function with or without the `read-only` flag.

linh2931: Think about it more. Probably I should rename the member field `is_caller_read_only` to `is_ancestor_read_only`?

arhag: @linh2931 I think your current approach is acceptable.

arhag: I don’t think those renames are necessary. If anything, maybe it could be renamed to force_read_only which would hopefully imply that even if false it doesn’t necessarily mean it isn’t (implicitly) read-only.

spoonincode: > So it will be `read_only: true` in all call_traces starting from the first one has `read_only` set.

I don't think that's how it works now though.

arhag: I didn’t verify if it works that way or not. But assuming the call traces work the way Matt described then I’m good with the enforcement delayed until the host functions.

linh2931: I will make sure the call traces flags are consistent when the two PRs are merged together.

Thanks to both of you.

linh2931: Oh, on the current call_trace branch, the read_only function does not look at the ancestors. But on the read_only brach, it does that. When the two PRs are merged, it will work as Matt expects.

spoonincode: I don't think it works the way described because in #1333 the call trace is populated from the flags, not the `is_read_only()`:

https://github.com/AntelopeIO/spring/blob/4fd5a8c34da0b89df36c38b54caef6d1a41e7bff/libraries/chain/host_context.cpp#L47-L48

imo this is an unintuitive way to deal with this vs doing in `execute_sync_call()`
```cpp
if(is_read_only())
   EOS_ASSERT(read_only /*(from flags)*/, "only can do read only from read only");
```
But if this is how you want it, then should make an issue about this problem so it isn't forgotten when the PRs are eventually merged

spoonincode (approved):

libraries/chain/sync_call_context.cpp
linh2931: https://github.com/AntelopeIO/spring/issues/1408

linh2931:
I think to merge this PR as is. The renaming to force_read_only is to be done in https://github.com/AntelopeIO/spring/issues/1408.

Issue #1408, opened 2025-04-18
SC: Make sure read-only flags in call_traces are consistent


linh2931:
The last of https://github.com/AntelopeIO/spring/pull/1333 and https://github.com/AntelopeIO/spring/pull/1404 should make sure read-only flags in call_traces are consistent.

It should also add a test for this.

https://github.com/AntelopeIO/spring/pull/1404#discussion_r2051126112

PR #1410, opened 2025-04-20
SC: Fix read_only flag in call traces and make code cleaner in read only processing

linh2931:
* Fix read_only flag in call traces such that `read_only` flag is `true` for all subsequent call traces after the first call whose `read_only` flag is `true`
* Rename `read_only` enum to `force_read_only` and introduce `read_only` to call_context to make code cleaner
* Add more comments explaining read_only request by the user and read_only state of the call_context
* Add tests to verify `read_only` flag values are correct in call traces

Resolves https://github.com/AntelopeIO/spring/issues/1408

### DIFF
--- a/libraries/chain/include/eosio/chain/sync_call_context.hpp
+++ b/libraries/chain/include/eosio/chain/sync_call_context.hpp
@@ -36,7 +36,6 @@ public:
    const uint32_t               ordinal = 1;
    action_trace&                current_action_trace;
    const account_name           sender{};
-   const uint64_t               flags = 0;
    // `read_only` represents what the read only status of the call context.
    // It tells the executing smart contract code whether or not it is in
    // read-only mode and therefore whether or not the system will enforce that it
--- a/libraries/chain/include/eosio/chain/sync_call_context.hpp
+++ b/libraries/chain/include/eosio/chain/sync_call_context.hpp
@@ -22,7 +22,7 @@ enum class sync_call_flags : uint64_t {
 class sync_call_context : public host_context {
 public:
 
-   sync_call_context(controller& con, transaction_context& trx_ctx, uint32_t ordinal, action_trace& current_action_trace, account_name sender, account_name receiver, bool privileged, uint32_t sync_call_depth, uint64_t flags, std::span<const char> data);
+   sync_call_context(controller& con, transaction_context& trx_ctx, uint32_t ordinal, action_trace& current_action_trace, account_name sender, account_name receiver, bool privileged, uint32_t sync_call_depth, bool read_only, std::span<const char> data);
 
    uint32_t get_call_data(std::span<char> memory) const override;
    void set_call_return_value(std::span<const char> return_value) override;
--- a/libraries/chain/include/eosio/chain/sync_call_context.hpp
+++ b/libraries/chain/include/eosio/chain/sync_call_context.hpp
@@ -8,14 +8,18 @@ namespace eosio { namespace chain {
 // When a new flag is added, its enum value must be 1 bit left shift from the last flag.
 // Update all_allowed_bits to include the newly added enum value.
 enum class sync_call_flags : uint64_t {
-   read_only        = 1ull<<0,
-   all_allowed_bits = read_only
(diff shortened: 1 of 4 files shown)


libraries/chain/include/eosio/chain/sync_call_context.hpp
```diff
@@ -24,21 +28,25 @@ class sync_call_context : public host_context {
...
-   const bool             is_caller_read_only = false;
-   std::span<const char>  data{}; // includes function name, arguments, and other information
-   std::vector<char>      return_value{};
-
-   bool has_recipient(account_name account)const override;
-   void update_db_usage( const account_name& payer, int64_t delta ) override;
+   const uint32_t               ordinal = 1;
+   action_trace&                current_action_trace;
+   const account_name           sender{};
+   const uint64_t               flags = 0;
```
spoonincode, line 39: It seems like don't really even need `flags` any more in the current implementation -- until some time in the future where maybe needing more flags.

libraries/chain/include/eosio/chain/sync_call_context.hpp
```diff
@@ -8,14 +8,18 @@ namespace eosio { namespace chain {
 // When a new flag is added, its enum value must be 1 bit left shift from the last flag.
 // Update all_allowed_bits to include the newly added enum value.
 enum class sync_call_flags : uint64_t {
-   read_only        = 1ull<<0,
-   all_allowed_bits = read_only
+   // `force_read_only` is a user's directive to the system, telling the system whether
+   // the new call context created must operate in read-only mode or if it is
+   // free to operate under its most permissible mode.
```
spoonincode, line 13: imo this description is missing a key point that with the flag clear it inherits the "readonlyness" from the calling context

spoonincode (approved):

libraries/chain/include/eosio/chain/sync_call_context.hpp
linh2931: Thanks. I will reword the comments.

libraries/chain/include/eosio/chain/sync_call_context.hpp
spoonincode: No, I'm referring to `flags` in `sync_call_context`. afaict it's currently unnecessary and actually burdens your implementation a little since you mutate the passed flags,
https://github.com/AntelopeIO/spring/blob/fd03667f24ad0e3a8a2237c0c637a9d2523ad385/libraries/chain/host_context.cpp#L47-L51

I approved because I don't think you _need_ to get rid of it but I think it would slightly reduce quantity of code

linh2931: oh I see. I removed it as it is not needed, and we can add it back when needed as it is an implementation thing and won't break consensus.

libraries/chain/include/eosio/chain/sync_call_context.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1410/commits/6c799e6cd55f01a59b9bc22e2e575cd243a15dac

libraries/chain/include/eosio/chain/sync_call_context.hpp
linh2931: https://github.com/AntelopeIO/spring/pull/1410/commits/2356b8fbb63a536a5e5fca17608bbb19c4bc5007

heifner (approved):

libraries/chain/include/eosio/chain/sync_call_context.hpp
linh2931: Thanks! I was a little hesitant to remove the flags in the constructor but that makes code even simpler. Let's do it.

linh2931: https://github.com/AntelopeIO/spring/pull/1410/commits/d7ecbe54979ba3f8507a5586ceeeff2e089c0d19

PR #1411, opened 2025-04-21
[2.0.0-dev] avoid copying action when computing action digest

spoonincode:
I was doing some research in this area and noticed that `generate_action_digest()` effectively copies an action's input and output (return) data. We can avoid the copy if we have a hash `encoder` that implements checktime. But, see comment about uncertainty of needing checktime here.

### DIFF
--- a/libraries/libfc/include/fc/crypto/digest.hpp
+++ b/libraries/libfc/include/fc/crypto/digest.hpp
@@ -12,4 +12,35 @@ namespace fc {
       fc::raw::pack( enc, value );
       return enc.result();
    }
+
+   template <typename DigestType, typename F>
+   class hash_encoder_with_checktime : public DigestType::encoder {
+      public:
+         hash_encoder_with_checktime(size_t checktime_block_size, F&& checktime) : DigestType::encoder(),
+                                                                                   checktime_block_size(checktime_block_size),
+                                                                                   checktime(std::forward<F>(checktime)) {}
+
+         void write(const char* d, uint32_t dlen) {
+            while (dlen > checktime_block_size) {
+               DigestType::encoder::write(d, checktime_block_size);
+               d    += checktime_block_size;
+               dlen -= checktime_block_size;
+               checktime();
+            }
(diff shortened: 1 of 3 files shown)


libraries/libfc/include/fc/crypto/digest.hpp
```diff
@@ -12,4 +12,35 @@ namespace fc {
...
+         }
+
+      private:
+         size_t encoded_since_checktime = 0;
+         const size_t checktime_block_size = 0;
+         F checktime;
+      };
+
+   template<typename F>
+   using sha256_encoder_with_checktime = hash_encoder_with_checktime<sha256, F>;
```
spoonincode, line 45: I am not entirely convinced we need checktime for this hash. Modern CPUs from Zen1 (2017) and Rocket Lake (2021) will generally be able to handle hashing 1MB[^1] sub 500µs. Unfortunately earlier Intel CPUs, which may still be in service, will breach 1ms. However Vaulta _won't_ actually need to hash 1MB since action return size is so constrained, and pre-RKL will probably still be able to hash 512KB under 1ms.

I realize an argument is that these sizes could be increased drastically. Nothing stopping the protocol or nodeos from having 1GB of return data. But turning knobs up like that exceeds target checktime constraints (1ms) in so many other places it's not clear the pedantry here for this specific case is worth any additional complexity or overhead.

So I'd like to hear other opinions on if we can skip checktime for this hash; which means removing this entire change block.

Also, I had to send this to 2.0 because clang18 seems to trip up on the CTAD alias here; weird! So if removing all this I can send it back to 1.x easier


[^1]: Considering 512KB of action input plus 512KB of return data.

heifner: Seems reasonable to me to drop the checktime here.

greg7mdp: I'm all for removing unnecessary pedantry.

spoonincode:
replaced with #1415 to 1.x

PR #1412, opened 2025-04-21
use a global merkle thread pool to avoid creation and destruction of threads every block

spoonincode:
Like #1398 but using this approach https://github.com/AntelopeIO/spring/pull/1398#issuecomment-2818894389

### DIFF
--- a/libraries/chain/include/eosio/chain/merkle.hpp
+++ b/libraries/chain/include/eosio/chain/merkle.hpp
@@ -13,7 +13,7 @@ namespace detail {
 
 struct merkle_pool {
    merkle_pool() {
-      thread_pool.start(4, [](const fc::exception&){}, [](size_t){});
+      thread_pool.start(4, thread_pool.make_on_except_abort());
    }
    named_thread_pool<struct merkle> thread_pool;
 };
--- a/libraries/chain/include/eosio/chain/merkle.hpp
+++ b/libraries/chain/include/eosio/chain/merkle.hpp
@@ -1,4 +1,6 @@
 #pragma once
+
+#include <eosio/chain/thread_utils.hpp>
 #include <eosio/chain/types.hpp>
 #include <fc/io/raw.hpp>
 #include <bit>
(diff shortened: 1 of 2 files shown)


libraries/chain/include/eosio/chain/merkle.hpp
```diff
@@ -9,6 +11,18 @@ namespace eosio::chain {
 
 namespace detail {
 
+struct merkle_pool {
+   merkle_pool() {
+      thread_pool.start(4, [](const fc::exception&){}, [](size_t){});
```
spoonincode, line 16: idk what to do with exceptions off hand here. Need to think about it

heifner: Log the exception and `std::exit(-2);`. Exceptions should be completely unexpected. If they can happen then they should be caught and handled as a bad hash in `calculate_merkle_pow2`.

greg7mdp:
> Like #1398 but using this approach [#1398 (comment)](https://github.com/AntelopeIO/spring/pull/1398#issuecomment-2818894389)

I like this better than the original.

I feel that a benefit of having the thread pool in appbase, and our custom `async` version, is that we can use this in other algorithms, without having to have another static `thread_pool`. So I think it would be a better solution, but I guess we can always move to it when (or if) the need arises.

libraries/chain/include/eosio/chain/merkle.hpp
spoonincode: I had to check it worked like I would have guess -- and yeah, when using `use_future` exceptions thrown inside of the dispatched work do get packaged in to the future. So if you're only post()ing `use_future`s then it is completely unexpected for anything to leak out.

It might be useful for `named_thread_pool` to provide some pre-defined functions one can pick up and use. `thread_pool.start(4, named_thread_pool::abort_on_exception,...)` or something. Will just do something locally here for now. I'm not sure `exit()` is the right choice since it seems like you'd want something more aborty

heifner: `std::abort()` does make sense here.
It does seem having a `named_thread_pool::abort_on_exception` would be handy.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -201,6 +205,13 @@ namespace eosio { namespace chain {
          }
       }
 
+      on_except_t make_on_except_abort() {
+         return [tn=thread_name_base_from_typeinfo(typeid(this))](const fc::exception& e) {
+            elog("Unexpected exception in a ${n} thread, aborting: ${e}", ("n", tn)("e", e.to_detail_string()));
+            abort();
+         };
+      }
+
```
spoonincode, line 214: I wanted to make a `make_on_except_quit_appbase()` and then replace a number of usages of that across the code but libchain does not depend on appbase so I can't do that here. oh well :disappointed:

heifner (approved):

greg7mdp (approved):

PR #1415, opened 2025-04-22
avoid copying action when computing action digest

spoonincode:
I was doing some research in this area and noticed that `generate_action_digest()` effectively copies an action's input and output (return) data. This is a refactor that avoids that. Discussion on removing checktime at https://github.com/AntelopeIO/spring/pull/1411#discussion_r2051923833. Not really expecting a notable perf improvement but a cleanup nonetheless.

### DIFF
--- a/libraries/chain/include/eosio/chain/action.hpp
+++ b/libraries/chain/include/eosio/chain/action.hpp
@@ -95,36 +95,21 @@ namespace eosio { namespace chain {
       }
    };
 
-   template <typename Hasher>
-   auto generate_action_digest(Hasher&& hash, const action& act, const vector<char>& action_output) {
-      using hash_type = decltype(hash(nullptr, 0));
-      hash_type hashes[2];
-      const action_base* base = &act;
-      const auto action_base_size   = fc::raw::pack_size(*base);
-      const auto action_input_size  = fc::raw::pack_size(act.data);
-      const auto action_output_size = fc::raw::pack_size(action_output);
-      const auto rhs_size           = action_input_size + action_output_size;
-      std::vector<char> buff;
-      buff.reserve(std::max(action_base_size, rhs_size));
-      {
-         buff.resize(action_base_size);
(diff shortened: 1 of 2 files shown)


libraries/chain/include/eosio/chain/action.hpp
```diff
@@ -95,41 +95,22 @@ namespace eosio { namespace chain {
...
 
-   struct action_notice : public action {
-      account_name receiver;
-   };
+      fc::sha256::encoder enc;
+      fc::raw::pack(enc, base);
+      hashes[0] = enc.result();
+
+      enc.reset();
+      fc::raw::pack(enc, std::make_pair(act.data, action_output));
```
heifner, line 107: I think you need `std::make_pair(std::cref(act.data), std::cref(action_output));` to avoid `make_pair` creating a copy.

heifner (changes requested):

libraries/chain/include/eosio/chain/action.hpp
spoonincode: whoops yep

greg7mdp:
Just noticed that, unfortunately,  `is_trivial_array` won't be true for `Hash` members, so we won't benefit of the optimization in:

https://github.com/AntelopeIO/spring/blob/436e188524168814fab5cb524c6f0c00d859a9f5/libraries/libfc/include/fc/io/raw.hpp#L672-L676

The previous code was iterating over the hashes anyway, so this version is not any slower, but:
1. the name `is_trivial_array` doesn't seem right to me, should probably be `is_trivial_array_element`
2. even better, just use: `std::is_trivially_copyable_v<T>`

heifner (approved):

greg7mdp:
> It does seem like `is_trivially_copyable_v` would be the right choice. But it seems more appropriate to touch raw.hpp in a different PR, if you want to take that on @greg7mdp

Will do, creating an issue, thanks!

greg7mdp (approved):

Issue #1221, opened 2025-02-27
SC: Add sync_call to Deep Mine logging


linh2931:

linh2931:
@matthewdarwin, any concerns if Deep Mind log includes sync calls? Thanks.

heifner:
@fschoell

fschoell:
can you elaborate this? Not sure what a sync call is and how this would be reflected in the deep mind logs

linh2931:
A sync call is a mechanism to enable an action to call a function in another contract synchronously. After an action makes a sync call, its execution is suspended, waiting for the function's execution and results, before continuing. It is unlike the existing inline actions, which are scheduled to be executed after the calling action finishes its execution.

Currently when an inline action is scheduled, we log `"CREATION_OP INLINE ${action_id}"` in deep mind log, where `action_id` is the caller action's ID. We could do the same when a sync call is made, like `"CREATION_OP CALL ${action_id}"`

fschoell:
ok yes please log it as well

linh2931:
@fschoell As sync calls can make nested sync calls, we realized we need to log more information to help Deepmind to correlate an action/sync call to table operations. We propose the following minimum changes to deepmind logging:
* Introduce a new `execution_id`. Whenever an action starts or a sync call is made, increment `execution_id` by 1.
* Replace every current use of `action_id` (including identifying table operations) with `execution_id`.

For the following action/call example,

```
action_1
    create table
    sync_call_1
         db_store
         sync_call_2
              db_update
         db_store
    remove table
```
you would see deepmind logging as (other parameters of table operations are omitted for simplicity)
```
   CREATION_OP ROOT 0
   TBL_OP INS 0
   CREATION_OP CALL 1
   DB_OP INS 1
   CREATION_OP CALL 2
   DB_OP UPD 2
   DB_OP INS 1
   TBL_OP REM 0
```

PR #1439, opened 2025-04-23
SC: Add sync calls to deep mind log

linh2931:
As sync calls can make nested sync calls,  we need to log more information to help Deepmind to correlate an action/sync call to operations. We implement the following minimum changes to deepmind logging:
* Introduce a new `execution_id`. Whenever an action starts or a sync call is made, increment `execution_id` by 1.
* Replace every current use of `action_id`  with `execution_id`.

For the following action/call example,

```
action_1
    create table
    sync_call_1
         db_store
         sync_call_2
              db_update
         db_store
    remove table
```
you would see deepmind logging as (other parameters of table operations are omitted for simplicity)
```
   CREATION_OP ROOT 0
   TBL_OP INS 0
   CREATION_OP CALL 1
   DB_OP INS 1
   CREATION_OP CALL 2
   DB_OP UPD 2
   DB_OP INS 1
   TBL_OP REM 0
```

Please note, the `action_traces` inside `transaction_trace` in `APPLIED_TRANSACTION` has been updated to include sync call traces. `action_traces` has two new fields:
```
struct action_trace {
    ... // existing fields
   std::vector<call_trace>              call_traces;
   std::vector<fc::unsigned_int>   console_markers;   // used for pretty printing console logs
}
```

See https://github.com/AntelopeIO/spring/blob/6cf29e7bbb9392f79fa1c245a0ab87422540e7dd/libraries/chain/include/eosio/chain/trace.hpp#L22 for definition of  `call_trace`

Resolves https://github.com/AntelopeIO/spring/issues/1221


libraries/chain/deep_mind.cpp
```diff
@@ -202,6 +202,12 @@ namespace eosio::chain {
    {
       ++_action_id;
    }
+   void deep_mind_handler::on_call()
+   {
+      fc_dlog(_logger, "CREATION_OP CALL ${action_id}",
+         ("action_id", _action_id)
+      );
+   }
```
heifner, line 210: This indicates which sync calls were called in the context of the current action. It does not provide any information on the nesting of the sync calls. Nor does it allow the deep-mind consumer to relate table actions to individual sync calls. 

Seems you need a mapping of sync calls into the transaction trace data. Currently `_action_id` provides that mapping for current `${action_id}` to a corresponding action trace of the transaction trace.

This might be as simple as having an `on_end_call() { ++_action_id; }` OR maybe it is having a completely separate `_call_id` and then modifying things like `on_create_table` to include a `${call_id}`

heifner (changes requested):

spoonincode:
Why do we not have the call data & call return value & readonlyness etc? Deepmind historically has been the most exhaustive history output so not sure why it has barely anything vs what we're doing in ship & trace

libraries/chain/deep_mind.cpp
heifner: This does raise the question if table changes are desired at the level of sync calls. Seems like they would be otherwise maybe we don't add sync calls to deep-mind at all.

heifner:
> Why do we not have the call data & call return value & readonlyness etc? Deepmind historically has been the most exhaustive history output so not sure why it has barely anything vs what we're doing in ship & trace

It correlates `action_id` to the `transaction_trace` which is exported in deepmind.

libraries/chain/deep_mind.cpp
heifner: Adding sync_call level info for all table modifications will potentially require work from all users of substreams. Or substreams will need to be able to allow the user to determine they want all sync call table modifications rolled-up to the action level.

spoonincode:
o yeah should have remembered that from #1333 :+1:

libraries/chain/deep_mind.cpp
linh2931: I initially thought about using call_id to show sync call hierarchy within an action and stated that in the issue description. But then when I looked into inline action and require_recipient, I saw they only log the `action_id`. That's why I changed to log `action_id` only.

linh2931:
> Need to add to PR description the change in `transaction_trace` for sync calls. This change also affects deepmind. https://github.com/AntelopeIO/spring/blob/main/libraries/chain/deep_mind.cpp#L157-L157

Thanks. Done.

spoonincode (approved):

PR #1443, opened 2025-04-24
add missing check for MAX_NUM_ARRAY_ELEMENTS in the std::array pack/unpack. Also c++20 syntax update.

### LINKED ISSUE #1422: Update `raw.hpp` to use appropriate traits like `std::is_trivially_copyable`
Some `enable_if` conditions in `raw.hpp` use custom traits that could be advantageously replaced by standard C++ traits, for ex:

https://github.com/AntelopeIO/spring/blob/436e188524168814fab5cb524c6f0c00d859a9f5/libraries/libfc/include/fc/io/raw.hpp#L672-L676

Would not select types like hashes. In this case replace `is_trivial_array_element<T>` with `std::is_trivially_copyable_v<T>`.

greg7mdp:
Add missing check for `MAX_NUM_ARRAY_ELEMENTS` in the `std::array` versions.
Also c++20 syntax update.

Resolves #1422.

### DIFF
--- a/libraries/libfc/include/fc/io/raw.hpp
+++ b/libraries/libfc/include/fc/io/raw.hpp
@@ -672,12 +672,14 @@ namespace fc {
     template<typename Stream, typename T, std::size_t S>
     inline auto pack( Stream& s, const std::array<T, S>& value ) -> std::enable_if_t<std::is_trivially_copyable_v<T>>
     {
+       static_assert( S <= MAX_NUM_ARRAY_ELEMENTS, "number of elements in array is too large" );
        s.write((const char*)value.data(), S * sizeof(T));
     }
 
     template<typename Stream, typename T, std::size_t S>
     inline auto pack( Stream& s, const std::array<T, S>& value ) -> std::enable_if_t<!std::is_trivially_copyable_v<T>>
     {
+       static_assert( S <= MAX_NUM_ARRAY_ELEMENTS, "number of elements in array is too large" );
        for( std::size_t i = 0; i < S; ++i ) {
           fc::raw::pack( s, value[i] );
        }
@@ -686,12 +688,14 @@ namespace fc {
     template<typename Stream, typename T, std::size_t S>
     inline auto unpack( Stream& s, std::array<T, S>& value )  -> std::enable_if_t<std::is_trivially_copyable_v<T>>
     {
+       static_assert( S <= MAX_NUM_ARRAY_ELEMENTS, "number of elements in array is too large" );
        s.read((char*)value.data(), S * sizeof(T));
     }
 
     template<typename Stream, typename T, std::size_t S>
     inline auto unpack( Stream& s, std::array<T, S>& value )  -> std::enable_if_t<!std::is_trivially_copyable_v<T>>
     {
+       static_assert( S <= MAX_NUM_ARRAY_ELEMENTS, "number of elements in array is too large" );
        for( std::size_t i = 0; i < S; ++i ) {
           fc::raw::unpack( s, value[i] );
        }
--- a/libraries/libfc/include/fc/io/raw.hpp
+++ b/libraries/libfc/include/fc/io/raw.hpp
@@ -145,31 +145,31 @@ namespace fc {
        usec = fc::microseconds(usec_as_int64);
     } FC_RETHROW_EXCEPTIONS( warn, "" ) }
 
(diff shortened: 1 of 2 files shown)


heifner (approved):

spoonincode (approved):
It might be nice, since you're touching them anyways, to update these to c++20 syntax. but obviously not critical and can go as-is

greg7mdp:
> It might be nice, since you're touching them anyways, to update these to c++20 syntax. but obviously not critical and can go as-is

Hopefully this is what you were thinking of.

greg7mdp:
Upon reflection, I don't think `std::is_trivially_copyable` is appropriate.

- pointers are `std::trivially_copyable`. We wouldn't want to allow serializing an `std::array` of pointers.
- we may have a type for which FC_REFLECT was defined, which we serialize without alignment padding
  between members. But that type can also be `std::trivially_copyable`, and we wouldn't want to
  serialize it with padding when in a `std::array` I think.

heifner:
My understanding is `std::array` can have padding because it is a struct. However we don't serialize the `std::array` we serialize the internal c-array of the `std::array`. So padding is not an issue.

spoonincode:
> My understanding is `std::array` can have padding because it is a struct. However we don't serialize the `std::array` we serialize the internal c-array of the `std::array`. So padding is not an issue.

I think the problem is if someone has
```cpp
struct foo {
   uint32_t bar;
   uint64_t baz;
};
FC_REFLECT(foo, (bar)(baz));
```
This is a trivially copyable type (std::is_trivially_copyable), but has padding. So if someone makes an `array<foo, 2>` we can't just write out the bytes of the entire array (like early change in this PR would have done).

I wonder how we could detect that something like `sha256` is safe to handle in this manner.

spoonincode (approved):

greg7mdp:
> My understanding is `std::array` can have padding because it is a struct. However we don't serialize the `std::array` we serialize the internal c-array of the `std::array`. So padding is not an issue.

Hum, that's not exactly what I meant. Consider `struct key_weight`, which has these members:

https://github.com/AntelopeIO/spring/blob/25978a6c2c2b24e3b217d65af80bf33c1a5907a1/libraries/chain/include/eosio/chain/authority.hpp#L95-L97

A `std::array<key_weight>` will have space paddings between the `key_weight` values, since `weight_type` is an `uint16_t`.
But it may be `std::is_trivially_copyable`, so we would write the memory representation of the array, spaces and all.

So that would be a consensus change from what we were writing before, as `key_weight` would not have been `is_trivial_array`, and we would have serialized the values individually without the padding between values.

Same issue if the members were in the opposite order, `weight` before `key`. We'd have internal padding between the two members, which would be written if we just write the memory representation of the std::array, but not if we write the values separately using:

https://github.com/AntelopeIO/spring/blob/25978a6c2c2b24e3b217d65af80bf33c1a5907a1/libraries/chain/include/eosio/chain/authority.hpp#L362

greg7mdp:
> I wonder how we could detect that something like `sha256` is safe to handle in this manner.

I had the exact same thought! We could detect it if `sha256` used a std::array, but since it is using a c-style array (which decays as pointer)

https://github.com/AntelopeIO/spring/blob/25978a6c2c2b24e3b217d65af80bf33c1a5907a1/libraries/libfc/include/fc/crypto/sha256.hpp#L108

I don't think we can safely match it, but then I didn't quite get what this does:

https://github.com/AntelopeIO/spring/blob/25978a6c2c2b24e3b217d65af80bf33c1a5907a1/libraries/libfc/include/fc/crypto/sha256.hpp#L158

heifner:
Thanks guys. I did misunderstand the issue at hand.



> I don't think we can safely match it, but then I didn't quite get what this does:
> 
> https://github.com/AntelopeIO/spring/blob/25978a6c2c2b24e3b217d65af80bf33c1a5907a1/libraries/libfc/include/fc/crypto/sha256.hpp#L158

That just flags that the type works with reflection. It specifies the `operator<<` & `operator>>` manually instead of using `FC_REFLECT`.

heifner (approved):

greg7mdp:
> That just flags that the type works with reflection. It specifies the `operator<<` & `operator>>` manually instead of using `FC_REFLECT`.

Oh, interesting, so it does write the four `uint64_t` in one shot. nice!

PR #1444, opened 2025-04-24
[1.1.5] Fix OC scheduling of compiles

### LINKED ISSUE #1441: OC compile scheduling
Fix issue with scheduling of OC compile.

heifner:
Use an unbounded queue for `_result_queue` to avoid results geting dropped because the queue is full. 
Also simplify thread usage by moving the call of `process_queued_compiles()` to the main thread. The mutex is no longer needed as `_compile_monitor_write_socket`, `_queued_compiles`, and `_outstanding_compiles_and_poison` are now only accessed from the main thread.

Resolves #1441


heifner:
The test failure points out that the `process_queued_compiles()` needs to be off the main thread. Closing this PR and will push up a new one with an alternative approach.

PR #1447, opened 2025-04-25
[1.1.5] Fix OC scheduling of compiles

### LINKED ISSUE #1441: OC compile scheduling

heifner:
Use an unbounded queue for `_result_queue` to avoid results getting dropped because the queue is full. 
Also call `free_code` when code is no longer used instead of waiting for LIB to reduce memory growth.

Resolves #1441

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -207,19 +207,30 @@ struct eosvmoc_tier {
          return it != wasm_instantiation_cache.end();
       }
 
-      void code_block_num_last_used(const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version, const uint32_t& block_num) {
+      void code_block_num_last_used(const digest_type& code_hash, uint8_t vm_type, uint8_t vm_version,
+                                    block_num_type first_used_block_num, block_num_type block_num_last_used)
+      {
          // The caller of this method apply_eosio_setcode has asserted that
          // the transaction is not read-only, implying we are
          // in write window. Read-only threads are not running.
          // Safe to update the cache without locking.
          wasm_cache_index::iterator it = wasm_instantiation_cache.find(boost::make_tuple(code_hash, vm_type, vm_version));
-         if(it != wasm_instantiation_cache.end())
-            wasm_instantiation_cache.modify(it, [block_num](wasm_cache_entry& e) {
-               e.last_block_num_used = block_num;
-            });
+         if(it != wasm_instantiation_cache.end()) {
+            if (first_used_block_num == block_num_last_used) {
+               // First used and no longer needed in the same block, erase immediately, do not wait for LIB.
+               // Since created and destroyed in the same block, likely will not be needed in a forked block.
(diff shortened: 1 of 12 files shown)


libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
```diff
@@ -205,19 +203,20 @@ code_cache_async::get_descriptor_for_code(mode m, const digest_type& code_id, co
...
    g.lock();
-   if(_outstanding_compiles_and_poison.size() >= _threads) {
+   if(_outstanding_compiles >= _threads) {
+      std::vector<char> code{codeobject->code.begin(), codeobject->code.end()};
       if (m.high_priority)
-         _queued_compiles.emplace_front(std::move(msg), std::move(fds_to_pass));
+         _queued_compiles.emplace_front(std::move(msg), std::move(code));
       else
-         _queued_compiles.emplace_back(std::move(msg), std::move(fds_to_pass));
+         _queued_compiles.emplace_back(std::move(msg), std::move(code));
```
greg7mdp, line 213: Shall we fix the typo in the type `queued_compilies_t`?

greg7mdp (approved):

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -227,10 +232,6 @@ struct eosvmoc_tier {
          // Anything last used before or on the LIB can be evicted.
          const auto first_it = wasm_instantiation_cache.get<by_last_block_num>().begin();
          const auto last_it  = wasm_instantiation_cache.get<by_last_block_num>().upper_bound(lib);
-#ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
-         if(eosvmoc) for(auto it = first_it; it != last_it; it++)
-            eosvmoc->cc.free_code(it->code_hash, it->vm_version);
-#endif
```
spoonincode, line 233: The changes aren't what we discussed. afaict the way we discussed would be 1) for this code to remain here, and 2) for the added lines above (where this here was moved to) would exist but would check that the code was set in the same block prior to doing the removal. That `first_block_used` is currently inaccessible from there but it would be trivial to plumb it through starting from here
https://github.com/AntelopeIO/spring/blob/504fa82dca9172a36056a67dc1932e063bdb6adf/libraries/chain/eosio_contract.cpp#L156-L163

Also, I realize now the discussion might have been ambiguous w.r.t. to us making this change strictly for OC or more universally. I believe the problem we are trying to solve applies for any runtime not just OC so I suspect we should be doing this universally. Maybe this aspect needs to be discussed idk

heifner: https://github.com/AntelopeIO/spring/pull/1447/commits/2a890fcfa4701cd4dd4b2658a4ee8c7b4e755bcd https://github.com/AntelopeIO/spring/pull/1447/commits/f6ad1370d6af23e3337bc4232dba40668417d379

spoonincode (approved):

greg7mdp (approved):

PR #1449, opened 2025-04-25
Un-submodule `chainbase`.

### LINKED ISSUE #1445: Un-submodule chainbase
- will live under new directory `library/chaindb`
- transfer existing [chainbase PR](https://github.com/AntelopeIO/chainbase/pull/54) to Spring PR #1070.

greg7mdp:
Resolves #1445.

fake-merged to preserve history (so that `git blame` still works), hence the many commits.


libraries/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ add_subdirectory( libfc )
 add_subdirectory( builtins )
 add_subdirectory( softfloat )
 add_subdirectory( wasm-jit )
-add_subdirectory( chainbase )
+add_subdirectory( chaindb )
```
heifner, line 18: Need to update `CMakeModules/EosioTesterBuild.cmake.in`

libraries/chaindb/.github/workflows/build.yaml
```diff
@@ -0,0 +1,46 @@
+name: "Build & Test"
```
heifner, line 1: Do we need this anymore? Seems like these should be just part of Spring now.

libraries/chaindb/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
+# Defines ChainBase library target.
+cmake_minimum_required( VERSION 3.12 )
+project( ChainBase )
```
heifner, line 3: Don't think we want `project` anymore or `cmake_minimum_required`.

heifner: Actually quite a bit of this file should probably be deleted.

libraries/chaindb/LICENSE.md
```diff
@@ -0,0 +1,14 @@
+AntelopeIO/chainbase
```
heifner, line 1: Do we just remove this file?

libraries/CMakeLists.txt
greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/3f7cfc65b654e4866522fd20ba4b86ded66b5387

libraries/chaindb/.github/workflows/build.yaml
greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/3f7cfc65b654e4866522fd20ba4b86ded66b5387

libraries/chaindb/LICENSE.md
spoonincode: No I don't think so. We can't remove the Steemit copyright notice

greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/7cc88c2b653e9b94a9da48408b431de6b2b0e51d

libraries/chaindb/CMakeLists.txt
greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/f2fba95810892649598c7f731f6b0fc0f5543461

libraries/chaindb/LICENSE.md
heifner: I tend to agree with Matt that we should keep it. I think maybe we should put fc's back in.

spoonincode: It does feel like fc should go back it now. It might have more defensible to remove it at the time since the base repo license was the same (MIT) and the fc license just say "Respective Authors" which is pretty much implied anyways so it wasn't like we were stripping someone's name from the license.

heifner: Does that mean we have to put the Business Source License txt above the chainbase/fc MIT license text?

spoonincode: That part I'm not confident about; we should ask. In the past we did this,
https://github.com/AntelopeIO/spring/blob/main/libraries/chain/webassembly/runtimes/eos-vm-oc/About%20WAVM
And in a bubble to myself that would be the pattern I'd reach for again. But we should double check

greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/de5c92fce62e500f4c29cb07177bb8a733e1fae5

libraries/chaindb/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Defines ChainBase library target.
+project( ChainBase )
+
+file(GLOB HEADERS "include/chainbase/*.hpp")
+add_library( chainbase src/chainbase.cpp src/pinnable_mapped_file.cpp ${HEADERS} )
+target_link_libraries( chainbase PUBLIC pthread Boost::system )
```
spoonincode, line 6: Instead of `pthread`, the portable way is to `Threads::Threads`. We also (I think) consistently set up options on `Threads::Threads` with:
https://github.com/AntelopeIO/spring/blob/504fa82dca9172a36056a67dc1932e063bdb6adf/libraries/libfc/CMakeLists.txt#L6-L8
(the first one, `CMAKE_THREAD_PREFER_PTHREAD`, is no longer documented, dunno what's up with that)

libraries/chaindb/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+if(TARGET Boost::asio)
```
spoonincode, line 8: No reason to keep the if afaict?

libraries/chaindb/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>"  )
+
+enable_testing()
+add_subdirectory( test )
+add_subdirectory( benchmark )
+
+if(CHAINBASE_INSTALL_COMPONENT)
+   set(INSTALL_COMPONENT_ARGS COMPONENT ${CHAINBASE_INSTALL_COMPONENT} EXCLUDE_FROM_ALL)
+endif()
```
spoonincode, line 23: You can remove the if and setting of `INSTALL_COMPONENT_ARGS`, and then hardcode `COMPONENT dev` like we do for fc,
https://github.com/AntelopeIO/spring/blob/504fa82dca9172a36056a67dc1932e063bdb6adf/libraries/libfc/CMakeLists.txt#L111-L114
I did notice that failed to remove line 2 here when doing fc way back. If you make this change it is fine to remove line 1 and 2
https://github.com/AntelopeIO/spring/blob/504fa82dca9172a36056a67dc1932e063bdb6adf/libraries/CMakeLists.txt#L1-L2

greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/31a08eee8c7b41ebad9bdd39bfcb9e7380c277f3

libraries/chaindb/CMakeLists.txt
greg7mdp: https://github.com/AntelopeIO/spring/pull/1449/commits/9516ea20b3f2ef5d493cf6d1d09ed0a8a1aee761

spoonincode (approved):

heifner (approved):

Issue #1393, opened 2025-04-17
SC: OC -- investigate how many slices are created for sync call memory


linh2931:
https://github.com/AntelopeIO/spring/pull/1378#discussion_r2045854875

PR #1451, opened 2025-04-25
SC: Implement memory pools with variable-sized memory slices for OC

linh2931:
To help reduce the number of memory slices pre-allocated at the startup and avoid using `mprotect` for top levels of sync calls, a set of memory pools with a variable-sized memory slices are implemented.

Top call levels use dedicated memory pools with greater than one slice. Deep levels share one pool with one slice. Level 1 is allocated highest slices, then each level decrements by one, until reaching one slice for all remaining levels.
 
For example, if max call depth is 10 and the first level is allocated 5 slices, we will have 5 memory pools:
 
* pool 1: 5 slices, for calls at level 1
* pool 2: 4 slices, for calls at  level 2
* pool 3: 3 slices, for calls at  level 3
* pool 4: 2 slices, for calls at  level 4
* pool 5: 1 slice,  for calls at  levels 5 to 10

Resolves https://github.com/AntelopeIO/spring/issues/1393


heifner (approved):

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/memory_pools.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+      // Update memory pools based on new max call depth
+      void set_max_call_depth(uint32_t depth);
+
+      // Slices for the first call depth.
+      static constexpr uint32_t first_call_depth_slices = 5;
+
+   private:
+      uint32_t num_threads = 1; // main thread + read-only threads. default 1 for main thread
+      std::vector<std::unique_ptr<call_resource_pool<eosvmoc::memory>>> pools;
```
spoonincode, line 48: bummer these need to be stuffed in a unique_ptr, but I see lockfree stack isn't moveable

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/memory_pools.hpp
```diff
@@ -0,0 +1,51 @@
...
+ * */
+class memory_pools {
+   public:
+      memory_pools();
+
+      // Acquire memory from the pool of `call_depth`
+      eosvmoc::memory* acquire_mem(uint32_t call_depth);
+
+      // Release memory `m` back to the pool of `call_depth`
+      void release_mem(uint32_t call_depth, eosvmoc::memory* m);
```
spoonincode, line 35: I wonder if it makes any sense for `memory` to be queryable for its slices so you don't need to parrot back the call_depth -- could just query `memory` for its slice count and know what pool to place it back it. Not really a big deal to parrot back call_depth though

spoonincode (approved):

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/memory_pools.hpp
linh2931: Do you want me to put a `sliced_pages` member to `memory` and query it in `release_mem` so we don't need `call_depth` parameter? I think it is cleaner and safer (not put back a memory into a wrong pool by accident)

spoonincode: I don't think it's necessary since we declare the scoped_exit so closely nearby to the aquire_mem, so it's up to you

linh2931: OK. I will keep it as is.

Issue #1392, opened 2025-04-17
SC: OC sync_call entry point validation


linh2931:
https://github.com/AntelopeIO/spring/pull/1378#discussion_r2045816980

linh2931:
It was agreed at the Engineering meeting to use code index table to store a flag indicating whether a contract exports a valid `sync_call` function, hence supporting sync calls or not.

PR #1455, opened 2025-04-28
SC: Sync call entry point validation for `eosvm`, `eosvm-jit`, and `eosvmoc`

linh2931:
This PR implements a new method to validate sync call entry point. It is applicable to `eosvm`, `eosvm-jit`, and `eosvmoc`. When an account code is deployed, check whether `sync_call` entry point exists. If it does, its signature is checked. A flag `sync_call_supported` is added to `code_object`. Before a sync call is executed, `sync_call_supported` is used to determine whether the call be proceeded.

`sync_call_supported` is not written into snapshots. It is populated from the code when reading from snapshots.

Resolves https://github.com/AntelopeIO/spring/issues/1219 and https://github.com/AntelopeIO/spring/issues/1392


libraries/chain/controller.cpp
```diff
@@ -2501,6 +2501,28 @@ struct controller_impl {
...
          }
 
+         // special case for in-place upgrade of code_index
+         if (std::is_same_v<value_t, code_object>) {
+            using v1 = snapshot_code_object_v1;
+            if (std::clamp(header.version, v1::minimum_version, v1::maximum_version) == header.version ) {
+               snapshot->read_section<code_object>([&db=this->db](auto &section) {
+                  v1 obj_v1;
+                  section.read_row(obj_v1, db);
+                  bool sync_call_supported = wasm_interface::is_sync_call_supported(obj_v1.code.data(), obj_v1.code.size());
```
linh2931, line 2511: Should we just set `sync_call_supported` to `false` for old snapshots? This way, we will not "accidentally" execute an exported `sync_call` function in old code if it has valid signature, because we don't know if the code owner wants the function to be called in the new sync call sense. If the owner wants it to be used by a sync call after the sync call protocol feature is activated, he should redeploy the contract.

But on the other hand, if a function is exported, it is expected to be used in any cases.

libraries/chain/include/eosio/chain/code_object.hpp
```diff
@@ -33,8 +35,23 @@ namespace eosio { namespace chain {
...
+                    "snapshot_code_object_v1 is no longer needed");
+
+      digest_type  code_hash;
+      shared_blob  code;
+      uint64_t     code_ref_count;
+      uint32_t     first_block_used;
+      uint8_t      vm_type = 0;
+      uint8_t      vm_version = 0;
+   };
+
```
spoonincode, line 51: I'm a little confused on this new `snapshot_code_object_v1` because it's the same as the `code_object` previously was serialized. Also, when we're converting from different chainbase<->snapshot types usually (always?) we use `snapshot_row_traits` which I don't see anywhere in here.

I'm suspicious we don't need any of these snapshot changes and can instead add a `code_object::reflector_init()` which just populates the new `sync_call_supported` field. (this would mean that we'd not include `sync_call_supported` as part of the `fc_reflect` which is a little weird but not unheard of)

I'm not 100% confident on the above but I'd suggest trying it

linh2931: The new `snapshot_code_object_v1` is for snapshot versions prior to sync call protocol feature. We have similar structs for block_state and chain_config.

Will try using `code_object::reflector_init()`. Thanks.

linh2931: Thanks!

`code_object::reflector_init()` works. We don't need to place `sync_call_supported` into snapshots.

libraries/chain/controller.cpp
linh2931: Turns out we don't need to put `sync_call_supported` into snapshots. OK to re-calculate it when a snapshot is read.

spoonincode (commented):
Looks really good

libraries/chain/code_object.cpp
```diff
@@ -0,0 +1,14 @@
+#include <eosio/chain/code_object.hpp>
+#include <eosio/chain/wasm_interface.hpp>
+
+namespace eosio { namespace chain {
+
+void code_object::reflector_init()
+{
+   static_assert(fc::raw::has_feature_reflector_init_on_unpacked_reflected_types,
+                 "code_object expects FC to support reflector_init");
```
heifner, line 9: We really don't need these `static_assert`s anymore now that `libfc` is no longer a submodule. It is fine to leave it, but we should probably clean those up at some point.

libraries/chain/wasm_interface.cpp
```diff
@@ -67,12 +67,13 @@ namespace eosio { namespace chain {
...
 
-   void wasm_interface::validate(const controller& control, const bytes& code) {
+   void wasm_interface::validate(const controller& control, const bytes& code, bool& sync_call_supported) {
+      sync_call_supported = false;
       const auto& pso = control.db().get<protocol_state_object>();
 
       if (control.is_builtin_activated(builtin_protocol_feature_t::configurable_wasm_limits)) {
          const auto& gpo = control.get_global_properties();
-         webassembly::eos_vm_runtime::validate( control, code, gpo.wasm_configuration, pso.whitelisted_intrinsics );
+         webassembly::eos_vm_runtime::validate( control, code, gpo.wasm_configuration, pso.whitelisted_intrinsics, sync_call_supported );
```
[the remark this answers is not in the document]
linh2931: I thought about it. Because `sync_call` depends on all previous protocol features, `CONFIGURABLE_WASM_LIMITS2` should have been activated. That's why I did not add to the other `validate()`. But will add for completeness and for not missing some corner cases.

linh2931:
> Looks really good

Thanks for all your suggestions

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -85,7 +85,19 @@ void validate(const bytes& code, const whitelisted_intrinsics_type& intrinsics)
    }
 }
 
-void validate( const controller& control, const bytes& code, const wasm_config& cfg, const whitelisted_intrinsics_type& intrinsics ) {
+static bool module_has_valid_sync_call(module& mod) {
```
heifner: Can this take a `const module& mod` ?

libraries/chain/host_context.cpp
```diff
@@ -101,6 +101,11 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
             return handle_call_failure();
          }
 
+         const code_object* const codeobject = db.find<code_object, by_code_hash>(boost::make_tuple(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version));
```
heifner, line 104: Since you already look it up here. Seems like it would be cleaner to pass `code_object` into `execute`. Then `code_hash`, `vm_type`, and `vm_version` do not have to be passed in.

libraries/chain/wasm_interface.cpp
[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/spring/pull/1455/commits/13e5f5e1db0763efdf95bc5beba7c11d0449935d

libraries/chain/eosio_contract.cpp
```diff
@@ -138,11 +138,12 @@ void apply_eosio_setcode(apply_context& context) {
 
    fc::sha256 code_hash; /// default is the all zeros hash
 
+   bool sync_call_supported = false;
    int64_t code_size = (int64_t)act.code.size();
 
    if( code_size > 0 ) {
      code_hash = fc::sha256::hash( act.code.data(), (uint32_t)act.code.size() );
-     wasm_interface::validate(context.control, act.code);
+     wasm_interface::validate(context.control, act.code, sync_call_supported);
```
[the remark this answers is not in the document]
heifner: Returning a struct of bools (only sync_call_supported for now) would be better I think.

heifner (changes requested):

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: It cannot take a `const`, as `get_exported_function()` is not defined as a `const` method.

libraries/chain/eosio_contract.cpp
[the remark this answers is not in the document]
linh2931: Initially I thought about returning a bool, but that does not sound right as the method name is `validate`. A struct of bools is a better choice. Thanks.

libraries/chain/host_context.cpp
linh2931: `apply_context` uses the same `execute()` and it does not dip `code_object`.  Even if we pass in `code_object` here, we still need to split it and pass in `code_hash`, `vm_type` and `vm_version` before calling `wasm_interface_private::execute()` to reuse the same code.

heifner: I think `apply_context` might as well look it up and pass it in. In either case it is needed inside `wasm_interface_private::execute()`.

linh2931: It only looks `code_object` when the code is not cached. We will need two `execute()`s with different signatures to avoid unnecessary lookup for `apply_context`. Can we do this refactor in a separate PR?

heifner: Good point about `code_object` not always needing to be looked up. Maybe we should cache the `sync_call_supported` in `wasm_cache_entry` and `code_descriptor`. That would mean it would have to validated in `execute()`.

linh2931: I think the current implementation is simpler and only checking one location for truth.

spoonincode: This extra lookup that must occur on every action bothered me too. It reminds me of EOSIO/eos#9740 since it's a lookup in to an ordered set so I'm worried it'll actually be a visible performance impact for heavy sync calls. 

> Good point about `code_object` not always needing to be looked up. Maybe we should cache the `sync_call_supported` in `wasm_cache_entry` and `code_descriptor`. That would mean it would have to validated in `execute()`.

crossed my mind as well. But that seemed to be _exactly_ what we were trying to avoid by doing it this way since this way is very simple :confused:

heifner: `code_descriptor` already has `call_offset` so just need to add to `wasm_cache_entry`. I agree the current implementation is simpler, but we avoid the lookup when using the caches so seems like we should continue to avoid the lookup. 

BTW, what happens if you attempt to call the sync call and it isn't there? Do we even need need the verification? Can we just tell by how it fails?

spoonincode: > BTW, what happens if you attempt to call the sync call and it isn't there? Do we even need need the verification? Can we just tell by how it fails?

JIT check parameters on every call, so you get some sort of exception out of it. OC does not do any validation so it'll do.. whatever. For example if the WASM call export takes four parameters instead of 3 I guess we'll leak some register's value in to the VM

heifner: We already have `call_offset` for oc we can check. Seems a bit fragile to rely on the JIT check, so probably should check it. So add it to the `wasm_cache_entry`?

linh2931: In current VM JIT implementation, it will crash if `sync_call()` is not exported. In OC, we have an assert if `code.call_offset` is 0.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -106,9 +126,27 @@ void validate( const controller& control, const bytes& code, const wasm_config&
...
+   return validate_result {
+      .sync_call_supported = sync_call_supported
+   };
+}
+
+bool is_sync_call_supported(const char* code_bytes, size_t code_size) {
+   wasm_code_ptr code_ptr((uint8_t*)code_bytes, code_size);
+   bool supported = false;
+   try {
+      eos_vm_null_backend_t<setcode_options> bkend(code_ptr, code_size, nullptr);
```
spoonincode, line 144: afaict this is good now, but it's something to be mindful of in the future -- if we ever relax WASM validation via some protocol feature in the future, this could cause problems.

spoonincode (approved):

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: What problems might be? Do you mean previously non-supported becomes supported?

linh2931: Thanks for the detailed explanation. If that happens, we will have to save `sync_call_supported` in the snapshot, so that the caller of `is_sync_call_supported()` can pass in `controller` to have access to protocol feature set. I will add a comment.

PR #1459, opened 2025-04-29
Updated docs with nodeos --help output

heifner:
Seems we have missed a number of updates.

Could look into "back-porting" these as needed to `1.2.0-rc2` if desired. Seems like these should be auto-generated instead of manually updating.

### DIFF
--- a/docs/01_nodeos/03_plugins/http_plugin/index.md
+++ b/docs/01_nodeos/03_plugins/http_plugin/index.md
@@ -29,6 +29,60 @@ Config Options for eosio::http_plugin:
                                         The local IP and port to listen for
                                         incoming http connections; set blank to
                                         disable.
+  --http-category-address arg           The local IP and port to listen for
+                                        incoming http category connections.
+                                        Syntax: category,address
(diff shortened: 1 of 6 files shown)


greg7mdp (approved):

docs/01_nodeos/03_plugins/http_plugin/index.md
```diff
@@ -29,6 +29,60 @@ Config Options for eosio::http_plugin:
...
+                                        ck
+                                            trace_api,:8086 # listen on all
+                                        network interfaces
+
+                                          Notice that the behavior for `[::1]`
+                                        is platform dependent. For system with
+                                        IPv4 mapped IPv6 networking
+                                          is enabled, using `[::1]` will listen
+                                        on both IPv4 and IPv6; other systems
+                                        like FreeBSD, it will only
```
spoonincode, line 81: I guess I never noticed this paragraph before. yuck, we should fix the discrepancy

spoonincode (approved):

PR #1499, opened 2025-05-05
P2P Feature: Transaction notice

heifner:
### P2P Transaction propagation enhancement

- No change for API RPC calls
- Transaction received over P2P connection
  - Record trx id and connection id and that we have the trx
  - If trx id already received then drop the transaction
  - If trx.size() > 200 bytes (may change this to 0, going to run some perf tests)
    - send to all connections
    ``` 
      transaction_notice_message { 
         transaction_id_type id; 
         time_point_sec      expiration;
      };
      ```
    - If transaction_notice_message or trx (pending tracked ids) exceeds 100K for connection, disconnect.
    - Also verify expiration not more than on-chain configured max.
    - clear all transaction ids from local cache when connection closes
  - Queue transaction for speculative execution
- If transaction fails to speculatively execute
  - Drop it
  - Do not immediately remove trx id from recorded transactions
    - Prevents immediately trying it again if received from another connection
- If transaction succeeds
  - Broadcast to all P2P connections not received trx or transaction-notify from
- On receive of `transaction_notice_message` record trx id and connection id but not that we have the transaction

#### Notes:

A transaction request message is not needed. Transactions are broadcast to nodes that are known not to have the transaction. A node knows a connection does not have a transaction because it has not received the trx or a `transaction_notify` from that connection.

`transaction_notice_message` is not propagated when received. `transaction_notice_message` is only sent when a transaction is received. Since transactions are verified (speculatively executed) before being broadcast this will normally provide enough time for `transaction_notice_message` to be received from nodes that already have the transaction.

Backward compatible with previous versions of net protocol. Will not send `transaction_notice_message` to any connections not at latest net protocol. Since no `transaction_notice_message` received from older versions all transactions will be broadcast to them if not received from them.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -79,12 +79,11 @@ namespace eosio {
       std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::milliseconds(config::block_interval_ms)).count();
 
    struct node_transaction_state {
-      transaction_id_type id;
-      time_point_sec  expires;        /// time after which this may be purged.
-      uint32_t        connection_id = 0;
+      transaction_id_type  id;
+      uint32_t             connection_id = 0;
+      time_point_sec       expires;              // time after which this may be purged.
+      mutable bool         have_trx = false;     // trx received, not just trx notice, mutable because not indexed
    };
-
-   struct by_expiry;
 
    typedef multi_index_container<
       node_transaction_state,
@@ -97,8 +96,12 @@ namespace eosio {
             >,
             composite_key_compare< std::less<transaction_id_type>, std::less<> >
          >,
+         hashed_non_unique<
+            tag< struct by_connection_id >,
+            member< node_transaction_state, uint32_t, &node_transaction_state::connection_id >
+         >,
          ordered_non_unique<
-            tag< by_expiry >,
+            tag< struct by_expiry >,
             member< node_transaction_state, fc::time_point_sec, &node_transaction_state::expires > >
          >
       >
@@ -220,9 +223,12 @@ namespace eosio {
       bool have_block(const block_id_type& blkid) const;
       void rm_block(const block_id_type& blkid);
 
-      bool add_peer_txn( const transaction_id_type& id, const time_point_sec& trx_expires, uint32_t connection_id,
-                         const time_point_sec& now = time_point_sec(time_point::now()) );
+      // returns the number of tracked ids for connecion_id
+      size_t add_peer_txn( const transaction_id_type& id, const time_point_sec& trx_expires,
+                           uint32_t connection_id, bool have_trx );
+      bool peer_has_txn( const transaction_id_type& id, uint32_t connection_id ) const;
       bool have_txn( const transaction_id_type& tid ) const;
+      void rm_txns( uint32_t connection_id );
       void expire_txns();
 
       void bcast_vote_msg( uint32_t exclude_peer, const send_buffer_type& msg );
@@ -244,6 +250,7 @@ namespace eosio {
    constexpr auto     def_sync_fetch_span = 1000;
    constexpr auto     def_keepalive_interval = 10000;
    constexpr auto     def_trx_notice_min_size = 100;
+   constexpr auto     def_max_trx_per_connection = 100000;
 
    class connections_manager {
    public:
@@ -1448,6 +1455,7 @@ namespace eosio {
       last_vote_received = time_point{};
       consecutive_blocks_nacks = 0;
       last_block_nack = block_id_type{};
+      my_impl->dispatcher.rm_txns(connection_id);
 
       uint32_t head_num = my_impl->get_chain_head_num();
       if (last_received_block_num >= head_num) {
@@ -2591,27 +2599,52 @@ namespace eosio {
       index.erase(p.first, p.second);
    }
 
-   bool dispatch_manager::add_peer_txn( const transaction_id_type& id, const time_point_sec& trx_expires,
-                                        uint32_t connection_id, const time_point_sec& now ) {
+   size_t
+   dispatch_manager::add_peer_txn( const transaction_id_type& id, const time_point_sec& trx_expires,
+                                   uint32_t connection_id, bool have_trx )
+   {
+      fc::time_point_sec now{fc::time_point::now()};
       fc::lock_guard g( local_txns_mtx );
-      auto tptr = local_txns.get<by_id>().find( std::make_tuple( std::ref( id ), connection_id ) );
-      bool added = (tptr == local_txns.end());
-      if( added ) {
+      auto& conn_idx = local_txns.get<by_connection_id>();
+      auto count = conn_idx.count( connection_id );
+
+      bool add = true;
+      if( count > 0 ) {
+         auto& id_idx = local_txns.get<by_id>();
+         if (auto tptr = id_idx.find( std::make_tuple( std::ref( id ), connection_id ) ); tptr != id_idx.end()) {
+            add = false;
+            if (have_trx && !tptr->have_trx)
+               tptr->have_trx = true;
+         }
+      }
+      if (add) {
          // expire at either transaction expiration or configured max expire time whichever is less
          time_point_sec expires{now.to_time_point() + my_impl->p2p_dedup_cache_expire_time_us};
          expires = std::min( trx_expires, expires );
          local_txns.insert( node_transaction_state{
             .id = id,
+            .connection_id = connection_id,
             .expires = expires,
-            .connection_id = connection_id} );
+            .have_trx = have_trx } );
+         ++count;
       }
-      return added;
+      return count;
+   }
+
+   bool dispatch_manager::peer_has_txn( const transaction_id_type& id, uint32_t connection_id ) const {
+      fc::lock_guard g( local_txns_mtx );
+      return local_txns.get<by_id>().contains( std::make_tuple( std::ref( id ), connection_id ) );
    }
 
    bool dispatch_manager::have_txn( const transaction_id_type& tid ) const {
       fc::lock_guard g( local_txns_mtx );
-      const auto tptr = local_txns.get<by_id>().find( tid );
-      return tptr != local_txns.end();
+      const auto r = local_txns.get<by_id>().equal_range( tid );
+      return std::ranges::any_of( r.first, r.second, []( const auto& t ) { return t.have_trx; } );
+   }
+
+   void dispatch_manager::rm_txns( uint32_t connection_id ) {
+      fc::lock_guard g( local_txns_mtx );
+      local_txns.get<by_connection_id>().erase( connection_id );
    }
 
    void dispatch_manager::expire_txns() {
@@ -2699,12 +2732,11 @@ namespace eosio {
    // called from any thread
    void dispatch_manager::bcast_transaction(const packed_transaction_ptr& trx) {
       trx_buffer_factory buff_factory;
-      const fc::time_point_sec now{fc::time_point::now()};
-      my_impl->connections.for_each_connection( [this, &trx, &now, &buff_factory]( const connection_ptr& cp ) {
+      my_impl->connections.for_each_connection( [this, &trx, &buff_factory]( const connection_ptr& cp ) {
          if( !cp->is_transactions_connection() || !cp->current() ) {
             return;
          }
-         if( !add_peer_txn(trx->id(), trx->expiration(), cp->connection_id, now) ) {
+         if( peer_has_txn(trx->id(), cp->connection_id) ) {
             return;
          }
 
@@ -2724,15 +2756,13 @@ namespace eosio {
       }
 
       trx_buffer_factory buff_factory;
-      const fc::time_point_sec now{fc::time_point::now()};
-      my_impl->connections.for_each_connection( [this, &trx, &now, &buff_factory]( const connection_ptr& cp ) {
+      my_impl->connections.for_each_connection( [this, &trx, &buff_factory]( const connection_ptr& cp ) {
          if( cp->protocol_version < proto_trx_notice || !cp->is_transactions_connection() || !cp->current() ) {
             return;
          }
-         // todo
-         // if( !add_peer_txn(trx->id(), trx->expiration(), cp->connection_id, now) ) {
-         //    return;
-         // }
+         if( peer_has_txn(trx->id(), cp->connection_id) ) {
+            return;
+         }
 
          const send_buffer_type& sb = buff_factory.get_notice_send_buffer( trx );
          fc_dlog( logger, "sending trx notice: ${id}, to connection - ${cid}", ("id", trx->id())("cid", cp->connection_id) );
@@ -3143,7 +3173,11 @@ namespace eosio {
          return true;
       }
       bool have_trx = my_impl->dispatcher.have_txn( ptr->id() );
-      my_impl->dispatcher.add_peer_txn( ptr->id(), ptr->expiration(), connection_id );
+      size_t connection_count = my_impl->dispatcher.add_peer_txn( ptr->id(), ptr->expiration(), true, connection_id );
+      if (connection_count > def_max_trx_per_connection) {
+         peer_wlog(this, "Max tracked trx reached ${c}, closing", ("c", connection_count));
+         close();
+      }
 
       if( have_trx ) {
          peer_dlog( this, "got a duplicate transaction - dropping" );
@@ -3173,7 +3207,13 @@ namespace eosio {
       transaction_notice_message msg;
       fc::raw::unpack( ds, msg );
 
-      // todo
+      size_t connection_count = my_impl->dispatcher.add_peer_txn( msg.id, msg.expiration, false, connection_id );
+      if (connection_count > def_max_trx_per_connection) {
+         peer_wlog(this, "Max tracked trx reached ${c}, closing", ("c", connection_count));
+         close();
+      }
+
+      handle_message( msg );
 
       return true;
    }
@@ -3793,7 +3833,7 @@ namespace eosio {
 
    // called from connection strand
    void connection::handle_message( const transaction_notice_message& msg ) {
-     // todo
+      peer_dlog( this, "received transaction_notice_message ${id}", ("id", msg.id) );
    }
 
    digest_type gossip_bp_peers_message::bp_peer::digest() const {
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -6,6 +6,7 @@
 #include <eosio/net_plugin/net_utils.hpp>
 #include <eosio/net_plugin/auto_bp_peering.hpp>
 #include <eosio/chain/types.hpp>
+#include <eosio/chain/global_property_object.hpp>
 #include <eosio/chain/controller.hpp>
 #include <eosio/chain/exceptions.hpp>
 #include <eosio/chain/block.hpp>
@@ -249,7 +250,7 @@ namespace eosio {
    constexpr auto     def_resp_expected_wait = std::chrono::seconds(5);
    constexpr auto     def_sync_fetch_span = 1000;
    constexpr auto     def_keepalive_interval = 10000;
-   constexpr auto     def_trx_notice_min_size = 100;
+   constexpr auto     def_trx_notice_min_size = 200; // transfer packed transaction is ~170 bytes, transaction notice is 41 bytes
    constexpr auto     def_max_trx_per_connection = 100000;
 
    class connections_manager {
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -79,13 +80,12 @@ namespace eosio {
       std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::milliseconds(config::block_interval_ms)).count();
 
    struct node_transaction_state {
-      transaction_id_type id;
-      time_point_sec  expires;        /// time after which this may be purged.
-      uint32_t        connection_id = 0;
+      transaction_id_type  id;
+      uint32_t             connection_id = 0;
+      time_point_sec       expires;              // time after which this may be purged.
+      mutable bool         have_trx = false;     // trx received, not just trx notice, mutable because not indexed
```
greg7mdp, line 86: iiuc, the `have_trx` member is specific to our node, not to the pair `<id, connection_id>`, and as such I feel it should not be stored here, but maybe is a separate `unordered_map<transaction_id_type, bool>`.

I think this would simplify the code in `add_peer_txn`.

heifner: Seems like too much overhead to store the trx id again. Also at some point we might want to know which connection the the trx was received on. At that point we could make the bool an uint8 enum with recieved_on_connection and received_on_other_connection and not_received.

greg7mdp: > at some point we might want to know which connection the the trx was received on

We know that we received the trx or the notice just from the fact that the pair `<id, connection_id>` is present.

> Seems like too much overhead to store the trx id again.

Hum, I don't feel this way. We already store the trx id for every connection that we received the trx or notice on. What's one more?

greg7mdp: It is interesting that the `connection_id` is a continuously increasing index. If we re-used the indices of old connections (we could keep a `std::vector<uint32_t> old_connection_ids;`), then we could use a dynamic_bitset instead of the `std::unordered_set`

greg7mdp: The `unordered_set` would work too, if we go that route I think the `boost::unordered_flat_set` would be a better choice (faster and more memory efficient, esp. for small values).

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2593,42 +2625,96 @@ namespace eosio {
...
+   }
+
+   size_t dispatch_manager::add_peer_txn_notice( const transaction_id_type& id, const time_point_sec& trx_expires, connection& c )
+   {
+      fc::lock_guard g( local_txns_mtx );
+
+      auto& id_idx = local_txns.get<by_id>();
+      if (auto tptr = id_idx.find( id ); tptr != id_idx.end()) {
+         if (tptr->connection_ids.insert(c.connection_id).second)
+            ++c.trx_entries_size;
```
greg7mdp, line 2660: Why increment the transaction count? We just received a notice.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2593,42 +2625,96 @@ namespace eosio {
...
+      } else {
+         // expire at either transaction expiration or configured max expire time whichever is less
+         time_point_sec expires{fc::time_point::now() + my_impl->p2p_dedup_cache_expire_time_us};
+         expires = std::min( trx_expires, expires );
+         local_txns.insert( node_transaction_state{
+            .id = id,
+            .expires = expires,
+            .connection_ids = {c.connection_id},
+            .have_trx = false } );
+         ++c.trx_entries_size;
```
greg7mdp, line 2670: same here

greg7mdp, line 2670: same here, why increment the transaction count?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2593,42 +2625,96 @@ namespace eosio {
...
-      auto tptr = local_txns.get<by_id>().find( std::make_tuple( std::ref( id ), connection_id ) );
-      bool added = (tptr == local_txns.end());
-      if( added ) {
+
+      auto& id_idx = local_txns.get<by_id>();
+      bool already_have_trx = false;
+      if (auto tptr = id_idx.find( id ); tptr != id_idx.end()) {
+         if (tptr->connection_ids.insert(c.connection_id).second)
+            ++c.trx_entries_size;
+         already_have_trx = tptr->have_trx;
```
greg7mdp, line 2637: I feel that maybe we should update `tptr->expires` if `tptr->have_trx == false`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2593,42 +2625,96 @@ namespace eosio {
...
-      auto ex_up = old.upper_bound( now );
-      old.erase( ex_lo, ex_up );
-      g.unlock();
+      flat_map<uint32_t, uint32_t> expired_trxs_for_connection;
+      {
+         fc::lock_guard g( local_txns_mtx );
+         start_size = local_txns.size();
+         auto& old = local_txns.get<by_expiry>();
+         auto ex_lo = old.lower_bound( fc::time_point_sec( 0 ) );
+         auto ex_up = old.upper_bound( fc::time_point_sec{now - fc::seconds(5)} ); // allow for some clock-skew
```
greg7mdp, line 2694: I think it should be at least 10s (prob. a little more) to match what we use when we accept transactions.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2593,42 +2625,96 @@ namespace eosio {
...
-      fc::time_point_sec now{time_point::now()};
-
-      fc::unique_lock g( local_txns_mtx );
-      start_size = local_txns.size();
-      auto& old = local_txns.get<by_expiry>();
-      auto ex_lo = old.lower_bound( fc::time_point_sec( 0 ) );
-      auto ex_up = old.upper_bound( now );
-      old.erase( ex_lo, ex_up );
-      g.unlock();
+      flat_map<uint32_t, uint32_t> expired_trxs_for_connection;
```
greg7mdp, line 2688: `boost::unordered_flat_map`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2593,42 +2625,96 @@ namespace eosio {
...
 
-      fc_dlog( logger, "expire_local_txns size ${s} removed ${r}", ("s", start_size)( "r", start_size - end_size ) );
+      if (!expired_trxs_for_connection.empty()) {
+         my_impl->connections.with_connection_lock([&](const connections_manager::connection_details_index& connections) {
+            // acquire connections_mtx before acquiring local_txns_mtx as connections_mtx can be held while acquiring local_txns_mtx
+            fc::lock_guard g( local_txns_mtx );
+            auto& conn_idx = connections.get<by_connection_id>();
+            for (auto [conn_id, count] : expired_trxs_for_connection) {
+               if (auto itr = conn_idx.find( conn_id ); itr != conn_idx.end()) {
+                  itr->c->trx_entries_size -= count;
```
greg7mdp, line 2710: I don't think the notices should affect `trx_entries_size`.
maybe instead of:

`mutable bool  have_trx = false;`

have:

~~`std::optional<uint32_t> trx_connection_source; // if we have received the trx, connection_id which sent it`~~.

actually, because we can receive the same transaction from multiple connections, we probably should have:

`mutable connection_id_set  connection_srcs; // connections which have sent us this transaction`

This is a lot of effort for managing this `trx_entries_size` count, I wonder if it is worth it?

greg7mdp (commented):
`unordered_flat_set` is much nicer :-)

plugins/net_plugin/net_plugin.cpp
heifner: The count is for the entries in this multiindex. Notice or trx both add to this multiindex.

plugins/net_plugin/net_plugin.cpp
heifner: Each node has a configurable amount of time it will keep trxs entries around. If we update it then we are not really honoring that limit if they keep being sent to us.

plugins/net_plugin/net_plugin.cpp
greg7mdp: Why do we need to keep that count?

plugins/net_plugin/net_plugin.cpp
greg7mdp: > Each node has a configurable amount of time it will keep trxs entries around. If we update it then we are not really honoring that limit if they keep being sent to us.

We store this expiration limit per transaction, not per connection. How can we honor the limit that each nodes sends?

plugins/net_plugin/net_plugin.cpp
greg7mdp: I must misunderstand the purpose of `c->trx_entries_size`? Why do we need to track this?

heifner: It is to prevent someone just spamming the node with transaction_notice_messages until it grinds to a halt.

plugins/net_plugin/net_plugin.cpp
heifner: Each node does not send us an expiration. Our node has an expiration for how long it will remember a trx. I don't think another coming into the node should reset the expiration.

greg7mdp: Why have the expiration in the notice then? It is somewhat irrelevant imo. I think it would make sense to have the expiration only when we have the transaction (so maybe make it a `std::optional`?

heifner: Often the expiration of the actual trx is going to be less than the configured node expiration. So it is sent along so that we can when appropriate remove it sooner.

greg7mdp: Yes, but why do we need it in the notice? I understand we need it in the transaction message.

heifner: The extra few bytes for the expiration in the notice I doubt would make little difference but it does allow the receiver of the notice to potentially remove it sooner. Note there can/will be many notices sent where a trx is never sent. If the trx fails to subjectively execute on the node then it will never be sent to its peers. So imagine a node that is being spammed with trxs that fail. That node is going to spam all its peers with trx notice messages.

greg7mdp: > The extra few bytes for the expiration in the notice I doubt would make little difference

yes, I agree

> but it does allow the receiver of the notice to potentially remove it sooner

Sooner is a doubtful benefit.  If a node is spammed with transactions that fail, why would the spammer provide a short expiration time. 
I think maybe we should have a member in `node_transaction_state` which would be: `timestamp first_notice_received;`, and in `expire_txns`, we would remove all transactions where `have_trxn == false` and for which `now() - first_notice_received > max_trx_lifetime `

plugins/net_plugin/net_plugin.cpp
heifner: `max_trx_lifetime` is 1hr currently and could be increased on-chain. If we wanted to allow for 10000 TPS on a connection (and I think we could argue that we would want even larger) then that is 600,000 entries. Note that BPs often have only one connection that transactions are flowing into the BP on. If they increased `max_trx_lifetime` to 8hrs (which I think has been discussed) then that would be 4.8 Million entries.

greg7mdp: I don't really see why your more complex solution alleviates this issue?

heifner: Because in practice most transactions are not going to have an expiration anywhere near 1hr. `Wharfkit` defaults to 120 seconds. `cleos` defaults to 30 seconds. So if users are using small expirations then the node can handle a much larger number of trxs.

plugins/net_plugin/net_plugin.cpp
heifner: In at least one common case of spam trxs they use a small (or default) expiration. If you are spamming a trx in as to hit some defi condition or win some game. You send many trx that fail. You are not trying to bring down the node, you are just trying to hit a condition on-chain.

heifner: Actually since there is a potential large delay in the notice and actually receiving the trx, I can see updating the expiration when actually receiving the trx. I'll make that change.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -540,9 +560,10 @@ namespace eosio {
    constexpr uint16_t proto_savanna = 9;                   // savanna, adds vote_message
    constexpr uint16_t proto_block_nack = 10;               // adds block_nack_message & block_notice_message
    constexpr uint16_t proto_gossip_bp_peers = 11;          // adds gossip_bp_peers_message
+   constexpr uint16_t proto_trx_notice = 12;               // adds transaction_notice_message
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1499/commits/8f314050a2d744e0478c0fea7b874ab421db933e

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3107,28 +3220,96 @@ namespace eosio {
...
+            last_dropped_trx_msg_time = fc::time_point::now();
+            peer_dlog(this, "${r}", ("r", std::move(reason)));
+         }
+         if (my_impl->increment_dropped_trxs) {
+            my_impl->increment_dropped_trxs();
+         }
+         return true;
+      }
+
+      auto[connection_count, have_trx] = my_impl->dispatcher.add_peer_txn( ptr->id(), ptr->expiration(), *this );
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1499/commits/48bc63a8828ad4cdc940ac8b797ace7fd48b0531

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2701,23 +2787,47 @@ namespace eosio {
...
+      std::optional<connection_id_set> trx_connections;
+      my_impl->connections.for_each_connection( [&]( const connection_ptr& cp ) {
+         if( cp->protocol_version < proto_trx_notice || !cp->is_transactions_connection() || !cp->current() ) {
+            return;
+         }
+         if (!trx_connections)
+            trx_connections = peer_connections(trx->id());
+         if( trx_connections->contains(cp->connection_id) ) {
+            return;
+         }
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1499/commits/48bc63a8828ad4cdc940ac8b797ace7fd48b0531

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1499/commits/446fc5ce49c6eb0d22bad78795c5bdc8de22d11e

plugins/net_plugin/net_plugin.cpp
heifner: Increased to 15 seconds.

heifner: https://github.com/AntelopeIO/spring/pull/1499/commits/9568000de1fda1ebdd8876f59a1e25b735f64bf7

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1499/commits/fe311644102e01eeaef033e9740ee9a746bf027b

plugins/net_plugin/net_plugin.cpp
greg7mdp: But you wrote:
`It is to prevent someone just spamming the node with transaction_notice_messages until it grinds to a halt.`

Why would you expect the spamming node to send notices with a short expiration time? It seems that a strategy that only works if that is true is not very sound.

heifner: The limit is to prevent true spamming. Just random generate trx id and send `transaction_notice_message` as quickly as possible. Either approach handles this case.

I'm arguing the current impl works better for the "accidental" spamming case where the user is not trying to take down the node (actually the opposite, he wants it to be available) but rather just generating a lot of trxs that fail.

greg7mdp: So we expect a honest node to send us 100,000 notices for transactions that fail within `max_trx_lifetime`? 

Maybe we should have a variant of the notice message to notify that the transaction failed and will never be sent, so we can clean it up immediately from our multiindex (in theory we whould have received it only from this peer, so the `connection_ids` unordered_flat_map should have only one entry.

greg7mdp: This would eliminate the need for the expitration time in the notice.

heifner: > So we expect a honest node to send us 100,000 notices for transactions that fail within `max_trx_lifetime`?

Or succeed. We don't remove them on success until expire either.

The failure notice is interesting. Lets see what @arhag thinks.

greg7mdp: > We don't remove them on success until expire either.

If they succeed, we will receive the transaction and update the expire in the multi-index accordingly, so we wouldn't rely on the expire from the notice anyways.

heifner: Thinking some more about this over the weekend. We already have `p2p-dedup-cache-expire-time-sec` which defaults to 10 seconds. So by default the longest a trx/trx-notice entry is kept is 10 seconds, not 1hr. 

> Why not just have a counter on the `connection`, which is incremented every time we receive a notice, and reset to 0 every `max_trx_lifetime` (we can check whenever we receive a notice on a connection whether the counter needs to be reset by keeping a `notice_counter_last_reset_time`).

This would be much simpler and faster. We could reset the connection entry counter every `p2p-dedup-cache-expire-time-sec`. We set the max on the counter to a hard-coded 200,000. By default that allows for 20,000 TPS. A user could increase that by decreasing the `p2p-dedup-cache-expire-time-sec`.

linh2931:
Should `transaction_notify`  be `transaction_notice_message`, in the 3rd paragraph from the bottom of the PR description ?

greg7mdp:
@heifner any thoughts about my telegram note:

Another idea which maybe is an alternative to the transaction notice:
For each connection:
1. every 1/10th of a second or so, send to the peer a vector of transaction ids (just the ids) that this node has validated since the last time, and that we are ready to propagate. Save the time the list was computed (we can use the same list for all peers).
2. the peer answers with, either a vector of transaction ids that it needs, or a bloom filter of the same to reduce traffic  (false positives just mean that we may send a few transactions that our peer already has)
3. we send to the peer, in a single i/o, a list of all the transactions it has requested.

This would reduce traffic and the number of i/o requests significantly, and, similarly as the notice, avoid sending a node a transaction it already has.

The more I think of it, the more I feel that this would be better than the transaction notice. It reduces the number of i/o requests drastically. Even if we have 2000 trx ids to send every 1/10th of a second (if we are doing 20,000 tps), that still a single message, of size only 64KB, and same for the answer. Also we send all the requested transactions batched together. 

It is also simpler than the notice feature I think, which requires updating info in the multiindex for every notice, and consuming memory for every transaction/notice. This new suggestion has very minimal overhead.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3105,30 +3198,76 @@ namespace eosio {
...
+         fc_dlog( logger, "send trx notice, trx size ${s}", ("s", message_length));
+         my_impl->dispatcher.bcast_transaction_notify(ptr);
+      }
+
       handle_message( ptr );
       return true;
    }
 
    // called from connection strand
+   bool connection::process_next_trx_notice_message(uint32_t message_length) {
```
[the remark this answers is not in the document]
heifner: If received, then it is fine to process it regardless of determined net version.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2558,77 +2574,133 @@ namespace eosio {
...
+   size_t dispatch_manager::add_peer_txn_notice( const transaction_id_type& id, connection& c )
+   {
       fc::lock_guard g( local_txns_mtx );
-      const auto tptr = local_txns.get<by_id>().find( tid );
-      return tptr != local_txns.end();
+
+      auto& id_idx = local_txns.get<by_id>();
+      if (auto tptr = id_idx.find( id ); tptr != id_idx.end()) {
+         if (tptr->connection_ids.insert(c.connection_id).second)
+            c.trx_entries_size += connection::trx_conn_entry_size;
```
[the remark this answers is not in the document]
heifner: We decided to go with size, so we could use diff values for a connection entry and a full entry. See:
```
      static constexpr uint32_t  trx_full_entry_size = sizeof(node_transaction_state);
      static constexpr uint32_t  trx_conn_entry_size = sizeof(connection_id_t);
```

greg7mdp (approved):

PR #1503, opened 2025-05-07
[1.2.0-rc2] P2P: Gossip BP peer enhancements

### LINKED ISSUE #1496: Gossip BP peer enhancements
- Add `expiration` to `gossip_bp_peers_message::bp_peer`
  - Part of the signed data
  - Used to time-out a `bp_peer` entry
  - Hard code expiration time to be 1hr.
  - Use head-block time + expiration time.
  - Configured `p2p-producer-peer` will send out an updated signed `bp_peer` entry every 1/2 of expiration time (30 minutes) which will be gossiped to all its bp gossip peers.
  - Error if expiration greater than 1.25 hr of current block time.
  - Error if expiration less than current block time.
  - Remove bp_peer entries when they expire
  - If more than 4/8 `bp_peer` entries received drop those with the shortest expiration. (4/8 - IPv4/IPv6 see below).
    - If 4/8 all have the same expiration then an Error.
  - Update existing expiration of bp_peers that match same [`producer_name`,`server_address`.]
 
- Add option `p2p-outbound-server-address`
  - Similar to how `p2p-server-address` correspond to `p2p-listen-endpoint`, `p2p-outbound-server-address` will correspond to `p2p-server-address`; defaulting to `p2p-server-address` if not specified. Note `p2p-server-address` defaults to `p2p-listen-address` if not specified.
  - Include `outbound_server_address` in `gossip_bp_peers_message::bp_peer` message
  - This is the address a peer will connect from so peer can enable firewall rules
  - Only communicated over bp gossip messages, not sent to non-bp-gossip peers.
 
- Add `:nobpgoss` as an option of `p2p-listen-endpoint`
  - Indicates on `p2p-producer-peer` configured nodes that this endpoint should not be bp auto gossiped.
  - Gossip all endpoints on `p2p-producer-peer` node unless they have this specifier.

- Increase from 4 to 8 allowed `bp_peer` entries if IPv4/IPv6.
  - 4 IPv4 entries allowed per BP.
  - 4 IPv6 entries allowed per BP.
  - Duplicate connections will naturally be de-dupped by peer.

- Add API `/v1/net/bp_gossip_peers` which will return the current set of known bp gossip peers.
  - Include all information in `gossip_bp_peers_message::bp_peer`
  - Can be queried to retrieve data for firewall

- Add `agent-name` as an option of `peer-log-format` so it can be included in every peer log statement if desired.
 
- Update p2p help/docs with valid examples.

heifner:
- Add new `p2p-bp-gossip-endpoint <producer account>,<advertised inbound connection endpoint>,<outbound connection IP address>`
    - Where `<producer account>` is the previous `p2p-producer-peer`. The BP producer account used for producing and used for registering an on-chain producer peer key.
    - Where `<advertised inbound connection endpoint>` is an externally reachable IP address or domain name with port.
    - Where `<outbound connection IP address>` is the IP address this node will connection from. This is provided so that BPs using a firewall can add this entry to allow connection. Required and must be a valid IP address.
    - Only communicated over bp gossip messages, not sent to non-bp-gossip peers.
    - `p2p-bp-gossip-endpoint` can be specified more than once if multiple endpoints for a node should be gossiped or if in a test network more than one BP peer key is handled by the nodeos.
    - `<producer account>`, `<advertised inbound connection endpoint>` combination must be unique. 8 unique bp gossip entries are allowed across all BP nodes for a specific producer account.
    - Examples for producer `defproda`:
      -  `p2p-bp-gossip-endpoint = defproda,p2p.proddomain.com:9876,173.194.219.100`
      -  `p2p-bp-gossip-endpoint = defproda,p2p.proddomain2.com:9875,173.194.219.99`

- Add `expiration` to `gossip_bp_peers_message::bp_peer`
  - Part of the signed data
  - Used to time-out a `bp_peer` entry
  - Hard code expiration time to be 1hr.
  - Use head-block time + expiration time.
  - Configured `p2p-bp-gossip-endpoint` will send out an updated signed `bp_peer` entry every 1/2 of expiration time (30 minutes) which will be gossiped to all its bp gossip peers.
  - Error if expiration greater than 1.25 hr of current block time.
  - Error if expiration less than current block time.
  - Remove bp_peer entries when they expire
  - If more than 8 `bp_peer` entries received from single producer than Error.
  - Update existing expiration of bp_peers that match same [`producer_name`,`server_endpoint`.]

- Increase from 4 to 8 allowed `bp_peer` entries per producer.
  - Duplicate connections will naturally be de-dupped by peer.

- Add API `/v1/net/bp_gossip_peers` which will return the current set of known bp gossip peers.
  - Include all information in `gossip_bp_peers_message::bp_peer`
  - Can be queried to retrieve data for firewall

- Add `agent-name` as an option of `peer-log-format` so it can be included in every peer log statement if desired.
 
- Update p2p help with valid examples.

```
  --p2p-bp-gossip-endpoint arg          The BP account, inbound connection
                                        endpoint, outbound connection IP
                                        address. The BP account is the producer
                                        peer name to retrieve peer-key from
                                        on-chain peerkeys table registered
                                        on-chain via regpeerkey action. The
                                        inbound connection endpoint is
                                        typically the listen endpoint of this
                                        node. The outbound connection IP
                                        address is typically the IP address of
                                        this node. Peer will use this value to
                                        allow access through firewall. Private
                                        key of peer-key should be configured
                                        via signature-provider.
                                         Syntax: bp_account,inbound_endpoint,ou
                                        tbound_ip_address
                                         Example:
                                           myprod,myhostname.com:9876,198.51.10
                                        0.1
                                           myprod,myhostname2.com:9876,[2001:0d
                                        b8:85a3:0000:0000:8a2e:0370:7334]
```

Resolves #1496

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3392,9 +3392,8 @@ struct controller_impl {
    void update_peer_keys() {
       if (!peer_keys_db.is_active())
          return;
-      // Since get_top_producer_keys() can't be called on startup, call it on first call to update_peer_keys().
       // if syncing or replaying old blocks don't bother updating peer keys.
-      if (peer_keys_db.last_block_num_update() > 0 && fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
+      if (fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
          return;
 
       try {
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3390,11 +3390,12 @@ struct controller_impl {
    } /// start_block
 
    void update_peer_keys() {
-      // if syncing or replaying old blocks don't bother updating peer keys
-      if (peer_keys_db.last_block_num_update() > 0) { // can't have been updated unless active
-         if (!peer_keys_db.is_active() || fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
-            return;
-      }
+      if (!peer_keys_db.is_active())
+         return;
+      // Since get_top_producer_keys() can't be called on startup, call it on first call to update_peer_keys().
+      // if syncing or replaying old blocks don't bother updating peer keys.
+      if (peer_keys_db.last_block_num_update() > 0 && fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
+         return;
 
       try {
          auto block_num = chain_head.block_num() + 1;
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3391,8 +3391,10 @@ struct controller_impl {
 
    void update_peer_keys() {
       // if syncing or replaying old blocks don't bother updating peer keys
-      if (!peer_keys_db.is_active() || fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
-         return;
+      if (peer_keys_db.last_update() > 0) { // can't have been updated unless active
+         if (!peer_keys_db.is_active() || fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
+            return;
+      }
 
       try {
          auto block_num = chain_head.block_num() + 1;
@@ -5864,7 +5866,7 @@ std::optional<peer_info_t> controller::get_peer_info(name n) const {
 }
 
 bool controller::configured_peer_keys_updated() {
-   return my->peer_keys_db.configured_peer_keys_updated();
+   return my->peer_keys_db.is_active() && my->peer_keys_db.configured_peer_keys_updated();
 }
 
 getpeerkeys_res_t controller::get_top_producer_keys() {
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3391,7 +3391,7 @@ struct controller_impl {
 
    void update_peer_keys() {
       // if syncing or replaying old blocks don't bother updating peer keys
-      if (peer_keys_db.last_update() > 0) { // can't have been updated unless active
+      if (peer_keys_db.last_block_num_update() > 0) { // can't have been updated unless active
          if (!peer_keys_db.is_active() || fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
             return;
       }
@@ -5857,7 +5857,7 @@ chain_id_type controller::get_chain_id()const {
    return my->chain_id;
 }
 
-void controller::set_peer_keys_retrieval_active(peer_name_set_t configured_bp_peers) {
+void controller::set_peer_keys_retrieval_active(name_set_t configured_bp_peers) {
    my->peer_keys_db.set_active(std::move(configured_bp_peers));
 }
 
--- a/libraries/chain/include/eosio/chain/peer_keys_db.hpp
+++ b/libraries/chain/include/eosio/chain/peer_keys_db.hpp
@@ -27,7 +27,7 @@ struct peer_info_t {
 };
 
 using peer_key_map_t = boost::unordered_flat_map<name, peer_info_t, std::hash<name>>;
-using peer_name_set_t = flat_set<name>;
+using name_set_t = flat_set<name>;
 
 /**
  * This class caches the on-chain public keys that BP use to sign the `gossip_bp_peers`
@@ -39,7 +39,7 @@ public:
    peer_keys_db_t() = default;
 
    // called on startup with configured bp peers of the node
-   void set_active(peer_name_set_t configured_bp_peers) {
+   void set_active(name_set_t configured_bp_peers) {
       _configured_bp_peers = std::move(configured_bp_peers);
       _active = true;
    }
@@ -48,7 +48,7 @@ public:
    bool is_active() const { return _active; }
 
    // must be called from the main thread
-   block_num_type last_update() const { return _last_block_num; }
+   block_num_type last_block_num_update() const { return _last_block_num; }
 
    // must be called from the main thread
    // return true if update_peer_keys should be called with new map of peer keys
@@ -67,7 +67,7 @@ public:
 private:
    bool               _active = false;       // if not active (the default), no update occurs
    block_num_type     _last_block_num = 0;
-   peer_name_set_t    _configured_bp_peers;  // no updates occurs
+   name_set_t         _configured_bp_peers;  // no updates occurs
    std::atomic<bool>  _configured_bp_peers_updated{false};
 
    mutable fc::mutex  _m;
--- a/libraries/chain/include/eosio/chain/peer_keys_db.hpp
+++ b/libraries/chain/include/eosio/chain/peer_keys_db.hpp
@@ -48,6 +48,9 @@ public:
    bool is_active() const { return _active; }
 
    // must be called from the main thread
+   block_num_type last_update() const { return _last_block_num; }
+
+   // must be called from the main thread
    // return true if update_peer_keys should be called with new map of peer keys
    bool should_update(block_num_type block_num) { return _active && (_last_block_num == 0 || block_num % 120 == 0); }
 
--- a/libraries/chain/include/eosio/chain/peer_keys_db.hpp
+++ b/libraries/chain/include/eosio/chain/peer_keys_db.hpp
@@ -48,9 +48,6 @@ public:
    bool is_active() const { return _active; }
 
    // must be called from the main thread
-   block_num_type last_block_num_update() const { return _last_block_num; }
-
-   // must be called from the main thread
    // return true if update_peer_keys should be called with new map of peer keys
    bool should_update(block_num_type block_num) { return _active && (_last_block_num == 0 || block_num % 120 == 0); }
 
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -23,6 +23,7 @@ class bp_connection_manager {
 
    static constexpr size_t           max_bp_peers_per_producer = 8;
    static constexpr fc::microseconds bp_peer_expiration = fc::hours(1);
+   static constexpr fc::microseconds bp_peer_expiration_variance = fc::hours(1) + fc::minutes(15);
 
    gossip_bp_index_t      gossip_bps;
 
@@ -356,13 +357,22 @@ public:
          return true;
       };
 
+      const auto head_block_time = self()->head_block_time.load();
+      const block_timestamp_type expire = head_block_time + bp_peer_expiration_variance;
+      auto is_expiration_valid = [&](const gossip_bp_peers_message::signed_bp_peer& peer) -> bool {
+         if (initial_msg)
+            return true; // initial message has no expiration
+         return peer.expiration > head_block_time && peer.expiration < expire;
+      };
+
       fc::lock_guard g(gossip_bps.mtx);
       auto& sig_idx = gossip_bps.index.get<by_sig>();
       for (auto i = msg.peers.begin(); i != msg.peers.end() && !invalid_message;) {
          const auto& peer = *i;
          bool have_sig = sig_idx.contains(peer.sig); // we already have it, already verified
-         if (!have_sig && !is_peer_key_valid(peer)) {
+         if (!have_sig && (!is_peer_key_valid(peer) || !is_expiration_valid(peer))) {
             // peer key may have changed or been removed on-chain, do not consider that a fatal error, just remove it
+            // may be expired, do not consider that fatal, just remove it
             i = msg.peers.erase(i);
          } else {
             ++i;
@@ -385,7 +395,8 @@ public:
          if (auto i = idx.find(boost::make_tuple(peer.producer_name, boost::cref(peer.server_address))); i != idx.end()) {
             if (i->sig != peer.sig) { // signature has changed, producer_name and server_address has not changed
                gossip_bps.index.modify(i, [&peer](auto& m) {
-                  m.sig = peer.sig; // update the signature, producer_name and server_address has not changed
+                  m.expiration = peer.expiration;
+                  m.sig = peer.sig; // update the signature, producer_name, server_address, outbound_address has not changed
                });
                diff = true;
             }
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -43,7 +43,7 @@ class bp_connection_manager {
       flat_map<net_utils::endpoint, account_name>   auto_bp_accounts;       // --p2p-auto-bp-peer endpoint->account
       // p2p-bp-gossip-endpoint
       name_set_t                                    my_bp_gossip_accounts;  // producer account of --p2p-bp-gossip-endpoint, for bp gossip
-      flat_set<bp_gossip_endpoint_t>                bp_gossip_endpoints;    // [inbound_endpoint,outbound_ip_address] to bp gossip
+      std::vector<bp_gossip_endpoint_t>             bp_gossip_endpoints;    // [inbound_endpoint,outbound_ip_address] to bp gossip
    } config; // thread safe only because modified at plugin startup currently
 
    // the following members are only accessed from main thread
@@ -193,7 +193,7 @@ public:
                         "Invalid p2p-bp-gossip-endpoint outbound ip address ${p}, syntax ip-address", ("p", outbound_ip_address));
 
             fc_dlog(self()->get_logger(), "Setting p2p-bp-gossip-endpoint ${a} -> ${i},${o}", ("a", account)("i", inbound_server_endpoint)("o", outbound_ip_address));
-            config.bp_gossip_endpoints.emplace(inbound_server_endpoint, outbound_ip_address);
+            config.bp_gossip_endpoints.emplace_back(inbound_server_endpoint, outbound_ip_address);
          } catch (chain::name_type_exception&) {
             EOS_ASSERT(false, chain::plugin_config_exception,
                        "The account ${a} supplied by --p2p-bp-gossip-endpoint option is invalid", ("a", aname));
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -63,6 +63,10 @@ class bp_connection_manager {
    static std::string to_string(const Container<account_name, Rest...>& peers) {
       return boost::algorithm::join(peers | boost::adaptors::transformed([](auto& p) { return p.to_string(); }), ",");
    }
+   template <template <typename...> typename Container, typename... Rest>
+   static std::string to_string(const Container<std::string, Rest...>& peers) {
+      return boost::algorithm::join(peers, ",");
+   }
 
    // Only called from main thread
    peer_name_set_t active_bp_accounts(const std::vector<chain::producer_authority>& schedule) const {
@@ -207,8 +211,10 @@ public:
       // normally only one bp peer account except in testing scenarios or test chains
       const controller& cc = self()->chain_plug->chain();
       block_timestamp_type expire = self()->head_block_time.load() + bp_peer_expiration;
+      fc_dlog(self()->get_logger(), "Updating BP gossip_bp_peers_message with expiration ${e}", ("e", expire));
       for (const auto& my_bp_account : config.my_bp_peer_accounts) { // my_bp_peer_accounts not modified after plugin startup
          for (const auto& le : config.bp_gossip_listen_endpoints) {
+            fc_dlog(self()->get_logger(), "Updating BP gossip_bp_peers_message for ${a} address ${s}", ("a", my_bp_account)("s", le.server_address));
             std::optional<peer_info_t> peer_info = cc.get_peer_info(my_bp_account);
             if (peer_info && peer_info->key) {
                if (!initial_updated) {
@@ -240,6 +246,8 @@ public:
                } else {
                   gossip_bps.index.emplace(peer);
                }
+            } else {
+               fc_wlog(self()->get_logger(), "On-chain peer key not found for configured BP ${a}", ("a", my_bp_account));
             }
          }
       }
@@ -478,7 +486,10 @@ public:
       {
          fc::lock_guard gm(mtx);
          active_bps = active_bp_accounts(active_schedule);
+         fc_dlog(self()->get_logger(), "active_bps: ${a}", ("a", to_string(active_bps)));
+
          addresses = find_gossip_bp_addresses(active_bps, "connect");
+         fc_dlog(self()->get_logger(), "active addresses: ${a}", ("a", to_string(addresses)));
       }
 
       for (const auto& add : addresses) {
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -28,22 +28,21 @@ class bp_connection_manager {
 
    gossip_bp_index_t      gossip_bps;
 
-   struct listen_endpoint_t {
-      std::string listen_endpoint;  // full host:port:type
-      std::string server_address;   // externally known host:port:type
-      std::string outbound_address; // externally known outbound address
+   struct bp_gossip_endpoint_t {
+      std::string server_endpoint;      // externally known inbound endpoint host:port
+      std::string outbound_ip_address;  // externally known outbound IP address
 
-      auto operator<=>(const listen_endpoint_t& rhs) const = default;
-      bool operator==(const listen_endpoint_t& rhs) const = default;
+      auto operator<=>(const bp_gossip_endpoint_t& rhs) const = default;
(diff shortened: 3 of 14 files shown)


plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -452,5 +579,21 @@ class bp_connection_manager {
...
       }
    }
+
+   // RPC called from http threads
+   vector<gossip_bp_peers_message::bp_peer> bp_gossip_peers() const {
+      fc::lock_guard g(gossip_bps.mtx);
+      vector<gossip_bp_peers_message::bp_peer> peers;
+      for (const auto& p : gossip_bps.index) {
+         // no need to include sig
+         peers.push_back(gossip_bp_peers_message::bp_peer{
```
linh2931, line 589: `emplace_back`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4285,15 +4300,24 @@ namespace eosio {
...
-         ( "agent-name", bpo::value<string>()->default_value("EOS Test Agent"), "The name supplied to identify this node amongst the peers.")
+           "The account and public p2p endpoint of a block producer node to automatically connect to when it is in producer schedule. Not gossipped.\n"
+           "  Syntax: bp_account,host:port\n"
+           "  Example:\n"
+           "    producer1,p2p.prod.io:9876\n"
+           "    producer2,p2p.trx.myprod.io:9876:trx\n"
+           "    producer3,p2p.blk.example.io:9876:blk\n")
+         ("p2p-bp-gossip-endpoint", boost::program_options::value<vector<string>>()->composing()->multitoken(),
+           "The BP account, inbound connection endpoint, outbound connection IP address. "
+           "The BP account is the producer peer name to retrieve peer-key from on-chain peerkeys table registered on-chain via regpeerkey action. "
```
linh2931, line 4311: `producer peer name` is confusing. It sounds like a peer name of the producer. Maybe `producer name for peering`.

There are other places in the code using the naming convention, like `my_bp_peer_accounts` in `config_t`. Maybe too late to change.

libraries/chain/include/eosio/chain/peer_keys_db.hpp
```diff
@@ -47,6 +47,9 @@ class peer_keys_db_t {
    // safe to be called from any thread, _active only modified on startup
    bool is_active() const { return _active; }
 
+   // must be called from the main thread
+   block_num_type last_update() const { return _last_block_num; }
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2236eb6badb3c46c08188e812cc6528bba341722

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -21,15 +21,28 @@ class bp_connection_manager {
  public:
 #endif
 
-   static constexpr size_t max_bp_peers_per_producer = 4;
+   static constexpr size_t           max_bp_peers_per_producer = 8;
+   static constexpr fc::microseconds bp_peer_expiration = fc::hours(1);
+   static constexpr fc::microseconds my_bp_peer_expiration = fc::minutes(30); // resend my bp_peer info every 30 minutes
+   static constexpr fc::microseconds bp_peer_expiration_variance = fc::hours(1) + fc::minutes(15);
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2236eb6badb3c46c08188e812cc6528bba341722

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -136,49 +161,102 @@ class bp_connection_manager {
...
+            auto comma_pos = entry.find(',');
+            EOS_ASSERT(comma_pos != std::string::npos, chain::plugin_config_exception,
+                       "p2p-bp-gossip-endpoint ${e} must consist of bp-account-name,inbound-server-endpoint,outbound-ip-address separated by commas", ("e", entry));
+            aname = entry.substr(0, comma_pos);
+            account_name account(aname);
+            config.my_bp_peer_accounts.emplace(account);
+            auto rest = entry.substr(comma_pos + 1);
+            comma_pos = rest.find(',');
+            EOS_ASSERT(comma_pos != std::string::npos, chain::plugin_config_exception,
+                       "p2p-bp-gossip-endpoint ${e} must consist of bp-account-name,inbound-server-endpoint,outbound-ip-address separated by commas", ("e", entry));
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2236eb6badb3c46c08188e812cc6528bba341722

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -136,49 +161,102 @@ class bp_connection_manager {
...
+            EOS_ASSERT( !host.empty() && !port.empty(), chain::plugin_config_exception,
+                        "Invalid p2p-bp-gossip-endpoint inbound server endpoint ${p}, syntax host:port", ("p", inbound_server_endpoint));
+            auto outbound_ip_address = rest.substr(comma_pos + 1);
+            EOS_ASSERT( outbound_ip_address.length() <= net_utils::max_p2p_address_length, chain::plugin_config_exception,
+                        "p2p-bp-gossip-endpoint outbound-ip-address ${a} too long, must be less than ${m}",
+                        ("a", outbound_ip_address)("m", net_utils::max_p2p_address_length) );
+            auto is_valid_ip_address = [](const std::string& ip_str) {
+               try {
+                  boost::asio::ip::address::from_string(ip_str);
+               } catch (const boost::system::system_error& ) {
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2236eb6badb3c46c08188e812cc6528bba341722

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2236eb6badb3c46c08188e812cc6528bba341722

plugins/net_plugin/net_plugin.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2236eb6badb3c46c08188e812cc6528bba341722

linh2931: I am glad it exposed an error. And the new names are much clearer and the code is easier to understand.

linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -3391,8 +3391,10 @@ struct controller_impl {
 
    void update_peer_keys() {
       // if syncing or replaying old blocks don't bother updating peer keys
-      if (!peer_keys_db.is_active() || fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))
-         return;
+      if (peer_keys_db.last_block_num_update() > 0) { // can't have been updated unless active
```
greg7mdp, line 3394: Test seems wrong. If not active, `peer_keys_db.last_block_num_update() ` will be 0 and we won't do the `return` at line 3396, and continue to execute the try block below.

greg7mdp: Also the test `fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))` seems wrong as well. Why would it ever be true? chain_head block more than 5 minutes old?

I think the test should be something like:

```
      if (!peer_keys_db.is_active() ||
          (peer_keys_db.last_block_num_update() > 0 && chain_head.block_num() - peer_keys_db.last_block_num_update() > 600)
            return;
      }
```

heifner: Good point about checking `is_active()`.

> Also the test `fc::time_point::now() - chain_head.timestamp() > fc::minutes(5))` seems wrong as well. Why would it ever be true? chain_head block more than 5 minutes old?

It is true when syncing or replaying blocks as indicated by the comment.

greg7mdp: So why the `peer_keys_db.last_block_num_update() > 0`? Do you want to do the update once at the beginning even if we are replaying?

greg7mdp: If we are replaying (and we don't know how far behind we are), we have an old state in chainbase, so why get old IP from chainbase and gossip them? Especially since we get the oldest peer IPs at startup, and then don't update them until we're fully caught up.

I think we should wait till we're caught up to start updating peer keys.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -326,17 +404,29 @@ class bp_connection_manager {
...
+               auto min_expiration_itr = r.first;
+               auto min_expiration = min_expiration_itr->expiration;
+               ++r.first;
+               for (; r.first != r.second; ++r.first) {
+                  if (r.first->expiration < min_expiration) {
+                     min_expiration = r.first->expiration;
+                     min_expiration_itr = r.first;
+                  }
+               }
+               gossip_bps.index.erase(min_expiration_itr);
```
greg7mdp, line 429: Suppose one node changes his `peer.server_endpoint` once. We will have two entries in our index, one with the old endpoint and one with the new one. 
When will we ever remove the entry with the old endpoint? Will we ever have `std::distance(r.first, r.second) >= max_bp_gossip_peers_per_producer`?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -300,13 +369,22 @@ class bp_connection_manager {
          return true;
       };
 
+      const auto head_block_time = self()->head_block_time.load();
+      const block_timestamp_type expire = head_block_time + bp_gossip_peer_expiration_variance;
+      auto is_expiration_valid = [&](const gossip_bp_peers_message::signed_bp_peer& peer) -> bool {
+         if (initial_msg)
+            return true; // initial message has no expiration
+         return peer.expiration > head_block_time && peer.expiration < expire;
+      };
```
greg7mdp, line 378: I don't understand this function.

I think the criteria for being `valid` should be `peer.expiration > expire`, not the opposite?

Why are we adding `bp_gossip_peer_expiration_variance` to `expire`?

greg7mdp (changes requested):

libraries/chain/controller.cpp
heifner: I was concerned about bootstrapping a dead/paused chain. But in that situation manual BP connections can be used and would likely be used anyway.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -139,18 +139,23 @@ namespace eosio {
...
+            // [producer_name, server_endpoint] is unique
+            return std::tie(producer_name, server_endpoint) < std::tie(rhs.producer_name, rhs.server_endpoint);
          }
       };
+      struct signed_bp_peer : bp_peer {
+         signature_type            sig;
+      };
 
-      std::vector<bp_peer> peers;
+      std::vector<signed_bp_peer> peers;
```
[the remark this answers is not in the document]
heifner: In the next PR we use a diff type for the http return info. The main reason for the separation is it makes digest() method very simple.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -136,49 +152,102 @@ class bp_connection_manager {
...
+      for (const auto& entry : bp_gossip_endpoints) {
+         std::string aname;
          try {
-            config.my_bp_peer_accounts.emplace(chain::name(entry));
+            auto comma_pos = entry.find(',');
+            EOS_ASSERT(comma_pos != std::string::npos, chain::plugin_config_exception,
+                       "p2p-bp-gossip-endpoint ${e} must consist of bp-account-name,inbound-server-endpoint,outbound-ip-address separated by commas", ("e", entry));
+            aname = entry.substr(0, comma_pos);
+            account_name account(aname);
+            config.my_bp_gossip_accounts.emplace(account);
```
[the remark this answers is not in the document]
heifner: All plugin_config_exceptions will terminate the program with an error here. Doesn't matter state of any partially populated data.

The way this was originally specified it there was no direct correlation between account and the addresses. Not that these are specified together it does seem like we should honor it.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: It will be expired in `expire_gossip_bp_peers()`. Until it expires we will have both.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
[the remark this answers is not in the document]
greg7mdp: Yes, I think so to. For example I believe you could have a node that gossips the addresses of 3 separate producers, each on his own node, so the addresses/name matches should be preserved I think.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: This is looking at the expiration that was communicated to us from another node. To be a valid expiration it has to be greater than the current head block time and also less than 1hr 15min past head block time. We want to limit how long of an expiration allowed. Currently that is 1hr 15 minutes from head block time.

greg7mdp: Thanks for the explanation! Can we rename `expire` to something like: `latest_acceptable_expiration_time`?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
greg7mdp: Why have this code here then? We could just do the `gossip_bps.index.insert(peer);` here, and rely on `expire_gossip_bp_peers()`  for retiring the expired ones.

heifner: It is considered invalid to gossip more than 8 bp entries. We remove here so we remain <= 8.

greg7mdp: Sure, but these are going to be removed anyways when they expire, won't they? Is there a reason to do it immediately at this point?

heifner: We need to remove it here or when we send out the bp gossip msgs otherwise other nodes will reject the bp gossip msg.

libraries/chain/controller.cpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/2c965d98b726d3da588dc728fb4317059c0ceef1

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -21,26 +21,40 @@ class bp_connection_manager {
...
-      flat_map<account_name, net_utils::endpoint>   bp_peer_addresses;
-      flat_map<net_utils::endpoint, account_name>   bp_peer_accounts;
-      peer_name_set_t                               my_bp_accounts;       // block producer --producer-name
-      peer_name_set_t                               my_bp_peer_accounts;  // peer key account --p2p-producer-peer
+      // p2p-auto-bp-peer
+      flat_map<account_name, net_utils::endpoint>   auto_bp_addresses;      // --p2p-auto-bp-peer account->endpoint
+      flat_map<net_utils::endpoint, account_name>   auto_bp_accounts;       // --p2p-auto-bp-peer endpoint->account
+      // p2p-bp-gossip-endpoint
+      name_set_t                                    my_bp_gossip_accounts;  // producer account of --p2p-bp-gossip-endpoint, for bp gossip
+      flat_set<bp_gossip_endpoint_t>                bp_gossip_endpoints;    // [inbound_endpoint,outbound_ip_address] to bp gossip
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/e549eb4da4f147a1a7d3a77a5626ab9da7320493 although now part of the map in https://github.com/AntelopeIO/spring/pull/1503/commits/771ead5d583e543326fde6a38410425e67a01aaa

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -49,26 +63,30 @@ class bp_connection_manager {
    static std::string to_string(const Container<account_name, Rest...>& peers) {
       return boost::algorithm::join(peers | boost::adaptors::transformed([](auto& p) { return p.to_string(); }), ",");
    }
+   template <template <typename...> typename Container, typename... Rest>
+   static std::string to_string(const Container<std::string, Rest...>& peers) {
+      return boost::algorithm::join(peers, ",");
+   }
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/c7f7fb4833dcf36b14a5441212643bbfa190f5ba

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/7f78ac327f1cac199d7c02d265721e1b6c1cd2b5

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -136,49 +152,102 @@ class bp_connection_manager {
...
-                  v.sig = peer.sig;
-               });
+      const controller& cc = self()->chain_plug->chain();
+      block_timestamp_type expire = self()->head_block_time.load() + bp_gossip_peer_expiration;
+      fc_dlog(self()->get_logger(), "Updating BP gossip_bp_peers_message with expiration ${e}", ("e", expire));
+      for (const auto& my_bp_account : config.my_bp_gossip_accounts) { // my_bp_gossip_accounts not modified after plugin startup
+         for (const auto& le : config.bp_gossip_endpoints) {
+            fc_dlog(self()->get_logger(), "Updating BP gossip_bp_peers_message for ${a} address ${s}", ("a", my_bp_account)("s", le.server_endpoint));
+            std::optional<peer_info_t> peer_info = cc.get_peer_info(my_bp_account);
+            if (peer_info && peer_info->key) {
```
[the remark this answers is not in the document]
heifner: Done. Thanks.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/771ead5d583e543326fde6a38410425e67a01aaa

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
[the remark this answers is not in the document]
greg7mdp: With the `requires`, the `Rest...` stuff is not needed anymore:
```
   template <typename Container>
   requires std::is_same_v<typename Container::value_type, account_name>
   static std::string to_string(const Container& peers) {
      return boost::algorithm::join(peers | boost::adaptors::transformed([](auto& p) { return p.to_string(); }), ",");
   }

   template <typename Container>
   requires std::is_same_v<typename Container::value_type, std::string>
   static std::string to_string(const Container& peers) {
      return boost::algorithm::join(peers, ",");
   }
```

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -21,54 +21,73 @@ class bp_connection_manager {
...
    struct config_t {
-      flat_map<account_name, net_utils::endpoint>   bp_peer_addresses;
-      flat_map<net_utils::endpoint, account_name>   bp_peer_accounts;
-      peer_name_set_t                               my_bp_accounts;       // block producer --producer-name
-      peer_name_set_t                               my_bp_peer_accounts;  // peer key account --p2p-producer-peer
+      // p2p-auto-bp-peer
+      flat_map<account_name, net_utils::endpoint>   auto_bp_addresses;      // --p2p-auto-bp-peer account->endpoint
+      flat_map<net_utils::endpoint, account_name>   auto_bp_accounts;       // --p2p-auto-bp-peer endpoint->account
+      // p2p-bp-gossip-endpoint, producer account -> [inbound_endpoint,outbound_ip_address] for bp gossip
+      flat_map<account_name, std::vector<bp_gossip_endpoint_t>>  my_bp_gossip_accounts;
```
greg7mdp, line 45: I'm worried about the `flat_map` containing a vector, and the insertion of new rows could cause possibly multiple vector copies (depending on the `flat_map` implementation).

`std::vector` can be trivially relocated to another memory position (just by copying the 3 pointers), but this relies on the [std::trivially_relocatable](https://isocpp.org/files/papers/P2786R13.html) trait which is not yet available, even in C++23.

It is possibly that boost's `flat_map` has its own trait for that, but I didn't see it with a cursory look. 

I think it would be safer in this case to use `std::map`, or even better `std::unordered_map` since I don't think ordered is a requirement here.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -136,50 +158,105 @@ class bp_connection_manager {
...
+                       .server_endpoint = le.server_endpoint,
+                       .outbound_ip_address = le.outbound_ip_address,
+                       .expiration = expire }
+                  };
+                  peer.sig = self()->sign_compact(*peer_info->key, peer.digest());
+                  EOS_ASSERT(peer.sig != signature_type{}, chain::plugin_config_exception,
+                             "Unable to sign bp peer ${p}, private key not found for ${k}", ("p", peer.producer_name)("k", peer_info->key->to_string({})));
+                  if (auto i = prod_idx.find(boost::make_tuple(bp_account, boost::cref(le.server_endpoint))); i != prod_idx.end()) {
+                     gossip_bps.index.modify(i, [&peer](auto& v) {
+                        v.outbound_ip_address = peer.outbound_ip_address;
```
greg7mdp, line 250: Interestingly, this means that we support different `server_endpoint` having the same `outbound_ip_address`, but not different `outbound_ip_address` having the same `server_endpoint`, so for example if we have:

```
p2p-bp-gossip-endpoint = defproda,p2p.proddomain.com:9876,173.194.219.100
p2p-bp-gossip-endpoint = defproda,p2p.proddomain.com:9875,173.194.219.99
```

the second entry will replace the first one in `gossip_bps`. I don't know if it is an issue though.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -192,7 +269,7 @@ class bp_connection_manager {
       std::string type;
       std::tie(e.host, e.port, type) = eosio::net_utils::split_host_port_type(conn->log_p2p_address);
 
-      if (config.bp_peer_accounts.count(e)) {
+      if (config.auto_bp_accounts.count(e)) {
```
greg7mdp, line 272: `contains`?

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: I think we should verify that and not allow it.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/07751bc9849efad8ac40e59ec425c37826477f71

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/07751bc9849efad8ac40e59ec425c37826477f71

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1503/commits/07751bc9849efad8ac40e59ec425c37826477f71

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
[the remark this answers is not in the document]
heifner: GCC doesn't like that: https://github.com/AntelopeIO/spring/actions/runs/15030364018/job/42241277512

greg7mdp (approved):

PR #1511, opened 2025-05-09
remove abieos usage from state history tests

spoonincode:
Removes abieos usage (and the submodule entirely); uses spring's abi_serializer instead.

### DIFF
--- a/tests/ship_client.cpp
+++ b/tests/ship_client.cpp
@@ -1,32 +1,30 @@
-#include <eosio/abi.hpp>
-#include <eosio/convert.hpp>
-#include <eosio/from_json.hpp>
-
-#include <rapidjson/document.h>
-#include <rapidjson/prettywriter.h>
-#include <rapidjson/stringbuffer.h>
-
 #include <boost/algorithm/string.hpp>
 #include <boost/asio.hpp>
 #include <boost/asio/local/stream_protocol.hpp>
 #include <boost/beast.hpp>
 #include <boost/program_options.hpp>
 
+#include <eosio/chain/abi_def.hpp>
+#include <eosio/chain/abi_serializer.hpp>
+#include <fc/io/json.hpp>
+
 #include <iostream>
+#include <regex>
 #include <string>
 
+using mvo = fc::mutable_variant_object;
 using tcp = boost::asio::ip::tcp;
 using unixs = boost::asio::local::stream_protocol;
 namespace ws = boost::beast::websocket;
 
 namespace bpo = boost::program_options;
 
-
 int main(int argc, char* argv[]) {
    boost::asio::io_context ctx;
    boost::asio::ip::tcp::resolver resolver(ctx);
    ws::stream<boost::asio::ip::tcp::socket> tcp_stream(ctx);
-   eosio::abi abi;
+   eosio::chain::abi_def abidef;
+   eosio::chain::abi_serializer abi;
 
    unixs::socket unix_socket(ctx);
    ws::stream<unixs::socket> unix_stream(ctx);
@@ -58,7 +56,7 @@ int main(int argc, char* argv[]) {
       statehistory_server = socket_address.substr(socket_address.find("unix://") + strlen("unix://") + 1);
    } else {
       std::string::size_type colon = socket_address.find(':');
-      eosio::check(colon != std::string::npos, "Missing ':' seperator in Websocket address and port");
+      FC_ASSERT(colon != std::string::npos, "Missing ':' seperator in Websocket address and port");
       statehistory_server = socket_address.substr(0, colon);
       statehistory_port = socket_address.substr(colon + 1);
    }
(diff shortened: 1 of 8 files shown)


tests/ship_client.cpp
```diff
@@ -97,37 +93,29 @@ int main(int argc, char* argv[]) {
...
-
-            request_writer.StartArray();
-               request_writer.String(request_result_types[num_requests%2].get_status_request.c_str());
-               request_writer.StartObject();
-               request_writer.EndObject();
-            request_writer.EndArray();
-
-            stream.write(boost::asio::buffer(request_type.json_to_bin(request_sb.GetString(), [](){})));
+            const eosio::chain::bytes get_status_bytes = abi.variant_to_binary("request",
+               fc::variants{request_result_types[num_requests%2].get_status_request, mvo()},
```
greg7mdp, line 97: I'm not quite understanding why we use a vector here:
`fc::variants{request_result_types[num_requests%2].get_status_request, mvo()}`

instead of:
`fc::variant{request_result_types[num_requests%2].get_status_request}`?

spoonincode: It's a variant: type is either `get_status_request_v0` or `get_status_request_v1` (the test just alternates between the two), but both types are just empty structs so just empty `mvo()`

greg7mdp: Thanks Matt!

spoonincode:
I don't know how to reasonably overcome the `MAX_SIZE_OF_BYTE_ARRAYS` limitation, so guess I gotta punt.

linh2931:
Can you keep the portion not related to the MAX_SIZE_OF_BYTE_ARRAYS limitation, so gradually removing the dependencies on abieos?

As it is only tests using abieos, can we write some mocking code to get around MAX_SIZE_OF_BYTE_ARRAYS?

spoonincode:
It might be possible to use `abi_serializer::add_specialized_unpack_pack()` and install a custom handler for the `bytes` type that avoids `MAX_SIZE_OF_BYTE_ARRAYS`. afaict on first glance it looks possible to "override" standard types that way .I will try it when I get a chance.

greg7mdp:
My suggestion would be update `raw.hpp` with serialization functions for `vector` and `strings` that don't check the size. 
The dispatch could be made by having a specific trait on the `Stream`, which would be checked with something like `fc::no_byte_array_size_limit<Stream>`.

It would be easy enough to add that trait to the streams used in the ship tests.

spoonincode:
I am not sure that's sufficient since I think you also will end up at
https://github.com/AntelopeIO/spring/blob/808f061b89a080fbf60cc0636ca0b38b134e523a/libraries/libfc/src/variant.cpp#L741-L744
Regardless, touching any of that stuff is way too invasive in consensus code for me to be comfortable doing the change myself right now (vs just hacking the tests up like this)

btw `abi_serializer::add_specialized_unpack_pack()` worked; so I will update PR with it (and state_history_tests.cpp) later

greg7mdp:
> I am not sure that's sufficient since I think you also will end up at

Ah, good point, what a mess!

linh2931 (approved):

tests/ship_client.cpp
```diff
@@ -58,32 +57,29 @@ int main(int argc, char* argv[]) {
...
-            std::string abi_string((const char*)abi_buffer.data().data(),
-                                   abi_buffer.data().size());
-            eosio::json_token_stream token_stream(abi_string.data());
-            eosio::abi_def abidef =
-                eosio::from_json<eosio::abi_def>(token_stream);
-            eosio::convert(abidef, abi);
+            std::string abi_string = boost::beast::buffers_to_string(abi_buffer.data());
+            //remove all tables since their names are invalid; tables not needed for this test
+            std::regex scrub_all_tables(R"(\{ "name": "[^"]+", "type": "[^"]+", "key_names": \[[^\]]*\] \},?)");
+            abi_string = std::regex_replace(abi_string, scrub_all_tables, "");
```
heifner, line 75: Why not use `ship_abi_without_tables()` here? You will need to link to state_history, but that seems ok.

tests/ship_streamer.cpp
```diff
@@ -66,18 +67,35 @@ int main(int argc, char* argv[]) {
...
+         std::string abi_string = boost::beast::buffers_to_string(abi_buffer.data());
+         //remove all tables since their names are invalid; tables not needed for this test
+         std::regex scrub_all_tables(R"(\{ "name": "[^"]+", "type": "[^"]+", "key_names": \[[^\]]*\] \},?)");
+         abi_string = std::regex_replace(abi_string, scrub_all_tables, "");
+
+         abi = eosio::chain::abi_serializer(fc::json::from_string(abi_string).as<eosio::chain::abi_def>(), null_yield_function);
+         //state history may have 'bytes' larger than MAX_SIZE_OF_BYTE_ARRAYS, so divert 'bytes' to an impl that does not have that check
+         abi.add_specialized_unpack_pack("bytes", std::make_pair<eosio::chain::abi_serializer::unpack_function, eosio::chain::abi_serializer::pack_function>(
+            [](fc::datastream<const char*>& stream, bool is_array, bool is_optional, const eosio::chain::abi_serializer::yield_function_t& yield) {
+               FC_ASSERT(!is_array, "sorry, this kludge doesn't support arrays");
```
heifner, line 79: `bytes` is a `vector<char>`, so shouldn't this assert say "bytes can't be an array" or something like that?

heifner (changes requested):

tests/ship_client.cpp
spoonincode: An important aspect of the test is consuming the ABI from the websocket. `ship_abi_without_tables()` will skip that.

tests/ship_streamer.cpp
```diff
@@ -66,18 +67,35 @@ int main(int argc, char* argv[]) {
...
          boost::beast::flat_buffer abi_buffer;
          stream.read(abi_buffer);
-         std::string abi_string((const char*)abi_buffer.data().data(), abi_buffer.data().size());
-         eosio::json_token_stream token_stream(abi_string.data());
-         eosio::abi_def abidef = eosio::from_json<eosio::abi_def>(token_stream);
-         eosio::convert(abidef, abi);
+         std::string abi_string = boost::beast::buffers_to_string(abi_buffer.data());
+         //remove all tables since their names are invalid; tables not needed for this test
+         std::regex scrub_all_tables(R"(\{ "name": "[^"]+", "type": "[^"]+", "key_names": \[[^\]]*\] \},?)");
+         abi_string = std::regex_replace(abi_string, scrub_all_tables, "");
```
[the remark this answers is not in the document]
spoonincode: Same as above; I want to use ABI from websocket

tests/ship_streamer.cpp
spoonincode: yeah  I guess I can change it to `doesn't support bytes[]` to be more specific

tests/ship_client.cpp
heifner: Oh right. Maybe have a `ship_abi_without_tables(const std::string& abi_string)` ?

tests/ship_streamer.cpp
heifner: Oh is `bytes[]` a supported type? I have used this streamer with production ship servers as a way of testing. If that is something the ship data can contain then it would be nice to support it.

tests/ship_client.cpp
spoonincode: Yeah I thought about that but where to put it? I liked not linking to state_history because it clearly indicates the full isolation from anything "libship" related. I could always make a new tiny lib just for it. Though, ship_client and ship_streamer already seem to be quite a bit of a copy paste in areas so eh..

tests/ship_streamer.cpp
spoonincode: Yes `bytes[]` is supported in ABIs

But you'll never encounter it in the response here since ship_streamer only parses (and then dumps out) the get_block_result i.e.
https://github.com/AntelopeIO/spring/blob/808f061b89a080fbf60cc0636ca0b38b134e523a/libraries/state_history/abi.cpp#L81-L92
(so `bytes?` did need support). ship_streamer doesn't do any deeper inspection of the block, traces, etc

tests/ship_client.cpp
heifner: Sounds fine, just leave as is.

tests/ship_streamer.cpp
heifner: Nice. Thanks.

heifner (approved):

PR #1512, opened 2025-05-09
[1.2.0-rc2] P2P: Versioned signed type for gossip_bp_peers_message::signed_bp_peer

### LINKED ISSUE #1495: Versioned signed bp gossip messages
How to version bp gossip messages or more generally how to version any signed message?

## Option map:
- Use a key->value map of data
- `struct message { int8_t version; std::map<std::string, std::string> data; };` 
  - `std::map<std::string, std::string>`
    - Clients can expect the `std::string` to be a simple string or it could be a `fc::raw::pack` sequence of `char`s.
      -  `Type t; datastream<const char*> ds(s.data(), s.size()); fc::raw::unpack(ds, t);`
      - Both reader/writer would know for any given key what structure the value is expected to be.
      - Data is verified in so much as unpack verifies expected fields
  - Signature is verified over the hash of the entire map
  - V3 creators would put all V3,V2,V1 data into the map
  - V2 creators would put all V2,V1 data into the map
  - V1 readers would look for V1 data in the map and ignore everything else
  - V2 readers would look for V1,V2 data in the map. If no V2 data provided then they use appropriate defaults.
    - If no appropriate defaults then that is a breaking change.

## Option Merkle tree over versions:
- Each version of a message is a node in the Merkle tree.
- Signature is over the root of the tree.
- When V2 sends message to V1 it must provide the hash of the V2 node data so V1 can verify the merkle tree.
- Clients verify tree creating hashes over nodes of data (or using provided) for signature verification

```
 root
  / \   
V3  V2
      \
      V1
```

V3 sending a message to V2 would send hash of V3 data along with V2 and V1.
V2 would validate the Merkle tree signature by hashing the V2 and V1 data, combining with the hash of the V3 data and verify the signature over the Merkle root.
V2 sending message to V3 would send V2 and V1 data. V3 knows the signature verification would be only over the V2 and V1 data merkle tree.

heifner:
Use a versioned signed type for `gossip_bp_peers_message::signed_bp_peer`. This will allow for future additions to the `signed_bp_peer` while maintaining backward compatibility with future signed data.

See #1495 for more details.

Resolves #1495

### DIFF
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -384,10 +384,12 @@ public:
 
       fc::lock_guard g(gossip_bps.mtx);
       auto& sig_idx = gossip_bps.index.get<by_sig>();
-      for (auto i = msg.peers.begin(); i != msg.peers.end() && !invalid_message;) {
+      for (auto i = msg.peers.begin(); i != msg.peers.end();) {
          const auto& peer = *i;
          bool have_sig = sig_idx.contains(peer.sig); // we already have it, already verified
          if (!have_sig && (!is_peer_key_valid(peer) || !is_expiration_valid(peer))) {
+            if (invalid_message)
+               return false;
             // peer key may have changed or been removed on-chain, do not consider that a fatal error, just remove it
             // may be expired, do not consider that fatal, just remove it
             i = msg.peers.erase(i);
@@ -395,9 +397,6 @@ public:
             ++i;
          }
       }
-
-      if (invalid_message)
-         return false;
 
       return true; // empty is checked by caller
    }
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -213,7 +213,7 @@ public:
                if (!initial_updated) {
                   // update initial so always an active one
                   gossip_bp_peers_message::signed_bp_peer signed_empty{{.producer_name = my_bp_account}}; // .server_endpoint not set for initial message
-                  signed_empty.sig = self()->sign_compact(*peer_info->key, signed_empty.digest());
+                  signed_empty.sig = self()->sign_compact(*peer_info->key, signed_empty.digest(self()->chain_id));
                   EOS_ASSERT(signed_empty.sig != signature_type{}, chain::plugin_config_exception,
                              "Unable to sign empty gossip bp peer, private key not found for ${k}", ("k", peer_info->key->to_string({})));
                   initial_gossip_msg_factory.set_initial_send_buffer(signed_empty);
@@ -224,7 +224,7 @@ public:
                gossip_bp_peers_message::signed_bp_peer peer{ {.producer_name = my_bp_account} };
                peer.cached_bp_peer_info.emplace(le.server_endpoint, le.outbound_ip_address, expire);
                peer.bp_peer_info = fc::raw::pack<gossip_bp_peers_message::bp_peer_info_v1>(*peer.cached_bp_peer_info);
-               peer.sig = self()->sign_compact(*peer_info->key, peer.digest());
+               peer.sig = self()->sign_compact(*peer_info->key, peer.digest(self()->chain_id));
                EOS_ASSERT(peer.sig != signature_type{}, chain::plugin_config_exception,
                           "Unable to sign bp peer ${p}, private key not found for ${k}", ("p", peer.producer_name)("k", peer_info->key->to_string({})));
                if (auto i = prod_idx.find(std::make_tuple(my_bp_account, std::cref(le.server_endpoint))); i != prod_idx.end()) {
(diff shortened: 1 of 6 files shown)


plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -137,22 +137,27 @@ namespace eosio {
...
-         bool operator<(const bp_peer& rhs) const {
-            // [producer_name, server_endpoint] is unique
-            return std::tie(producer_name, server_endpoint) < std::tie(rhs.producer_name, rhs.server_endpoint);
-         }
       };
       struct signed_bp_peer : bp_peer {
-         signature_type            sig;
+         signature_type  sig; // signature over bp_peer
+
+         std::optional<bp_peer_info_v1> cached_bp_peer_info; // not serialized
```
linh2931, line 156: Why not follow the pattern in SHiP and TraceAPI that use `std::variant` to represent versions?

heifner: We can do that when we have another version. No reason for a variant now.

linh2931 (approved):
Is it possible to add a unit test for versions, even though we only have 1 version now?

heifner:
> Is it possible to add a unit test for versions, even though we only have 1 version now?

Added a test.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -220,7 +225,8 @@ FC_REFLECT( eosio::request_message, (req_trx)(req_blocks) )
 FC_REFLECT( eosio::sync_request_message, (start_block)(end_block) )
 FC_REFLECT( eosio::block_nack_message, (id) )
 FC_REFLECT( eosio::block_notice_message, (previous)(id) )
-FC_REFLECT( eosio::gossip_bp_peers_message::bp_peer, (producer_name)(server_endpoint)(outbound_ip_address)(expiration) )
+FC_REFLECT( eosio::gossip_bp_peers_message::bp_peer_info_v1, (server_endpoint)(outbound_ip_address)(expiration) )
+FC_REFLECT( eosio::gossip_bp_peers_message::bp_peer, (producer_name)(bp_peer_info) )
```
greg7mdp, line 229: missing `version`.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -404,11 +401,11 @@ class bp_connection_manager {
       auto& idx = gossip_bps.index.get<by_producer>();
       bool diff = false;
       for (const auto& peer : msg.peers) {
-         if (auto i = idx.find(boost::make_tuple(peer.producer_name, boost::cref(peer.server_endpoint))); i != idx.end()) {
-            if (i->sig != peer.sig && peer.expiration >= i->expiration) { // signature has changed, producer_name and server_endpoint has not changed
+         if (auto i = idx.find(std::make_tuple(peer.producer_name, std::cref(peer.server_endpoint()))); i != idx.end()) {
```
greg7mdp, line 404: Did we unpack `peer.cached_bp_peer_info` so we can access `peer.server_endpoint()`?

I don't see where, but if we did, why do we inpack it again on line 408?

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
heifner: Nice catch. Thanks!

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: It is unpacked as part of `validate_gossip_bp_peers_message()`.
You are correct, no reason to unpack it again here.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1512/commits/424962d8d0b13e4a5771eeefbbfd7b5c7664a65d

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -315,34 +305,41 @@ class bp_connection_manager {
...
          return false;
       // initial case, no server_addresses to validate
-      bool initial_msg = msg.peers.size() == 1 && msg.peers[0].server_endpoint.empty();
+      bool initial_msg = msg.peers.size() == 1 && msg.peers[0].bp_peer_info.empty();
       if (!initial_msg) {
          // validate structure and data of msg
-         auto valid_address = [](const std::string& addr) -> bool {
+         auto valid_endpoint = [](const std::string& addr) -> bool {
             const auto& [host, port, type] = net_utils::split_host_port_type(addr);
             return !host.empty() && !port.empty();
```
greg7mdp, line 313: Should we make sure that type is empty?
`return !host.empty() && !port.empty() && type.empty();`

Same at line 143.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -402,7 +399,7 @@ class bp_connection_manager {
       if (invalid_message)
          return false;
```
greg7mdp, line 400: We could move these two lines right before line 391: `// peer key may have changed ... `

greg7mdp (approved):

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: Probably should since it is the value on their `p2p-server-address` or `p2p-listen-endpoint` that matters.

heifner: However, 143 is for `p2p-auto-bp-peer` which is outbound and `type` is valid there.

Issue #1523, opened 2025-05-15
SC: support new sync_call() entry point signature that returns execution status


linh2931:
`sync_call()` performs the validation of the called function name. Currently `sync_call()` entry point is a void function. It needs to be changed to return a status about the execution. Return
* `0` if the function is executed
* `-2` if the version in the data header is not supported (currently it is `0`)
* `-3` if the called function is not found

This issue tracks changes on the Protocol side, including, passing along the return value to the `call` host function, modifying entry point signature validation, and updating tests.

PR #1525, opened 2025-05-16
SC: Support new sync_call entry point function signature which returns an int64_t

linh2931:
`sync_call()` entry point needs to return an `int64_t` to indicate the status of its execution:
* `0`: success
* `-2`:  `version` field in the data header is not supported
* `-3`:  called function is not found in the receiver contract

This PR builds all the plumbing to pass the execution status from WASM execution all the way to host function `call`.

This is required by CDT data validation work.

Resolves https://github.com/AntelopeIO/spring/issues/1523

### DIFF
--- a/libraries/chain/host_context.cpp
+++ b/libraries/chain/host_context.cpp
@@ -66,13 +66,13 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
       throw;
    };
 
-   auto handle_call_failure = [&]()
+   auto handle_call_failure = [&](int64_t error_id)
    {
       auto& call_trace = get_call_trace(ordinal);
-      call_trace.error_id = -1;
+      call_trace.error_id = error_id;
       finalize_call_trace(call_trace, start);
       trx_context.checktime();
-      return -1;
+      return error_id;
    };
 
    last_sync_call_return_value.clear(); // reset for current sync call
@@ -98,12 +98,12 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
 
          const account_metadata_object* receiver_account = &db.get<account_metadata_object, by_name>( call_receiver);
          if (receiver_account->code_hash.empty()) {
-            return handle_call_failure();
+            return handle_call_failure(static_cast<int64_t>(call_error_code::empty_receiver));
          }
 
          const code_object* const codeobject = db.find<code_object, by_code_hash>(boost::make_tuple(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version));
          if (!codeobject || !codeobject->sync_call_supported) {
-            return handle_call_failure();
+            return handle_call_failure(static_cast<int64_t>(call_error_code::sync_call_not_supported_by_receiver));
          }
 
          // use a new sync_call_context for next sync call
@@ -111,7 +111,10 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
 
          try {
             // execute the sync call
-            control.get_wasm_interface().execute(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
+            auto status = control.get_wasm_interface().execute(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
+            if (status < 0) {
+               return handle_call_failure(status);
+            }
          } catch( const wasm_exit&) {}
 
          // Store return value here for the case when the contract sets the
--- a/libraries/chain/include/eosio/chain/host_context.hpp
+++ b/libraries/chain/include/eosio/chain/host_context.hpp
@@ -565,7 +565,13 @@ public:
 
    /// Sync call methods:
 
-   // sync calls can be initiated from actions or other sync calls
+   enum class call_error_code : int64_t {
+      sync_call_not_supported_by_receiver = -1, // receiver contract does not have sync_call entry point
+      unsupported_data_version            = -2, // version in the data header is not supported
+      called_function_not_found           = -3, // called function is not found in the receiver contract
+      empty_receiver                      = -4  // receiver contract code is empty
+   };
+   // Negative status indicates failure, positive or 0 is the size of return value of the called function
    int64_t execute_sync_call(name receiver, uint64_t flags, std::span<const char> data);
    uint32_t get_call_return_value(std::span<char> memory) const;
 
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
@@ -165,7 +165,7 @@ executor::executor(const code_cache_base& cc) {
    mapping_is_executable = true;
 }
 
-void executor::execute(const code_descriptor& code, memory& mem, host_context& context) {
+int64_t executor::execute(const code_descriptor& code, memory& mem, host_context& context) {
    if(mapping_is_executable == false) {
       mprotect(code_mapping, code_mapping_size, PROT_EXEC|PROT_READ);
       mapping_is_executable = true;
@@ -251,6 +251,7 @@ void executor::execute(const code_descriptor& code, memory& mem, host_context& c
 
    context.trx_context.checktime(); //catch any expiration that might have occurred before setting up callback
 
+   int64_t retval = 0;
    switch(sigsetjmp(*cb->jmp, 0)) {
       case 0:
          stack.run([&]{
@@ -272,7 +273,7 @@ void executor::execute(const code_descriptor& code, memory& mem, host_context& c
             } else if (code.call_offset) {
                const auto& ctx = static_cast<eosio::chain::sync_call_context&>(context);
                int64_t(*call_func)(uint64_t, uint64_t, uint32_t) = (int64_t(*)(uint64_t, uint64_t, uint32_t))(cb->running_code_base + *code.call_offset);
-               call_func(ctx.sender.to_uint64_t(), ctx.receiver.to_uint64_t(), static_cast<uint32_t>(ctx.data.size()));
+               retval = call_func(ctx.sender.to_uint64_t(), ctx.receiver.to_uint64_t(), static_cast<uint32_t>(ctx.data.size()));
             } else {
                assert(false);
             }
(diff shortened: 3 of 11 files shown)


libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
```diff
@@ -251,6 +251,7 @@ void executor::execute(const code_descriptor& code, memory& mem, host_context& c
 
    context.trx_context.checktime(); //catch any expiration that might have occurred before setting up callback
 
+   int64_t retval = 0;
```
heifner, line 254: It is not clear to me what `retval` should be for `EOSVMOC_EXIT_CHECKTIME_FAIL`.

heifner (changes requested):
Update PR description for `-4`.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
linh2931: Thanks for the observation.

The `retval` here is used to populate the return code of the `call()` host function, where the caller of the sync call uses it to decide whether to abort the call or to let it as a no-op. That is why the value of `retval` must come from the `sync_call()` entry point; in all other failures it must be set to default `0` and the caller would handle them in other means.

I have added a comment to the code here.

libraries/chain/host_context.cpp
```diff
@@ -98,20 +98,23 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
...
          // use a new sync_call_context for next sync call
          sync_call_context call_ctx(control, trx_context, ordinal, get_current_action_trace(), get_sync_call_sender(), call_receiver, receiver_account->is_privileged(), depth, is_next_call_read_only, data);
 
          try {
             // execute the sync call
-            control.get_wasm_interface().execute(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
+            auto status = control.get_wasm_interface().execute(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
+            if (status < 0) {
+               return handle_call_failure(status);
+            }
```
spoonincode, line 117: There isn't any validation that only `-2` or `-3` are returned. It seems given the current design it really ought to be enforced.

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -565,7 +565,13 @@ class host_context {
...
    /// Sync call methods:
 
-   // sync calls can be initiated from actions or other sync calls
+   enum class call_error_code : int64_t {
+      sync_call_not_supported_by_receiver = -1, // receiver contract does not have sync_call entry point
+      unsupported_data_version            = -2, // version in the data header is not supported
+      called_function_not_found           = -3, // called function is not found in the receiver contract
+      empty_receiver                      = -4  // receiver contract code is empty
+   };
+   // Negative status indicates failure, positive or 0 is the size of return value of the called function
```
spoonincode, line 574: btw I noticed now that maybe these should be `FC_REFLECT_ENUM`ed so they are output pretty in the trace instead of just a number. Assuming we can `FC_REFLECT_ENUM` negatives.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
spoonincode: actually this is very similar context to my other comment,
https://github.com/AntelopeIO/spring/pull/1525#discussion_r2094599332
if the VM exits due to checktime (`EOSVMOC_EXIT_CHECKTIME_FAIL` here) it seems like `context.trx_context.checktime();` _must_ throw, otherwise we have problems.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -244,14 +246,22 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
+            return bkend.timed_run(std::move(wd), std::move(fn));
          } catch(eosio::vm::timeout_exception&) {
             context.trx_context.checktime();
          } catch(eosio::vm::wasm_memory_exception& e) {
             FC_THROW_EXCEPTION(wasm_execution_error, "access violation: ${d}", ("d", e.detail()));
          } catch(eosio::vm::exception& e) {
             FC_THROW_EXCEPTION(wasm_execution_error, "eos-vm system failure: ${d}", ("d", e.detail()));
          }
+
+         // This is to get around `control reaches end of non-void function` compile warning.
```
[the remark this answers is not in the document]
linh2931: Thanks for your detailed analysis!

https://github.com/AntelopeIO/spring/pull/1525/commits/e356fe0028d0fcd907d8294df6c09b861e6d5285

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1525/commits/e356fe0028d0fcd907d8294df6c09b861e6d5285

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -196,20 +196,20 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
...
-            bkend.call(
-                iface, "env", "sync_call",
-                context.sender.to_uint64_t(),
-                context.receiver.to_uint64_t(),
-                static_cast<uint32_t>(context.data.size()));
+            return bkend.call_with_return(
+               iface, "env", "sync_call",
+               context.sender.to_uint64_t(),
+               context.receiver.to_uint64_t(),
+               static_cast<uint32_t>(context.data.size()))->to_i64();
```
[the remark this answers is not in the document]
linh2931: `call_with_return` returns an optional. The `->to_i64()` is to convert it to the expected type. Otherwise compile would fail:

```
.../libraries/chain/webassembly/runtimes/eos-vm.cpp:202:42: error: cannot convert ‘std::optional<eosio::vm::operand_stack_elem>’ to ‘int64_t’ {aka ‘long int’} in return
  202 |             return bkend.call_with_return(
      |                    ~~~~~~~~~~~~~~~~~~~~~~^
      |                                          |
      |                                          std::optional<eosio::vm::operand_stack_elem>
  203 |                iface, "env", "sync_call",
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~
  204 |                context.sender.to_uint64_t(),
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  205 |                context.receiver.to_uint64_t(),
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  206 |                static_cast<uint32_t>(context.data.size()));
```

libraries/chain/include/eosio/chain/host_context.hpp
linh2931: I tested. Negative enums worked with FC_REFLECT_ENUM.

https://github.com/AntelopeIO/spring/pull/1525/commits/92175f6147f18d4cd81a8535288003a35ae20545

libraries/chain/host_context.cpp
```diff
@@ -98,20 +98,25 @@ int64_t host_context::execute_sync_call(name call_receiver, uint64_t flags, std:
...
 
          // use a new sync_call_context for next sync call
          sync_call_context call_ctx(control, trx_context, ordinal, get_current_action_trace(), get_sync_call_sender(), call_receiver, receiver_account->is_privileged(), depth, is_next_call_read_only, data);
 
          try {
             // execute the sync call
-            control.get_wasm_interface().execute(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
+            auto status = control.get_wasm_interface().execute(receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, call_ctx);
+            if (status < 0) {
+               EOS_ASSERT((status == -2 || status == -3), sync_call_invalid_status_exception, // unsupported_data_version or called_function_not_found
```
spoonincode, line 116: This still allows returning positive values

linh2931: The entry point returns 0 or positive values for success, and only -2 or -3 for failure. This EOS_ASSERT enforces the negative return can only be -2 or -3. Do I miss something?

spoonincode: Yeah if you want to allow >=0 to be success this is fine, but the value is just being thrown away

spoonincode (approved):

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
```diff
@@ -281,6 +284,12 @@ void executor::execute(const code_descriptor& code, memory& mem, host_context& c
       //case 1: clean eosio_exit
       case EOSVMOC_EXIT_CHECKTIME_FAIL:
          context.trx_context.checktime();
+
+         // If EOSVMOC_EXIT_CHECKTIME_FAIL is caught, context.trx_context.checktime()
+         // must throw. Otherwise we would have interrupted contract execution
+         // at some unknown time but still considered it completed successfully.
+         assert(false);
+         __builtin_unreachable();
```
spoonincode, line 292: btw if we don't want to put this noise here (and below for eos-vm too), maybe we could make a `[[noreturn]] checktime_must_throw()` to be used in these two locations, and move this check inside of `checktime_must_throw()`. Not required; just thinking about a way to deduplicate.,

libraries/chain/host_context.cpp
linh2931: Thanks for above comment. Actually `sync_call()` entry point can only return the status of `0`, `-2`, and `-3`. If `0`, the status is dicarded and the host function will return the called function's return value size, if `-2` or `-3`, the host function passes the negative value along to the caller. I will add the validation of `0` too.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
linh2931: https://github.com/AntelopeIO/spring/pull/1525/commits/7db87ec5a9cc5a1af9fb4943c9f53b0e9c0c61a5

heifner (approved):

PR #1526, opened 2025-05-16
allow `hash()` to take multiple arguments similar to `pack()`

spoonincode:
This change allows passing `hash()` multiple arguments to concatenate together similar to how `pack()` already does. This can allow some more concise usage of `hash()` without resorting to `make_tuple/pair()` and remembering to avoid the trap of needing to use `cref` along with `make_tuple/pair()` to avoid copies.

I updated `generate_action_digest()` as a good demonstration of the improved conciseness. Of course there are other locations we could potentially change too.

Unfortunately there were a few gnarly quirks that came out of adding just this "simple" variadic template that I will describe below. If it is too gnarly (i.e. too risky) we can explore other approaches that aren't so invasive. For example I have a different branch https://github.com/AntelopeIO/spring/compare/main...vardigest that just creates a new variadic `digest()` call so as not to conflict with any existing `hash()` usages at all.

### DIFF
--- a/libraries/libfc/include/fc/crypto/sha1.hpp
+++ b/libraries/libfc/include/fc/crypto/sha1.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include <fc/fwd.hpp>
 #include <fc/string.hpp>
+#include <fc/io/raw.hpp>
 
 namespace fc{
 
@@ -20,11 +21,11 @@ class sha1
     static sha1 hash( const char* d, uint32_t dlen );
     static sha1 hash( const std::string& );
 
-    template<typename T>
-    static sha1 hash( const T& t ) 
+    template<typename... T>
+    static sha1 hash( const T&... t )
     { 
       sha1::encoder e; 
-      e << t; 
+      fc::raw::pack(e,t...);
       return e.result(); 
     } 
 
--- a/libraries/libfc/include/fc/crypto/sha1.hpp
+++ b/libraries/libfc/include/fc/crypto/sha1.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include <fc/fwd.hpp>
 #include <fc/string.hpp>
+#include <fc/crypto/hash_concepts.hpp>
 #include <fc/io/raw.hpp>
 
 namespace fc{
@@ -22,6 +23,7 @@ class sha1
     static sha1 hash( const std::string& );
 
     template<typename... T>
+    requires NotTwoArgsCharUint32<T...>
     static sha1 hash( const T&... t )
     { 
       sha1::encoder e; 
--- a/libraries/libfc/include/fc/crypto/sha256.hpp
+++ b/libraries/libfc/include/fc/crypto/sha256.hpp
@@ -4,7 +4,7 @@
 #include <fc/fwd.hpp>
 #include <fc/string.hpp>
 #include <fc/platform_independence.hpp>
-#include <fc/io/raw_fwd.hpp>
+#include <fc/io/raw.hpp>
 #include <boost/functional/hash.hpp>
 
 namespace fc
@@ -36,11 +36,11 @@ class sha256
     static sha256 hash( const std::string& );
     static sha256 hash( const sha256& );
 
-    template<typename T>
-    static sha256 hash( const T& t ) 
+    template<typename... T>
+    static sha256 hash( const T&... t )
     { 
       sha256::encoder e; 
-      fc::raw::pack(e,t);
+      fc::raw::pack(e,t...);
       return e.result(); 
     } 
 
--- a/libraries/libfc/include/fc/crypto/sha256.hpp
+++ b/libraries/libfc/include/fc/crypto/sha256.hpp
@@ -4,13 +4,14 @@
 #include <fc/fwd.hpp>
 #include <fc/string.hpp>
 #include <fc/platform_independence.hpp>
+#include <fc/crypto/hash_concepts.hpp>
 #include <fc/io/raw.hpp>
 #include <boost/functional/hash.hpp>
 
 namespace fc
 {
 
-class sha256 
+class sha256
 {
   public:
     sha256();
@@ -37,6 +38,7 @@ class sha256
     static sha256 hash( const sha256& );
 
     template<typename... T>
+    requires NotTwoArgsCharUint32<T...>
     static sha256 hash( const T&... t )
     { 
       sha256::encoder e; 
--- a/libraries/libfc/include/fc/crypto/sha3.hpp
+++ b/libraries/libfc/include/fc/crypto/sha3.hpp
@@ -2,7 +2,7 @@
 #include <fc/fwd.hpp>
 #include <fc/string.hpp>
 #include <fc/platform_independence.hpp>
-#include <fc/io/raw_fwd.hpp>
+#include <fc/io/raw.hpp>
 #include <boost/functional/hash.hpp>
 
 namespace fc
@@ -32,12 +32,23 @@ public:
 	static sha3 hash(const std::string& s, bool is_nist=true) { return hash(s.c_str(), s.size(), is_nist); }
 	static sha3 hash(const sha3& s, bool is_nist=true) { return hash(s.data(), sizeof(s._hash), is_nist); }
 
-	template <typename T>
-	static sha3 hash(const T &t, bool is_nist=true)
+	struct keccak {};
+	struct nist {};
+
+	template<typename... T>
+	static sha3 hash( keccak, const T&... t )
 	{
 		sha3::encoder e;
-		fc::raw::pack(e, t);
-		return e.result(is_nist);
+		fc::raw::pack(e,t...);
+		return e.result(false);
+	}
+
+	template<typename... T>
+	static sha3 hash( nist, const T&... t )
+	{
+		sha3::encoder e;
+		fc::raw::pack(e,t...);
+		return e.result(true);
 	}
 
 	class encoder
--- a/libraries/libfc/include/fc/io/raw.hpp
+++ b/libraries/libfc/include/fc/io/raw.hpp
@@ -176,6 +176,7 @@ namespace fc {
     } FC_RETHROW_EXCEPTIONS( warn, "fc::array<${type},${length}>", ("type",fc::get_typename<T>::name())("length",N) ) }
 
     template<typename Stream, typename T, size_t N>
+    requires (!std::is_same_v<std::remove_cv_t<T>, char>)
     inline void pack( Stream& s, T (&v)[N]) {
       fc::raw::pack( s, unsigned_int((uint32_t)N) );
       for (uint64_t i = 0; i < N; ++i)
@@ -183,6 +184,7 @@ namespace fc {
     }
 
     template<typename Stream, typename T, size_t N>
+    requires (!std::is_same_v<std::remove_cv_t<T>, char>)
     inline void unpack( Stream& s, T (&v)[N])
     { try {
       unsigned_int size; fc::raw::unpack( s, size );
--- a/libraries/chain/include/eosio/chain/action.hpp
+++ b/libraries/chain/include/eosio/chain/action.hpp
@@ -96,20 +96,8 @@ namespace eosio { namespace chain {
    };
 
    inline digest_type generate_action_digest(const action& act, const vector<char>& action_output) {
-      std::array<digest_type,2> hashes;
       const action_base& base = act;
-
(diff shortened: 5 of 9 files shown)


libraries/libfc/include/fc/io/raw.hpp
```diff
@@ -176,13 +176,15 @@ namespace fc {
...
     template<typename Stream, typename T, size_t N>
+    requires (!std::is_same_v<std::remove_cv_t<T>, char>)
     inline void pack( Stream& s, T (&v)[N]) {
       fc::raw::pack( s, unsigned_int((uint32_t)N) );
       for (uint64_t i = 0; i < N; ++i)
          fc::raw::pack(s, v[i]);
     }
 
     template<typename Stream, typename T, size_t N>
+    requires (!std::is_same_v<std::remove_cv_t<T>, char>)
```
spoonincode, line 187: When changing from `raw_fwd.hpp` to `raw.hpp` (a required change afaict), I would get ambiguous call errors between the `T(&v)[N]` (only present in `raw.hpp`) and `char*` overloads, due to a call such as `sha256::hash("foo")`. This seems like an appropriate change to keep C strings using the `char*` overload.

libraries/libfc/include/fc/crypto/sha256.hpp
```diff
@@ -36,11 +37,12 @@ class sha256
     static sha256 hash( const std::string& );
     static sha256 hash( const sha256& );
 
-    template<typename T>
-    static sha256 hash( const T& t ) 
+    template<typename... T>
+    requires NotTwoArgsCharUint32<T...>
```
spoonincode, line 41: The various hash types have some existing overloads of `hash()` such as 
https://github.com/AntelopeIO/spring/blob/ac0abffa690bb963bcbd5e297c8d12c50ee035ac/libraries/libfc/include/fc/crypto/sha256.hpp#L35-L37
This is pretty gnarly on its own because the `std::string` overload doesn't follow the same rules as if sending a `std::string` through the `encoder`. i.e.
```c++
   sha256 a = sha256::hash(std::string("banana"));
   sha256::encoded enc;
   fc::raw::pack(enc, std::string("banana"));
   sha256 b = enc.result();
   // a != b
```
but the real problem is the `hash( const char* d, uint32_t dlen )` overload that will be used, for example,
```c++
std::vector<char> foo;
sha256::hash(foo.data(), foo.size())
```
I need to ensure that calls like `sha256::hash(foo.data(), foo.size())` continue to go to the existing overload, so I need to remove the variadic function from consideration in such cases (it _is_ picked without this).

libraries/libfc/include/fc/crypto/sha3.hpp
```diff
@@ -32,12 +32,23 @@ class sha3
 	static sha3 hash(const std::string& s, bool is_nist=true) { return hash(s.c_str(), s.size(), is_nist); }
 	static sha3 hash(const sha3& s, bool is_nist=true) { return hash(s.data(), sizeof(s._hash), is_nist); }
 
-	template <typename T>
-	static sha3 hash(const T &t, bool is_nist=true)
+	struct keccak {};
+	struct nist {};
```
spoonincode, line 36: There didn't appear to be any existing usages of `sha3::hash()`. I opted to use tag dispatching to select between the two sha3 options and still have multiple arguments. The downside with this is that can't use sha3's hash() somewhere the hash type itself is templated. e.g.
```
template <typename HashType>
HashType do_something(unsigned x) {
   return HashType::hash(x);
}
```

libraries/libfc/include/fc/crypto/sha1.hpp
```diff
@@ -20,11 +22,12 @@ class sha1
     static sha1 hash( const char* d, uint32_t dlen );
     static sha1 hash( const std::string& );
```
heifner, line 23: Personally I think we should rename these to `hash_str` and remove the complicated template concepts.

heifner: Maybe safest to rename the variadic one to `hash_values` so it is clear which one is in-use.

spoonincode: could be a viable option; I was reluctant to make changes across existing code but I can try and see just how pervasive those changes would be. (many tests did fail without the complicated concepts though it's possible only due to a small number root causes)

greg7mdp: Alternatively, you could just add:

```
   template<typename T1, typename T2, typename... T>
   static sha256 hash( const T1& t1, const T2& t2, const T&... t )
    {
      sha256::encoder e;
      fc::raw::pack(e, t1, t2, t...);
      return e.result();
    }
```

to support the case a `hash` called with two or more parameters. I believe it would require no change to existing code or new concepts.

spoonincode: I don't think that solves the problem though. Need std::vector<char> .data() and .size() to go through the 'old' overload. e.g.
```c++
#include <iostream>
#include <vector>
#include <string>
#include <cstddef>
#include <cstdint>
#include <typeinfo>

struct sha256 {};

struct hasher {
    static sha256 hash(const char* d, uint32_t dlen) {
        std::cout << "CALLED: hash(const char*, uint32_t)" << std::endl;
        return {};
    }

    static sha256 hash(const std::string& s) {
        std::cout << "CALLED: hash(const std::string&)" << std::endl;
        return {};
    }

    static sha256 hash(const sha256& h) {
        std::cout << "CALLED: hash(const sha256&)" << std::endl;
        return {};
    }

    template<typename T1, typename T2, typename... T>
    static sha256 hash( const T1& t1, const T2& t2, const T&... t )
    {
        std::cout << "CALLED: Variadic hash"<< std::endl;
        return {};
    }
};

int main() {
    std::vector<char> wif_bytes = {'t', 'e', 's', 't'};
    hasher::hash(wif_bytes.data(), wif_bytes.size());
    return 0;
}
```
when running this get `CALLED: Variadic hash` but we need `CALLED: hash(const char*, uint32_t)`

greg7mdp: That's because the arguments don't match exactly.

If we add the following:

```
    static sha256 hash(char* d, size_t dlen) {
        std::cout << "CALLED: hash(char*, size_t)" << std::endl;
        return {};
    }
```

then it is called. C++ will prefer the non-template function, but the types of the parameters have to match exactly.

spoonincode: It's a little more than that because you need both `hash(char* d, size_t dlen)` and `hash(const char* d, size_t dlen)` (to handle `const std::vector<char>`)

And in for the `char*` case you need to cast it back to const before calling original impl. So for example you'd do something like this

```cpp
    static sha256 hash(const char* d, size_t dlen) {
        std::cout << "CALLED: hash(const char*, size_t)" << std::endl;
        return hash(d, static_cast<uint32_t>(dlen));
    }

    static sha256 hash(char* d, size_t dlen) {
        std::cout << "CALLED: hash(char*, size_t)" << std::endl;
        return hash(static_cast<const char*>(d), static_cast<uint32_t>(dlen));
    }
```
Doing it this way doesn't require the 2+ parameters variant, nor the concepts. But it's 2 more overloads. So would you and @heifner prefer this way or `hash_str()` sort of thing?

heifner: In my opinion, KISS and just have two diff function names.

greg7mdp: Personally, I think it is better this way because:
1. we have the variadic `hash` without having to introduce an alternate name
2. we have the proper, non-template hash function that will accept calls with `vec.data(), vec.size()` parameters

I think it would be better to have the `hash(connst char*, size_t)` being the one with the implementation, esp since I see that for example sha256 ends up calling `OPENSSL_EXPORT int SHA256_Update(SHA256_CTX *sha, const void *data, size_t len);` which takes a size_t.

heifner: I think it is fragile to depend on size_t, what it int or ssize_t is used. My vote is to be explicit.

spoonincode: actually that makes me realize that instead of making concepts to _exclude_ calling of the variatic version, can make concepts to have the "old" one be more _inclusive_. This reads a little better than the original concept stuff. For example this seems to operate correctly
```c++
#include <iostream>
#include <vector>
#include <string>
#include <cstddef>
#include <cstdint>
#include <typeinfo>

struct sha256 {};

template <typename T>
concept CStyleStringPtr = std::is_same_v<T, char*> || std::is_same_v<T, const char*>;

template <typename T>
concept UU = std::is_convertible_v<std::decay_t<T>, size_t>;

struct hasher {
    template<CStyleStringPtr D, UU U>
    static sha256 hash(D d, U dlen) {
        std::cout << "CALLED: legacy hash CStyleStringPtr/UU" << std::endl;
        return {};
    }

    static sha256 hash(const std::string& s) {
        std::cout << "CALLED: hash(const std::string&)" << std::endl;
        return {};
    }

    static sha256 hash(const sha256& h) {
        std::cout << "CALLED: hash(const sha256&)" << std::endl;
        return {};
    }

    template<typename... T>
    static sha256 hash( const T&... t )
    {
        std::cout << "CALLED: Variadic hash";
        return {};
    }
};

int main() {

    const std::vector<char> wif_bytes = {'t', 'e', 's', 't'};
    hasher::hash(wif_bytes.data(), wif_bytes.size());

    uint32_t test = 4;
    std::vector<char> wif2_bytes = {'t', 'e', 's', 't'};
    hasher::hash(wif2_bytes.data(), test);

    return 0;
}
```

greg7mdp: > actually that makes me realize that instead of making concepts to _exclude_ calling of the variatic version, can make concepts to have the "old" one be more _inclusive_. 

I like this! I was somewhat more confident that the version with the two additional string overloads was not changing existing behavior though.

spoonincode: Possibly the `std::is_convertible_v<std::decay_t<T>, size_t>` is too loose, since even an `fc::unsigned_int` will match it (due to it having `operator uint32_t()`), so you can end up with a nasty surprise.

greg7mdp: This is a valid concern for the `CStyleStringPtr ` version.

In my comment I was more worried about the variadic hash version (`hash( const T&... t )`) catching unexpected calls. One way to check that it doesn't happen would be to add an `assert(0)` in its implementation, and run the ci/cd on code that is not yet using the variadic version. If the `asserton` build passes, it will be a good indication that existing uses of `hash` didn't migrate to using the variadic version.

Upon reflection, I think the pattern of `hash` accepting a `char *` + `length` is not a good one, it probably would have been better to always have a single parameter to specify an object to be hashed, whether it is
- a `char*` ( expected to be null-terminated)
- a `std::string` or `std::string_view`
- a `vector<T>` where T is a hashable type (or maybe any iterable container where the `value_type` is hashable)
- a `std::span<T>` where T is a hashable type
- etc...

which would avoid the ambiguity of a two argument `hash` call specifying one or two values to be hashed.

heifner: That can all be avoided by renaming the existing string ones to hash_chars()

spoonincode: It does increasingly feel like renaming the old overloads is the best option, provided there is not too much fallout

spoonincode: I hadn't originally noticed but,
https://github.com/AntelopeIO/spring/blob/ac0abffa690bb963bcbd5e297c8d12c50ee035ac/libraries/libfc/include/fc/crypto/sha256.hpp#L16-L18
so could just eliminate the `hash_chars()` entirely and make these usages use the ctor. idk how I feel about that yet

spoonincode: So the number of calls to the non-templated hash() was substantially more than may have guessed. I prepared a commit where I replace all them to (I hope; didn't audit yet) 1:1 equivalent of new `hash_raw()`, c893000

But a substantial number of these don't need to be 1:1 exact as they were before (like in the benchmark, or many/most of tests), so I guess I can build on top of the commit a new commit that migrates some back to plain `hash()` to limit the number of calls `hash_raw()`

PR #1549, opened 2025-05-26
[2.0.0] P2P: Add {version} support to agent-name

### LINKED ISSUE #1537: Support dynamic version number inclusion in agent name
Example "Spring {version}"

Make this versioned string the default value.

Document this capability in help text

heifner:
Add ability to include the Spring version in the `agent-name` via `{version}`. When `{version}` is included in the `agent-name`, `{version}` will be replaced with the version of `nodeos`.  For example `"Spring {version}"` would generate `Spring 2.0.0`.

```
  --agent-name arg (=Spring {version})  The name supplied to identify this node
                                        amongst the peers. {version} will be
                                        replaced with the version of this node.
```

Also increase the `_agent` logged string size of `peer-log-format` from 15 to 20 characters.

Resolves #1537

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3293,7 +3293,7 @@ namespace eosio {
             return;
          }
 
-         short_agent_name = msg.agent.substr( msg.agent.size() > 1 && msg.agent[0] == '"' ? 1 : 0, 15);
+         short_agent_name = msg.agent.substr( msg.agent.size() > 1 && msg.agent[0] == '"' ? 1 : 0, 20);
          log_p2p_address = msg.p2p_address;
          fc::unique_lock g_conn( conn_mtx );
          p2p_address = msg.p2p_address;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -4317,7 +4317,8 @@ namespace eosio {
            "   myprod,myhostname.com:9876,198.51.100.1\n"
            "   myprod,myhostname2.com:9876,[2001:0db8:85a3:0000:0000:8a2e:0370:7334]"
            )
-         ( "agent-name", bpo::value<string>()->default_value("Vault Agent"), "The name supplied to identify this node amongst the peers.")
+         ( "agent-name", bpo::value<string>()->default_value("Spring {version}"),
```
linh2931, line 4320: Should we append `Spring version` to the agent name if the user does not use the default?

heifner: I don't think so. If someone doesn't want the version in their `agent-name` then they can just leave off `{version}`.

heifner: https://github.com/AntelopeIO/spring/pull/1549/commits/79c63901f50221a306a08aae6f3e5bcbe7309b76

linh2931 (approved):

PR #1551, opened 2025-05-27
[1.2.0-rc3] P2P: agent-name

heifner:
Change default `agent-name` to `Spring Agent".

Related to #1537

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -4317,7 +4317,7 @@ namespace eosio {
            "   myprod,myhostname.com:9876,198.51.100.1\n"
            "   myprod,myhostname2.com:9876,[2001:0db8:85a3:0000:0000:8a2e:0370:7334]"
            )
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -4317,7 +4317,7 @@ namespace eosio {
            "   myprod,myhostname.com:9876,198.51.100.1\n"
            "   myprod,myhostname2.com:9876,[2001:0db8:85a3:0000:0000:8a2e:0370:7334]"
            )
-         ( "agent-name", bpo::value<string>()->default_value("Vault Agent"), "The name supplied to identify this node amongst the peers.")
+         ( "agent-name", bpo::value<string>()->default_value("Spring Agent"), "The name supplied to identify this node amongst the peers.")
```
linh2931, line 4320: Can this be just `Spring`? Or hardcoded to `Spring 2.0` but it causes trouble when merged to `main` branch.

linh2931: Yes, that's what I was worried we need to keep changing the version. Just keep it as is.

spoonincode (approved):

PR #1561, opened 2025-05-29
[2.0.0-dev] when outputting K1 public keys, always use `PUB_K1_` format

spoonincode:
Change nodeos/cleos/keosd to output K1 public keys in the `PUB_K1_` format instead of the "legacy" `EOS...` format. Legacy K1 public keys that begin with `EOS` are still accepted as input, but spring will only output these keys as `PUB_K1...`. This effects everything from `cleos wallet keys` to nodeos' RPC replies (which means even an old cleos will still display keys as `PUB_K1...` since nodeos will have returned those part of its reply).

Thoughts on what else we should do, or if this is "too breaking"? I was thinking maybe a cleos (spring-util?) command to convert between the two formats might be useful. Adding a `--legacy` as suggested in one of the issues below will be tricky, I think.

Work on #1518

Related: VaultaFoundation/mandel#44 EOSIO/eos#5155


linh2931:
> This LGTM.
> 
> An ability to convert between the two formats seems like it would be useful. I would expect people to want a easy way to convert the `EOS...` to `PUB_K1...`. Not sure anyone would want the other way. Seems like a `cleos` feature, not a `spring-util`.

Why not `spring-util`? The conversion does not need nodeos to be running.

heifner:
> Why not `spring-util`? The conversion does not need nodeos to be running.

I see `spring-util` as a tool for node operators. I see `cleos` as a tool for general users. Converting of keys is a general user feature.

spoonincode:
I tend to think cleos is the right spot too. Interestingly we already have a `convert` subcommand group that is named appropriately for where we could stuff it in

programs/cleos/main.cpp
```diff
@@ -2958,6 +2958,24 @@ int main( int argc, char** argv ) {
...
+   auto public_key_to_legacy = convert->add_subcommand("print_public_key_as_legacy", localized("Output a public key in legacy format (EOS, PUB_R1, PUB_WA)"));
+   public_key_to_legacy->add_option("public_key", public_key_to_convert, localized("Public key to output in legacy format"))->required();
+   public_key_to_legacy->callback([&] {
+      try {
+         std::cout << fc::crypto::public_key(public_key_to_convert).to_legacy_string({}) << std::endl;
+      } EOS_RETHROW_EXCEPTIONS(public_key_type_exception, "Failed to parse public key");
+   });
+   // output public key in new format
+   auto public_key_to_new = convert->add_subcommand("print_public_key_as_new", localized("Output a public key in new format (PUB_K1, PUB_R1, PUB_WA)"));
+   public_key_to_new->add_option("public_key", public_key_to_convert, localized("Public key to output in new format"))->required();
```
spoonincode, line 2972: I wanted to come up with something better then 'new format' but :woman_shrugging:

heifner: I think `print_` is a bit verbose. Maybe `convert` `public_key_as_legacy`. Seems like print is implied.

heifner:
```
Usage: ./cleos convert [OPTIONS] SUBCOMMAND

Options:
  -h,--help              Print this help message and exit
  --help-all             Show all help


Subcommands:
  pack_transaction       From plain signed JSON to packed form
  unpack_transaction     From packed to plain signed JSON form
  pack_action_data       From JSON action data to packed form
  unpack_action_data     From packed to JSON action data form
```

Existing `convert` is what it is converting to.
To be consistent I think it would be:

```
legacy_public_key       From existing public key to legacy form, i.e. EOS..
public_key              From existing public key to canonical form, i.e. PUB_..
```

Other ideas instead of `canonical`: `new`, `generic`, `standard`, `current`

heifner (approved):

spoonincode:
huh well yeah we could still split out the cleos convert functionality to main to be able to point to that as part of 1.3 deprecation notice. But I'm not sure it's really worth it.

linh2931 (approved):
Please update the PR comment, as now the commands are `cleos convert legacy_public_key` and `cleos convert public_key`.

linh2931:
Maybe it is too late. I feel the command names `cleos convert legacy_public_key` and `cleos convert public_key` do not tell whether they are converting to or from `legacy_public_key` / `public_key`. `cleos convert to legacy_public_key` and `cleos convert to public_key` more clear? Or `cleos convert to legacy_public_key` and `cleos convert to public_key`? But the `to` adds an extra level of `convert`. We can even have a `convert from` for other use cases in the future.

linh2931:
> Not too late for changes I still need to go through and handle this (I expected to do this sooner which is why I didn't move to draft but now I'm not sure when I'll come back around to finish it off)
> 
> > There are quite a few `EOS...` keys in the documentation; and in bios boot tut. I should modify these as part of this PR
> 
> > Maybe it is too late. I feel the command names `cleos convert legacy_public_key` and `cleos convert public_key` do not tell whether they are converting to or from `legacy_public_key` / `public_key`. `cleos convert to_legacy_public_key` and `cleos convert to_public_key` more clear? Or `cleos convert to legacy_public_key` and `cleos convert to public_key`? But the `to` adds an extra level of `convert`. We can even have a `convert from` for other use cases in the future.
> 
> Yeah it's tough... I feel like currently it's no worse than existing subcommands here though. `convert pack_transaction` -- without knowing anything else does that convert from or to a packed transaction? The help text clarifies what exactly is from and to. And the name of the subcommand is what it is converted to. So we're consistent in that regard.

Yes, the existing `convert` commands do not indicate `to/from` which sometimes is confusing. Maybe for new commands we introduce a new `convert_to` subcommand?

spoonincode:
But subcommands are shown in `cleos --help` so having both a `convert` and `convert_to` listed there seems bloaty and noisy. I thought it was really nice how we were able to stuff these new public key conversions in to an existing top level subcommand that makes contextual sense for them -- it didn't expand the top level subcommand list any

PR #1578, opened 2025-06-04
cleos convert unpack_hex

heifner:
Add new `cleos` `convert` functionality of `unpack_hex`.

I have created a one-off exec one too many times and decided to just add this functionality to `cleos`. It is handy for decoding the deep-mind output that is hex encoded.

### DIFF
--- a/libraries/chain/include/eosio/chain/thread_utils.hpp
+++ b/libraries/chain/include/eosio/chain/thread_utils.hpp
@@ -341,3 +341,31 @@ namespace eosio { namespace chain {
    }
 
 } } // eosio::chain
+
+namespace fc {
+
+template<typename T> struct get_typename<eosio::chain::copyable_atomic<T>> {
+   static const char* name()  {
+      static std::string n = std::string("copyable_atomic<") + get_typename<T>::name() + ">";
+      return n.c_str();
+   }
+};
+
+template<typename T>
+void to_variant(const eosio::chain::copyable_atomic<T>& e, fc::variant& v) {
+   T t = e.load();
+   if constexpr (std::is_same_v<T, bool>) {
+      v = t;
+   } else {
+      to_variant( t, v );
+   }
+}
+
+template<typename T>
+void from_variant(const fc::variant& v, eosio::chain::copyable_atomic<T>& e) {
+   T t;
+   from_variant( v, t );
+   e.store(t);
+}
+
+} // namespace fc
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -2957,6 +2957,71 @@ int main( int argc, char** argv ) {
       std::cout << fc::json::to_pretty_string(unpacked_action_data_json) << std::endl;
    });
 
+   // unpack hex
+   string packed_hex;
+   auto unpack_hex = convert->add_subcommand("unpack_hex", localized("From packed HEX to JSON form"));
+   unpack_hex->add_option("hex", packed_hex, localized("The packed HEX of built-in types including: signed_block, transaction/action_trace, transaction, action, abi_def"))->required();
+   unpack_hex->callback([&] {
+      EOS_ASSERT( packed_hex.size() >= 2, transaction_type_exception, "No packed HEX data found" );
+      vector<char> packed_blob(packed_hex.size()/2);
+      fc::from_hex(packed_hex, packed_blob.data(), packed_blob.size());
+      bool success = false;
+      try {
+         auto v = fc::raw::unpack<block_state_legacy>( packed_blob );
+         std::cout << fc::json::to_pretty_string(v) << std::endl;
+         success = true;
+      } catch (...) {}
+      try {
+         auto v = fc::raw::unpack<signed_block>( packed_blob );
+         std::cout << fc::json::to_pretty_string(v) << std::endl;
+         success = true;
+      } catch (...) {}
+      try {
+         auto v = fc::raw::unpack<transaction_trace>( packed_blob );
+         std::cout << fc::json::to_pretty_string(v) << std::endl;
+         success = true;
+      } catch (...) {}
+      try {
+         auto v = fc::raw::unpack<action_trace>( packed_blob );
+         std::cout << fc::json::to_pretty_string(v) << std::endl;
+         success = true;
+      } catch (...) {}
+      try {
+         auto v = fc::raw::unpack<transaction_receipt>( packed_blob );
+         std::cout << fc::json::to_pretty_string(v) << std::endl;
+         success = true;
+      } catch (...) {}
+      try {
+         auto v = fc::raw::unpack<packed_transaction>( packed_blob );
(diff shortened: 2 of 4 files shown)


ericpassmore:
Nice addition! I've been using to abieos tools to do the same.

ericpassmore (approved):

ericpassmore (approved):

ericpassmore:
> > Nice addition! I've been using to abieos tools to do the same.
> 
> Any missing types I should add?

No. Could be extended to pull contract specific ABI and unpack that. For example `action_results`.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -341,3 +341,31 @@ namespace eosio { namespace chain {
    }
 
 } } // eosio::chain
+
```
linh2931, line 344: Just curious. Why is this put in thread_utils header file?

programs/cleos/main.cpp
```diff
@@ -2957,6 +2957,97 @@ int main( int argc, char** argv ) {
...
+   string type;
+   auto unpack_hex = convert->add_subcommand("unpack_hex", localized("From packed HEX to JSON form"));
+   unpack_hex->add_option("hex", packed_hex, localized("The packed HEX of built-in types including: signed_block, transaction/action_trace, transaction, action, abi_def"))->required();
+   unpack_hex->add_option("--type", type, (localized("Type of the HEX data, if not specified then every type is attempted")));
+   unpack_hex->callback([&] {
+      EOS_ASSERT( packed_hex.size() >= 2, transaction_type_exception, "No packed HEX data found" );
+      vector<char> packed_blob(packed_hex.size()/2);
+      fc::from_hex(packed_hex, packed_blob.data(), packed_blob.size());
+      bool success = false;
+      if (type.empty() || type == "block_state_legacy") {
```
linh2931, line 2971: If `type` is specified and upacking fails, it should not be attempted on other types. Should just return the failure and return.

libraries/chain/include/eosio/chain/thread_utils.hpp
linh2931: I was asking why it was not in a file in FC directory because I saw it was defined in FC name space.

programs/cleos/main.cpp
```diff
@@ -2957,6 +2957,97 @@ int main( int argc, char** argv ) {
...
+         try {
+            auto v = fc::raw::unpack<block_state_legacy>( packed_blob );
+            std::cout << fc::json::to_pretty_string(v) << std::endl;
```
linh2931, line 2974: Not a big deal. If you define `v` at the the beginning, you can do `std::out` only once if `success` is `true`.

libraries/chain/include/eosio/chain/thread_utils.hpp
linh2931: Then no need to change.

linh2931 (approved):

spoonincode:
as is no problems, LGTM to merge as is if you want, but fwiw personally I like the feature set of abieos' tool better. I like that I must specify a type and that built in types work (e.g. I can specify `bool[]` or `public_key?` and it works), and that I can use an abi file to supplement the built in types.

And of course I think we should do a pack_hex if doing this.

programs/cleos/main.cpp
```diff
@@ -2957,6 +2961,201 @@ int main( int argc, char** argv ) {
...
+         auto v = s.variant_to_binary(type, unpacked_json, fc::microseconds::maximum());
+         std::cout << fc::json::to_pretty_string(v) << std::endl;
+         success = true;
+      } catch (...) {}
+      if (!success) {
+         try {
+            if (type == "block_state_legacy") {
+               block_state_legacy t;
+               from_variant(unpacked_json, t);
+               auto v = fc::raw::pack<block_state_legacy>(t);
```
greg7mdp, line 2992: I don't think you need to specify the template parameter (like  `<block_state_legacy>`) in these `pack()` calls.

programs/cleos/main.cpp
```diff
@@ -2957,6 +2961,201 @@ int main( int argc, char** argv ) {
...
+         EOS_ASSERT( abi_file.empty(), misc_exception, "--type required if --abi-file specified");
+      }
+      if (type.empty() || type == "block_state_legacy") {
+         try {
+            auto v = fc::raw::unpack<block_state_legacy>( packed_blob );
+            std::cout << fc::json::to_pretty_string(v) << std::endl;
+            success = true;
+         } catch (...) {}
+      }
+      if (type.empty() || type == "signed_block") {
```
greg7mdp, line 3086:
```suggestion
      if (!success && (type.empty() || type == "signed_block")) {
```

otherwise when `type.empty() ` we will try all `unpacks` even if the first one succeeds.

programs/cleos/main.cpp
```diff
@@ -2957,6 +2961,201 @@ int main( int argc, char** argv ) {
...
+
+      if (!success) {
+         std::cerr << "ERROR: Unable to convert the packed hex";
+         if (!type.empty()) {
+            std::cerr << " to type: " << type;
+         }
+         std::cerr << std::endl;
+      }
+   });
+
```
spoonincode, line 3158: I'd encourage us to seek a way to eliminate the repetitiveness above. A first thought might be to
```cpp
using try_types = std::tuple<block_state_legacy, signed_block, transaction_trace, etc>;

template <typename... Types>
bool try_each_type(const std::tuple<Types...>&, auto&& func) {
    return [&]<std::size_t... Idxs>(std::index_sequence<Idxs...>) {
        return (func.template operator()<std::tuple_element_t<Idxs, std::tuple<Types...>>>() || ...);
    }(std::make_index_sequence<sizeof...(Types)>{});
}

bool success = try_each_type(try_types{}, [&]<typename Type>() {
   try {
      auto v = fc::raw::unpack<Type>( packed_blob );
      std::cout << fc::json::to_pretty_string(v) << std::endl;
      return true;
   } catch(...) {}
   return false;
});
```
though this only works for the automatic case.

For the non-automatic case, maybe you could instantiate an abi_serializer and then load in these types with `add_specialized_unpack_pack()`. So it'd be something like
```cpp
template <typename... Types>
void for_each_type(const std::tuple<Types...>&, auto&& func) {
    [&]<std::size_t... Idxs>(std::index_sequence<Idxs...>) {
        (func.template operator()<std::tuple_element_t<Idxs, std::tuple<Types...>>>(), ...);
    }(std::make_index_sequence<sizeof...(Types)>{});
}

for_each_type(try_types{}, [&]<typename Type>() {
   std::string fully_qualified_name  = boost::core::demangle(typeid(Type).name());
   //XXX strip off everything before last :: here, we don't have something that does it already I guess?
   my_abi_serializer.add_specialized_unpack_pack(stripped_name, pack_unpack<Type>());
});
```

Disclaimer: just improvising -- code above may not work or be ideal.

spoonincode (approved):

PR #1597, opened 2025-06-06
[1.2.0-rc3] P2P: Trim addresses

### LINKED ISSUE #1585: P2P: Trim p2p entries
Trim p2p entries: `p2p-server-address`, `p2p-peer-address`, `p2p-listen-endpoint`, `p2p-auto-bp-peer`, `p2p-bp-gossip-endpoint` so that they are easier to query in `/v1/net/status name`

heifner:
Remove leading-trailing spaces from `p2p-bp-gossip-endpoint` values to avoid peers attempting to connect to a host with a leading or trailing space. Also trim all host names when using them to avoid attempting to use a host from a peer with a leading or trailing space.

Resolves #1585

### DIFF
--- a/plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
@@ -1,8 +1,8 @@
 #pragma once
 
-#include <ostream>
 #include <eosio/chain/exceptions.hpp>
 #include <boost/numeric/conversion/cast.hpp>
+#include <boost/algorithm/string.hpp>
 
 #include <string>
 #include <sstream>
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -51,18 +51,19 @@ namespace detail {
    }
 
    /// @return host, port, remainder
-   inline std::tuple<std::string, std::string, std::string> split_host_port_remainder(const std::string& peer_add, bool should_throw) {
+   inline std::tuple<std::string, std::string, std::string> split_host_port_remainder(const std::string& peer_add_input, bool should_throw) {
       using std::string;
       // host:port[:trx|:blk][:<rate>]
-      if (peer_add.empty()) {
-         EOS_ASSERT(!should_throw, chain::plugin_config_exception, "Address specification is empty" );
+      if (peer_add_input.size() > max_p2p_address_length) {
```
linh2931, line 57: Why not trimming before this?

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
```diff
@@ -51,18 +51,19 @@ namespace detail {
    }
 
    /// @return host, port, remainder
-   inline std::tuple<std::string, std::string, std::string> split_host_port_remainder(const std::string& peer_add, bool should_throw) {
+   inline std::tuple<std::string, std::string, std::string> split_host_port_remainder(const std::string& peer_add_input, bool should_throw) {
```
linh2931, line 54: Not from this PR. It would be more clearer if those `add`s are `addr`s. `add` sounds like the verb "add". Not need to change. Just a thought.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: Because the string could potentially be very large (larger than max_p2p_address_length).

linh2931: I think if a string after trimmed is less than the max size, it should be allowed.

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
heifner: Renamed

plugins/net_plugin/include/eosio/net_plugin/net_utils.hpp
linh2931: This is not a big concern.

linh2931 (approved):

PR #1602, opened 2025-06-10
[2.0.0-dev] replace all legacy keys from docs, tests, tutorials, etc with their canonical form

spoonincode:
Tests in `test_cypher_suites.cpp` and `cli_test.py` retain legacy keys as needed for testing legacy keys.

I guess we will have to be mindful on merges from main that we don't reintroduce legacy keys; we'll probably just have to make another pass closer to 2.0 release


spoonincode:
> thanks for the approvals; I really want to remove `accounts.json` before it becomes part of the main branch history. I don't see any reason we can't just generate all those on the fly. So I'm going to hold on this a little until I can find some time to try and remove the file

I see now, after investigating, this isn't trivial to do because bios_boot_tutorial allows piecemeal starting of various components. So if I generate the keys on the fly with cleos, the keys won't be the same across different innovations of bios_boot_tutorial. I originally didn't expect that to be a problem but it seems like it will be :disappointed:

PR #1606, opened 2025-06-11
[1.1.6] Ensure callback complete before exiting platform_timer stop

### LINKED ISSUE #1601: Test failure: transition_to_if_lr
https://github.com/AntelopeIO/spring/actions/runs/15538267467/job/43743364453

```
  File "/__w/spring/spring/build/tests/transition_to_if.py", line 106, in <module>
    assert cluster.biosNode.waitForBlock(currentHead, blockType=BlockType.lib), "LIB did not advance for setfinalizers to activate"
  File "/__w/spring/spring/build/tests/TestHarness/Node.py", line 208, in waitForBlock
    ret=Utils.waitForBool(lam, timeout, reporter=reporter)
  File "/__w/spring/spring/build/tests/TestHarness/testUtils.py", line 260, in waitForBool
    ret=Utils.waitForObj(myLam, timeout, sleepTime, reporter=reporter)
  File "/__w/spring/spring/build/tests/TestHarness/testUtils.py", line 250, in waitForObj
    raise RuntimeError('waitForObj reached 60 second timeout')
RuntimeError: waitForObj reached 60 second timeout
2025-06-09T15:51:36.665609 Shutting down wallet manager process 598
```

heifner:
Ensure any inflight `platform_timer` callbacks are complete before exiting `platform_timer::stop()`. This makes sure the correct transaction is interrupted when the callback is called.

Resolves #1601

### DIFF
--- a/libraries/chain/platform_timer_asio_fallback.cpp
+++ b/libraries/chain/platform_timer_asio_fallback.cpp
@@ -4,6 +4,7 @@
 #include <fc/fwd_impl.hpp>
 #include <fc/log/logger_config.hpp> //set_os_thread_name()
 
+#include <boost/core/yield_primitives.hpp>
 #include <boost/asio.hpp>
 
 #include <mutex>
@@ -22,6 +23,7 @@ struct platform_timer::impl {
 };
 
 platform_timer::platform_timer() {
+   static_assert(std::atomic<timer_state_t>::is_always_lock_free, "Only lock-free atomics AS-safe.");
    static_assert(sizeof(impl) <= fwd_size);
 
    std::lock_guard guard(timer_ref_mutex);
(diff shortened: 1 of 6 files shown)


libraries/chain/platform_timer_asio_fallback.cpp
```diff
@@ -78,25 +80,30 @@ void platform_timer::start(fc::time_point tp) {
...
       call_expiration_callback();
+      _state.store(timer_state_t{state_t::interrupted, false});
    }
 }
 
 void platform_timer::stop() {
-   const state_t prior_state = _state;
-   if(prior_state == state_t::stopped)
+   while (_state.load().callback_in_flight)
+      boost::core::sp_thread_pause();
```
spoonincode, line 100: I'm worried this might be too early. Is it possible for a thread to `interrupt_timer()` but then call `stop()` before the CAS in `interrupt_timer()` occurs? If so it seems like we might have the same problem still? Because `stop()` would skip the the busy loop since `false` but then the CAS is entered which starts the callback.

So maybe we need another CAS here in `stop() `prior to busy waiting on `callback_in_flight`. The CAS purpose is to atomicly prevent `expire_now()` & `interrupt_timer()` from entering the `call_expiration_callback()`

heifner: Very good point. Thanks.

heifner: https://github.com/AntelopeIO/spring/pull/1606/commits/c509c731ae023543c79cffb710a7df9281b130dd

greg7mdp:
Why not just have a `std::mutex`  in the `platform_timer` ensuring that callbacks are performed atomically with state changes? I think the implementation with multiple atomics is fragile and harder to reason with.

heifner:
> Also, it is not clear to me why we use a single `platform_timer`? Why not have one per block (maybe stored in `building_block`?

One per block does not solve this issue. I could see potentially having one per transaction, but not sure the overhead of creating a new timer for each transaction. Also if each `platform_timer` is ultimately going to SEGFAULT *the* running WASM of the thread then having a new `platform_timer` doesn't really help at all. We would need to somehow tie the `platform_timer` to a particular running set of WASMs associated with a particular transaction and its actions.

greg7mdp:
> One per block does not solve this issue.

Yes, that's not what I meant to imply. It just would have seemed more natural to me to have one per block instead of reusing a global one, but maybe it is a moot point.

> Also if each platform_timer is ultimately going to SEGFAULT the running WASM of the thread

In some cases (like starting production), we want to interrupt all the WASMs threads that are running for this block, right?

greg7mdp:
> Can't safely use mutex from `expire_now()` which is called from signal handling context.

Couldn't we use a mutex safely with `try_lock`, i.e. ignoring the interrupt if the mutex is already locked?
I think in essence this is what we are doing with the `std::atomic` in `expire_now` or `interrupt_timer` (we ignore the calls if the timer is not in the expected `running` state`).

greg7mdp (approved):

PR #1616, opened 2025-06-16
[1.2.0] P2P: BP gossip connections

### LINKED ISSUE #1584: P2P: Do not disconnect from BP gossip peers if manually configured
In `1.2.0-rc2` peers that are BP gossiped and are configured via `p2p-peer-address` are disconnected when the peer is no longer in the top 21 proposers. It will auto-reconnect on the next connection interval.

Instead do not BP auto-disconnect from peers that are manually configured.

heifner:
- Do not disconnect from BP gossip peers if they are manually configured.
- Disconnect from BP gossip peers when no longer in the active proposer schedule.

Resolves #1584

### DIFF
--- a/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
@@ -575,7 +575,7 @@ public:
 
          flat_set<std::string> addresses = find_gossip_bp_addresses(peers_to_drop, "disconnect");
          for (const auto& add : addresses) {
-            self()->connections.disconnect(add);
+            self()->connections.disconnect_gossip_connection(add);
          }
 
          active_schedule_version = schedule.version;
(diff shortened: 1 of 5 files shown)


plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -489,6 +489,17 @@ class bp_connection_manager {
       return addresses;
    }
 
+   flat_set<std::string> all_gossip_bp_addresses(const char* desc) const {
```
greg7mdp, line 492: I think that `boost::unordered_flat_set` would be better here and for `find_gossip_bp_addresses` unless there is a reason to use `boost::flat_set`?

greg7mdp (approved):

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
heifner: https://github.com/AntelopeIO/spring/pull/1616/commits/81183efcf570c98ffdeb4e71df8c61f3a3fbc101

greg7mdp: I see the test fails because it is not ordered... maybe you can convert to a `std::set` for the comparison?

greg7mdp (approved):

PR #1632, opened 2025-06-20
[1.1.6] link with `libatomic` when using `clang`

### LINKED ISSUE #1635: Build issue with Spring's `main` branch
When building locally using clang-18, I get the following issues:

```
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x23
libraries/chain/libeosio_chain.a(platform_timer_posix.cpp.o): in function `std::atomic<eosio::chain::platform_timer::timer_state_t>::compare_exchange_strong(eosio::chain::platform_timer::timer_state_t&, eosio::chain::platform_timer::timer_state_t, std::memory_order, std::memory_order)':
platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x1c1): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x21d): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x276): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x330): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x38c): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: libraries/chain/libeosio_chain.a(platform_timer_posix.cpp.o):platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x3e5): more undefined references to `__atomic_compare_exchange' follow
clang++-20: error: linker command failed with exit code 1 (use -v to see invocation)
```

Linking with `libatomic` seems to fix the issue.

greg7mdp:
Resolves #1635

Fix issue when building `main` with `clang`:

```
/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x23
libraries/chain/libeosio_chain.a(platform_timer_posix.cpp.o): in function `std::atomic<eosio::chain::platform_timer::timer_state_t>::compare_exchange_strong(eosio::chain::platform_timer::timer_state_t&, eosio::chain::platform_timer::timer_state_t, std::memory_order, std::memory_order)':
platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x1c1): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x21d): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x276): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x330): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x38c): undefined reference to `__atomic_compare_exchange'
/usr/bin/ld: libraries/chain/libeosio_chain.a(platform_timer_posix.cpp.o):platform_timer_posix.cpp:(.text._ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_[_ZNSt6atomicIN5eosio5chain14platform_timer13timer_state_tEE23compare_exchange_strongERS3_S3_St12memory_orderS6_]+0x3e5): more undefined references to `__atomic_compare_exchange' follow
clang++-20: error: linker command failed with exit code 1 (use -v to see invocation)
```


CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ else()
     if( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" )
         if( CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 4.0.0 OR CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.0.0 )
             set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-invalid-partial-specialization" )
+            set( CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -latomic")
```
heifner, line 112: Not clear to me if `target_link_libraries( eosio_chain PRIVATE -latomic)` in `libraries/chain/CMakeLists.txt` would be better.

spoonincode, line 112: What is the rationale for doing this only when using Clang? Are you trying to link to it only when it is required? But it's only needed currently when using Clang and libstdc++, not just simply when using Clang. Also there are probably other variables in play as well, such as the platform CPU.

I went down the rabbit hole a little and didn't really find anything conclusive on how to know when linking to libatomic is required/necessary. Some projects attempt to detect its necessity during configure time (like a cmake `try_compile()`) by compiling some atomic-using code. But that's fragile because you would need to exhaustively test all permutations of your atomic usages to be sure. For example,
```cpp
struct big {
    uint64_t a;
    uint64_t b;
};
void one(std::atomic<big>& x) {
   x.store({2, 5});
}
void two(std::atomic<__int128>& x) {
   x.store(0);
}
```
Both of these will generate a call to libatomic with gcc or clang, even though I'm pretty sure both of them are nominally lockfree even in libatomic's impl for x86-64. I'm guessing it defers to libatomic because the strategy to use for a 128-bit atomic differs based on x86 features (and at one point even the manufacturer of the x86 core).

So I wonder what the downside is to just always link to libatomic on Linux[^1]?

If we want to use the library only when required it seems like `--push-state --as-needed -latomic --pop-state` works well. For example,
```
g++     -O3 --std=c++20                 atom.cpp main.cpp  -o gcc      -Wl,--push-state -Wl,--as-needed -latomic -Wl,--pop-state
clang++ -O3 --std=c++20                 atom.cpp main.cpp  -o clang    -Wl,--push-state -Wl,--as-needed -latomic -Wl,--pop-state
clang++ -O3 --std=c++20 --stdlib=libc++ atom.cpp main.cpp  -o clangc++ -Wl,--push-state -Wl,--as-needed -latomic -Wl,--pop-state

ldd gcc | grep atomic
(nothing)
ldd clang | grep atomic
	libatomic.so.1 => /usr/lib/libatomic.so.1 (0x00007fd2e2aa0000)
ldd clangc++ | grep atomic
(nothing)
```

However this will still error out if libatomic is nonexistent (for example on macos). So I wonder if the ultra paranoid and generic approach is to first detect presence of libatomic during configure time, and then only if it exists do a `--push-state --as-needed -latomic --pop-state`. 

[^1]: Perhaps rule ought be non-mac? Because if you're on freebsd or cross-compiling win32 with gcc you'd still want libatomic? But what if you're actually using gcc on macos? Do we need libatomic there? ugh this is the wild uncertainty with when or when not to link with it.

greg7mdp: I just pushed a commit to use `-latomic` on linux regardless of the compiler, just to see what happens in ci/cd. 
I tend to prefer simple solutions unless there is a problem with them, but I'm happy to go with what you think is best.

spoonincode: That would probably be more idiomatic, especially as long as we're just linking libatomic and not doing more hairy linker commands. Wouldn't it need to be `PUBLIC` though :thinking:

heifner (approved):

PR #1640, opened 2025-06-23
Add stricter abi version checkes in `cleos` and `abi_serializer`

### LINKED ISSUE #1580: cleos: Respect ABI semantic versioning and preserve version when writing out ABI
With regards to cleos...

### Respect Semantic Versioning
The version of the ABI being read must be less than or equal to the version of ABIs that the software supports, otherwise reject the ABI with an appropriate error message.

Example: the version of the software supports ABIs up to 1.2, it can read ABIs with versions 1.1 or 1.2, but it cannot read ABIs with version 1.3+

### Preserve Version
Any tool that is reading an ABI and then writing it back out (example JSON to Binary), must preserve the original version that the ABI wrote.

greg7mdp:
Resolves #1580.

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -136,11 +136,9 @@ namespace eosio { namespace chain {
       auto abi_version = abi.get_version();
       EOS_ASSERT(!!abi_version, unsupported_abi_version_exception, "Unrecognized abi version: ${v}",
                  ("v", abi.version));
-#if 0
       EOS_ASSERT(*abi_version <= current_abi_support, unsupported_abi_version_exception,
                  "ABI has an unsupported version: ${v}, current nodeos supports up to: ${vc}",
-                 ("v", std::string(version_str))("vc", current_abi_support.str()));
-#endif
+                 ("v", abi.version)("vc", current_abi_support.str()));
 
       size_t types_size = abi.types.size();
       size_t structs_size = abi.structs.size();
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -133,18 +133,14 @@ namespace eosio { namespace chain {
 
    void abi_serializer::set_abi(abi_def abi, const yield_function_t& yield) {
       impl::abi_traverse_context ctx(yield, fc::microseconds{});
-
-      static constexpr std::string version_header = "eosio::abi/";
-      EOS_ASSERT(starts_with(abi.version, version_header), unsupported_abi_version_exception,
-                 "Unexpected abi version string, should start with: " + version_header);
-
-      std::string_view version_str(abi.version.c_str() + version_header.size(), abi.version.size() - version_header.size());
-      version_t abi_version(version_str);
-      EOS_ASSERT(abi_version.is_valid(), unsupported_abi_version_exception, "Unrecognized abi version: ${v}",
-                 ("v", std::string(version_str)));
-      EOS_ASSERT(abi_version <= current_abi_support, unsupported_abi_version_exception,
+      auto abi_version = abi.get_version();
+      EOS_ASSERT(!!abi_version, unsupported_abi_version_exception, "Unrecognized abi version: ${v}",
+                 ("v", abi.version));
+#if 0
+      EOS_ASSERT(*abi_version <= current_abi_support, unsupported_abi_version_exception,
                  "ABI has an unsupported version: ${v}, current nodeos supports up to: ${vc}",
                  ("v", std::string(version_str))("vc", current_abi_support.str()));
+#endif
 
       size_t types_size = abi.types.size();
       size_t structs_size = abi.structs.size();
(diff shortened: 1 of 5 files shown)


libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -1,8 +1,37 @@
...
+         valid = (std::from_chars(ptr+1, last, minor).ec == std::errc());
+   }
+
+   friend auto operator<=>(const version_t&, const version_t&) = default;
+   friend bool operator==(const version_t&, const version_t&) = default;
+
+   std::string str() const {
+      std::ostringstream oss;
+      oss << std::to_string(major) << '.' << std::to_string(minor);
+      return oss.str();
```
heifner, line 28: `ostringstring` seems like a lot when this could just be `return std::to_string(major) + "." + std::to_string(minor);`

libraries/chain/abi_serializer.cpp
```diff
@@ -133,8 +133,12 @@ namespace eosio { namespace chain {
...
    void abi_serializer::set_abi(abi_def abi, const yield_function_t& yield) {
       impl::abi_traverse_context ctx(yield, fc::microseconds{});
-
-      EOS_ASSERT(starts_with(abi.version, "eosio::abi/1."), unsupported_abi_version_exception, "ABI has an unsupported version");
+      auto abi_version = abi.get_version();
+      EOS_ASSERT(!!abi_version, unsupported_abi_version_exception, "Unrecognized abi version: ${v}",
+                 ("v", abi.version));
+      EOS_ASSERT(*abi_version <= current_abi_support, unsupported_abi_version_exception,
+                 "ABI has an unsupported version: ${v}, current nodeos supports up to: ${vc}",
+                 ("v", abi.version)("vc", current_abi_support.str()));
```
heifner, line 141: I think we need some tests that verify get table rows handles this.
https://github.com/AntelopeIO/spring/blob/a0a56fbbd8abc39d267e6aaa6b299817c998d97d/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp#L688-L688
https://github.com/AntelopeIO/spring/blob/a0a56fbbd8abc39d267e6aaa6b299817c998d97d/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp#L800-L800

heifner (changes requested):

libraries/chain/include/eosio/chain/abi_def.hpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1640/commits/4df6c8ceeab86c3126153a8740ab4d404bb73874

libraries/chain/abi_serializer.cpp
greg7mdp: handle in what sense? The ` get_table_row` in chain_api_plugin uses `CALL_WITH_400` which does handle exceptions occuring during the call.

heifner: I think get table row should return the data un-deserialized if it can't use the ABI. Not just fail completely.

libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -131,14 +163,27 @@ struct abi_def {
    extensions_type                           abi_extensions;
    may_not_exist<vector<variant_def>>        variants;
    may_not_exist<vector<action_result_def>>  action_results;
+
+   std::optional<version_t> get_version() const {
+      using boost::algorithm::starts_with;
```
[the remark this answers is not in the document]
greg7mdp: Thanks, I think it is better to avoid using boost here, avoids adding an additional include in `abi_def.hpp`.

libraries/chain/abi_serializer.cpp
greg7mdp: I'm not sure, I think it might be better to fail completely, so that the user realizes that there is a problem and uses a up-to-date API provider.

heifner: Reminder: If `json` is not requested don't call `set_abi` and avoid exception.

greg7mdp: https://github.com/AntelopeIO/spring/pull/1640/commits/31beefa74f4a0395c75434bd8c56a90f0ac5c7f0

heifner (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -684,13 +684,14 @@ class read_only : public api_base {
...
-         abi_serializer abis;
-         abis.set_abi(std::move(abi), abi_serializer::create_yield_function(abi_serializer_max_time));
-         auto table_type = abis.get_table_type(p.table);
-         
+
          for (auto& row : p.rows) {
             fc::variant data_var;
             if( p.json ) {
+               abi_serializer abis;
+               abis.set_abi(std::move(abi), abi_serializer::create_yield_function(abi_serializer_max_time));
```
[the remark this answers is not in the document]
greg7mdp: Duh! Thanks for catching this!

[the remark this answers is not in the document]
greg7mdp: https://github.com/AntelopeIO/spring/pull/1640/commits/08a40c6766236216e1ec59408b4dfa37c9bd5fd6

heifner (approved):

spoonincode (approved):

PR #1645, opened 2025-06-23
Use `std::string::start_with()` instead of the boost version.

greg7mdp:


heifner:
Any reason this is `release/2.0` instead of `main`?

greg7mdp:
> Any reason this is `release/2.0` instead of `main`?

Not really, I can change it if you think it is better.

spoonincode (approved):

PR #1679, opened 2025-07-03
SHiP: Verify last log index entry matches last log entry

### LINKED ISSUE #990: SHiP: should check last entry's position in log file matches last entry in index file
Chosen approach
=============
SHiP should compare last entries of log and index files. In the event that they don't match, it should shutdown with a descriptive error message.

Lin's original report below
===================
This is related to test failure 1 of https://github.com/AntelopeIO/spring/pull/983

Currently SHiP verifies the number of entries in the index file matches the number of entries in the log file. It does not verifies the value of the last entry's position in log file when validating the last entry.

heifner:
Add additional sanity check at startup of last entry's position in log file matches last entry in index file. 

Resolves #990


heifner:
Diff approach needed.

PR #1680, opened 2025-07-03
SHiP: Verify last log index entry matches last log entry

### LINKED ISSUE #990: SHiP: should check last entry's position in log file matches last entry in index file

heifner:
Add additional sanity check at startup of last entry's position in log file matches last entry in index file. 

Resolves #990

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -434,7 +434,7 @@ private:
          //verify last index position matches last log entry
          const uint64_t index_pos = get_pos(_end_block-1);
          FC_ASSERT(index_pos == last_header_pos, "Last index position ${ip} does not match last entry in log ${lp}", ("ip", index_pos)("lp", last_header_pos));
-      } EOS_RETHROW_EXCEPTIONS(chain::plugin_exception, "${name} is corrupted and cannot be repaired", ("name", index.display_path()));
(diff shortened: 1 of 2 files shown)


libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -419,6 +420,23 @@ class state_history_log {
       ilog("${name} regeneration complete", ("name", index.display_path()));
    }
 
+   void check_log_and_index_on_init() {
+      if(log.size() == 0)
+         return;
+
+      try {
+         log_header first_header = log.unpack_from<decltype(first_header)>(0);
```
greg7mdp, line 428: is the `decltype` needed? Why not just `log.unpack_from<log_header>`?

spoonincode: yeah that's what something like clang-tidy will suggest. But generally I don't like using `auto` where unnecessary, especially for such a simple case

heifner: Personally I would prefer just `log_header first_header = log.unpack_from<log_header>(0);`
I don't see the point of changing it as it matches other usages in the file.

PR #1700, opened 2025-07-09
[1.2.1] TraceAPI: Fix for finding trx after starting from snapshot

### LINKED ISSUE #1693: Trace API can't find transactions if started from snapshot

heifner:
When starting from a snapshot, not all stride files are present. Instead of attempting to find the first stride file and failing, loop through the trx id files on disk. Also now searches in reverse order assuming the trx searched for is more likely to be in a recent block.

TODO: add unittest.

Resolves #1693


greg7mdp:
Is that new solution slower than using the stride files if present? Should we try the stride files first, and in case of failure use this new implementation?

greg7mdp (approved):

PR #1668, opened 2025-07-01
[2.0.0-dev2] Protocol feature: packed_transaction_restrictions

heifner:
- Adds protocol feature: `PACKED_TRANSACTION_RESTRICTIONS`.
  - Enforces no extra-data in `packed_transaction` in blocks
  - Enforces no compressed `packed_transaction` in blocks
- Subjectively enforces no extra-data in `packed_transaction` via P2P and API.
- Subjectively normalizes all `packed_transaction` to non-compressed for transfer on P2P and in blocks.

Resolves #1472

PR #1708, opened 2025-07-10
[2.0.0-dev] Protocol Feature: Remove Canonical Signature Check

### LINKED ISSUE #1106: Protocol Feature: Remove Canonical Signature Check
Requiring signatures to be canonical is necessary to prevent Signature Malleability attacks.

However, in Antelope-based chains, this requirement serves no functional purpose since the signature does not affect the transaction hash generation.

Consequently, even if an attacker alters the signature, they cannot exploit it.

The primary problem with canonical checks is that they require the signer to repeat the entire signing process until a canonical signature is found. This can be quite time-consuming for applications that need high-speed signing.

greg7mdp:
Resolves #1106.

In `apply_block()`,  if the `allow_non_canonical_signatures` protocol feature has not been activated yet, we validate that all transaction and block signatures included in the incoming `signed_block` are canonical.

We also disabled all canonical signature checks done when an incoming block is added to the fork database, as we don't know at the time if one of its ancestor block will activate the `allow_non_canonical_signatures` protocol feature.

This PR also, incidentally:
- updates the `public_key` constructor in libfc to take a non-optional `check_canonical_t` parameter (as opposed to the previous `bool check_canonical = true` optional parameter.
- add a new way to specify a set of protocol features to activate in tester (`setup_action::activate_features_up_to`) and use it to simplify the tester configs: `setup_policy::old_wasm_parser` and `setup_policy::before_disable_deferred_trx`.
- add a new `savanna` `setup_policy` in tester which activates and transitions to `savanna`, but doesn't activate any `Spring 2.0` features.
- simplify some tests to remove duplicated tests doing exactly the same work. For example, using `testers` or `validating_testers` (`using testers = boost::mpl::list<legacy_tester, savanna_tester>`) makes sense only if the tester is constructed with `setup_policy::full` because that's what `legacy_tester` checks to change the policy, otherwise both testers are constructed with the same policy and are identical.
- updated tester and unittests to avoid signing with canonical signatures if tester is initialized with `allow_non_canonical_signatures` activated.


libraries/chain/controller.cpp
```diff
@@ -4307,7 +4315,8 @@ struct controller_impl {
                     const flat_set<digest_type>& cur_features,
                     const vector<digest_type>& new_features )
             { check_protocol_features( timestamp, cur_features, new_features ); },
-            skip_validate_signee
+            skip_validate_signee,
+            check_canonical()
```
heifner, line 4319: Can't do this here. This is off the main thread. `is_builtin_activated` is not thread safe.

libraries/chain/webassembly/crypto.cpp
```diff
@@ -50,7 +50,7 @@ namespace eosio::chain::webassembly {
          EOS_ASSERT(s.variable_size() <= context.control.configured_subjective_signature_length_limit(),
                     sig_variable_size_limit_exception, "signature variable length component size greater than subjective maximum");
 
-      auto check = fc::crypto::public_key( s, *digest, false );
+      auto check = fc::crypto::public_key( s, *digest, fc::check_canonical_t::no );
```
heifner, line 53: Don't we want to relax this when the protocol feature is enabled? Same for `recover_key`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1012,6 +1012,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                  transaction_metadata_ptr trx_meta;
                  try {
                     trx_meta = transaction_metadata::recover_keys(trx, chain.get_chain_id(), time_limit, trx_type,
+                                                                  chain.check_canonical(),
```
heifner, line 1015: `check_canonical()` can't be called off the main thread.

heifner (changes requested):

libraries/chain/webassembly/crypto.cpp
greg7mdp: It was not doing the canonical check even when before the protocol feature.

greg7mdp:
> Why? [098b83d](https://github.com/AntelopeIO/spring/commit/098b83d4d8cc0c94fb7c5feeffafeb059c1c1d28) Legacy is pre-savanna which means this new protocol feature can't be active.

My thoughts originally!
But then I thought that even though savanna was activated on our chain soon after the feature was activated, the software does not guarantee it, and some other chain could activate the `savanna` feature and the `allow_non_canonical_sigs` feature *before* the `setfinalizers` causing the transition to savanna.

libraries/chain/controller.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1708/commits/74ac3486d588cdc62211abbb3eab839b1b5f5c0f

great that the tests failed because of this.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1708/commits/74ac3486d588cdc62211abbb3eab839b1b5f5c0f

heifner:
Bad news: I think this complete approach is not right.

- We need to validate trxs have canonical signatures when validating a block before `builtin_protocol_feature_t::allow_non_canonical_signatures`
- Current impl does not do this.

Good news: I think the correct approach is fairly trivial.

Goals: Allow non-canonical signatures after protocol feature ALLOW_NON_CANONICAL_SIGNATURES is activated.

I believe you can just use `check_canonical_t::no` everywhere. Then add something like the following to `transaction_context::init_for_input_trx` (see https://github.com/AntelopeIO/spring/pull/1668)

```
      if( !control.is_builtin_activated(builtin_protocol_feature_t::allow_non_canonical_signatures) ) {
         packed_trx.validate_canonical_signatures();
      }
```

Where `validate_canonical_signatures` will loop through the signatures on the transaction and throw if any are not `public_key::is_canonical`.

unittests/protocol_feature_tests.cpp
```diff
@@ -2315,6 +2315,60 @@ BOOST_AUTO_TEST_CASE(set_finalizers_test) { try {
...
+   c.create_accounts( {alice} );
+   auto b1 = c.produce_block();\
+   t.push_block(b1);
+
+   signed_transaction trx;
+   for (size_t i=0; i<10; ++i) {
+      trx = c.create_dummy_transaction(alice, std::to_string(i));
+      auto sig = trx.sign(get_private_key(alice, "active"), c.control->get_chain_id(), fc::require_canonical_t::no);
+      if (!sig.is_ecc_canonical()) {
+         std::cout << "Found non-canonical signature after " << i << " iterations\n";
```
heifner, line 2334: Use `dlog` so not always printed.

unittests/protocol_feature_tests.cpp
```diff
@@ -2315,6 +2315,60 @@ BOOST_AUTO_TEST_CASE(set_finalizers_test) { try {
...
+   c.create_accounts( {alice} );
+   auto b1 = c.produce_block();
+   t.push_block(b1);
+
+   signed_transaction trx;
+   for (size_t i=0; i<10; ++i) {
+      trx = c.create_dummy_transaction(alice, std::to_string(i));
+      auto sig = trx.sign(get_private_key(alice, "active"), c.control->get_chain_id(), fc::require_canonical_t::no);
+      if (!sig.is_ecc_canonical()) {
+         std::cout << "Found non-canonical signature after " << i << " iterations\n";
```
heifner, line 2362: Use `dlog` so not always printed.

heifner (changes requested):

unittests/protocol_feature_tests.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1708/commits/2e32b6abf14446735f21dcece3d8530565b69b68

heifner (approved):

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -120,8 +120,10 @@ namespace eosio { namespace chain {
       vector<signature_type>    signatures;
       vector<bytes>             context_free_data; ///< for each context-free action, there is an entry here
 
-      const signature_type&     sign(const private_key_type& key, const chain_id_type& chain_id);
-      signature_type            sign(const private_key_type& key, const chain_id_type& chain_id)const;
+      const signature_type&     sign(const private_key_type& key, const chain_id_type& chain_id,
+                                     fc::require_canonical_t require_canonical = fc::require_canonical_t::yes);
```
linh2931, line 124: Can this not use a default value?

libraries/chain/transaction.cpp
```diff
@@ -70,7 +70,7 @@ fc::microseconds transaction::get_signature_keys( const vector<signature_type>&
          auto now = fc::time_point::now();
          EOS_ASSERT( now < deadline, tx_cpu_usage_exceeded, "transaction signature verification executed for too long ${time}us",
                      ("time", now - start)("now", now)("deadline", deadline)("start", start) );
-         auto[ itr, successful_insertion ] = recovered_pub_keys.emplace( sig, digest );
+         auto[ itr, successful_insertion ] = recovered_pub_keys.emplace( sig, digest, fc::check_canonical_t::no );
```
linh2931, line 73: Do you need to check `allow_non_canonical_signatures` is true?

spoonincode:
Sorry I've only taken a brief look -- I think this PR is likely missing the change from the proposal,
> R1 signatures will no longer require the low-s check

Because I see this line here unchanged,
https://github.com/AntelopeIO/spring/blob/2e32b6abf14446735f21dcece3d8530565b69b68/libraries/libfc/src/crypto/elliptic_r1.cpp#L509-L510

This low-s check actually came up recently as an annoyance for exsat when (I think if I understood correctly) doing webauth verification inside a contract. So I think we do really want to remove the canonical check for r1 signatures.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -66,8 +66,10 @@ namespace eosio::testing {
    enum class setup_action {
       preactivate_protocol_feature,
       set_before_producer_authority_bios_contract,
-      activate_features,     // activate features listed in `setup_policy::features`. enforce dependencies.
-      activate_all_features, // activate all available features
+      activate_features,       // activate features listed in `setup_policy::features`. enforce dependencies.
+      activate_features_up_to, // activate all features up to the first one specified in `setup_policy::features`
+                               // if `setup_policy::features` contains more than one the remaining are also activated
```
linh2931, line 71: How do you determine the set of protocol features before a particular one? By dependencies?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4310,7 +4310,7 @@ namespace eosio {
          }
          chain::public_key_type peer_key;
          try {
-            peer_key = crypto::public_key(msg.sig, msg.token, true);
+            peer_key = crypto::public_key(msg.sig, msg.token, fc::check_canonical_t::no);
```
linh2931, line 4313: We just assume the protocol feature has already activated?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1152,8 +1152,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          auto private_key_itr = _signature_providers.find(key);
          EOS_ASSERT(private_key_itr != _signature_providers.end(), producer_priv_key_not_found,
                     "Local producer has no private key in config.ini corresponding to public key ${key}", ("key", key));
-
-         return private_key_itr->second(digest);
+         return private_key_itr->second(digest, fc::require_canonical_t::no);
```
linh2931, line 1155: No need to check whether the protocol feature has been activated?

libraries/libfc/test/crypto/test_cypher_suites.cpp
```diff
@@ -60,7 +60,7 @@ BOOST_AUTO_TEST_CASE(test_r1_recovery) try {
    auto pub = key.get_public_key();
    auto sig = key.sign(digest);
 
-   auto recovered_pub = public_key(sig, digest);
+   auto recovered_pub = public_key(sig, digest, fc::check_canonical_t::yes);
```
spoonincode, line 63: It seems like these might be the only cases of `fc::check_canonical_t::yes`?

Does it make sense to just remove this ability from the `public_key` ctor entirely? Because other then these two lines[^1] the approach seems to be "always not care" about canonical at recovery time, and then check `is_ecc_canonical()` as appropriate (based on protocol feature activation). And this approach is really nice and clean.

So these two usages of `fc::check_canonical_t::yes` and all the plumbing needed in the ctor to support it sticks out a little.

But I can understand not wanting to make more changes than necessary.

[^1]: Which can be replaced with a `sig.is_ecc_canonical(); public_key(... fc::check_canonical_t::no); `

spoonincode:
> Thanks for catching that Matt! I saw that the parameter `check_canonical` was not used, so I assumed that there was no canonical check for r1 signatures.

Right, frustratingly the r1 canonical check was always clutched in. That also means this https://github.com/AntelopeIO/spring/pull/1708#discussion_r2198076146 will have to be revisited too with r1's behavior being different post-activation.

> So I assume I should make the check you highlighted conditional on `check_canonical` for this code to be correct, right?

If I'm understanding right I think you're going to have to plumb it through to `is_ecc_canonical().` The problem is.. loading up these bigints to do comparison on them is going to be kind of a bummer here :disappointed:  There are probably more clever approaches we could do (precompute the half order or something) but maybe it's not worth it.

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -371,6 +371,20 @@ host function call will trigger a transition to the Savanna consensus algorithm.
 Builtin protocol feature: SYNC_CALL
 
 Enables synchronous calls to functions in other and own contracts.
+*/
+            {builtin_protocol_feature_t::savanna}
+         } )
+         (  builtin_protocol_feature_t::allow_non_canonical_signatures, builtin_protocol_feature_spec{
+            "ALLOW_NON_CANONICAL_SIGNATURES",
```
linh2931, line 378: I feel the `non` in the name causing some confusions when reading code together with `require_canonical_t::{yes, no}`. Like `if (!allow_non_canonical_signatures) use require_canonical_t::yes; ... `

greg7mdp:
> I think you're going to have to plumb it through to `is_ecc_canonical().`

yes.

> That also means this [#1708 (comment)](https://github.com/AntelopeIO/spring/pull/1708#discussion_r2198076146) will have to be revisited too with r1's behavior being different post-activation.

I think the way the code is now the issue in the comment is addressed for both cases.

spoonincode:
> Although unfortunately there is not a version of `is_canonical()` already coded for r1.

Right, though it shouldn't be too hard to graft out
https://github.com/AntelopeIO/spring/blob/2e32b6abf14446735f21dcece3d8530565b69b68/libraries/libfc/src/crypto/elliptic_r1.cpp#L497-L510
it's just a "lot" of math relative to the simple bit checks of k1 :(

> I think the way the code is now the issue in the comment is addressed for both cases.

The problem is you'd need either a `fc::check_canonical_t::no` or `fc::check_canonical_t::yes` there based on the PF activation, because otherwise the behavior of r1 keys will be different. (or, alternatively, always `fc::check_canonical_t::no` and do the is `is_ecc_canonical()` based on PF)

greg7mdp:
> always `fc::check_canonical_t::no` and do the is `is_ecc_canonical()` based on PF)

Yes, I think I need to do that.

greg7mdp:
> If I'm understanding right I think you're going to have to plumb it through to `is_ecc_canonical().` The problem is.. loading up these bigints to do comparison on them is going to be kind of a bummer here 😞

Hey @spoonincode , inplemented what I think is correct in https://github.com/AntelopeIO/spring/pull/1708/commits/ea0bdaa32278168040d95d2e3035abdc1f9a78af. Seems like, unlike `ecc` signatures where the `canonical` can be checked directly on the signature, for `r1` signatures we have to recover the public key and check that the public key is canonical. Did I understand correctly?

libraries/chain/block_header_state_legacy.cpp
```diff
@@ -429,10 +429,10 @@ namespace eosio::chain {
 
       std::set<public_key_type> keys;
       auto digest = sig_digest();
-      keys.emplace(fc::crypto::public_key( header.producer_signature, digest, true ));
+      keys.emplace(fc::crypto::public_key( header.producer_signature, digest, fc::check_canonical_t::no ));
```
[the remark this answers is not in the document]
greg7mdp: See this comment https://github.com/AntelopeIO/spring/pull/1708#issuecomment-3057983817

libraries/chain/transaction.cpp
greg7mdp: No, we don't check here anymore, we do the check when the block is applied, and we know whether the protocol feature is activated.

libraries/libfc/src/crypto/elliptic_r1.cpp
```diff
@@ -521,6 +524,26 @@ namespace fc { namespace crypto { namespace r1 {
...
+    bool public_key::is_canonical( const compact_signature& c ) {
+        int nV = c.data[0];
+        if (nV<27 || nV>=35)
+            FC_THROW_EXCEPTION( exception, "unable to reconstruct public key from signature" );
+
+        ecdsa_sig sig = ECDSA_SIG_new();
+        BIGNUM *r = BN_new(), *s = BN_new();
+        BN_bin2bn(&c.data[1],32,r);
+        BN_bin2bn(&c.data[33],32,s);
+        ECDSA_SIG_set0(sig, r, s);
```
spoonincode, line 536: afaict all you need from the cluster above is just
```cpp
BIGNUM *s = BN_new();
BN_bin2bn(&c.data[33],32,s);
```
`sig` is not used in the rest of the function, ergo `r` is not used either

libraries/chain/include/eosio/chain/transaction.hpp
greg7mdp: restore default param in https://github.com/AntelopeIO/spring/pull/1708/commits/c1df814ba177648935a81701fca1574d8240d19f
because of usage in system-contracts repo.
```
/__w/spring/spring/system-contracts/tests/eosio.system_tester.hpp:1553:18: error: no matching function for call to 'eosio::chain::signed_transaction::sign(fc::crypto::private_key, eosio::chain::chain_id_type)'
2440
 1553 |          trx.sign( get_private_key( "producer1111"_n, "active" ), control->get_chain_id()  );
2441
      |          ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

libraries/chain/protocol_feature_manager.cpp
greg7mdp: This is the clearest name I can think of for the protocol feature. Do you have a suggestion?

libraries/libfc/src/crypto/elliptic_r1.cpp
```diff
@@ -521,6 +528,27 @@ namespace fc { namespace crypto { namespace r1 {
         FC_THROW_EXCEPTION( exception, "unable to reconstruct public key from signature" );
     }
 
+    bool public_key::is_canonical( const compact_signature& c ) {
+        int nV = c.data[0];
+        if (nV<27 || nV>=35)
+            FC_THROW_EXCEPTION( exception, "unable to reconstruct public key from signature" );
+
+        BIGNUM s;
```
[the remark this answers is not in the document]
greg7mdp: Yes, you're right! https://github.com/AntelopeIO/spring/pull/1708/commits/c3cb16529d9c5aee3ecc6aad23bd8a3e9d465c4f

plugins/net_plugin/net_plugin.cpp
spoonincode: There might be a compatibility problem here though. Say you have a 2.0 node connected to a 1.2 node, with peer keys configured. The 2.0 node when creating the handshake signature will,
https://github.com/AntelopeIO/spring/blob/c3cb16529d9c5aee3ecc6aad23bd8a3e9d465c4f/plugins/net_plugin/net_plugin.cpp#L4366
which goes to
https://github.com/AntelopeIO/spring/blob/c3cb16529d9c5aee3ecc6aad23bd8a3e9d465c4f/plugins/net_plugin/net_plugin.cpp#L4337-L4345
Now this function above is a little wonky to me: if it takes line 4341 then it goes to
https://github.com/AntelopeIO/spring/blob/c3cb16529d9c5aee3ecc6aad23bd8a3e9d465c4f/libraries/libfc/include/fc/crypto/private_key.hpp#L30
which makes a canonical signature (default param). But if it takes line 4343 then we'll go to
https://github.com/AntelopeIO/spring/blob/c3cb16529d9c5aee3ecc6aad23bd8a3e9d465c4f/plugins/producer_plugin/producer_plugin.cpp#L1150-L1159
This implementation does _not_ make a canonical signature.

So if the 2.0 node takes 4343 it will make a handshake signature that the 1.2 node will not accept, since in 1.2 it did require the signature to be canonical.

I can't immediately tell why net_plugin has this funky thing of trying to merge the private keys from 2 separate locations -- so maybe taking 4343 isn't going to happen for peer handshakes but just the BP peering thing? Maybe @heifner can shed some light.

spoonincode:
I think everything is good; let me just sit on it one more night and I'll approve tomorrow

plugins/net_plugin/net_plugin.cpp
heifner: It appears, according to the help text, that the authenticate peer option, `allowed-connection`, is suppose to support: `any`, `producers`, `specified`, or `none`. So I think the intention was for `net_plugin::sign_compact` to use either the explicit configured `peer-private-key` or a producer key according to the configuration. Unless I'm looking at that wrong, it appears that doesn't work correctly and if `specified` or `producers` is configured then it will allow either a producer key or an explicit configured key to work.

Regardless, the `peer-private-key` is not used for bp gossip, so that will take the producer plugin route which uses the `signature_provider_plugin`.

As for compatibility, it is true there is a compatibility issue with 1.2 and 2.0. However, the peer authentication feature is not widely used. I know of only one entity that has used it recently and I don't think they even use it anymore. Anyone that is using it can easily re-generate a key. In my mind it is not worth the effort to maintain compatibility.

libraries/libfc/include/fc/crypto/public_key.hpp
```diff
@@ -28,7 +30,8 @@ namespace fc { namespace crypto {
          public_key( const public_key& ) = default;
          public_key& operator= (const public_key& ) = default;
 
-         public_key( const signature& c, const sha256& digest, bool check_canonical = true );
+         public_key( const signature& c, const sha256& digest );
+         public_key( const signature& c, const sha256& digest, bool check_r1_canonical ); // potential canonical check only for r1 sigs
```
heifner, line 34: Seems like this should get the `enum` treatment like the other changes.

libraries/libfc/src/crypto/signature.cpp
```diff
@@ -53,6 +53,20 @@ namespace fc { namespace crypto {
       }, _storage);
    }
 
+   bool signature::is_ecc_canonical() const {
```
heifner, line 56: With recent changes, should this be renamed back to `is_canonical()`?

libraries/libfc/test/crypto/test_webauthn.cpp
```diff
@@ -52,7 +52,7 @@ BOOST_AUTO_TEST_CASE(good) try {
 
    memcpy(auth_data.data(), origin_hash.data(), sizeof(origin_hash));
 
-   BOOST_CHECK_EQUAL(wa_pub, make_webauthn_sig(priv, auth_data, json).recover(d, true));
+   BOOST_CHECK_EQUAL(wa_pub, make_webauthn_sig(priv, auth_data, json).recover(d, check_canonical_t::yes));
```
heifner, line 55: Since `webauthn` are not canonical, seems like we should change all of these to pass `no`.

programs/cleos/main.cpp
```diff
@@ -4279,7 +4279,7 @@ int main( int argc, char** argv ) {
          try {
             priv_key = private_key_type(str_private_key);
          } EOS_RETHROW_EXCEPTIONS(private_key_type_exception, "Invalid private key")
-         trx.sign(priv_key, *chain_id);
+         trx.sign(priv_key, *chain_id, fc::require_canonical_t::yes);
```
heifner, line 4282: Should we consider making `canonical` an option to `cleos sign` ?
What about `keosd`, should `keosd` have a `config.ini` option to sign `canonical` or not?

heifner (changes requested):

libraries/libfc/include/fc/crypto/public_key.hpp
greg7mdp: I thought it was used in just one location, and where it is used we are passing an expression (`!is_builtin_activated(builtin_protocol_feature_t::allow_non_canonical_signatures)`), not passing a blind constant like true/false), so I think bool is fine here even if it seems admitedly inconsistent.

libraries/libfc/src/crypto/signature.cpp
greg7mdp: haha yes will do.

libraries/libfc/include/fc/crypto/public_key.hpp
greg7mdp: That would makke the call location change from:
```
      auto recovered = fc::crypto::public_key(
         s, *digest, !context.control.is_builtin_activated(builtin_protocol_feature_t::allow_non_canonical_signatures));
```

to

```
      auto recovered = fc::crypto::public_key(
         s, *digest,
         context.control.is_builtin_activated(builtin_protocol_feature_t::allow_non_canonical_signatures)
            ? fc::check_canonical_t::no
            : fc::check_canonical_t::yes);
```

which doesn't seem better to me.

programs/cleos/main.cpp
greg7mdp: What about we discuss this in standup and create another issue if we decide to do it. I feel it doesn't have to be included in this PR, which is already plenty big.

libraries/libfc/src/crypto/signature.cpp
greg7mdp: https://github.com/AntelopeIO/spring/pull/1708/commits/3b85a2184061016dbe3cd312f2238773963bb291

heifner (approved):

plugins/net_plugin/net_plugin.cpp
spoonincode: Okay thanks for helping with that. Totally agree with the rarity of its use and being non-consensus this isn't a blocker.

libraries/chain/protocol_feature_manager.cpp
spoonincode: fwiw I usually dislike "logic puzzle" names (which this one certainly classifies as :laughing: ) but my vote would be to keep it. Only other thing I could immediately think of is `NO_CANONICAL_SIGNATURE_CHECK` or `REMOVE_CANONICAL_SIGNATURE_CHECK` (at least these have only 1 positive/negative).

libraries/libfc/include/fc/crypto/public_key.hpp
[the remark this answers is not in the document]
greg7mdp: Interesting, let me look into it. You are right that the `public_key` contructor with a bool is new just for that r1 check.

libraries/chain/protocol_feature_manager.cpp
linh2931: `REMOVE_CANONICAL_SIGNATURE_CHECK` is better. But no need to change it now.

programs/cleos/main.cpp
greg7mdp: https://github.com/AntelopeIO/spring/issues/1757

linh2931 (approved):

PR #1711, opened 2025-07-10
[2.0.0-dev] Enforce gxx version 11

ericpassmore:
Update top level CMakeLists.txt to check for and require gnu version 11 or better. Clear error message for someone with older build, for example Ubuntu 20, that may pick up earlier version of the compilers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,8 +37,8 @@ set( SPRING_UTIL_EXECUTABLE_NAME spring-util )
 
 # http://stackoverflow.com/a/18369825
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -37,8 +37,8 @@ set( SPRING_UTIL_EXECUTABLE_NAME spring-util )
 
 # http://stackoverflow.com/a/18369825
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.2)
-        message(FATAL_ERROR "GCC version must be at least 10.2.0!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0)
+        message(FATAL_ERROR "GCC version must be at least 11.0.0!")
```
greg7mdp, line 41: I think we should require 11.4 which is what is on ubuntu22 and is checked on ci/cd.

greg7mdp (changes requested):

PR #1719, opened 2025-07-10
try 2: allow `hash()` to take multiple arguments similar to `pack()`

spoonincode:
This is another attempt to allow passing `hash()` multiple arguments to concatenate together similar to how `pack()` already does. This can allow some more concise usage of `hash()` without resorting to `make_tuple/pair()` and remembering to avoid the trap of needing to use `cref` along with `make_tuple/pair()` to avoid copies.

Based on this comment chain https://github.com/AntelopeIO/spring/pull/1526#discussion_r2093536377 I have modified the hash impls such that,
* `hash_raw()` takes a `ContiguousCharSource` that it hashes directly -- this is the previous behavior when using these nontemplated call forms:
https://github.com/AntelopeIO/spring/blob/e3caa447b9742bb1bf190f53315cb34522542917/libraries/libfc/include/fc/crypto/sha1.hpp#L20-L21
* `hash()` now is variadic and hashes with the same rules of our serialization format

Of course this means that usage of `hash(std::string())` now has different output between these two revisions so any usages of that which matter (many don't!) need to be migrated to `hash_raw(std::string())`.

So the way I approached this was to first convert all calls to the two forms above to `hash_raw()`, and then go back and make a determination for sites that are fine to transition to simple `hash()` in order to limit the number of `hash_raw()` calls. But that was a huge pain: so this PR just leaves all calls to the two forms above as `hash_raw()` but that means the change set is large.

Ultimately I'm not really all that confident in making such a huge change.

### DIFF
--- a/benchmark/hash.cpp
+++ b/benchmark/hash.cpp
@@ -25,62 +25,62 @@ void hash_benchmarking() {
    }
 
    auto sha1_small_msg = [&]() {
-      fc::sha1::hash(small_message);
+      fc::sha1::hash_raw(small_message);
    };
    benchmarking("sha1 (" + std::to_string(small_message.length()) + " bytes)", sha1_small_msg);
 
    auto sha1_large_msg = [&]() {
-      fc::sha1::hash(large_message);
+      fc::sha1::hash_raw(large_message);
    };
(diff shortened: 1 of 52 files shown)


heifner (approved):

spoonincode:
going with #1785

PR #1723, opened 2025-07-11
[2.0.0-dev] bump cmake requirement to 3.21; remove <3.21 workaround

spoonincode:
Remove a workaround for cmake <3.21. There doesn't seem to be a compelling reason to keep support for <3.21 since Ubuntu22 is on cmake 3.22. Debian11 has 3.18 out of the box but ships with gcc10 which isn't going to cut it -- so no real argument that keeping support for <3.21 helps Debian any either. EL9 ships with gcc11 and has cmake 3.26 available so that's fine too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required( VERSION 3.16...4.0 )
+cmake_minimum_required( VERSION 3.21...4.0 )
 
 project( antelope-spring )
 include(CTest)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -263,26 +263,14 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tests/TestHarness DESTINATION ${CM
         PATTERN "__pycache__" EXCLUDE
         PATTERN "CMakeFiles" EXCLUDE)
 
-if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21)
-   # Cmake versions < 3.21 did not support installing symbolic links to a directory via install(FILES ...)
-   # Cmake 3.21 fixed this (https://gitlab.kitware.com/cmake/cmake/-/commit/d71a7cc19d6e03f89581efdaa8d63db216184d40)
-   # If/when the lowest cmake version supported becomes >= 3.21, the else block as well as the postinit and prerm scripts
-   # would be a good option to remove in favor of using the facilities in cmake / cpack directly.
```
spoonincode, line 270: I'm not sure what exactly this is suggesting we do, particularly since the install script does a runtime discovery of where to install the package. I don't think there is anything analogous in cpack. So idk.. I just removed the comment.. if someone has something they want back we can put something back.

heifner (approved):

Issue #1628, opened 2025-06-20
SC: Cleos support for Sync Calls


linh2931:
`cleos` needs to be updated to support sync calls:
1. Print out `call_traces` inside `action_trace` for `get_actions` command
2. Print out `call_traces` inside `action_trace` when printing out push transaction and push action
3. Print out console output following the call hierarchy.

PR #1724, opened 2025-07-11
SC: Cleos support of displaying call traces and console logs

linh2931:
Add `cleos` support to display call traces and console logs.

Please note: in dev-preview-2,  `expand_console()` will be combined with the one in `trace.cpp` class.

Resolves https://github.com/AntelopeIO/spring/issues/1628

An example `cleos` session looks like

```
./cleos push action caller addaction '[10, 20]' -p caller@active
executed transaction: 7845a14b0cb16071f858c65a5dc7ebee104e26a27b4b1e1e4cfc4b098fbc0ead  104 bytes  682 us
#                   caller <= caller::addaction         {"x":10,"y":20}
#              callee::add <= caller                    {"x":10,"y":20}
=>                                return value: 30

CONSOLE OUTPUT BEGIN =====================
before add()
[caller->(callee,add)]: CALL BEGIN ======
inside add()
[caller->(callee,add)]: CALL END   ======
after add()
CONSOLE OUTPUT END   =====================
warn  2025-07-11T13:04:52.128 cleos     main.cpp:935                  print_result        warning: transaction executed locally, but may not be confirmed by the network yet
```

The `addaction` action is
```
[[eosio::action]]
void addaction(uint32_t x, uint32_t y) {
   sync_callee::add_func add{ "callee"_n };
   eosio::print("before add()");
   eosio::check(add(x, y) == x + y, "x + y was not correct");
   eosio::print("after add()");
}
```
The `add` sync call function is

```
[[eosio::call]]
uint32_t sync_callee::add(uint32_t x, uint32_t y) {
   eosio::print("inside add()");
   return x + y;
}
```


programs/cleos/main.cpp
```diff
@@ -606,6 +606,137 @@ void print_return_value( const fc::variant& at ) {
    }
 }
 
+constexpr uint32_t receiver_width = 14;
+constexpr uint32_t function_width = 30;
```
heifner, line 610: I doubt it matters but this was 28 before.

linh2931: Because of potential long names, I have made them all 30.

spoonincode (approved):

PR #1726, opened 2025-07-14
[1.2.2] P2P: Reset sync on rejected block

### LINKED ISSUE #1673: sync issue (`block not accepted`) when syncing in to 2.0 dev1 chain
It recovered but shouldn't have had this glitch in the first place right?

[dev1huh.txt](https://github.com/user-attachments/files/21007539/dev1huh.txt)

### LINKED ISSUE #1716: Test failure: nodeos_read_terminate_at_block_if_lr_test
https://github.com/AntelopeIO/spring/actions/runs/16200499530/job/45738833997

```
  File "/__w/spring/spring/build/tests/nodeos_read_terminate_at_block_test.py", line 363, in <module>
    assert cluster.getNode(nodeId).waitForLibToAdvance(), f"LIB did not advance for {nodeId}"
AssertionError: LIB did not advance for 8
```

heifner:
Reverts #1676. A recent test failure #1716 demonstrates the need for the reset. The test failed because it timed out waiting for LIB to advance. The node did actually advance LIB but not in the expected 30 second timeout. The issue was that an unlinkable block at transition from LIB catchup to HEAD catchup didn't reset sync and therefore the node didn't know it needed to continue syncing. The handshake was not sufficient because the other node thought it had caught up. This was exasperated by the node syncing was in IRREVERSIBLE mode. It is possible that we could live with only resetting when running in IRREVERSIBLE mode. However, I think the safest and most direct approach is to go back to resetting the sync, but guard against resetting the sync when know we have already reset the sync. Therefore, fixing #1673 in a different way.

Resolves #1716  
Resolves #1673

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2400,7 +2400,7 @@ namespace eosio {
       {
          // reset sync on rejected block
          fc::lock_guard g( sync_mtx );
-         if (blk_num <= sync_next_expected_num-1) { // no need to reset if we already reset and are syncing again
+         if (sync_last_requested_num != 0 && blk_num <= sync_next_expected_num-1) { // no need to reset if we already reset and are syncing again
             sync_last_requested_num = 0;
             sync_next_expected_num = my_impl->get_fork_db_root_num() + 1;
          }


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2398,6 +2397,14 @@ namespace eosio {
    // called from connection strand
    void sync_manager::rejected_block( const connection_ptr& c, uint32_t blk_num, closing_mode mode ) {
       c->block_status_monitor_.rejected();
+      {
+         // reset sync on rejected block
+         fc::lock_guard g( sync_mtx );
+         if (sync_last_requested_num != 0 && blk_num <= sync_next_expected_num-1) { // no need to reset if we already reset and are syncing again
```
greg7mdp, line 2403: it is not clear to me why this check indicates that we `already reset`?

PR #1740, opened 2025-07-21
Only attempt trx if previous execution time less than remaining block time

### LINKED ISSUE #1736: Do not attempt trx if previous execution time > remaining block time
We already track previously billed amount for a transaction if it was speculatively executed on a node. We use this to factor into subjective billing. 

https://github.com/AntelopeIO/spring/blob/1c6c379af188fb615c17de0a2e2aed45c590ff53/plugins/producer_plugin/producer_plugin.cpp#L2580-L2590

We could also use this to skip scheduling the trx for execution in the current block if not enough block time remains to allow for the trx execution assuming it executes for the same amount of time.

heifner:
- If a transaction has been previously executed, then only schedule it to execute if there is enough time left in the block for it to finish. Assuming it takes the same amount of time as last execution.
  - Evaluated both remaining time of a block and remaining billable CPU of a block.
- Adds `transaction_metadata.elapsed_time_us`
  - Since billed time is not always the same as elapsed time and there are plans to increase these differences.
  - Use new `elapsed_time_us` for existing `max_trx_time` calculation since there are plans to use CPU for other billing outside execution time.
- Removes unused `transaction_metadata.accepted`.

Resolves #1736

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -2584,7 +2584,7 @@ producer_plugin_impl::push_result producer_plugin_impl::push_transaction(const f
       const uint64_t block_cpu_limit = rl.get_block_cpu_limit();
       const fc::microseconds block_time_remaining_us = block_deadline - start;
 
-      fc_tlog(_log, "prev cpu ${pc}us, prev elapsed ${p}us, cpu left ${c}us, time left ${t}us, tx: ${txid}",
+      fc_tlog(_log, "prev cpu ${pc}us, prev elapsed ${p}us, block cpu limit ${c}us, time left ${t}us, tx: ${txid}",
               ("pc", prev_billed_cpu_time_us)("p", prev_elapsed_time_us)("c", block_cpu_limit)("t", block_time_remaining_us)("txid", trx->id()));
       // no use attempting to execute if not enough time left in block for what it took previously
       if (block_time_remaining_us.count() < prev_elapsed_time_us || block_cpu_limit < prev_billed_cpu_time_us ) {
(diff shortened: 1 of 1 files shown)


unittests/chain_tests.cpp
```diff
@@ -198,4 +198,73 @@ BOOST_AUTO_TEST_CASE_TEMPLATE( signal_validated_blocks, T, testers ) try {
...
+      BOOST_REQUIRE(empty_block);
+      BOOST_TEST(empty_block->transactions.empty());
+      last_trace = nullptr;
+      signed_block_ptr block = chain.produce_block();
+      BOOST_REQUIRE(block);
+      BOOST_REQUIRE(last_trace);
+      BOOST_TEST(create_account_trace->id == last_trace->id);
+      BOOST_TEST(create_account_trace->block_num < last_trace->block_num); // different block
+      BOOST_TEST(trx_meta->elapsed_time_us == std::max(last_trace->elapsed.count(), create_account_trace->elapsed.count())); // verify trx_meta updated
+      BOOST_TEST(block->block_num() == last_trace->block_num);
```
greg7mdp, line 255: Is the mix of `BOOST_TEST` and `BOOST_REQUIRE` intentional? I'm not seeing the benefit of using `BOOST_TEST`?

greg7mdp: `BOOST_REQUIRE` checks a bool, so I don't think it requires a conversion to string. Maybe you're thinking of `BOOST_REQUIRE_EQUAL` of `BOOST_REQUIRE_LE`. These offer a nicer output, so for example I think:
` BOOST_REQUIRE_LT(create_account_trace->block_num, last_trace->block_num);`
would have better output than:
`BOOST_TEST(create_account_trace->block_num < last_trace->block_num);`

heifner: Examples of failures:
```
/home/heifner/ext/leap/unittests/chain_tests.cpp(253): error: in "chain_tests/signal_applied_transaction<eosio__testing__tester>": check create_account_trace->block_num < last_trace->block_num-10 has failed [13 >= 4]
/home/heifner/ext/leap/unittests/chain_tests.cpp(254): error: in "chain_tests/signal_applied_transaction<eosio__testing__legacy_tester>": check trx_meta->elapsed_time_us != std::max(last_trace->elapsed.count(), create_account_trace->elapsed.count()) has failed [218 == 218]
```

greg7mdp: Wow! Thanks I learned something today :-).

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2577,13 +2577,31 @@ producer_plugin_impl::push_result producer_plugin_impl::push_transaction(const f
...
+            pr.block_exhausted = block_is_exhausted(); // smaller trx might fit
+         pr.trx_exhausted = true;
+         fc_dlog(trx->is_transient() ? _transient_trx_failed_trace_log : _trx_failed_trace_log,
+                 "[TRX_TRACE] Block ${bn} for producer ${prod} COULD NOT FIT, prev elapsed ${e}us, block cpu limit ${bl}, tx: ${txid} RETRYING ",
+                 ("bn", chain.head().block_num() + 1)("prod", get_pending_block_producer())("e", prev_elapsed_time_us)("bl", block_cpu_limit)("txid", trx->id()));
+         return pr;
+      }
+      if (prev_billed_cpu_time_us > 0 && !subjective_bill.is_account_disabled(first_auth) && !rl.is_unlimited_cpu(first_auth)) {
+         int64_t prev_time_us = std::max<int64_t>(prev_billed_cpu_time_us, prev_elapsed_time_us);
+         int64_t prev_time_plus100_us = prev_time_us + EOS_PERCENT(prev_time_us, 100 * config::percent_1);
```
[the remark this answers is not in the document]
heifner: https://github.com/AntelopeIO/spring/pull/1740/commits/6d55de396202a2cce1049d2db9437fb2bcdb1be1

linh2931 (approved):

PR #1785, opened 2025-07-25
try 3: add `packhash()` to take multiple arguments similar to `pack()` and hash together

spoonincode:
This is the 'throw in the towel' variant of #1526 & #1719. This version just adds a new `packhash()` for each hash type that takes variable number of args. So no headache about potentially changing some behavior of existing `hash()` and its odd overloads.


libraries/libfc/include/fc/crypto/packhash.hpp
```diff
@@ -0,0 +1,24 @@
...
+template<typename Derived>
+struct add_packhash_to_hash {
+   template<typename... T>
+   requires ( sizeof...(T) > 0 )
+   static auto packhash(const T&... args) {
+      return fc::packhash<typename Derived::encoder>(args...);
+   }
+};
+
+}
```
spoonincode, line 24: I am willing to hear that we should just copy/paste `packhash()` in to every hash class instead of doing the funky CRTP which injects it. I felt like `packhash()` being right alongside `hash()` was important. (i.e. `sha256::hash()` and `sha256::packhash()`)

libraries/libfc/include/fc/io/raw.hpp
```diff
@@ -176,13 +176,15 @@ namespace fc {
...
     template<typename Stream, typename T, size_t N>
+    requires (!std::is_same_v<std::remove_cv_t<T>, char>)
     inline void pack( Stream& s, T (&v)[N]) {
       fc::raw::pack( s, unsigned_int((uint32_t)N) );
       for (uint64_t i = 0; i < N; ++i)
          fc::raw::pack(s, v[i]);
     }
 
     template<typename Stream, typename T, size_t N>
+    requires (!std::is_same_v<std::remove_cv_t<T>, char>)
```
spoonincode, line 187: Remember this change from https://github.com/AntelopeIO/spring/pull/1526#discussion_r2093475193. Need to pull in raw.hpp to get the variadic pack(), but that causes an ambiguity with c strings.

libraries/libfc/include/fc/crypto/sha3.hpp
```diff
@@ -56,6 +65,13 @@ class sha3
 		fc::fwd<impl, 1016> my;
 	};
 
+	struct nist_encoder : public encoder {
+		sha3 result() { return encoder::result(true); }
+	};
+	struct keccak_encoder : public encoder {
+		sha3 result() { return encoder::result(false); }
+	};
+
```
spoonincode, line 74: sha3 is funky because of this nist vs keccak stuff going on. I opted to keep with the tag dispatching but it means more bespoke plumbing for this hash type & packhash()

greg7mdp (commented):
Since we are retiring `digest()`, I wonder if we could use `digest` instead of `packhash`, I feel it is a nicer name.

libraries/libfc/include/fc/crypto/packhash.hpp
greg7mdp: I like it as it is personally.

heifner (approved):

PR #1786, opened 2025-07-28
[1.2.2] P2P: Block nack reset fix

### LINKED ISSUE #1784: P2P: close not sending block nack reset for active connection
Currently on `net_plugin` close we have:

https://github.com/AntelopeIO/spring/blob/3182f7d4c369616f32a125b78d823c61f939e624/plugins/net_plugin/net_plugin.cpp#L1477-L1480

The intention is that if we are closing a connection that we are actively receiving blocks from we should reset block nack to all other connections so they are sending us blocks again instead of block nacks.

The problem with this approach is the delay between chain head and received blocks can be such that we can miss sending out a block nack reset when we should. 

Instead seems like we should track last block received across all connections and compare that with the last block received on the connection at close to know if a block nack reset should be sent to all other connections.

This appears to be what caused the fork of https://github.com/AntelopeIO/spring/issues/1753

heifner:
Track `last_block_num_received` across all connections and use that instead of chain head for determination if block nack reset should be sent. Also reset `consecutive_blocks_nacks` when block nacks are not consecutive.

Resolves #1784

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -407,7 +407,7 @@ namespace eosio {
       std::atomic<fc::time_point>           head_block_time;
 
       alignas(hardware_destructive_interference_sz)
-      std::atomic<block_num_type>           last_block_num_received{0};
+      std::atomic<fc::time_point>           last_block_received_time{};
 
       struct chain_info_t {
          block_id_type fork_db_root_id;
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3019,6 +3020,7 @@ namespace eosio {
       fc::raw::unpack( peek_ds, bh );
       const block_id_type blk_id = bh.calculate_id();
       const uint32_t blk_num = last_received_block_num = block_header::num_from_id(blk_id);
+      my_impl->last_block_num_received = blk_num;
```
greg7mdp, line 3023: shouldn't it be something like:
`my_impl->last_block_num_received = std::max(blk_num, my_impl->last_block_num_received);`

or are we guaranteed to receive blocks from only one connection at a time?

PR #1788, opened 2025-07-28
some initial very minor fc-ification of chaindb - logging & DB header pack vs cast

spoonincode:
Does some very minor fc-ification of chaindb now that they can be used together.

First is to replace all the cerr output with proper fc `log()`s. The benefit of this is consistency, and not just the output appearance but also that chaindb logs go to the configured logger (like GELF). Pretty uncontentious I think.

Second is to replace the casting of the DB header & environment, along with the `__attribute__ ((packed))` structs, and just use fc's `pack()` & `unpack()`. Unfortunately this ends up being unwieldy in a number of places because of trying to maintain compatibility with prior 1.x DB layouts. So unwieldy it probably just bloats the code more. So I'm willing to consider ripping this out and replacing it with something cleaner in 2.0 world where strict backwards compatibility isn't as needed.

### DIFF
--- a/libraries/chaindb/src/pinnable_mapped_file.cpp
+++ b/libraries/chaindb/src/pinnable_mapped_file.cpp
@@ -523,10 +523,10 @@ std::ostream& operator<<(std::ostream& osm, pinnable_mapped_file::map_mode m) {
 namespace fc {
 //Reconsider post-CHAINB01, when compiler can be stored as a proper string
 void to_variant(const chainbase::environment& bi, variant& v) {
-      v = fc::mutable_variant_object()("debug", bi.debug)
-                                      ("os", bi.os)
-                                      ("arch", bi.arch)
-                                      ("boost_version", bi.boost_version)
-                                      ("compiler", bi.compiler.data());
-   }
+   v = fc::mutable_variant_object()("debug", bi.debug)
+                                   ("os", bi.os)
+                                   ("arch", bi.arch)
+                                   ("boost_version", bi.boost_version)
+                                   ("compiler", bi.compiler.data());
+}
 }


libraries/chaindb/include/chainbase/environment.hpp
```diff
@@ -1,17 +1,29 @@
...
+
+consteval uint64_t operator""_u64str(const char* s, const size_t size) {
+   assert(size == 8);
+
+   uint64_t result = 0;
+   for(size_t i = 0; i < 8; ++i)
+      result |= static_cast<uint64_t>(s[i]) << (i * 8);
+
+   return result;
+}
```
heifner, line 15: Nice.

libraries/chaindb/src/pinnable_mapped_file.cpp
```diff
@@ -538,15 +539,29 @@ static std::string print_arch(environment::arch_t arch) {
...
-   return os;
+std::string environment::as_old_string() const {
+   std::stringstream ss;
+   ss << std::right << std::setw(17) << "Compiler: " << compiler.data() << '\n';
+   ss << std::right << std::setw(17) << "Debug: " << (debug ? "Yes" : "No") << '\n';
+   ss << std::right << std::setw(17) << "OS: " << print_os(os) << '\n';
+   ss << std::right << std::setw(17) << "Arch: " << print_arch(arch) << '\n';
+   ss << std::right << std::setw(17) << "Boost: " << boost_version/100000 << "."
+                                                  << boost_version/100%1000 << "."
+                                                  << boost_version%100;
```
heifner, line 550: Any reason to maintain this? Seems like pretty print of JSON of the variant object would be just fine.

heifner (approved):

libraries/chaindb/include/chainbase/environment.hpp
```diff
@@ -54,28 +66,68 @@ struct environment  {
...
+   }
 
-   bool operator==(const environment& other) const {
-      return !memcmp(this, &other, sizeof(environment));
-   } 
-   bool operator!=(const environment& other) const {
-      return !(*this == other);
+   template <typename DataStream>
+   friend DataStream& operator>>(DataStream& ds, environment& e) {
+      fc::raw::unpack(ds, e.debug);
```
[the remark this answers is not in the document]
linh2931: You are right, no better options than re-throwing here.

linh2931 (approved):

libraries/chaindb/src/pinnable_mapped_file.cpp
```diff
@@ -520,48 +518,15 @@ std::ostream& operator<<(std::ostream& osm, pinnable_mapped_file::map_mode m) {
...
-                                                  << boost_version/100%1000 << "."
-                                                  << boost_version%100;
-   return ss.str();
-}
-
 }
 
 namespace fc {
+//Reconsider post-CHAINB01, when compiler can be stored as a proper string
 void to_variant(const chainbase::environment& bi, variant& v) {
```
greg7mdp, line 525: very minor indentation

greg7mdp (approved):

PR #1793, opened 2025-07-29
always rely on JSON string format for outputting chaindb environment

spoonincode:
Removes the old pretty stringified method of outputting chaindb's environment and just leans on the JSON method. See conversation here https://github.com/AntelopeIO/spring/pull/1788#discussion_r2239859778 Seems like a nice cleanup even though it slightly reduces readability of the error.


heifner (commented):
Nice. Particularly like the removal of `as_old_string()`

PR #1804, opened 2025-07-31
[1.2.2] Add exhausted count and time to block report

### LINKED ISSUE #1798: Block report of exhausted trx
Currently block reports in producer plugin report exhausted trxs that will be retried as `success`.
For example:
```
debug 2025-07-29T19:35:59.174 nodeos    producer_plugin.cpp:306       report               ] Block #133 defproducerc trx idle: 0us out of 873428us, success: 2, 873197us, fail: 0, 0us, transient: 0, 0us, other: 231us
```
The 2 success here is actually one trx that was retried in this block and one trx that was retried in the previous block.
Seems we should report these as `fail` or add an `exhausted` category.

Also the 2 were reported in the the same block report because of a bug in how we restart exhausted blocks and reset the `_time_tracker`.
Specifically: https://github.com/AntelopeIO/spring/blob/12ef1e243cff2f16cc53df4d5743303be66463bd/plugins/producer_plugin/producer_plugin.cpp#L1059-L1074
The destructor of `block_time_tracker` which does the accounting is called after the call to `restart_speculative_block()` which clears the `block_time_tracker`.

heifner:
Add exhausted count and time to block report.
Also fixes an issue with restarting a block not reporting trx time in the correct block report.

This does change the debug log output of `producer_plugin` `debug` output which many block producers do have enabled. If that is seen as an issue we could target this to `main` instead. If we did that, not sure if we would just leave the bug in 1.2.x that reports exhausted trx in wrong block or just fix that. Of course, "fixing" that without adding `exhausted` would mean putting `exhausted` in `success` report.

Resolves #1798

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -209,7 +209,7 @@ private:
 struct block_time_tracker {
 
    struct trx_time_tracker {
-      enum class time_status { success, fail, other };
+      enum class time_status { success, fail, exhausted, other };
 
       trx_time_tracker(block_time_tracker& btt, bool transient)
           : _block_time_tracker(btt), _is_transient(transient) {}
@@ -222,6 +222,7 @@ struct block_time_tracker {
       trx_time_tracker& operator=(trx_time_tracker&&) = delete;
 
       void trx_success() { _time_status = time_status::success; }
+      void trx_exhausted() { _time_status = time_status::exhausted; }
 
       // Neither success nor fail, will be reported as other
       void cancel() { _time_status = time_status::other; }
@@ -234,6 +235,9 @@ struct block_time_tracker {
             break;
          case time_status::fail:
             _block_time_tracker.add_fail_time(_is_transient);
+            break;
+         case time_status::exhausted:
+            _block_time_tracker.add_exhausted_time(_is_transient);
             break;
          case time_status::other:
             _block_time_tracker.add_other_time();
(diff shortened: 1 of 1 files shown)


linh2931:
I think we should just put this fix on `main` without breaking backward compabability  to `1.2`. The bug is about debug block report and has been there since day one. It might be OK to wait until next major release to fix.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -303,20 +307,20 @@ struct block_time_tracker {
...
                   ("o", other_time)("rest", diff.count() > 5 ? ", diff: "s + std::to_string(diff.count()) + "us"s : ""s ) );
       }
    }
 
    void clear() {
       assert(!paused);
-      block_idle_time = trx_fail_time = trx_success_time = transient_trx_time = other_time = fc::microseconds{};
-      trx_fail_num = trx_success_num = transient_trx_num = 0;
+      block_idle_time = trx_success_time = trx_exhausted_time = trx_fail_time = transient_trx_time = other_time = fc::microseconds{};
+      trx_success_num = trx_exhausted_num = trx_fail_num = transient_trx_num = 0;
```
greg7mdp, line 323: nit: not requesting a change, just thinking that, at this point, I think it would be cleaner to have an array of `struct trx_result { fc::microseconds time_spent, uint32_t num_trx };` indexed by `time_status`.

heifner:
Going to target this fix to `main` as to not change the output of a release branch.

PR #1807, opened 2025-08-05
avoid accessing currently in progress `emplace()`ing `node_t`

spoonincode:
I am encountering a bizarre problem locally with the savanna cluster. Bizarre in that I can't explain why we're not seeing it in CI etc

When a savanna cluster is created in reserves some space in `_nodes` and then `emplace_back()` the `node_t`s,
https://github.com/AntelopeIO/spring/blob/ecc902d5e54d25ceec8ebde9b2c45e7078421ffb/unittests/savanna_cluster.hpp#L376-L385

During the `node_t` ctor, toward the end it,
https://github.com/AntelopeIO/spring/blob/ecc902d5e54d25ceec8ebde9b2c45e7078421ffb/unittests/savanna_cluster.cpp#L49-L57

So notice it will `cluster.get_new_blocks_from_peers(node_idx)` before completing the ctor. And look what `get_new_blocks_from_peers()` does,
https://github.com/AntelopeIO/spring/blob/ecc902d5e54d25ceec8ebde9b2c45e7078421ffb/unittests/savanna_cluster.hpp#L626-L632
It accesses the currently emplacing `node_t` -- or rather where it _will_ be -- in `_nodes`. But it isn't done being created yet.

Locally on a debug build via gcc I fail at line 627 -- but it's not the `assert()` that fails rather it fails earlier due to the `vector::operator[]` internally asserting because it believes its size is still 0 at this time (indeed looking at vector's source shows it increments its size after the ctor of the emplaced object completes). 

Locally on a release build via the pinned compiler I fail by what I'm pretty sure is `LIBCXX_HARDENING_MODE=fast` asserting.

So I'm just completely perplexed why this is not failing in CI via the pinned builds. Or why it's not failing for anyone else.

I moved the code around a little to not call back in to `cluster` during the ctor to resolve the problem. Not sure if this is best approach or not.

### DIFF
--- a/unittests/savanna_cluster.cpp
+++ b/unittests/savanna_cluster.cpp
@@ -46,15 +46,17 @@ node_t::node_t(size_t node_idx, cluster_t& cluster, setup_policy policy /* = set
       }
    };
 
-   auto node_initialization_fn = [&, node_idx]() {
+   auto node_initialization_fn = [&]() {
       [[maybe_unused]] auto _a = control->voted_block().connect(_voted_block_cb);
       [[maybe_unused]] auto _b = control->accepted_block().connect(_accepted_block_cb);
       tester::set_node_finalizers(_node_finalizers);
-      cluster.get_new_blocks_from_peers(node_idx);
    };
 
-   node_initialization_fn();                  // initialize the node when it is first created
-   set_open_callback(node_initialization_fn); // and do the initalization again every time `open()` is called
+   node_initialization_fn();                                    // initialize the node when it is first created
+   set_open_callback([&, node_initialization_fn, node_idx]() {  // and do the initialization again every time `open()` is called + simulate blocks from peers
+      node_initialization_fn();
+      cluster.get_new_blocks_from_peers(node_idx);
+   });
 }
 
 node_t::~node_t() {}


greg7mdp:
> `vector::operator[]` internally asserting because it believes its size is still 0 at this time

My glibc version doesn't check the size:
```
      _GLIBCXX_NODISCARD _GLIBCXX20_CONSTEXPR
      reference
      operator[](size_type __n) _GLIBCXX_NOEXCEPT
      {
	__glibcxx_requires_subscript(__n);
	return *(this->_M_impl._M_start + __n);
      }
```

and the `reserve`  allocated a buffer large enough for all the `node_t` and `this->_M_impl._M_start` should point to that buffer.

spoonincode:
> My glibcxx version doesn't check the size:

The `__glibcxx_requires_subscript` should be the check; but I think the library needs to have assertions enabled somehow and I'm not sure off hand how (it is quite possible my local gcc has hardening enabled by default via some mechanism).

greg7mdp:
Yes, in my case `_GLIBCXX_ASSERTIONS` doesn't appear to be defined (maybe because I'm building with clang), so I get the empty checks (even with a debug build).

```
#ifndef _GLIBCXX_ASSERTIONS
# define __glibcxx_requires_non_empty_range(_First,_Last)
# define __glibcxx_requires_nonempty()
# define __glibcxx_requires_subscript(_N)
#else

// Verify that [_First, _Last) forms a non-empty iterator range.
# define __glibcxx_requires_non_empty_range(_First,_Last)	\
  __glibcxx_assert(_First != _Last)
# define __glibcxx_requires_subscript(_N)	\
  __glibcxx_assert(_N < this->size())
// Verify that the container is nonempty
# define __glibcxx_requires_nonempty()		\
  __glibcxx_assert(!this->empty())
#endif
```

spoonincode:
ah yeah this seems to be a new gcc15 thing:
https://gcc.gnu.org/gcc-15/changes.html#libstdcxx
> **Runtime Library (libstdc++)**
> * Debug assertions are now enabled by default for unoptimized builds. Use -D_GLIBCXX_NO_ASSERTIONS to override this.

greg7mdp:
>  but I'm still not seeing it -- `get_new_blocks_from_peers()` runs at the very end of `node_t`'s ctor so nothing internally to `node_t` should be in a wonky state at that point

You're right! Besides the fact that it is not very clean to access an object before the constructor is done,  I think the only issue is the `vector::operator[]` bounds check that fails because the `vector::_M_finish` member is only updated after the object being constructed.

spoonincode:
Probably in the flurry of troubleshooting the problem I got a little mixed up: it looks like I was still locally doing a debug build with pinned toolchain, and it was the expression inside of the `assert()` that was getting flagged by `LIBCXX_HARDENING_MODE=fast` similar to how gcc did. A local `Release` build with the pinned toolchain does indeed complete successfully like it does in CI. So no weird CI vs non-CI discrepancy after all.

heifner:
Not backporting to 1.2.2 ?

heifner:
> (and it only affects debug builds)

Makes sense. Thanks.

Issue #1810, opened 2025-08-06
BP stops making progress


heifner:
Reported by a Vaulta BP.

```
info  2025-08-06T04:16:22.274 nodeos    controller.cpp:3621           log_applied          ] Received block 8daf25d6358d8b64... #450381336 @ 2025-08-06T04:16:22.500 signed by starteosiobp [trxs: 7, lib: 450381334, net: 1120, cpu: 27451 us, elapsed: 9964 us, applying time: 17764 us, latency: -225 ms]
debug 2025-08-06T04:16:22.274 nodeos    producer_plugin.cpp:2957      schedule_delayed_pro ] Scheduling Speculative/Production Change at 2025-08-06T04:16:22.774
debug 2025-08-06T04:16:22.274 nodeos    producer_plugin.cpp:2258      start_block          ] Starting block #450381337 2025-08-06T04:16:23.000 producer starteosiobp, deadline 2025-08-06T04:16:22.774
debug 2025-08-06T04:16:22.274 nodeos    producer_plugin.cpp:167       report               ] Failed 13 trxs, account: miner2.sx, sub bill: 1388653us, reason: tx_cpu_usage
debug 2025-08-06T04:16:22.274 nodeos    producer_plugin.cpp:2422      remove_expired_trxs  ] Processed 27 expired transactions of the 426 transactions in the unapplied queue.
debug 2025-08-06T04:16:22.274 nodeos    subjective_billing.hpp:186    remove_expired       ] Processed 426 subjective billed transactions, Expired 27
debug 2025-08-06T04:16:22.274 nodeos    producer_plugin.cpp:2903      schedule_production_ ] Speculative Block Created; Scheduling Speculative/Production Change
debug 2025-08-06T04:16:22.274 nodeos    producer_plugin.cpp:2969      schedule_delayed_pro ] Not Scheduling Speculative/Production, no local producers had valid wake up times
debug 2025-08-06T04:16:22.719 nodeos    producer_plugin.cpp:312       report               ] Block #450381337 starteosiobp trx idle: 441471us out of 445342us, success: 7, 3436us, fail: 10, 178us, transient: 0, 0us, other: 257us
debug 2025-08-06T04:16:22.722 nodeos    producer_plugin.cpp:853       on_accepted_block    ] Removed applied transactions before: 406, after: 402
debug 2025-08-06T04:16:22.722 nodeos    subjective_billing.hpp:160    on_block             ] Subjective billed accounts 169 removed 0
[...]

PR #1812, opened 2025-08-07
[experiment] protect against timer "rearm race"

spoonincode:
There appears to be an ABA-like defect in how the timer is managed. I found this while scrutinizing the code more due to #1810. I am not sure it is a smoking gun for #1810 though.

The problem goes something like this,
1) `start()` a timer which starts the kernel timer (i.e. not a 'forever timer'); this also sets `_state = running`
2) the kernel timer expires, and makes some progress between queuing the signal, waking a thread (could be any), and `expire_now()` may even start. But critically `expire_now()` does not reach the CAS yet. https://github.com/AntelopeIO/spring/blob/f38100e4b523c36afb43ade0ff298ea18b98a107/libraries/chain/platform_timer_posix.cpp#L81-L83
3) `stop()` is called. `stop()` identifies that  `_state = running`, sets `_state = interrupted` and disarms the kernel timer. However note that the kernel timer has already expired at this moment due to the criteria specified in line 2.
4) `start()` is called. This sets `_state = running` and potentially arms a kernel timer if non-max
5) The thread in line 2 continues to make progress and reaches the CAS. Here it sees that `_state == running` and triggers the timer expiry via `_state = timed_out` and calls the expiry callbacks. Thus the problem: this expiration was intended for the timer started on line 1, but instead we trigger the timer started on line 4.

Fixing this was a bit elusive. Probably the classical method would be adding a 'generation' for each timer start and verify the expiry matches the current generation. But `timer_settime()` doesn't let us stash anything away indicating what timer generation an expiry corresponds to. The only information we can deliver on expiry signal is the 64-bit value given on timer creation (and the signal number). So we could potentially create and destroy the timer each time we start/stop it. That way we could encode some generation in the 64-bit payload that is delivered to the signal (smuggle the generation in some unused address bits (still need `this`) or something). But I wanted to avoid that it seemed like more headache.

So this approach is to identify if the kernel timer has already expired when stopping the timer, and if that is the case wait for the signal handler to complete. The principle is that if the kernel timer expires the signal handler must eventually be run and we can wait for that to complete as a synchronization mechanism that ensures no pending signals or half-run `expire_now()`s are lingering around after `stop()` to cause potential grief later on.

Unfortunately I fear I may be too trusting of the kernel here. Is it _truly guaranteed_ that if a timer returns a 0 value that a signal was queued? Is it _truly guaranteed_ that if a timer returns a non-0 value that a signal will _not_ be queued? My hunch is these are probably unrealistic expectations and this approach is not viable.

### DIFF
--- a/libraries/chain/include/eosio/chain/platform_timer.hpp
+++ b/libraries/chain/include/eosio/chain/platform_timer.hpp
@@ -58,7 +58,7 @@ private:
    bool timer_running_forever = false;
 
    struct impl;
-   constexpr static size_t fwd_size = 8;
+   constexpr static size_t fwd_size = 16;
    fc::fwd<impl,fwd_size> my;
 
    void call_expiration_callback() {
--- a/libraries/chain/platform_timer_posix.cpp
+++ b/libraries/chain/platform_timer_posix.cpp
@@ -20,6 +20,7 @@ static_assert(std::atomic_bool::is_always_lock_free, "Only lock-free atomics AS-
 
 struct platform_timer::impl {
    timer_t timerid;
+   std::atomic<bool> clear_on_signal_handler_complete;
 
    static void sig_handler(int, siginfo_t* si, void*) {
       platform_timer* self = (platform_timer*)si->si_value.sival_ptr;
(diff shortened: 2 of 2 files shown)


greg7mdp:
I wonder if, within the platform_timer class, we could use two separate posix timers in a flip-flop mode (each with their own `state_t` tracking variable) to solve the issue (so in step 5, the sig_handler (holding a pointer to one of the internal timer state) would see `_state == stopped` and would not trigger an interrupt)?

spoonincode:
Yep one way of doing that would be to have two kernel timers: one that fires on SIGRTMIN, and one that fires on SIGRTMIN+1, and then treat that bit of information as a comparison to the LSb of your current generation. Of course this mutates the problem from ABA to... ABCA?

I am not sure the "depth" of lagging timer notifications we are willing to accept. We could even round robin across 16 of them (and compare to 4 LSb of generation) . Is that enough :woman_shrugging: 

Forcing the signal to arrive on the same thread as start/stop doesn't seem to completely absolve this issue either, but it might increase the success rate. The additional complication now is that in theory the signals could be delivered to a non-scheduled thread. At least if we force the signal to be delivered to the same thread that is calling start/stop you know the signal will be promptly delivered.... the kernel won't hold on to a pending signal across multiple syscalls etc.

heifner (approved):

Issue #1809, opened 2025-08-06
[v2.0.0-dev1.1] nodeos crashes when a contract tries to remove a row that is removed in sync call.


taokayan:
version:
commit 9f942af6cc1ba91fdcce1d2b0e798c4df7ec4692 (HEAD, tag: v2.0.0-dev1.1, origin/release/2.0.0-dev1)

```
info  2025-08-06T09:33:05.964 nodeos    controller.cpp:3634           log_applied          ] Produced block a9f5848a48e70a14... #364 @ 2025-08-06T09:33:06.000 signed by initb [trxs: 0, lib: 351, confirmed: 12, net: 0, cpu: 100 us, elapsed: 383 us, producing time: 463747 us]
info  2025-08-06T09:33:06.426 nodeos    controller.cpp:3634           log_applied          ] Produced block 066ee655d63074db... #365 @ 2025-08-06T09:33:06.500 signed by initb [trxs: 0, lib: 351, confirmed: 0, net: 0, cpu: 100 us, elapsed: 314 us, producing time: 461529 us]
nodeos: /home/kkk3/workspaces/spring/libraries/boost/libs/intrusive/include/boost/intrusive/avltree_algorithms.hpp:558: static void boost::intrusive::avltree_algorithms<NodeTraits>::rebalance_after_erasure_restore_invariants(boost::intrusive::avltree_algorithms<NodeTraits>::node_ptr, boost::intrusive::avltree_algorithms<NodeTraits>::node_ptr, boost::intrusive::avltree_algorithms<NodeTraits>::node_ptr) [with NodeTraits = chainbase::offset_node_traits<boost::multi_index::ordered_unique<boost::multi_index::tag<by_id>, boost::multi_index::member<eosio::chain::key_value_object, chainbase::oid<eosio::chain::key_value_object>, &eosio::chain::key_value_object::id> > >; boost::intrusive::avltree_algorithms<NodeTraits>::node_ptr = chainbase::offset_node_base<boost::multi_index::ordered_unique<boost::multi_index::tag<by_id>, boost::multi_index::member<eosio::chain::key_value_object, chainbase::oid<eosio::chain::key_value_object>, &eosio::chain::key_value_object::id> > >*]: Assertion `false' failed.
[...]

PR #1814, opened 2025-08-07
[2.0.0] SC: Broadcast erasure of an iterator object and invalidate the object in all the iterator caches along sync call path

linh2931:
Erasing a row object in a sync call can make iterators for the object in iterator caches acquired in ancestor sync_call_context/apply_context invalid. The erasure needs to be broadcast to all the prior iterator caches and invalidate the iterator if it is there.

Added a number of tests for the fix.

Resolves https://github.com/AntelopeIO/spring/issues/1809


libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -81,7 +87,19 @@ class host_context {
...
+
+            EOS_ASSERT(0 <= itr->second && (size_t)itr->second < _iterator_to_object.size(), invalid_table_iterator, "iterator out of range");
+
+            _object_to_iterator.erase((const T*)obj_ptr);
+            _iterator_to_object[itr->second] = nullptr;
+         }
+
       private:
+         host_context&                                   _context;
          map<table_id_object::id_type, pair<const table_id_object*, int>> _table_cache;
```
spoonincode, line 103: Do we need to do anything similar with the table cache? I would have guessed so but I haven't studied it yet

linh2931: I think it should be OK not to update the table cache. When all rows in a table are removed, even though the table cache still has an entry for the table, that entry will not be used anymore because no iterators exist for that table.

libraries/chain/include/eosio/chain/host_context.hpp
```diff
@@ -81,7 +87,19 @@ class host_context {
...
 
+         // Invalid the iterator entry for the  given object pointer
+         void invalidate(const void* obj_ptr) {
+            auto itr = _object_to_iterator.find((const T*)obj_ptr);
+            if (itr == _object_to_iterator.end()) return;
+
+            EOS_ASSERT(0 <= itr->second && (size_t)itr->second < _iterator_to_object.size(), invalid_table_iterator, "iterator out of range");
+
+            _object_to_iterator.erase((const T*)obj_ptr);
+            _iterator_to_object[itr->second] = nullptr;
```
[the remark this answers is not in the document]
linh2931: Thanks! It's safer to swap the lines.

spoonincode:
> Why does every context in the sync call chain have its own iterator caches? Couldn't the called context point to the iterator caches of the callers when sync calls?

If I'm understanding what you're suggesting, I think the main problem is that a callee could move iterators of its callers. It would effectively mean callers could not trust their iterators after making a sync call.

greg7mdp:
> > Why does every context in the sync call chain have its own iterator caches? Couldn't the called context point to the iterator caches of the callers when sync calls?
> 
> If I'm understanding what you're suggesting, I think the main problem is that a callee could move iterators of its callers. It would effectively mean callers could not trust their iterators after making a sync call.

Ah, OK, I definitely do not understand that part very well. Thanks for the comment.

spoonincode (approved):
This implementation seems correct to me; I don't have strong feelings about the specific approach or requirement.

linh2931:
> still wondering why it would not be possible to share the iterator cache along the sync call call chain?

I think for now we get the current design in to fix dev1.1. I still think it would be better to have separate iterator caches for sync calls. Not sure how often the use cases for a chain of sync calls touching the same table. 

@arhag ?

PR #1819, opened 2025-08-08
protect against timer "rearm race": more timers approach

spoonincode:
A different solution for the problem described in #1812

This approach uses more kernel timers -- each with a different signal number -- and round robins across them each time the platform_timer is started. It is kind of like storing a per-timer-start payload in the timer. The signal number upon expiration is used as validation that the signal handler corresponds to the currently running timer before taking action to expire it. This solves the ABA like problem, but turns it more in to an ABCDA like problem which is better but not perfect.

The choice of 4 timers is arbitrary. The more used the more resilient against mixups, in theory anyways. I actually don't initially see any technical reasons to not go with something like 16 because these timer handles don't consume e.g. file handle space.

### DIFF
--- a/libraries/chain/include/eosio/chain/platform_timer.hpp
+++ b/libraries/chain/include/eosio/chain/platform_timer.hpp
@@ -62,7 +62,7 @@ private:
    generation_t generation = 0;
 
    struct impl;
-   constexpr static size_t fwd_size = 32;
+   constexpr static size_t fwd_size = 64;
    fc::fwd<impl,fwd_size> my;
 
    void call_expiration_callback() {
--- a/libraries/chain/platform_timer_posix.cpp
+++ b/libraries/chain/platform_timer_posix.cpp
@@ -19,7 +19,7 @@ namespace eosio::chain {
 static_assert(std::atomic_bool::is_always_lock_free, "Only lock-free atomics AS-safe.");
 
 struct platform_timer::impl {
-   constexpr static unsigned num_timers = 4;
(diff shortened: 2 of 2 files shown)


spoonincode:
I don't think anything in particular. Yeah we could have separate structs for each timer, like
```cpp
struct {
   generation_t timer_generation;
   platform_timer* timer;
}
```
To avoid the indirection could also stuff the timer_generation in to the unused high bits of the pointer. I can try it that way

linh2931 (approved):
Why not try `8`? The likelihood that abcdefgha happens would be close to 0.

heifner:
I personally like #1812 better as it is not acceptable to ABCDA.

spoonincode:
Me too but I don't trust the conditions the last paragraph goes in to
> Unfortunately I fear I may be too trusting of the kernel here. Is it _truly guaranteed_ that if a timer returns a 0 value that a signal was queued? Is it _truly guaranteed_ that if a timer returns a non-0 value that a signal will _not_ be queued? My hunch is these are probably unrealistic expectations and this approach is not viable.

Because if we get it wrong worst case we'll spin lock forever waiting for a signal that never comes.

spoonincode:
Another consideration may be which approach is safer to send immediately to stable, if we chose to do so (an in-the-wild BP failure is pretty serious..). Personally I feel more comfortable sending the timer generation change to stable vs #1812. Discussion to be had this week.

spoonincode:
Will be going with #1820 approach

PR #1825, opened 2025-08-11
[experiment] use submoduled LLVM library

spoonincode:
An experiment to bundle LLVM as a submodule. Works.. not so bad. And doesn't even double the build time (for me anyways). Still lots that needs to be cleaned up (libtester!). But a good spot to start discussion.


linh2931:
I tried llvmsybmod branch on my ubuntu24 box. The build was successful without me modifying CMakeList.txt and took about only 2 minutes longer.

PR #1828, opened 2025-08-14
[2.0.0-dev] upgrade to boost 1.89.0

spoonincode:
Released today, I'm eager to update to clear up some warnings. Unfortunately boostorg/container#300 remains and is still very noisy.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -19,6 +19,7 @@
 #include <fc/time.hpp>
 
 #include <boost/asio.hpp>
+#include <boost/asio/deadline_timer.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/signals2/connection.hpp>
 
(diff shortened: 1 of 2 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -19,6 +19,7 @@
 #include <fc/time.hpp>
 
 #include <boost/asio.hpp>
+#include <boost/asio/deadline_timer.hpp>
```
spoonincode, line 22: deadline_timer has been deprecated since 1.87, and in 1.89 it has been removed from the `<boost/asio.hpp>` include. We should spend a few minutes some time and refactor to system_timer (I don't think too difficult).

linh2931 (approved):

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
spoonincode: okay I made #1830 but I don't think we'll forget forever since eventually deadline_timer will just be removed :grinning:

PR #1834, opened 2025-08-20
[1.2.3] use debian's package snapshot repo for old buster packages needed in reproducible build

spoonincode:
Unfortunately it is no longer possible to build the 1.x pinned/reproducible package from scratch because it is based on Debian 10 which has had its packages removed from Debian's package repo around July 12 (it has been out of support for about a year now).

This moves the build to use Debian's package snapshot repo which contains all prior releases. This was previously plumbed through here (since pinning to exact packages would presumably be even more reproducible), though I had disabled it in the past since the package snapshot repo can be kind of slow.

Taking this approach will ensure no operational changes to any users who are using 1.2.x on an old distribution like Ubuntu 18. Though we didn't actually claim support for these old distributions, so it's not clear how aggressively we should maintain this compatibility. Maybe in 1.3 we could increase the base build to Debian 11 (it already is on our 2.0 branch).

### DIFF
--- a/tools/reproducible.Dockerfile
+++ b/tools/reproducible.Dockerfile
@@ -1,21 +1,20 @@
 # syntax=docker/dockerfile:1
-# debian:buster on Sep 20 2023
-FROM debian@sha256:d774a984460a74973e6ce4d1f87ab90f2818e41fcdd4802bcbdc4e0b67f9dadf AS builder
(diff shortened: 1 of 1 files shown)


linh2931 (approved):
Should the PR title be `[1.2.2]`?

PR #1845, opened 2025-08-25
validate LLVM & cmake tarball hashes in reproducible build

spoonincode:
we already validate the GPG signatures on LLVM & cmake in our reproducible build, but similar to reasoning of #1841 maybe we'd like to validate the exact hashes as well to prevent some attack vectors.


spoonincode:
ah this is busted until #1843 goes in, I should have chained it off that branch

PR #1868, opened 2025-09-03
[2.0.0-dev] add xxhash

spoonincode:
Adds new non-cryptographic hash type to libfc

### DIFF
--- a/libraries/libfc/include/fc/crypto/xxh3.hpp
+++ b/libraries/libfc/include/fc/crypto/xxh3.hpp
@@ -1,0 +1,62 @@
+#pragma once
+
+#include <fc/fwd.hpp>
+#include <fc/crypto/packhash.hpp>
+#include <fc/io/raw_fwd.hpp>
+
+#include <xxHash/xxhash.h>
+
+namespace fc {
+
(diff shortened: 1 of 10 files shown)


libraries/libfc/include/fc/crypto/xxh3.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace detail {
+struct alignas(64) alignmeto64 {};
+
+template <typename T>
+concept ContiguousCharSource = requires(const T& obj) {
+   { obj.data() } -> std::convertible_to<const char*>;
+   { obj.size() } -> std::unsigned_integral;
+};
+}
```
spoonincode, line 19: We really need a better place for stuff like this -- even just putting in `detail` here doesn't protect against against ODR oopsies since it's just in `fc::detail` still.

greg7mdp (approved):

libraries/libfc/include/fc/crypto/xxh3.hpp
linh2931: Put in a separate file with a different name space?

PR #1877, opened 2025-12-04
Test: Support multiple threads writing to subprocess_results.log

heifner:
The read only tests use multiple threads to push transactions. These end up trying to write to the `subprocess_results.log` at the same time. Protect the write with a lock.

### DIFF
--- a/tests/TestHarness/testUtils.py
+++ b/tests/TestHarness/testUtils.py
@@ -93,15 +93,15 @@ class Utils:
     @staticmethod
     def checkOutputFileWrite(time, cmd, output, error):
         stop=Utils.timestamp()
-        os.makedirs(Utils.TestLogRoot, exist_ok=True)
-        os.makedirs(Utils.DataPath, exist_ok=True)
-
-        # Ensure filename is set
-        if not hasattr(Utils, "checkOutputFilename"):
-            Utils.checkOutputFilename=f"{Utils.DataPath}/subprocess_results.log"
-
         # Serialize concurrent writes and open file per write to avoid sharing closed handles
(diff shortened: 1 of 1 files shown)


greg7mdp:
Wouldn't the following work and avoid re-opening the file every time `checkOutputFileWrite` is called?

```python
        with Utils._check_output_lock:
            if not hasattr(Utils, "checkOutputFile"):
                Utils.checkOutputFilename=f"{Utils.DataPath}/subprocess_results.log"
                if Utils.Debug: Utils.Print("opening %s in dir: %s" % (Utils.checkOutputFilename, os.getcwd()))
                Utils.checkOutputFile=open(Utils.checkOutputFilename,"w")

        with Utils._check_output_lock:
            Utils.checkOutputFile.write(Utils.FileDivider + "\n")
            Utils.checkOutputFile.write("start={%s}\n" % (time))
            Utils.checkOutputFile.write("cmd={%s}\n" % (" ".join(cmd)))
            Utils.checkOutputFile.write("cout={%s}\n" % (output))
            Utils.checkOutputFile.write("cerr={%s}\n" % (error))
            Utils.checkOutputFile.write("stop={%s}\n" % (stop))
```

heifner:
> Wouldn't the following work and avoid re-opening the file every time `checkOutputFileWrite` is called?

I kind of like opening to append each time. The close makes sure it is flushed. Not like performance matters here.

greg7mdp:
Sound fine. Still I think we should take the lock for the open as well to avoid contention when writing into `Utils.checkOutputFilename`.

greg7mdp (approved):

PR #1879, opened 2025-12-05
P2P: Signal controller to apply blocks if sync interrupted

### LINKED ISSUE #1878: P2P sync issue
Node got stuck syncing:
```
info  2025-11-30T02:14:02.787 oc-monito wasm_interface_private:125    operator()           ] EOS VM OC tier up interrupting 8826a349eea89556a508ef62fc754c8c8a7a1a9e58c5a8368604ed5801a1c835
info  2025-11-30T02:14:02.789 nodeos    controller.cpp:3895           apply_block          ] Interrupt of trx id: 4e5506d3f6af9abdef9c32ebe97b79785002e70ef60b00859e8cc5322eb808a6
info  2025-11-30T02:14:02.789 nodeos    controller.cpp:4580           operator()           ] interrupt while applying block 30346002 : 01cf0b12485d52f25c4197c3767611f4d0cdb6cffdb4a77f0a9fd43a55756e16
```
... then node got stuck
`sync-fetch-span = 200` which is less than LIB range as this was WAX.

```
debug 2025-11-30T09:49:49.039 net-2     net_plugin.cpp:2184           is_sync_request_ahea ] sync ahead not allowed. block 30356901, sync_last_requested_num 30357032, sync-fetch-span 200
```
Can't make progess.

heifner:
It is possible due to an interrupt that the controller has stopped applying blocks. Give it a kick to make sure it continues.

I was not able to reproduce the problem, but looking at the logs I think this will fix the issue. Hopeful, to get feedback from reporter if this fixes the issue.

Resolves #1878

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2178,6 +2178,10 @@ namespace eosio {
          fc_dlog(p2p_blk_log, "sync ahead not allowed. block ${bn}, head ${h}, fhead ${fh}, fhead->lib ${fl}, sync-fetch-span ${sp}, fork_db size ${s}",
                  ("bn", blk_num)("h", head_num)("fh", cc.fork_db_head().block_num())("fl", cc.fork_db_head().irreversible_blocknum())
                  ("sp", sync_fetch_span)("s", cc.fork_db_size()));
+
+         // make sure controller is processing blocks, might have been interrupted and need a kick
+         fc_dlog(p2p_blk_log, "sync post process_incoming_block to app thread, block ${n}", ("n", blk_num));
+         my_impl->producer_plug->process_blocks();
       }
 
       fc_dlog(p2p_blk_log, "sync ahead not allowed. block ${bn}, sync_last_requested_num ${lrn}, sync-fetch-span ${s}",


greg7mdp:
hum, looking at the code I'm wondering if `producer_plugin::process_blocks()` has some possible unbounded recursion (probably not likely at all though).

heifner:
> hum, looking at the code I'm wondering if `producer_plugin::process_blocks()` has some possible unbounded recursion (probably not likely at all though).

The `post` starts a new thread stack. If it did `dispatch` then it could potentially be unbounded recursion.

greg7mdp:
> The `post` starts a new thread stack. If it did `dispatch` then it could potentially be unbounded recursion.

Yes, but the post calls `process_incoming_blocks(process_incoming_blocks);` which [calls itself recursively](https://github.com/AntelopeIO/spring/blob/8a7415a70ea8f764f240a7b192fba406007ebdff/plugins/producer_plugin/producer_plugin.cpp#L3051),  so just executing the posted task could cause a stack overflow right?

heifner:
> > The `post` starts a new thread stack. If it did `dispatch` then it could potentially be unbounded recursion.
> 
> Yes, but the post calls `process_incoming_blocks(process_incoming_blocks);` which calls itself recursively, so just executing the posted task could cause a stack overflow right?

No.

process_incoming_blocks -> calls on_incoming_block, then calls post( process_incoming_blocks ), and then current stack ends

The post'ed task is picked up to execute:

process_incoming_blocks -> calls on_incoming_block, then calls post( process_incoming_blocks ), and then current stack ends

etc.

greg7mdp:
How did I miss that. Thanks for pointing it out!

